PR #14, opened 2021-09-05
Add comments to issue templates

marko-radosavljevic:
The main improvement is adding comments to issue template instructions, so they don't leak if the user forgets to delete them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,27 +1,30 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: ''
-labels: bug
+title: "[BUG]"
+labels: 'Type: Bug'
 assignees: ''
 
 ---
 
 **Describe the bug**
-A clear and concise description of what the bug is.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,27 +1,30 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: ''
-labels: bug
+title: "[BUG]"
+labels: 'Type: Bug'
```
marko-radosavljevic, line 5: This is something I would like opinion on.

I like "[TITLE]" because it shows clearly that the user is using our template, and it's clear what the issue is about.
I like "Type: Issue" as a label, because sometimes issues have many tags, and this way it's easier to spot the Type and Priority, and other more important labels.

Zabuzard: Same for `Feature` then?

Zabuzard: NVM, you did 👌

Zabuzard (approved):

Tais993 (approved):

PR #29, opened 2021-09-07
Added ".editorconfig"

Zabuzard:
- mostly IJ default settings, tweaked a bit.
- spaces, CRLF, 120, wrap, strict

Try it out and propose change, if any. Otherwise we will start with that as a base.

(also applied auto formatting on all files, so some changes were made - all reviewed by me)


Zabuzard:
Not sure if it works in other IDEs than Intellij. Probably not correctly, maybe in Eclipse but I did not test it.

Ideally we should provide a proper setup for Intellij, Eclipse and Netbeans, as much as possible.

Anyways, this is an Intellij setup for now.

Tais993 (approved):

.editorconfig
```diff
@@ -0,0 +1,536 @@
...
+ij_java_keep_simple_methods_in_one_line = false
+ij_java_label_indent_absolute = false
+ij_java_label_indent_size = 0
+ij_java_lambda_brace_style = end_of_line
+ij_java_layout_static_imports_separately = true
+ij_java_line_comment_add_space = false
+ij_java_line_comment_at_first_column = true
+ij_java_method_annotation_wrap = split_into_lines
+ij_java_method_brace_style = end_of_line
+ij_java_method_call_chain_wrap = split_into_lines
```
I-Al-Istannen, line 134: I'd only chop down if necessary. I think that'd be 
```suggestion
ij_java_method_call_chain_wrap = on_every_item
```

I-Al-Istannen (changes requested):

marko-radosavljevic (commented):
Do we need configs for each IDE? What about vscode or vim?

We should have auto-formatting inside the pipeline imo, so one tool that takes care of everything. Seems like a much more universal solution that will potentially save us a lot of hassle with different IDE's.

Also, it provides much more flexibility and comfort for everyone working on the project, because they can enjoy their own formatting how they like it, and on push it gets formatted for the repo.

We should also have .gitattributes with LF/crlf normalization for each file type, git gets confused on some file types, and it's an annoying thing to deal with.

Zabuzard:
There are ideas of going with a different setup that allows all (most) IDEs.
Some sort of Google Style with Spotless or similar (@I-Al-Istannen tries out some stuff).

Holding back this PR for now.

.editorconfig
```diff
@@ -0,0 +1,536 @@
+[*]
+charset = utf-8
+end_of_line = crlf
```
d3vel0per, line 3: Shouldnt this be just linefeed as our deployments would be on a linux box. Eg. an executable script will most likely error out with crlf 🤔
IDE's are anyways okay with both formats.

```suggestion
end_of_line = lf
```

Zabuzard:
The project went into a different direction and we now have an Eclipse Google Style configuration that can be imported into Intellij, as well as SonarLint that is setup nicely. Stuff is still going on but we definitely will not use an `.editorconfig` as suggested here.

Hence, closing this.

Issue #10, opened 2021-09-05
How to design a basic command system?


Tais993:
By making usage of a command system we can make the creation of commands really easy.

A user only needs to create 1 class that extends a "Command" class.
This automatically requires the override of multiple methods that create the help command, and more.

### We would have to decide whenever we want **reflection** or **hardcoded commands**

By making usage of reflection, the user only needs to extend 1 class. (Which would also be required when hardcoded commands)
The command will automatically be added, and automatically run.
Disabling commands can be done by overriding a (non abstract) method (isDisabled a.e) that by default returns false.
This is really easy for everyone.

I can't comment on advantages of hardcoded commands, if anyone has any I'll add those here.

Feel free to add counter arguments against either in the commands, those will be added to this message.

### React with the corresponding emoji to vote

- 🚀 Reflection
- 🥇 Hardcoded commands

illuminator3:
I don't think reflection would be good, since we aim to be a beginner friendly and reflection isn't really beginner friendly at all. It could also cause portability issues over different Java versions.

Tais993:
How would it not be beginner friendly?
I feel like it's even more beginner friendly, they don't have to add it to any List, or anything.
Just extending a class would be enough.

>  could also cause portability issues over different Java versions.

I'd assume we will use the latest Java, but feel free to discuss this in #13

d3vel0per:
We really need not use reflection as such. We can leverage something like a service registry pattern.
Commands would have various classes of interfaces based on the kind of work they are supposed to do. And all command implementations will register in a central command registry at compile time which by definition would be a hardcoded list. It would be a ticking timebomb if any new command added dynamically gets deployed in production just because it adheres to a "command interface". We should have some control over the which commands are OK to run in production vs which aren't. Either it starts as a static list, can become a dynamic list managed in the database or if we have very fancy use cases rather use a rules engine. This way apart from the primary deployment even the command deployments can be rolled out in a controlled manner.

Its another case if we want to leverage a DI framework which of course will use reflection/proxying behind the scenes but does the use case ask for those level of abstractions [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

Additionally since the new bot is supposed to be an aggregation of a whole lot of existing features I feel it would be good if the commands are isolated from each other in the runtime ie. a single bad command shouldn't bring down the whole bot engine.

Tais993:
[...]

PR #30, opened 2021-09-07
CommandHandler

Tais993:
Will be the PR for the command handler and more.
Currently this is only the Command class, open for feedback!

Issue #10

### DIFF
--- a/src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
+++ b/src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
@@ -1,0 +1,53 @@
+package org.togetherjava.tjbot.commands;
+
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.requests.restaction.CommandCreateAction;
+
+/**
+ * The class all commands have to abstract.
+ * Look at {@link org.togetherjava.tjbot.commands.example.ExampleCommand} for an example.
+ * A second example will come for sub-commands.
+ */
+public abstract class AbstractCommand {
+
+    /**
+     * The commands name.
+     * @return
+     *  The commands name as a {@link String}
+     */
+    public abstract String getCommandName();
+
+    /**
+     * The commands description.
+     * @return
+     *  The commands description as a {@link String}
+     */
+    public abstract String getDescription();
+
+    /**
+     * Whenever the command is only for guilds, optional method.
+     * @return
+     *  Whenever the command is only for guilds as a {@link Boolean}
+     */
+    public boolean isGuildOnly() {
+        return false;
+    }
+
+    /**
+     * This command will add all the options to the slash-command.
+     * Look at {@link org.togetherjava.tjbot.commands.example.ExampleCommand#addOptions(CommandCreateAction)} for an example.
+     * @param commandCreateAction
+     *  The {@link CommandCreateAction} where the options need to be added.
+     * @return
+     *  The changed {@link CommandCreateAction}
+     */
+    public abstract CommandCreateAction addOptions(CommandCreateAction commandCreateAction);
+
+    /**
+     * The execute method for when the command gets executed.
+     * Check <a href="https://github.com/DV8FromTheWorld/JDA/wiki/Interactions#slash-commands">JDA's slash-commands Wiki article</a> for more information.
+     * @param event
+     *  The relating {@link SlashCommandEvent}
+     */
+    public abstract void execute(SlashCommandEvent event);
+}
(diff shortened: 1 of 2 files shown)


Nincodedo:
Just curious, why an abstract class and not an interface?

java-coding-prodigy:
I too don't see the benefit of using an abstract class over an interface here.

Tais993:
You didn't look at my code in that case.
This allows us to add optional options.

Nincodedo:
How does an interface prevent us from adding optional options?

Tais993:
There's no way to have a default method that can be overridden? 
At least, as far as I know.

java-coding-prodigy:
Well if you take the example of `isGuildOnly` you could have it as a regular method in the interface. When you implement all the methods, a boolean method will return false, at least in IntelliJ.

Tais993:
I'm confused? Can you show me an example if how it should be implemented.

java-coding-prodigy:
> I'm confused? Can you show me an example if how it should be implemented.

I meant that, say you have this code
`interface AnInterface{
        boolean booleanMethod();
}
class ExtendingClass implements AnInterface{

}
`Obviously, this won't compile. On IntelliJ, it will directly give the option to add unimplemented methods. Once you click on it, this is what the class will look like.
`class ExtendingClass implements AnInterface{
        public boolean booleanMethod(){
                return false;
        }
}`

Tais993:
Howly fuck, my mind just exploded.

On that note, fields might end up being useful.
And since all commands extenders are related (atleast from my point of view), it doesn't make sense to have an interface.

Tais993:
and @java-coding-prodigy, I meant having default methods that can be overriden.
But Nincodedo already showed that

Nincodedo:
If we add fields, then yes an abstract class would make more sense.

java-coding-prodigy:
It is possible to have fields in interfaces, just that they will be the same throughout all implementing classes, either that or we could have a method in the interface called `getVar`, which will be different for different classes. 
As we don't have abstract fields in java, we would have to do the same thing for abstract classes, so unless I am missing something here, in this scenario too, interfaces come out on top.

Tais993:
I'm not sure whenever we will actually add fields, so if we eventually end up not using fields this will be changed to a interface.

And yes, an interface allows constants, but not non-public or non-static variables.
Before this gets merged, we will look back and change to an interface if no features of abstract classes will be used.
That's alright to everyone?

src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.requests.restaction.CommandCreateAction;
+
+/**
+ * The class all commands have to abstract.
+ * Look at {@link org.togetherjava.tjbot.commands.example.ExampleCommand} for an example.
+ * A second example will come for sub-commands.
+ */
+public abstract class AbstractCommand {
```
I-Al-Istannen, line 11: This might be able to be an interface if we never need local state:
```suggestion
public interface Command
```

src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
```diff
@@ -0,0 +1,53 @@
...
+    public abstract String getDescription();
+
+    /**
+     * Whenever the command is only for guilds, optional method.
+     * @return
+     *  Whenever the command is only for guilds as a {@link Boolean}
+     */
+    public boolean isGuildOnly() {
+        return false;
+    }
```
I-Al-Istannen, line 34:
```suggestion
default isGuildOnly() {
  return false;
}
```

Tais993:
Will create a new PR from feature/command-system to develop.
This way everyone can contribute to the command-system.

I-Al-Istannen (commented):
I think we should have a *basic* story for menus / selects / buttons. How do they fit in here, if the command can only be triggered by a `SlashCommandEvent`? Are they a separate system? Hw could a button in message trigger a command? How could we edit the message if a button is clicked?

Tais993:
> Are they a separate system?
I'd say so yeah.

We can also make all execute methods optional.
And then add 2 context menu execute methods.

This way everything can be done with extending 1 class.

We could also separate it, not sure what everyone thinks.

src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
Tais993: > Before this gets merged to develop, we will look back and change to an interface if no features of abstract classes will be used.
[My last message](https://github.com/Together-Java/TJ-Bot/pull/30#issuecomment-915054339)

marko-radosavljevic (approved):
Since this goes into a feature branch, I'm fine with it for now.

Tais993:
Thought I merged it, mistake on my side

PR #32, opened 2021-09-08
Add spotless integration and an Eclipse XML configuration

I-Al-Istannen:
This PR introduces an eclipse xml formatter configuration and configures spotless to use it.

The formatter is based on the google code style, but chops down method chains if they are too long and uses four spaces, but that can be changed :)

You can run spotless with `./gradlew spotlessApply` to automatically format your files and we can use `./gradlew spotlessCheck` in CI to ensure everything is formatted properly.

The XML file is understood by Eclipse (and therefore should also work in VS Code) and you can import it into IntelliJ. If you do the latter, you need to tweak the annotation wrapping, as that is not import correctly for some reason.


Zabuzard (commented):
What is that empty `gradlew` file, why is it there?

java-coding-prodigy:
> What is that empty `gradlew` file, why is it there?

Maybe we can edit that file suitably later?

build.gradle
```diff
@@ -24,4 +25,10 @@ dependencies {
 
 test {
     useJUnitPlatform()
-}
\ No newline at end of file
+}
+
+spotless {
+    java {
+        eclipse().configFile('meta/formatting/google-style-eclipse.xml')
```
marko-radosavljevic, line 32: What do you think about adding `removeUnusedImports()`?

marko-radosavljevic (approved):

I-Al-Istannen:
> What is that empty `gradlew` file, why is it there?

That is not empty, it is a mode change. The previous file was not executable (maybe because it was created on windows and the concept doesn't exist over there?)

The git CLI displays this:
```patch
commit d0a87cab1908b64572c1700f6c07aab77462512e (HEAD -> feature/formatting, origin/feature/formatting)
Author: I-Al-Istannen <i-al-istannen@users.noreply.github.com>
Date:   Wed Sep 8 12:15:47 2021 +0200

    Make gradelew executable

diff --git a/gradlew b/gradlew
old mode 100644
new mode 100755
```

build.gradle
I-Al-Istannen: Sounds fine :)

marko-radosavljevic (approved):

illuminator3 (approved):

PR #37, opened 2021-09-08
Command-system pr

Tais993:
This will be the main PR for the command-system.
See the relating issue #10 for more information.

There's still a few things we have to do, therefor:

- [X] Add handling for buttons and selection menus in the CommandHandler class
- [x] Util for generating component ID's of commands
- [x] Util for converting component ID's of commands to a list - will be pushed when internet is available
- [x] Add command which updates slash-commands
- [x] Command handler 

more in the future..

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
@@ -1,0 +1,101 @@
+package org.togetherjava.tjbot.commands;
+
+import net.dv8tion.jda.api.events.ReadyEvent;
+import net.dv8tion.jda.api.events.interaction.ButtonClickEvent;
+import net.dv8tion.jda.api.events.interaction.GenericComponentInteractionCreateEvent;
+import net.dv8tion.jda.api.events.interaction.SelectionMenuEvent;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
+import net.dv8tion.jda.api.interactions.commands.build.CommandData;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
+
+/**
+ * The command handler
+ *
+ * Commands need to be added to the commandList
+ */
+public class CommandHandler extends ListenerAdapter {
+    private final static Logger logger = LoggerFactory.getLogger(CommandHandler.class);
+
+    private final List<Command> commandList = new ArrayList<>();
+    private final Map<String, Command> commandMap;
+
+    public CommandHandler() {
+        commandList.addAll(List.of(new ReloadCommand(this)
+        // add your command here
+        ));
+
+        commandMap = commandList.stream()
+            .collect(Collectors.toMap(Command::getCommandName, command -> command));
+    }
+
+    @Override
+    public void onReady(@NotNull ReadyEvent event) {
+        event.getJDA().getGuildCache().forEach(guild -> {
+            guild.retrieveCommands().queue(commands -> {
+                boolean hasReloadCommand =
+                        commands.stream().anyMatch(command -> command.getName().equals("reload"));
+
+                if (!hasReloadCommand) {
+                    Command command = commandMap.get("reload");
+                    guild.upsertCommand(command.addOptions(
+                            new CommandData(command.getCommandName(), command.getDescription())))
+                        .queue();
+                }
+            });
+        });
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull SlashCommandEvent event) {
+        Command command = commandMap.get(event.getName());
+
+        if (command != null) {
+            command.onSlashCommand(event);
+        } else {
+            logger.error("Command '{}' doesn't exist?", event.getName());
+        }
+    }
+
+    @Override
+    public void onButtonClick(@NotNull ButtonClickEvent event) {
+        getArgumentsAndCommandByEvent(event, (command, args) -> command.onButtonClick(event, args));
+    }
+
+    @Override
+    public void onSelectionMenu(@NotNull SelectionMenuEvent event) {
+        getArgumentsAndCommandByEvent(event,
+                (command, args) -> command.onSelectionMenu(event, args));
+    }
+
+    /**
+     * Gets the arguments as a {@link List} and gets the {@link Command} by the event.
+     * 
+     * @param event a {@link GenericComponentInteractionCreateEvent}
+     * @param onSucceed a {@link BiConsumer} that takes the command and the arguments
+     */
+    private void getArgumentsAndCommandByEvent(GenericComponentInteractionCreateEvent event,
+            BiConsumer<Command, List<String>> onSucceed) {
+        String[] argsArray = event.getComponentId().split("-");
+        Command command = commandMap.get(argsArray[0]);
+
+        if (command != null) {
+            List<String> args = new ArrayList<>(Arrays.asList(argsArray));
+            args = args.subList(2, args.size());
+            onSucceed.accept(command, args);
+        }
+    }
+
+    public List<Command> getCommandList() {
+        return commandList;
+    }
+}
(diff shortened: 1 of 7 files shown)


Tais993:
While I originally wanted to add functionality so it's easier to work with slash-command permissions.
I myself will wait for the rework of the slash-command permissions.

Current way slash-command permissions work is shit, and Discord has confirmed on reworking them.
See [this discussion](https://github.com/discord/discord-api-docs/discussions/3581), the moderation section 
"Revisit the permissions system to make it more flexible"

So anyone is free to create it, but most likely you will have to revamp it when the reworked permission system comes.

illuminator3 (changes requested):
Please apply _our_ formatting by running `gradle spotlessApply` :)

application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public void onButtonClick(@NotNull ButtonClickEvent event) {
+        String[] argsArray = event.getComponentId().split("-");
+        Command command = commandMap.get(argsArray[0]);
+
+        if (command != null) {
+            List<String> args = new ArrayList<>(Arrays.asList(argsArray));
+            args.remove(0);
+            args.remove(0);
+            command.onButtonClick(event, args);
```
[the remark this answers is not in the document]
Tais993: Didn't think about that, and that's why reviews are always appreciated :p

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
```diff
@@ -0,0 +1,93 @@
...
+                    event.reply("Next time use the right command!").queue();
+                    event.getMessage().editMessageComponents().queue();
+                }
+
+                case SUCCESS -> {
+                    event.deferReply().queue();
+                    List<Command> commands = commandHandler.getCommandList();
+
+                    CommandListUpdateAction globalCommandsUpdateAction = event.getJDA().updateCommands();
+                    CommandListUpdateAction guildCommandsUpdateAction = event.getJDA().getGuilds().get(0).updateCommands();
```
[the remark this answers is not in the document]
Tais993: I should just loop through the guilds and add commands, always works.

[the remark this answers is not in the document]
Tais993: Updated it, code looks less readable but it works in every guild now.

illuminator3 (approved):

Tais993:
I'll accept after 2 reviews, not the required 2 since I find this really important.

Tais993:
> Apart from the added review comments. I think extracting out some of very large lambdas into well named functions will aid in readability as well as testability.

You'll have to quote some exact lambdas, I find it myself readable and maintainable this way.

application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
```diff
@@ -0,0 +1,75 @@
...
+        Command command = commandMap.get(event.getName());
+
+        if (command != null) {
+            command.onSlashCommand(event);
+        } else {
+            logger.error("Command '{}' doesn't exist?", event.getName());
+        }
+    }
+
+    @Override
```
illuminator3, line 48: You're repeating yourself [here](https://github.com/Together-Java/TJ-Bot/pull/37/files#diff-e5f33a13deeba755c63c3da6a06821b5b943b870e8a92c7b05ed588da51ba6dbR48-R70), I'd say extract this to a separate method.

.github/workflows/docker-publish.yaml
```diff
@@ -3,17 +3,19 @@ name: Docker Publish
 on:
   push:
     branches:
-      - 'master'
```
d3vel0per, line 6: Why did we remove master or are we saying we will just use images from develop instead ie. for prod deployments?

I-Al-Istannen: That's not part of this PR but incidental due to the git structure of it. I removed it as we need two different image tags (to ensure we can run both develop and prod in parallel) and therefore need to overhaul that anyways.

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                case SUCCESS -> {
+                    event.deferReply().queue();
+                    List<Command> commands = commandHandler.getCommandList();
+                    List<RestAction<List<net.dv8tion.jda.api.interactions.commands.Command>>> restActions =
+                            new ArrayList<>();
+
+                    // * adds all the global commands
+                    CommandListUpdateAction globalCommandsUpdateAction =
```
d3vel0per, line 77: This is with regards commentary around extracting methods from large lambdas eg.
```java
private final RestAction<List<net.dv8tion.jda.api.interactions.commands.Command>> restActionsForGlobalCommandUpdate(Event, List<Command>)
```

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
```diff
@@ -0,0 +1,116 @@
...
+                            globalCommandsUpdateAction.addCommands(
+                                    command.addOptions(new CommandData(command.getCommandName(),
+                                            command.getDescription())));
+                        }
+                    });
+
+                    restActions.add(globalCommandsUpdateAction);
+
+                    // * adds all the guild commands to each guild
+                    event.getJDA().getGuildCache().forEach(guild -> {
```
d3vel0per, line 91: Similar to the previous comment a similar one for guild specific commands

application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        getArgumentsAndCommandByEvent(event, (command, args) -> command.onButtonClick(event, args));
+    }
+
+    @Override
+    public void onSelectionMenu(@NotNull SelectionMenuEvent event) {
+        getArgumentsAndCommandByEvent(event,
+                (command, args) -> command.onSelectionMenu(event, args));
+    }
+
+    public void getArgumentsAndCommandByEvent(GenericComponentInteractionCreateEvent event,
```
illuminator3, line 79: this is missing javadoc

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
```diff
@@ -0,0 +1,137 @@
...
+                            .editOriginal(
+                                    "Commands successfully reloaded! *Global commands can take upto 1 hour to load")
+                            .queue();
+                    });
+                }
+            }
+        }
+    }
+
+    public CommandListUpdateAction getGlobalCommandUpdateRestAction(Event event,
```
illuminator3, line 89: same as above

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
```diff
@@ -0,0 +1,137 @@
...
+            List<Command> commands) {
+        return addCommandsToUpdateAction(event.getJDA().updateCommands(), commands, false);
+    }
+
+    public List<CommandListUpdateAction> getGuildCommandUpdateRestActions(Event event,
```
illuminator3, line 94: same as above

illuminator3 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
Tais993: While the methods were public, they aren't for use, it's a command not something public.
So I don't see any points in making docs, the code itself is extremely simple,
It feels like creating docs for methods like this is just a waste of everybody's time

application/src/main/java/org/togetherjava/tjbot/commands/ReloadCommand.java
Tais993: See comment above

Tais993:
2 things need to be fixed atm

Naming of the AvstractCommand is incorrect and bad, alternatives?

Docs mention using the interface still

I-Al-Istannen (approved):
Looks good for now, we'll see where we might need to take it :)

marko-radosavljevic (approved):

PR #41, opened 2021-09-08
Add formatter

illuminator3:
The goal of this DPR is to create a formatter. An example of this would be:
```
public static
void main          (String[]args){
    System.out.println("Hello World"
    );
    }
```
should get formatted into
```
\`\`\`java
public static void main(String[] args) {
    System.out.println("Hello World");
}
\`\`\`
```


I-Al-Istannen:
At least most of this can be written in a way that is completely independent from the Bot (i.e. as a `String -> String` function) so they could be fine if they keep that in mind :)

illuminator3:
> Is this a command request for the bot? I.e. actual functionality instead of meta-work?
> Did we agree on a command and general code architecture already? I dont think so - just mentioning cause I dont want you to have to rewrite all your code later on again.
> 
> Related: #37

This is basically just meta work for an API

AzatothTheAwakened:
> This is basically just meta work for an API

Is there any related feature issue? So everyone knows what you are doing instead of just jumping right into the code without people being able to validate the code

src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static Lexer getInstance() {
+        return instance;
+    }
+
+    public List<Token> tokenize(String input) {
+        return tokenize(Arrays.asList(input.split("\n")));
+    }
+
+    public List<Token> tokenize(List<String> lines) {
```
AzatothTheAwakened, line 20: Break this method up into multiple methods, the cyclomatic complexity is too high. You can put that lambda in its own method and use a method reference.

src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,48 @@
+package org.togetherjava.formatter.tokenizer;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.stream.Collectors;
+
+public class Lexer {
```
AzatothTheAwakened, line 9: I dont yet see the reason for it to be a Singleton, please reconsider this.

src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final Lexer instance = new Lexer();
```
AzatothTheAwakened, line 10: I dont see a reason yet to make this a Singleton, please reconsider your design choice here.

src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
AzatothTheAwakened, line 20: Break this up into multiple functions, the cyclomatic complexity is pretty high here. Put that lambda in its own method and reference it. Maybe break even that more up into methods to keep the complexity low.

src/main/java/org/togetherjava/formatter/tokenizer/Token.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.List;
+
+public record Token(String content, TokenType type) {
+    private static final List<TokenType> displayTypes =
+            List.of(TokenType.IDENTIFIER, TokenType.UNKNOWN, TokenType.STRING);
+
+    @Override
+    public String toString() {
+        return type().name() + (displayTypes.contains(type()) ? "(" + content() + ")" : "");
```
AzatothTheAwakened, line 11: I personally am not a big fan of ternary operators, but that might be me nitpicking, they often make it harder to read than necessary.

AzatothTheAwakened (dismissed):
Also missing is Javadoc, please change what I commented and resubmit it for a review.

src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,48 @@
...
+                            break tokenCheck;
+                        }
+                    }
+
+                    throw new TokenizationException("Token not found for '" + content + "'");
+                }
+            }
+
+            return tokens;
+        }).flatMap(List::stream).collect(Collectors.toList());
```
d3vel0per, line 46: JDK 16 already has `toList()` instead of `.collect(Collectors.toList())`

src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,87 @@
+package org.togetherjava.formatter.tokenizer;
+
+import java.util.regex.Pattern;
+
+public enum TokenType {
+    // keywords
```
d3vel0per, line 6: `yield` is missing

illuminator3:
> Also missing is Javadoc, please change what I commented and resubmit it for a review.

this is a draft pr, it's under heavy development and not submitted for a review yet :)

src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,87 @@
...
+    NOT("!"),
+    ASSIGN("="),
+    PLUS("+"),
+    MINUS("-"),
+    DIVIDE("/"),
+    MULTIPLY("*"),
+    DOT("."),
+    // other
+    ANNOTATION(Pattern.compile("^@[a-zA-Z][a-zA-Z0-9_]*")),
+    NUMBER(Pattern.compile("^(0[xb])?\\d+[dDfFlL]?")),
```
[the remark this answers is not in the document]
illuminator3: > Any number of underscores can appear in numeric literals excluding the first and last positions.

> Also the missing decimal point (but may be you are just handling it as 2 separate tokens)

yea I forgot about those two

Zabuzard:
I supposed this is somewhat linked to #66 .

illuminator3:
The PR is up for review now, although the sectionization doesn't work _as good_ as I imagined it (might have to fix it later).

marko-radosavljevic:
Please unit test this thoroughly, there is not much of a review from me until you do that, since I can't compile and check all this in my head. ^^

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * Formatter which can format a given string into a string which contains code blocks etc
+ *
+ * @author illuminator3
+ */
+public class Formatter {
+    /**
+     * List of tokens who should not be put after a space
```
I-Al-Istannen, line 19: which and this could be a Set?

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+     * @param lexer lexer to use
+     * @return resulting code
+     * @author illuminator3
+     */
+    public String format(String input, Lexer lexer) {
+        return format(lexer.tokenize(input));
+    }
+
+    /**
+     * Reverts the tokens into the "original state" before the tokenization occurred
```
I-Al-Istannen, line 67: So it just joins them? Maybe "join" would be a better name? And what is the original state? Would it ever return something like `public               int foo;` or is whitespace still normalized?

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+
+    /**
+     * Writes and formats a given code section (in form of a list of tokens) into a StringBuilder
+     *
+     * @param tokens tokens to write
+     * @return written code sections
+     * @author illuminator3
+     */
+    private StringBuilder writeCodeSection(List<Token> tokens) {
+        purgeWhitespaces(tokens = makeMutable(tokens));
```
I-Al-Istannen, line 85: This is quite hard to read, could you split this into two lines and make a new variable for it? Ideally purgeWhitspace would return a new list, so that would be:
```java
List<Token> normalizedTokens = purgeWhitespaces(tokens);
```

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+        for (Token token : tokens) {
+            TokenType type = token.type();
+
+            if (lastNewLine && type != TokenType.CLOSE_BRACES) {
+                putIndentation(indentation, result);
+            }
+
+            lastNewLine = false;
+
+            iteration: {
```
I-Al-Istannen, line 103: If you need to break an if you should **seriously** reconsider the code!

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+                    putIndentation(--indentation, result);
+
+                    append(result, token);
+
+                    result.append("\n");
+
+                    lastNewLine = true;
+
+                    break iteration;
+                } else if (type == TokenType.IDENTIFIER && (lastToken == TokenType.CLOSE_BRACES
```
I-Al-Istannen, line 136: Those conditions are really hard to read, can you split them up into a function?
```java
else if(whateverThatIfMeans(lastToken, token)
```

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+                    append(result, token);
+
+                    if (forClosed) {
+                        result.append("\n");
+
+                        lastNewLine = true;
+                    }
+
+                    break iteration;
+                } else if ((type == TokenType.OPEN_PARENTHESIS && lastToken != TokenType.IDENTIFIER
```
I-Al-Istannen, line 158: This is *definitely* too much for an if. As above, turn this into one (or more!) *aptly named* function calls describing what it does.

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+                        || lastToken == TokenType.RETURN || lastToken == TokenType.COMMA
+                        || lastToken == TokenType.CATCH
+                        || (lastToken == TokenType.CLOSE_PARENTHESIS
+                                && type != TokenType.CLOSE_PARENTHESIS && type != TokenType.DOT
+                                && type != TokenType.COMMA)
+                        || lastToken == TokenType.PACKAGE) {
+                    result.append(" ");
+                }
+
+                if (!forClosed) {
```
I-Al-Istannen, line 172: I'd rather call a different format method like `formatFor()` in this case, that cuts down on the amount of variables and makes it easier to read.

Keep in mind that you can make the whole class stateful, i.e. have a token queue as a *field* you walk through - then you do not need to have as many variables and can decompose the formatting into (possibly recursive) functions where needed. Or you just have helper functions like `formatFor`, that parse a few tokens and then yield control to the caller again.

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * Puts the needed indentation into a StringBuilder
+     *
+     * @param indentation indentation level
+     * @param sb string builder
+     * @author illuminator3
+     */
+    private void putIndentation(int indentation, StringBuilder sb) {
```
I-Al-Istannen, line 208: I am not sure what this is but there is `String#repeat` and `String#indent` in the stdlib. One of those probably does what you want.

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * Removes every whitespace from a given list of tokens
+     *
+     * @param tokens tokens to remove whitesapces from
+     * @author illuminator3
+     */
+    private void purgeWhitespaces(List<Token> tokens) {
+        tokens.removeIf(t -> t.type() == TokenType.WHITSPACE);
```
I-Al-Istannen, line 241: Why not `return tokens.stream().filter(it -> it.type() != WHITESPACE)).collect(toCollection(ArrayList::new))` (you are missing an `E` in Whitespace btw).

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * Indexes tokens to contain information about whether they are code tokens or not
+     *
+     * @param tokens not-indexed tokens
+     * @return indexed tokens
+     * @author illuminator3
+     */
+    private List<IndexedToken> indexTokens(List<Token> tokens) {
```
I-Al-Istannen, line 251: This might be easier to read as
```java
return tokens.stream()
             .map(token -> new IndexedToken(token, isCodeToken(token)))
             .toList();
```

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+
+        for (Token token : tokens) {
+            result.add(new IndexedToken(token, isCodeToken(token)));
+        }
+
+        return result;
+    }
+
+    /**
+     * Checks if a given token is a code token
```
I-Al-Istannen, line 262: What is a "code token"? Maybe this should be called `hasRecognizedTokenType` or `hasKnownTokenType` or `isTokenTypeKnown`?

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,319 @@
...
+     *
+     * @param tokens tokens to remove whitesapces from
+     * @author illuminator3
+     */
+    private void purgeWhitespaces(List<Token> tokens) {
+        tokens.removeIf(t -> t.type() == TokenType.WHITSPACE);
+    }
+
+    /**
+     * Indexes tokens to contain information about whether they are code tokens or not
```
I-Al-Istannen, line 245: I am not sure why that is called `indexed`. What do you use the information for why don't you check that in the token? E.g. add an `isKnown()` method to `Token` and remove the whole method?

formatter/src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Tokenizer that can turn a list of strings (or a string) into a list of tokens
+ *
+ * @author illuminator3
+ */
+public class Lexer {
+    private static final Pattern commentPatcherRegex =
```
I-Al-Istannen, line 15: What does this do? A complicated Regex without any example/explanation is quite hard to understand.

formatter/src/main/java/org/togetherjava/formatter/tokenizer/Lexer.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return resulting tokens
+     * @author illuminator3
+     */
+    private List<Token> tokenizeLine(String line) {
+        List<Token> tokens = new ArrayList<>();
+        int index = 0;
+        String content;
+
+        while (!(content = line.substring(index)).isEmpty()) {
+            tokenCheck: {
```
I-Al-Istannen, line 53: Why do you have a random block here with a name? Why not just use `break`? It currently is harder to read than necessary IMHO.

You would need a nested break here, which indicates that the inner loop should probably be a method somewhere and not inlined here. Maybe sth like:
```java
while (!(content = line.substring(index)).isEmpty()) {
    // can throw or return an optional and you throw in here
    Token matchingToken = findTokenInString(content);
    index += matchingToken.length();
    tokens.add(matchingToken);
}
```

formatter/src/main/java/org/togetherjava/formatter/tokenizer/Token.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+
+/**
+ * Class representing a Token with a given content and a type
+ *
+ * @author illuminator3
+ */
+public record Token(String content, TokenType type) {
+    private static final List<TokenType> displayTypes =
```
I-Al-Istannen, line 11: Why not a Set?

formatter/src/main/java/org/togetherjava/formatter/tokenizer/Token.java
```diff
@@ -0,0 +1,32 @@
...
+        return type().name() + displayMe();
+    }
+
+    /**
+     * Returns a non-empty string if this token has something to display
+     *
+     * @return the displayed value
+     * @author illuminator3
+     */
+    private String displayMe() {
```
I-Al-Istannen, line 25: `formatContentForDisplay`? `formatContent`? Not sure what a good name is, but I am not a huge fan of `displayMe`. If you want to keep that terminology `display` would be better, but `display` in general sounds quite active - it might put sth on the screen. This doesn't happen here. If you are consistent with the use of the term that's fine though.

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,127 @@
...
+    NOT("!"),
+    ASSIGN("="),
+    PLUSPLUS("++"),
+    PLUS("+"),
+    MINUSMINUS("--"),
+    MINUS("-"),
+    BOOL_AND("&&"),
+    BOOL_OR("||"),
+
+    // this has to be other DIVIDE or else it'll be seen as two divides
```
I-Al-Istannen, line 84: I am not sure what that comment says, but that might be on me.

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,127 @@
...
+
+    DIVIDE("/"),
+    MULTIPLY("*"),
+    DOT("."),
+    // other
+    ANNOTATION(Pattern.compile("^(@[a-zA-Z][a-zA-Z0-9_]*)")),
+    NUMBER(Pattern.compile("^((0[xb])?([0-9_]+|[0-9_]*(\\.\\d*))[dDfFlL]?)")),
+    STRING(Pattern.compile("^(\"[^\"]*\")")),
+    IDENTIFIER(Pattern.compile("^([a-zA-Z][a-zA-Z0-9_]*)")),
+    WHITSPACE(Pattern.compile("^(\s+)")),
```
I-Al-Istannen, line 95: WHIT**E**SPACE

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,127 @@
...
+    UNKNOWN(Pattern.compile("^(.)"));
+
+    private final Pattern regex;
+    private final boolean isKeyword;
+
+    TokenType(String spattern, boolean isKeyword) {
+        if (!isKeyword) {
+            this.regex = Pattern.compile("^(\\Q" + spattern + "\\E)");
```
I-Al-Istannen, line 103: Prefer properly named methods: `this.regex = Pattern.compile(Pattern.quote(spattern))` is easier to understand.

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,127 @@
...
+    NUMBER(Pattern.compile("^((0[xb])?([0-9_]+|[0-9_]*(\\.\\d*))[dDfFlL]?)")),
+    STRING(Pattern.compile("^(\"[^\"]*\")")),
+    IDENTIFIER(Pattern.compile("^([a-zA-Z][a-zA-Z0-9_]*)")),
+    WHITSPACE(Pattern.compile("^(\s+)")),
+    UNKNOWN(Pattern.compile("^(.)"));
+
+    private final Pattern regex;
+    private final boolean isKeyword;
+
+    TokenType(String spattern, boolean isKeyword) {
```
I-Al-Istannen, line 101: Why is this called `spattern`? Is there a clash with `pattern`?

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,127 @@
...
+        if (!isKeyword) {
+            this.regex = Pattern.compile("^(\\Q" + spattern + "\\E)");
+        } else {
+            this.regex = Pattern.compile("^(\\Q" + spattern + "\\E)[^a-zA-Z]");
```
I-Al-Istannen, line 105: Why is the end different here than above? Why does a keyword need this?

I-Al-Istannen (changes requested):
Looks nice, but I would factor out some things :lady_beetle:

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
illuminator3: typo, I meant "over"

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
illuminator3: smh another typo

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
illuminator3: because I want to mutate the existing list, not create a new one

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
illuminator3: the lexer should be independant from the formatter

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
illuminator3: so that e.g. `for` doesn't match `format`

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
java-coding-prodigy: You forgot increment and decrement. Also what about punctuators and separators?

formatter/src/test/java/org/togetherjava/formatter/tokenizer/LexerTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @DisplayName("Given input returns expected result")
+    void testInput() {
+        String testString =
+            "class enum record interface import if for while null extends implements new return package this yield catch try else if else public private protected static sealed non-sealed final abstract void int long short byte boolean float double char ( ) { } [ ] ; :: : , += -= *= /= %= &= |= == != >= > <= < ! = ++ + -- - && || /* a */ / * . @annotation 0 \"string\" identifier ` ";
+        List<TokenType> tokens = lexer.tokenize(testString).stream().map(Token::type).toList(),
+                        expected = new ArrayList<>();
+
+        for (TokenType type : TokenType.values()) {
+            if (type == TokenType.UNKNOWN) {
+                System.out.println("hi");
```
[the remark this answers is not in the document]
illuminator3: oops, that was a debug thing I accidentally left there :p

illuminator3:
Don't worry guys -- I'll add javadoc to everything tomorrow

Zabuzard:
Why are there changes in the `meta/formatting/google-style-eclipse.xml` file? Is this not up2date with `develop`?

illuminator3:
> Why are there changes in the `meta/formatting/google-style-eclipse.xml` file? Is this not up2date with `develop`?

It's up-to-date. I forgot you can cherry-pick, so I just copy-pasted the config from `develop` :P

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -71,6 +72,8 @@
     PLUS("+"),
     MINUSMINUS("--"),
     MINUS("-"),
+    BOOL_AND("&&"),
+    BOOL_OR("||"),
 
     // this has to be other DIVIDE or else it'll be seen as two divides
     COMMENT(Pattern.compile("^(//.*|/\\*[^*]*\\*+(?:[^/*][^*]*\\*+)*/)")),
```
java-coding-prodigy, line 89: A comment can begin in the same line after a statement 2, provided the statement has terminated, but the `^` forbids that

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Sectionizes a given list of tokens
+     *
+     * @param checkedTokens checked tokens
+     * @return list of sections
+     * @author illuminator3
+     */
+    private List<Section> sectionize(List<CheckedToken> checkedTokens) {
+        CheckedToken first = checkedTokens.get(0);
+        Section currSec = new Section(new ArrayList<>(), first.isCode());
```
marko-radosavljevic, line 113:
```suggestion
        Section currentSection = new Section(new ArrayList<>(), first.isCode());
```

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param token token to check
+     * @return true if it's a code token, false if not
+     * @author illuminator3
+     */
+    private boolean isTokenPartOfCode(Token token) {
+        return token.type() != TokenType.UNKNOWN;
+    }
+
+    /**
+     * Sectionizes a given list of tokens
```
marko-radosavljevic, line 105: Improve a bit on this javadoc, what does the 'sectionize list of tokens' means exactly. I had to take a look at the implementation to see what it was doing.

It's a private method, so you can leave it.

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
```diff
@@ -0,0 +1,396 @@
...
+        if (type == TokenType.COMMA || type == TokenType.WILDCARD || type == TokenType.EXTENDS
+                || type == TokenType.SUPER) {
+            result.append(' ');
+        }
+
+        // ------------------------------------------------
+
+        switch (type) {
+            case SMALLER -> genericDepth++;
+            case BIGGER -> genericDepth--;
```
marko-radosavljevic, line 191: Same as below, needs a default case. Because this will probably happen more, won't repeat it every time, you just check every switch.

Also took me some time to figure out what `SMALLER` means, it should be [`LESS_THAN`](https://en.wikipedia.org/wiki/Less-than_sign)

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
```diff
@@ -0,0 +1,138 @@
...
+    DIVIDE_EQUALS("/=", false, true),
+    MODULO_EQUALS("%=", false, true),
+    AND_EQUALS("&=", false, true),
+    OR_EQUALS("|=", false, true),
+    EQUALS("==", false, true),
+    NOT_EQUALS("!=", false, true),
+    BIGGER_OR_EQUALS(">=", false, true),
+    BIGGER(">", false, true),
+    SMALLER_OR_EQUALS("<=", false, true),
+    SMALLER("<", false, true),
```
marko-radosavljevic, line 78:
```suggestion
    GREATER_THAN_OR_EQUALS(">=", false, true),
    GREATER_THAN(">", false, true),
    LESS_THAN_OR_EQUALS("<=", false, true),
    LESS_THAN("<", false, true),
```

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
```diff
@@ -0,0 +1,396 @@
...
+            result.append(token.content());
+
+            return;
+        }
+
+        if (genericDepth != 0) {
+            handleGeneric(token);
+
+            return;
+        }
```
marko-radosavljevic, line 80: I dislike your usage of whitespace here, I think my eyes prefer this:
```suggestion
        if (checkGeneric(type)) {
            result.append(token.content());
            return;
        }

        if (genericDepth != 0) {
            handleGeneric(token);
            return;
        }
```

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param tokens tokens to write
+     * @return written code sections
+     * @author illuminator3
+     */
+    private StringBuilder writeCodeSection(List<Token> tokens) {
+        CodeSectionFormatter formatter = new CodeSectionFormatter(tokens);
+
+        formatter.format();
+
+        return formatter.result();
```
marko-radosavljevic, line 77: Why new lines after each line in here, it's all the same context, it should be in the same section.
```suggestion
    private StringBuilder writeCodeSection(List<Token> tokens) {
        CodeSectionFormatter formatter = new CodeSectionFormatter(tokens);
        formatter.format();
        return formatter.result();
```

marko-radosavljevic (commented):
Good job, I like it! :relaxed:

What I do dislike is your usage of whitespace, so at some point I stopped commenting on that :p

Also, it's annoying that all our pipeline integrations are not working on your branch, and I know they would be failing, at least on a couple of things.. :/

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
```diff
@@ -0,0 +1,396 @@
...
+        return isKeyword(token) || isOperator(token) || isParenthesisRule(token)
+                || type == TokenType.CLOSE_BRACKETS // put a space after e.g. 'try' or 'else' and
+                                                    // ] or if it's an operator
+                || type == TokenType.COMMA // for e.g. multiarg method calls or method parameters
+                || (type == TokenType.IDENTIFIER && !queue.isEmpty()
+                        && queue.peek().type() == TokenType.IDENTIFIER); // for double identifier
+                                                                         // in e.g. method
+                                                                         // declaration or enhanced
+                                                                         // for loops
+    }
```
marko-radosavljevic, line 135: Also, I dislike inline comments, especially when they do this. :shrug:

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Updates the enhanced for loop level
+     *
+     * @param type current token type
+     * @author illuminator3
+     */
+    private void updateEnhancedLevel(TokenType type) {
```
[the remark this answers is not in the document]
illuminator3: I removed that entirely since it wasn't doing anything anyway

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
```diff
@@ -0,0 +1,396 @@
...
+    /**
+     * Updates the indentation based on the current token type
+     *
+     * @param type current token type
+     * @author illuminator3
+     */
+    private void updateIndentation(TokenType type) {
+        switch (type) {
+            case OPEN_BRACES -> indentation++;
+            case CLOSE_BRACES -> indentation--;
```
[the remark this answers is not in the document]
illuminator3: I did it like that, since I don't want to do
```java
if (token.type() == TokenType.OPEN_BRACES || token.type() == TokenType.CLOSE_BRACES) {
    updateIndentation(token.type());
}
```
everywhere, so I just ignore anything other than those who are actually used.

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
illuminator3: Improved it

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
illuminator3: (same as above)

formatter/src/main/java/org/togetherjava/formatter/tokenizer/TokenType.java
illuminator3: Changed the names

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
illuminator3: subjective.

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
illuminator3: Where else am I supposed to put those comments? They explain what each part of the conditional does.

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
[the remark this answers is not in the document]
marko-radosavljevic: The switch is fine, it's missing a default case tho. ^^

Something like `default -> throw new IllegalStateException("Invalid token type " + type) `
Or at least a logger if you don't want to throw.

formatter/src/main/java/org/togetherjava/formatter/Formatter.java
illuminator3: subjective.

formatter/src/main/java/org/togetherjava/formatter/CodeSectionFormatter.java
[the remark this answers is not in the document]
illuminator3: The method is also called if the type is anything other than those two, so throwing an exception wouldn't make any sense.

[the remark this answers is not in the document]
marko-radosavljevic: sure, then log it ^^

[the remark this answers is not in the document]
marko-radosavljevic: This is a critical issue in sonar, which means it wouldn't let you merge.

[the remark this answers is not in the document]
Tais993: > This is a critical issue in sonar

image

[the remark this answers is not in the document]
Tais993: I don't see the critical issue in sonar, only issue I can see is incorrectly formatted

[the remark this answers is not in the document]
illuminator3: Logging it would also not make sense, since again, it's not an error.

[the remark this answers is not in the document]
marko-radosavljevic: sure, then add the comment explaining that ^^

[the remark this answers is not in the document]
marko-radosavljevic: image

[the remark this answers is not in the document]
illuminator3: I'll change it into an if-elseif to satisfy sonar.

marko-radosavljevic (approved):

PR #42, opened 2021-09-08
Removing redundant .idea exclusions + a few other unused ones ✂

d3vel0per:
- Removed a few redundant .idea sub directory exclusions as the parent one is already excluded
- Removed exclusions for C, Scala, PHP. Which probably will never be a part of the codebase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,23 +23,8 @@ local.properties
 # PyDev specific (Python IDE for Eclipse)
 *.pydevproject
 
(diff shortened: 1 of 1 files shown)


I-Al-Istannen (approved):

Zabuzard:
Note that the gitignore file was autogenerated by www.gitignore.io. Not sure if a cleanup is actually meaningful. That said, I also dont really have to say anything against it.

Zabuzard (approved):

PR #46, opened 2021-09-08
Added SQLite and simple example commands

Zabuzard:
Since the bot will be using a database soon anyways, I decided to add SQLite as dependency and also a simple command to showcase how it can be used (in order to get started).

The added commands are:

- `!dbput <key> <value>`
- `!dbget <value>`

and it acts basically as a simple persisted `Map<String, String>`.

Command usage example


Tais993 (approved):

marko-radosavljevic (dismissed):

marko-radosavljevic:
Maybe add some tests to this before merging, if you think it will stay.

d3vel0per:
I was holding off for the command structure story. I think Zabu's intention here was to just set an example and not for the long term. The last discussion I had on Discord was both Istannen and Tijs are against text commands. Probably slash commands are the way to go.

marko-radosavljevic:
Yeah, I'm just reminding for every case. It's easy to forget about testing, and then it's too late. ^^

It's also something we didn't discuss, how much testing we want, etc. Makes sense that only tested things are merged into develop.

PR #49, opened 2021-09-08
Basic command interface and examples

Tais993:
Thought again about interface vs abstract class.
And ended up with an interface.

I'm not sure how it exactly ends like, so maybe it will become an abstract class again.
We'll find out

Added ICommand, the basic command interface all commands have to implement.
Added SubCommandExample, it's an example for a command with subcommands.
Added CommandExample, just a normal command example.

Opinions on the docs, or the code itself are welcome!
Hopefully this is alright so we will have this as basis for the branch.

### DIFF
--- a/src/main/java/org/togetherjava/tjbot/commands/ICommand.java
+++ b/src/main/java/org/togetherjava/tjbot/commands/ICommand.java
@@ -13,16 +13,26 @@ import org.togetherjava.tjbot.commands.example.CommandExample;
 public interface ICommand {
 
     /**
-     * The commands name.
+     * <p> The command's name.
      *
+     * <p> <b>Requirements!</b>
+     *  <ul>
+     *      <li>Lowercase</li>
+     *      <li>Alphanumeric (with dash)</li>
+     *      <li>1 to 32 characters long</li>
+     *  </ul>
      * @return
      *  The commands name as a {@link String}
      */
     String getCommandName();
 
     /**
-     * The command's description.
+     * <p> The command's description.
      *
+     * <p> <b>Requirements!</b>
+     *  <ul>
+     *      <li>1 to 100 characters long</li>
+     *  </ul>
      * @return
      *  The commands description as a {@link String}
      */
--- a/src/main/java/org/togetherjava/tjbot/commands/ICommand.java
+++ b/src/main/java/org/togetherjava/tjbot/commands/ICommand.java
@@ -1,0 +1,62 @@
+package org.togetherjava.tjbot.commands;
+
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.commands.build.CommandData;
+import org.togetherjava.tjbot.commands.example.CommandExample;
+
+/**
+ * The class all commands have to abstract.
+ * <br> Look at {@link CommandExample} for an example.
+ * <br> A second example will come for sub-commands.
+ *
+ */
+public interface ICommand {
+
+    /**
+     * The commands name.
+     *
+     * @return
+     *  The commands name as a {@link String}
+     */
+    String getCommandName();
+
+    /**
+     * The command's description.
+     *
+     * @return
+     *  The commands description as a {@link String}
+     */
+    String getDescription();
+
+    /**
+     * Whenever the command is only for guilds, optional method.
+     *
+     * @return
+     *  Whenever the command is only for guilds as a {@link Boolean}
+     */
+    default boolean isGuildOnly() {
+        return false;
+    }
+
+    /**
+     * This command will add all the options to the slash-command.
+     * <br> Look at {@link CommandExample#addOptions(CommandData)} for an example.
+     *
+     * @param commandData
+     *  The {@link CommandData} where the options need to be added.
+     * @return
+     *  The changed {@link CommandData}
+     */
+    default CommandData addOptions(CommandData commandData) {
+        return commandData;
+    }
+
+    /**
+     * The execute method for when the command gets executed.
+     * <br> Check <a href="https://github.com/DV8FromTheWorld/JDA/wiki/Interactions#slash-commands">JDA's slash-commands Wiki article</a> for more information.
+     *
+     * @param event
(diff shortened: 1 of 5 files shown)


src/main/java/org/togetherjava/tjbot/commands/ICommand.java
```diff
@@ -0,0 +1,62 @@
...
+import net.dv8tion.jda.api.interactions.commands.build.CommandData;
+import org.togetherjava.tjbot.commands.example.CommandExample;
+
+/**
+ * The class all commands have to abstract.
+ * <br> Look at {@link CommandExample} for an example.
+ * <br> A second example will come for sub-commands.
+ *
+ */
+public interface ICommand {
```
Nincodedo, line 13:
```suggestion
public interface Command {
```
Starting interfaces with I is an old standard that I don't think we need to follow. Feel free to disregard if the group feels otherwise.

src/main/java/org/togetherjava/tjbot/commands/example/CommandExample.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param commandData
+     *  The thing we will add our options to!
+     * @return
+     *  A {@link CommandCreateAction} with the required options for this command
+     */
+    @Override
+    public CommandData addOptions(CommandData commandData) {
+        return commandData.addOptions(
+                // * add the first option
+                new OptionData(OptionType.USER, "User", "User that will be pinged!", true),
```
Nincodedo, line 64: I think option names have to be all lower case. I don't quite remember.

Tais993: Could be, command names and the description also have some limits so this will have to be added to the docs.
But didn't do that currently since it's more of a draft.

Tais993: Fixed it and added requirements section to 2 ICommand methods

src/main/java/org/togetherjava/tjbot/commands/ICommand.java
Tais993: We don't need to follow it yeah, following it has no downsides.
While it does have the upside of telling the user it's a interface.
Which is nice, since commands have to implement it, not extend it.

java-coding-prodigy: We should anyway not make changes for such minor things, but I thing we should stick with a naming convention throughout the project.

RicardoMonteiroSimoes: I'd also suggest sticking to common _Java_ conventions. IIRC using I and comparable as prefix is more a thing for C-based languages.

RicardoMonteiroSimoes:
What speaks against using already existing libraries that solve this problem? Out of my head I know of one, the thing is it uses annotations for it.

The next question would be if all commands are created at compile time or if you want to be able to load them at runtime? As in, you have the bot framework that handles everything, and i.e. the "Ban"-Command could be updated outside of the bot and _dynamically_ loaded.

src/main/java/org/togetherjava/tjbot/commands/ICommand.java
Tais993: There's no Java convention telling you to not use or or use it ¯\_(ツ)_/¯
> Interface names should be capitalized like class names.
So by adding an I to the name, we are following Java conventions.

Tais993:
> What speaks against using already existing libraries that solve this problem? Out of my head I know of one, the thing is it uses annotations for it
Yes, Chewtils a.e

But those are fun for people that don't know how to code, so everything is done for them.
In short, you lack control.

By creating something like this yourself, it's first of all educational for everyone, and it's fun.

> could be updated outside of the bot and dynamically loaded.
I'm not gonna bother creating anything like that, just restarting also works and is way simpler.

src/main/java/org/togetherjava/tjbot/commands/ICommand.java
RicardoMonteiroSimoes: Well yes, everything falls under that classification. Just saying that I've yet to see anyone code Java like that in the industry/higher education

Tais993: I don't really care so ig I'll change it

RicardoMonteiroSimoes:
Suggestion:

Add a `getCommandExplanation` or comparable function that allows the bot to print out what the command even does and how it works.

Tais993:
I want this PR merged asap so there's a basis for people to work on.
The Command interface isn't fully finished, but this introduces a basis so the comandhandler can be created

RicardoMonteiroSimoes:
Has it been defined how you handle commands that are locked behind roles? Maybe thats a feature that you could include, or make it a separate interface

Tais993:
I thought I mentioned this somewhere..

And yes I thought about it, I'm planning for making it easier to add command permissions in the future.

Currently slash-command permissions are pretty shit, and Discord has [confirmed ](https://github.com/discord/discord-api-docs/discussions/3581) on reworking them. (See Moderation -> "Revisit the permissions system to make it more flexible")

So I'll wait until the rework, when the rework has been completed this will be added.
But not going to spend time on a bad system that will need to be fully reworked once the new system is out

RicardoMonteiroSimoes:
Yes, I've read that comment, but wasn't aware it would be a limitation on role based filters? The bot would (probably) filter those, so such a system would be independent of discords slash command API? Unless, of course, you can define roles in the API for them

Tais993:
The current slash-command permissions system works fully on roles.
Which works yes, but it'd be nicer to have the clear command for everyone with the manage message permission.
Instead of having to hardcode the roles and more.

So right now you can just do Member#hasPermission instead

Well, I don't see a reason to use roles besides because of Discord limitations.

Zabuzard:
May I ask, @Tais993 why did you not migrate the existing command to the "new" system?

Tais993:
What existing commands?

PR #55, opened 2021-09-09
Sonarcloud integration

marko-radosavljevic:
## Usage

In a typical development process:
1. Developers develop and merge code in an IDE (preferably using SonarLint to receive immediate feedback in the editor) and check-in their code to their ALM.
2. An organization’s continuous integration (CI) tool checks out, builds, and runs unit tests, and an integrated SonarCloud scanner analyzes the results.
3. The scanner posts the results to the SonarCloud server which provides feedback to developers through the SonarCloud interface, email, in-IDE notifications (through SonarLint), and decoration on pull or merge requests (when using Developer Edition and above).

### External Analyzers
Extenrnal analyzers can be integrated into our SonarCloud, supported external analyzers are:

1. PMD
2. SpotBugs
3. CheckStyle

### Info

1. >With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
Because of this, we are using [pull_request_target](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target), so our sonar_token can work when we create PR form a fork.
2. Sonar workflow should be run last, so it can collect all the useful data. So for example, after testing, code coverage and other similar tasks. We can configure it in build.gradle like this: `project.tasks["sonarqube"].dependsOn "anotherTask"`


d3vel0per (approved):
Looks good to me 👍

PR #67, opened 2021-09-09
Added handling for buttons and 

Tais993:
I'll just squash and create a summary when PR'ing to development.
Gotta improve at handling my commits and git


Zabuzard:
Btw you do not have to work on a fork. You also do not need to create a PR for feature-branches. You could just push directly and also work on the real repository directly without fork, if you want.

Issue #73, opened 2021-09-10
[FEATURE] Build and publish OCI images from Github Actions


d3vel0per:
**Problem**
The CI/CD pipeline to build and produce OCI artifacts as the same will be used for deployments

**Solution**
Add github actions for building and pushing OCI images to container registry

**Implementation**

- on a pull_request it will just trigger a docker build ie. build OCI image and push it to local docker daemon
- on push to master and develop, it will result in a docker build + push to docker registry

PR #77, opened 2021-09-10
Patch fix for broken docker builds 💉

d3vel0per:
Related to #73 

This just fixes null property assignment but need to figure out a way to pass env vars from github to gradle. Locally a -D seems to work fine, not sure why the same is not picked up in the CI/CD pipeline


d3vel0per:
Observation from the last time. The new workflow files only take effect from subsequent builds.

I-Al-Istannen (approved):

PR #79, opened 2021-09-10
Fix spotless submodules

### LINKED ISSUE #76: [BUG] Spotless not working in submodules
**Describe the bug**
Spotless doesn't "automatically" also work in submodules. Doing `gradle spotlessApply` only does it for the top module, doing e.g. `gradle application:spotlessApply` gives you an error that a) the plugin isn't added for that module and b) the meta/formatting/<formatting config>.xml doesn't exist.

marko-radosavljevic:
Fixes #76

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -3,9 +3,6 @@ plugins {
     id 'com.google.cloud.tools.jib' version '3.1.4'
 }
 
-group 'org.togetherjava'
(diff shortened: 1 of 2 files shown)


illuminator3 (approved):

illuminator3:
apparently GH broke I think? It tells me I still have to merge even though the commit is already there

PR #86, opened 2021-09-11
Various improvements to the DB code

Zabuzard:
Mostly shown by linter, some are:
* gitignore generated DB files (schema, ...)
* try-with-resources for SELECT query
* more generalized generics (? super ...)
* locks have to be opened before the try-block
* converted new Object() {...} into named helper class

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/DatabaseListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/DatabaseListener.java
@@ -81,9 +81,9 @@ public final class DatabaseListener extends ListenerAdapter {
 
         try {
             Optional<StorageRecord> foundValue = database.read(context -> {
-                try (var query = context.selectFrom(Storage.STORAGE)) {
-                    return query.where(Storage.STORAGE.KEY.eq(key)).stream().findFirst();
-                }
+                return Optional.ofNullable(context.selectFrom(Storage.STORAGE)
+                                                  .where(Storage.STORAGE.KEY.eq(key))
+                                                  .fetchOne());
             });
             if (foundValue.isEmpty()) {
                 event.getChannel().sendMessage("Nothing found for the key '" + key + "'").queue();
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/DatabaseListener.java
```diff
@@ -79,11 +80,10 @@ private void handleGetMessage(String message, MessageReceivedEvent event) {
...
         try {
-            Optional<StorageRecord> foundValue = database.read(ctx -> {
-                return ctx.selectFrom(Storage.STORAGE)
-                          .where(Storage.STORAGE.KEY.eq(key))
-                          .stream()
-                          .findFirst();
+            Optional<StorageRecord> foundValue = database.read(context -> {
+                try (var query = context.selectFrom(Storage.STORAGE)) {
+                    return query.where(Storage.STORAGE.KEY.eq(key)).stream().findFirst();
+                }
```
I-Al-Istannen, line 86:
```suggestion
                return Optional.ofNullable(context.selectFrom(Storage.STORAGE)
                                                  .where(Storage.STORAGE.KEY.eq(key))
                                                  .fetchOne());
```
Maybe? That way we can avoid the annoying try there.

application/src/main/java/org/togetherjava/tjbot/db/Database.java
```diff
@@ -109,13 +110,12 @@ public void write(CheckedConsumer<DSLContext, DataAccessException> action) {
      * @return whatever the handler returned#
```
I-Al-Istannen, line 110:
```suggestion
     * @return whatever the handler returned
```

application/src/main/java/org/togetherjava/tjbot/DatabaseListener.java
Zabuzard: The linter says that `context.selectFrom(Storage.STORAGE)` opens a resource that has to be closed. Does `fetchOne` avoid that?

I-Al-Istannen: Yes https://blog.jooq.org/tag/autocloseable/

marko-radosavljevic (approved):

I-Al-Istannen (approved):

PR #116, opened 2021-09-14
Modularize database

illuminator3:
This will put everything in the db package into its own gradle module, possibly removing circular dependencies in the future.


marko-radosavljevic (dismissed):

Zabuzard:
@marko-radosavljevic why do the checks fail and how to fix it?

marko-radosavljevic:
@Zabuzard

I e believe because it's ran on `pull_request_target`, it runs in the context of the target repository of the PR, rather than in the merge commit.

Zabuzard:
> @Zabuzard
> 
> I believe because it's ran on `pull_request_target`, it runs in the context of the target repository of the PR, rather than in the merge commit. Which means, it checks out code from the dev branch, instead of the code from this PR.
> 
> And because code structure is changed, db code is extracted to a new module, generated code is not where the workflow expects it.

Sounds like a bug to me. Do we have an issue and/or someone planning a fix already? @marko-radosavljevic

PR #117, opened 2021-09-14
Add tophelpers command 🏆

### LINKED ISSUE #65: Migrate TopHelper system
We have to migrate the TopHelper system. The details are up to be decided by the community.

`/tophelpers <limit>` by **TogetherJava** to display the ranking used to determine the **Top Helpers** each month.

It counts messages send by people in help channels and ranks them filtered by the message count from the last 30 days, descending.

`<limit>` is optional and controls the max amount of users to show in the ranking.

d3vel0per:
Fixes #65


application/src/test/java/org/togetherjava/tjbot/commands/TopHelpersTest.java
```diff
@@ -0,0 +1,71 @@
...
+        Mockito.when(slashCommandEvent.getGuild().getMemberById(1L).getEffectiveName())
+            .thenReturn("ALPHA");
+        Mockito.when(slashCommandEvent.getGuild().getMemberById(2L).getEffectiveName())
+            .thenReturn("BETA");
+        Mockito.when(slashCommandEvent.getGuild().getMemberById(3L).getEffectiveName())
+            .thenReturn("GAMMA");
+    }
+
+    @Test
+    void get_top_helpers() {
```
illuminator3, line 57: don't use snake_case here, Java methods should always be camelCase

application/src/test/java/org/togetherjava/tjbot/db/MessageMetadataListenerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Mockito.when(messageReceivedEvent.getMessage().getIdLong()).thenReturn(MESSAGE_ID);
+        Mockito.when(messageReceivedEvent.getGuild().getIdLong()).thenReturn(GUILD_ID);
+        Mockito.when(messageReceivedEvent.getChannel().getIdLong()).thenReturn(CHANNEL_ID);
+        Mockito.when(messageReceivedEvent.getAuthor().getIdLong()).thenReturn(USER_ID);
+        Mockito.when(messageReceivedEvent.getMessage().getTimeCreated())
+            .thenReturn(OffsetDateTime.now());
+    }
+
+    @Test
+    void received_message_author_is_bot() {
```
illuminator3, line 42: same as above

application/src/test/java/org/togetherjava/tjbot/db/MessageMetadataListenerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Mockito.when(messageReceivedEvent.getAuthor().isBot()).thenReturn(true);
+        messageMetadataListener.onMessageReceived(messageReceivedEvent);
+        var count = database.read(ctx -> {
+            return ctx.fetchCount(MessageMetadata.MESSAGE_METADATA);
+        });
+        Assertions.assertEquals(0, count);
+    }
+
+    @Test
+    void received_message_is_a_webhook() {
```
illuminator3, line 52: same as above

application/src/test/java/org/togetherjava/tjbot/db/MessageMetadataListenerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Mockito.when(messageReceivedEvent.isWebhookMessage()).thenReturn(true);
+        messageMetadataListener.onMessageReceived(messageReceivedEvent);
+        var count = database.read(ctx -> {
+            return ctx.fetchCount(MessageMetadata.MESSAGE_METADATA);
+        });
+        Assertions.assertEquals(0, count);
+    }
+
+    @Test
+    void received_message_is_private() {
```
illuminator3, line 62: same as above

application/src/test/java/org/togetherjava/tjbot/db/MessageMetadataListenerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Mockito.when(messageReceivedEvent.isFromType(ChannelType.PRIVATE)).thenReturn(true);
+        messageMetadataListener.onMessageReceived(messageReceivedEvent);
+        var count = database.read(ctx -> {
+            return ctx.fetchCount(MessageMetadata.MESSAGE_METADATA);
+        });
+        Assertions.assertEquals(0, count);
+    }
+
+    @Test
+    void received_message_is_not_of_type_text() {
```
illuminator3, line 72: same as above

application/src/test/java/org/togetherjava/tjbot/db/MessageMetadataListenerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Mockito.when(messageReceivedEvent.isFromType(ChannelType.TEXT)).thenReturn(false);
+        messageMetadataListener.onMessageReceived(messageReceivedEvent);
+        var count = database.read(ctx -> {
+            return ctx.fetchCount(MessageMetadata.MESSAGE_METADATA);
+        });
+        Assertions.assertEquals(0, count);
+    }
+
+    @Test
+    void received_message_is_persisted() {
```
illuminator3, line 82: same as above

build.gradle
```diff
@@ -34,3 +34,7 @@ subprojects {
         }
     }
 }
+
+dependencies {
```
illuminator3, line 38: don't add this here, it belongs to the application build.gradle

illuminator3 (changes requested):
also maybe put the things into a different gradle module?

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+import org.togetherjava.tjbot.util.PresentationUtils;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.jooq.impl.DSL.*;
+import static org.togetherjava.tjbot.db.generated.tables.MessageMetadata.MESSAGE_METADATA;
+import static org.togetherjava.tjbot.util.MessageTemplate.PLAINTEXT_MESSAGE_TEMPLATE;
+
+public final class TopHelpers implements Command {
```
Zabuzard, line 24: This class should use `AbstractCommand` as base class to get rid of the code duplication with the method implementation pattern (as flagged by SonarCloud).

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public @NotNull CommandData addOptions(@NotNull CommandData commandData) {
+        return commandData.addOptions(new OptionData(OptionType.STRING, COUNT_OPTION,
+                "Count of top helpers to be retrieved (default is 10 and capped at 30)", false));
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandEvent event) {
+        var guildId = event.getGuild().getIdLong();
+        var count = getBoundedCount(event.getOption(COUNT_OPTION));
```
Tais993, line 68: I don't recommend using var here

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+        database.read(dsl -> {
+            var records = dsl.with("TOPHELPERS")
```
Tais993, line 70: See above, no idea what var is

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+                .select(rowNumber().over(orderBy(field(name("COUNT")).desc())).as("#"),
+                        field(name("USER_ID"), Long.class), field(name("COUNT"), Long.class))
+                .from(table(name("TOPHELPERS")))
+                .fetch();
+            event.reply(prettyPrintOutput(buildTopHelpersDataFrame(event, records))).queue();
+        });
+    }
+
+    private long getBoundedCount(OptionMapping count) {
+        return count == null ? MIN_COUNT : Math.min(count.getAsLong(), MAX_COUNT);
```
Tais993, line 85: OptionMapping#getAsLong throws when it's not a number, so maybe add errorhandler if you don't change the type

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+        return Arrays.stream(records.intoArrays())
+            .map(objects -> List.of(objects[0].toString(),
+                    getMemberNameFromEvent(event, objects[1]), objects[2].toString()))
+            .toList();
+    }
+
+    private String getMemberNameFromEvent(SlashCommandEvent event, Object member) {
+        String memberName;
+        try {
+            memberName = event.getGuild().getMemberById((Long) member).getEffectiveName();
```
Tais993, line 99: Doesn't work, there's a reason why I told to run the command instead of using tests.
Regarding why? Running it yourself and the docs explain why.

Tais993, line 99: Use user instead, always works when the ID exists

application/src/main/java/org/togetherjava/tjbot/db/MessageMetadataListener.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Stores the relevant message metadata for on message received event.
+     * 
+     * @param event incoming message received event.
+     */
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event) {
+        if (!event.getAuthor().isBot() && !event.isWebhookMessage()
+                && !event.isFromType(ChannelType.PRIVATE) && event.isFromType(ChannelType.TEXT)) {
```
Tais993, line 35: > !event.isFromType(ChannelType.PRIVATE)
Why listen to a MessageReceivedEvent when you don't want private channels?

Use the GuildMessageReceivedEvent

application/src/main/java/org/togetherjava/tjbot/db/MessageMetadataListener.java
```diff
@@ -0,0 +1,52 @@
...
+            var messageId = event.getMessage().getIdLong();
+            var guildId = event.getGuild().getIdLong();
```
Tais993, line 37: Unnecessary, you already have channel ID

application/src/main/java/org/togetherjava/tjbot/util/PresentationUtils.java
```diff
@@ -0,0 +1,40 @@
...
+     * 
+     * @param dataFrame dataframe represented as List<List<String>> where List<String> represents a
+     *        single row
+     * @param headers column headers for the table
+     * @param horizontalAligns column alignment for the table
+     * @return String representation of the dataFrame in tabular form
+     */
+    public static String dataFrameToAsciiTable(List<List<String>> dataFrame, String[] headers,
+            HorizontalAlign[] horizontalAligns) {
+        Objects.requireNonNull(dataFrame, "DataFrame cannot be null");
```
Tais993, line 29: Use the @NotNull annotation from Jetbrains
And add to the docs that it has to be non-null

application/src/main/resources/db/V2__table_message_metadata.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE MESSAGE_METADATA(
+    MESSAGE_ID       BIGINT,
+    GUILD_ID         BIGINT,
+    CHANNEL_ID       BIGINT,
+    USER_ID          BIGINT,
+    CREATE_TIMESTAMP BIGINT
```
Tais993, line 6: While there's nothing wrong with using this, for your info, Snowflake ID's contain when it was created.
See TimeUtil#getTimeCreated(long) in JDA

So you honestly don't even need this

Tais993 (commented):
+1 for using longs for ID's!

application/src/test/java/org/togetherjava/tjbot/commands/TopHelpersTest.java
d3vel0per: Can be done but JUnits are places where snake case naming has been accepted to aid readability. My bad I should have added a DisplayNameGenerator which would have highlighted the benefits.
If you have good naming, its better than keeping docs or comments in sync and is the very reason you would many a times see things like `fun 'some descriptive test name'()` in languages like Kotlin specifically used in test cases

References:
https://softwareengineering.stackexchange.com/questions/237561/naming-test-methods-in-java
https://github.com/junit-team/junit5/pull/1588 (JUnit 5 SPI for display name generation)
https://github.com/junit-team/junit5/issues/1595 (JUnit 5, failed attempt to create a camelCase based implementation as thre are too many edge cases)

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public boolean isGuildOnly() {
+        return true;
+    }
+
+    @Override
+    public @NotNull CommandData addOptions(@NotNull CommandData commandData) {
+        return commandData.addOptions(new OptionData(OptionType.STRING, COUNT_OPTION,
```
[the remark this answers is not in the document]
d3vel0per: No reason. Just lazy, I just looked at example code from one of the other implementations :P
Fixing this

application/src/main/resources/db/V2__table_message_metadata.sql
d3vel0per: We may need time travel queries in the database like `WHERE CREATE_TIMESTAMP > XXX`
A database doesn't understand snowflake hence its existence is justified.

application/src/main/java/org/togetherjava/tjbot/db/MessageMetadataListener.java
d3vel0per: Again, used from one of the started examples. No idea what a GuildMessageReceivedEvent was.
Seems it would eliminate some of the checks so thats a yes to replacement.

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: I assume that's another API but the check here was for if the userId doesnt exists. So how does that help?

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: I'll change the type.

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: Worked on my test server 🤷‍♂️ but I will check the doc

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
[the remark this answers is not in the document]
Tais993: Next time maybe look further than your nose's length :p

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: Think I saw that but didnt use it as it was under examples. Let me refractor it in

application/src/main/resources/db/V2__table_message_metadata.sql
Tais993: Not sure how the database exactly works, but I'd assume there's a way to look at the start of the ID
[Here](https://discord.com/developers/docs/reference#snowflakes) you can see how snowflake ID's are structured, start is timestamp

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: Can replace but looks perfectly readable to me.

var **guildId** = event.getGuild().getId**Long**();
Its a guildId of type long similarly you always count in numbers

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: Yes you said it's your recommendation. Here records are well records it's an intermediary state. A type of `Result<Record3<Integer, Long, Long>>` isnt very helpful.

I mean you wouldn't be concerned about what's the type of `dsl.with()` or `select(..).as(...)` you are just making sense based on their names. Plus everything is in your IDE

application/src/main/java/org/togetherjava/tjbot/util/PresentationUtils.java
d3vel0per: This I'll check but I mostly tend to rely on JDK libs if they already provide something similar.

build.gradle
d3vel0per: Yeah, I ll rebase based on your PR

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
Tais993: Guild#getMemberById, like the docs say, only check cached members.
[See the default caching policy](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/utils/MemberCachePolicy.html#DEFAULT)
And, people that left the guild display as [UNKNOWN MEMBER] now.

So instead, JDA#retreiveUserById solves both of those.
Only thing required is a valid account.

Yes you lose effective nicknames, so if you really want that you're forced to retrieve the user and member, or to cache all users.
Both not really ideal, but regardless I'd recommend retrieving myself.

application/src/main/java/org/togetherjava/tjbot/db/MessageMetadataListener.java
Tais993: And again, examples can't show everything possible, look further then your nose.

application/src/main/java/org/togetherjava/tjbot/util/PresentationUtils.java
Tais993: Annotations exist for a reason, and Objects#requireNonNull lacks of a lot.

Documentation state nothing, and your IDE probably doesn't either.
By having annotation, docs add it automatically when generating, and you don't need to read source.

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: Yeah looks like I have to make it blocking to maintain simplicity of the codebase.

Tais993: Then write it differently, async can still be simple.
All you do is copy paste the logic into a lambda when done correctly.
Guild#retrieveMembersByIds is what you should use instead

```java
Collection<Long> memberIds = List.of(1000L, 200L);
guild.retrieveMembersByIds(memberIds).onSuccess(list -> {
    Map<Long, String> memberNicknameMap = list.stream().collect(Collectors.toMap(Member::getIdLong, Member::getEffectiveName));
    
    memberIds.forEach(memberId -> {
        System.out.println(memberNicknameMap.get(memberId));
    });
```

Already does everything for you, you just need to loop through all members.
You get their nickname, and format it to however you want.

And after further notice, I just noticed Discord has no retrieve users endpoint so you can just forget about that honestly.

application/src/main/resources/db/V2__table_message_metadata.sql
Tais993: In that case I forgive you, the guild ID is still useless though

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
d3vel0per: No, that's a consumer all I want is a `RestAction<User>` or a `RestAction<String>` this way I can extract logical blocks into specific methods. Fact is JDA type system isn't rich enough to allow for this. It's a collision of wannabe reactive RestAction with CompletableFuture as far as I can see

Tais993: There's map and flatmap, why cant you use those? It works literally the same as your reactive libraries, just less methods.
And retrieving users one by one is pretty bad, so I'd say retrieveMembersById

d3vel0per: 1. You are jumping back and forth between your suggestions. retrieveMembersById is again for cached members only
2. There is nothing about `map/flatMap` in this discussion. The reactive type RestAction is only bound to well REST actions ie. web endpoints and one you get out of that world you need to figure out to map those to other types that Java has. Just like your wouldn't mix and match nio with io(which you can but you shouldnt ideally) or streams with loops

There isn't a reactive Producer type that you can use across your API's which is bad. It's a mismatch of sorts.

Tais993: How is retrieveMembersByIds only for cached users?
Did you even read the docs?

Yes if you're using retrieveMembersByIds there's no RestAction. You mentioned a RestAction<User> and that implies you're retrieving users manually.

Still, I showed you a simple example of how it can be done

Don't make it more complicated then it is.

Yes there's no reactor like thing, if you want reactive programming this is the wrong library, this just allows mapping.

d3vel0per: Apologies probably my IDE autocompleted but all of getMemberByXXX methods of Guild retrieve data from cache.
image

Tais993: Where did I mention getMembersByIds
image
image
image

d3vel0per: My bad. But now it has a new type Task lol!!

Tais993: I gave you code above, in how you can do it, assuming you're alright with using consumers.

Tais993: And yeah, it's a Task not a RestAction.
Why? Because it's a gateway request, not a rest request.

d3vel0per:
Changed to draft. Should not be merged.
Waiting for a few other PR's to go in.

Zabuzard:
@d3vel0per Any news on this?

d3vel0per:
> @d3vel0per Any news on this?

Yes, I am waiting for:
#119 - Separate thread pool for command system + AbstractCommand
#135 - Database migration and segregation of module

Since both will need merges at my end.

Zabuzard:
Tell me if you need help with rebasing.

By the way, the database related things you have been waiting for with #135 found their way into `develop` already via different PRs. I am unblocking this.

java-coding-prodigy:
We seem to be having some trouble with the formatting, I think we should wait till @I-Al-Istannen and @borgrel fix it.

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.jooq.impl.DSL.*;
+import static org.togetherjava.tjbot.db.generated.tables.MessageMetadata.MESSAGE_METADATA;
+import static org.togetherjava.tjbot.util.MessageTemplate.PLAINTEXT_MESSAGE_TEMPLATE;
+
+public final class TopHelpers extends SlashCommandAdapter {
+
+    private static final String COUNT_OPTION = "count";
```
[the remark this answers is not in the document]
Tais993: Zabuzard is also referring to the variable's name if I understand it correctly

application/src/main/java/org/togetherjava/tjbot/util/PresentationUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.IntStream;
+
+/**
+ * Utility class for representation of data in various formats
+ */
+public final class PresentationUtils {
+    private PresentationUtils() {}
```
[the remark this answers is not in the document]
Tais993: He said "or" 

So, why did you ignore his second statement?

application/src/main/java/org/togetherjava/tjbot/commands/TopHelpers.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String NO_ENTRIES = "No entries";
+
+    private static final long MIN_COUNT = 10L;
+    private static final long MAX_COUNT = 30L;
+
+    private record TopHelperRow(Integer serialId, Long userId, Long messageCount) {
```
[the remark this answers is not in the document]
Tais993: I'd use the primitives instead

Tais993 (changes requested):

Zabuzard:
@d3vel0per How is it going? Still planning on finishing this? Otherwise, should we maybe assign it to someone else to help out?

Zabuzard:
Seems @d3vel0per is a bit busy right now (which is totally fair). Unassigned. Someone else can take over, if wanted 👍

marko-radosavljevic:
Since origin branch of this PR is on a fork, people cannot work on it.

Moved to #282. :relaxed:

Thanks, @d3vel0per :heart:

PR #123, opened 2021-09-15
Add sonarlint

marko-radosavljevic:
Adds sonarlint so we have basic linting inside PRs from forks.

### DIFF
--- a/.github/workflows/basic-checks.yml
+++ b/.github/workflows/basic-checks.yml
@@ -15,8 +15,6 @@ jobs:
         with:
           java-version: ${{ env.JAVA_VERSION }}
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 2
       - name: Check
         run: ./gradlew spotlessCheck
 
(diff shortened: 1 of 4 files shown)


marko-radosavljevic:
The issue with it is that it fails the build on every violation, which is not how it's currently working.

To suppress a violation this can be done: [link](https://stackoverflow.com/a/39109612).

I-Al-Istannen:
> The issue with it is that it fails the build on every violation, which is not how it's currently working.
> 
> To suppress a violation this can be done: [link](https://stackoverflow.com/a/39109612).

I think you can disable them with a comment / SuppressWarnings?

marko-radosavljevic:
Yup, that's on the link I posted.

If you have the id of the rule you want to ignore, annotation:
`@SuppressWarnings("squid:S0016")`

Or a comment:
`//NOSONAR`

> The issue with it is that it fails the build on every violation, which is not how it's currently working.

This issue can also be seen as a feature, because you get to see all the violations on build, and you can fix them before pushing. I can also implement a flag to turn it on/off on build too, will do that later today.

I-Al-Istannen:
I would probably refrain from merging this until the issues it fails on are resolved? I am not sure whether a failing develop build is a good idea?

marko-radosavljevic:
Yes, that's the idea, wait for violations to be fixed before merging. ^^

Are you working on that @I-Al-Istannen, or you want someone else to pick it up?

Zabuzard (approved):

Zabuzard:
> 
> 
> Yes, that's the idea, wait for violations to be fixed before merging. ^^
> 
> Are you working on that @I-Al-Istannen, or you want someone else to pick it up?

If this takes too long, I would vote for merging just to put more pressure on it and to prevent that the state gets worse the more is merged.

I-Al-Istannen:
Ah I missed that comment and I wasn't, but I can give it a shot on the current develop branch.

I-Al-Istannen (approved):

PR #135, opened 2021-09-18
Tag system

### LINKED ISSUE #61: Migrate tag system
We have to migrate the existing tag system. The details are up to be decided by the community.

### `?tag` system by **Dyno**

People with the role **Tag Manager** can create, delete and edit custom message which Dyno can then post.

Available commands are:

#### General commands
- `?tag` - Shows a help dialog
- `?tags` - Shows all user created tags

#### Tag commands

- `?tag raw <name>` - Get the raw tag for use copying/editing.
- `?tag create <name> <message>` - Create a new tag.
- `?tag edit <name> <message>` - Edit an existing tag.
- `?tag delete <name>` - Delete an existing tag.

#### Category commands (we never used these)
- `?tag category <name>` - Create a tag category.
- `?tag categories` - List tag categories.
- `?tag delcat <name>` - Delete a category.

#### All tags
The following are all user created tags so far:
> xy, w3, visibility-chart, visibility, unsupported, typeparamnaming, typeparam, this, syntax, student, static, stacktrace, spring-password-handling, simplenet, share, scanner, regex, question, projects, premade, ping, person, parseLine, ownership, ossu, oracle, operator, nio, naming, mooc, mockdata, method, mcve, markdown, manual, lsp, js, jpackage, javascript, javaisnotjs, javaisnotjavascript, javafxpom, javafx-swing, javadoc, interrupt, interfacesimple, interface-abstract2, interface-abstract, interface, instance, install, ide, helsinki, git, gist, getting-started, free, framework, executors, equals, eoa, doesntwork, docker, dfs, debugger, debug, constructor, code, busychannel, busy, bfs, ask, arch, android-getstarted, ad, abstract, URI, SRP, ++

illuminator3:
This PR adds a tag system & commands to use it. Closes #61.


illuminator3:
Seems like I forgot to add javadoc, will add it now

Tais993 (changes requested):
Well done.

While I would strongly prefer it if you made usage of sub-commands instead.
Makes it easier for the user and will allow a ton more commands.

This in total is 8 commands while it could be 1.

application/src/main/java/org/togetherjava/tjbot/commands/tag/CreateIdTagCommand.java
```diff
@@ -0,0 +1,77 @@
...
+    public CreateIdTagCommand(TagSystem tagSystem) {
+        super("createidtag", "Create a tag using a message id", true);
+
+        this.tagSystem = tagSystem;
+    }
+
+    @Override
+    public CommandData addOptions(CommandData data) {
+        return data.addOption(OptionType.STRING, "tag-id", "Tag id", true)
+            .addOption(OptionType.STRING, "message-id", "Message id", true);
```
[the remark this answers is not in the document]
illuminator3: I was looking for something like that but only found INTEGER which wouldn't work for longs. Thanks!

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void onButtonClick(ButtonClickEvent event, List<String> idArgs) {
+        String userId = idArgs.get(0);
+
+        if (event.getUser().getId().equals(userId)
+                || event.getMember().hasPermission(Permission.MESSAGE_MANAGE)) {
+            event.getMessage().delete().queue();
+        } else {
+            event.reply(":police_car: Button theft is illegal").setEphemeral(true).queue();
```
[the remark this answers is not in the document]
illuminator3: Isn't that only for server emotes?

illuminator3:
> While I would strongly prefer it if you made usage of sub-commands instead.

I first thought about that as well, but it would be kinda confusing for new players to do `/tag view <id>` instead of just `/tag <id>`. I could maybe do sub commands with a `/tagmanage <create|edit|delete>` or `/managetag`.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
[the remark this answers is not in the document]
Tais993: Wops, didn't think about their existence.
Yes that's only for server emotes.

Tais993:
> > While I would strongly prefer it if you made usage of sub-commands instead.
> 
> I first thought about that as well, but it would be kinda confusing for new players to do `/tag view <id>` instead of just `/tag <id>`. I could maybe do sub commands with a `/tagmanage <create|edit|delete>` or `/managetag`.

Do like the idea of a separate tagmanage command, even better then one for all

illuminator3:
> > > While I would strongly prefer it if you made usage of sub-commands instead.
> > 
> > 
> > I first thought about that as well, but it would be kinda confusing for new players to do `/tag view <id>` instead of just `/tag <id>`. I could maybe do sub commands with a `/tagmanage <create|edit|delete>` or `/managetag`.
> 
> Do like the idea of a separate tagmanage command, even better than one for all

Yea, I'll implement it!

Tais993 (dismissed):
I approve it.

Still I'll mention that it's maybe nice to replace some of the code in the switch with methods?
Just a suggestion, the code is good. 👍

Tais993 (approved):
You're cute keep it up

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,90 @@
...
+            String tagId = idOption.getAsString();
+
+            TagUtility.replyTag(event, tagId, event.getUser().getAsTag(), tagSystem,
+                    event.getOption("raw") != null, event.getUser().getId());
+        } else {
+            SelectionMenu.Builder menu =
+                    SelectionMenu.create(generateComponentId(event.getUser().getId()))
+                        .setRequiredRange(1, 1);
+
+            for (String tag : tagSystem.retrieveIds()) {
```
I-Al-Istannen: If there are more than 25 tags, this will breaks as a menu can only have 25 entries.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,90 @@
...
+                                "Delete", Emoji.fromUnicode("\uD83D\uDDD1")))
+                .queue();
+        }
+    }
+
+    @Override
+    public void onButtonClick(ButtonClickEvent event, List<String> idArgs) {
+        String userId = idArgs.get(0);
+
+        if (event.getUser().getId().equals(userId)
```
I-Al-Istannen, line 64: Not to future us: This might be nice as a utility method?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,231 @@
...
+            event
+                .replyEmbeds(new EmbedBuilder().setColor(Color.GREEN)
+                    .setTitle(SUCCESS)
+                    .setTimestamp(LocalDateTime.now())
+                    .setFooter(event.getUser().getAsTag())
+                    .setDescription("Successfully edited tag '" + tagId + "'")
+                    .build())
+                .queue();
+        }, failure -> {
+            if (failure instanceof ErrorResponseException ex
```
I-Al-Istannen, line 126: Does this silently swallow other error conditions?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,231 @@
...
+
+        if (!tagSystem.exists(tagId)) {
+            event.reply(THIS_TAG_DOES_NOT_EXIST).setEphemeral(true).queue();
+
+            return;
+        }
+
+        event.reply("You sure? Confirming this will delete the tag '" + tagId + "'")
+            .addActionRow(
+                    Button.of(ButtonStyle.SUCCESS, generateComponentId(member.getId(), tagId),
```
I-Al-Istannen, line 144: Switch them around. Deleting is the destructive button and should be red, aborting is the safe option and should be green/neutral.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,231 @@
...
+                            "Of course!"),
+                    Button.of(ButtonStyle.DANGER, generateComponentId(member.getId()), "Abort"))
+            .queue();
+    }
+
+    private void createTag(SlashCommandEvent event) {
```
I-Al-Istannen, line 150: What is this versus an "id tag"? Both seem to have an id?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,231 @@
...
+        });
+    }
+
+    @Override
+    public void onButtonClick(ButtonClickEvent event, List<String> idArgs) {
+        String userId = idArgs.get(0);
+
+        if (event.getUser().getId().equals(userId)
+                || event.getMember().hasPermission(Permission.MESSAGE_MANAGE)) {
+            if (event.getButton().getStyle() == ButtonStyle.SUCCESS
```
I-Al-Istannen, line 207: I'd rather you give the delete button a name (maybe "DELETE") than only switching on the style of it.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -0,0 +1,88 @@
...
+            .setTimestamp(LocalDateTime.now())
+            .setFooter(user)
+            .setTitle("All available tags");
+
+        for (Map.Entry<String, String> entry : tagSystem.retrieve().entrySet()) {
+            String id = entry.getKey();
+            String text = entry.getValue();
+            String preview = text.substring(0, Math.min(text.length(), 50));
+
+            builder.addField(id, preview + (text.length() > 50 ? "..." : ""), true);
```
I-Al-Istannen, line 74: The field count is limited.

settings.gradle
```diff
@@ -3,3 +3,4 @@ rootProject.name = 'TJ-Bot'
 include 'application'
 include 'database'
 include 'formatter'
+include 'tag-system'
```
I-Al-Istannen, line 6: A newline is missing.

tag-system/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+    implementation 'org.jooq:jooq:3.15.2'
+
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
+}
+
+test {
+    useJUnitPlatform()
+}
```
I-Al-Istannen, line 23: A newline is missing.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,80 @@
...
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.Tags;
+import org.togetherjava.tjbot.db.generated.tables.records.TagsRecord;
+
+import java.util.*;
+
+public class TagSystem {
+    private final Database database;
+    private boolean changed = true;
+    private Map<String, String> cache;
```
I-Al-Istannen, line 13: A non-threadsafe cache sounds a bit annoying. I would just not cache at all until profiling shows it would be beneficial enough.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
```diff
@@ -0,0 +1,33 @@
+package org.togetherjava.tjbot.tag;
+
+import org.togetherjava.tjbot.db.Database;
+
+public class TagSystemInstanceHolder {
+    private TagSystemInstanceHolder() {}
+
+    private static TagSystem instance;
+
+    public static TagSystem getOrCreate(Database database) {
```
I-Al-Istannen, line 10: I am not sure why you need all these methods, where do you use them?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: "id tag" lets you specify the content of the tag using a message id

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: To create a lazy-singleton that lets you specify the database instance.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -127,7 +127,7 @@ private void editIdTag(SlashCommandEvent event) {
                     && ex.getErrorResponse() == ErrorResponse.UNKNOWN_MESSAGE) {
                 event.reply("This message doesn't exist").setEphemeral(true).queue();
             } else {
-                throw new RuntimeException(failure);
+                throw new RethrownRuntimeException(failure);
```
Tais993, line 130: Instead of adding a useless custom exception you can also just tell Sonarcloud to ignore the warning

illuminator3: How?

Tais993:
image
Make sure you're logged in when trying

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
marko-radosavljevic: https://stackoverflow.com/questions/39109228/how-can-we-ignore-some-sonarqube-rules-in-java/39109612#39109612

database/src/main/resources/db/V2__Tag_System.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE tags
```
d3vel0per, line 1: Should the file name be representative of the action eg. V2__Create_table_tags.sql? That way looking at a resource folder would give a clear idea on what each of the migrations represent.
Just a thought

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
...
+        database.writeTransaction(ctx -> {
+            ctx.deleteFrom(Tags.TAGS).where(Tags.TAGS.ID.eq(tag)).execute();
+        });
+    }
+
+    public void put(String tag, String content) {
+        database.writeTransaction(ctx -> {
+            TagsRecord tagsRecord = ctx.newRecord(Tags.TAGS).setId(tag).setContent(content);
+
+            if (tagsRecord.update() == 0) {
```
d3vel0per, line 35: Most databases provide a facility of upserting records. That would make it an atomic statement as well as a single database call. Putting links below, I am assuming jooq merge will change it to sqlite upsert.

Ref:
https://www.jooq.org/doc/3.1/manual/sql-building/sql-statements/merge-statement/
https://www.sqlite.org/lang_UPSERT.html

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        });
+    }
+
+    public String get(String tag) {
+        return database.readTransaction(ctx -> {
+            return Optional
+                .ofNullable(ctx.selectFrom(Tags.TAGS).where(Tags.TAGS.ID.eq(tag)).fetchOne())
+                .map(TagsRecord::getContent)
+                .orElse(null);
```
d3vel0per, line 46: Returning a null Optional is weird. Should we return "Tag not found" or something similar?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
d3vel0per: This isnt safe btw. May be just use an enum instead
Visibility of instance isn't guaranteed without it being volatile.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+            return;
+        }
+
+        switch (event.getSubcommandName()) {
+            case "edittag" -> editTag(event);
+            case "editidtag" -> editIdTag(event);
+            case "deletetag" -> deleteTag(event, member);
+            case "createtag" -> createTag(event);
+            case "createidtag" -> createIdTag(event);
+            default -> throw new IllegalStateException();
```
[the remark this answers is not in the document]
Tais993: It's impossible to throw so no

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: No, gigo (garbage in garbage out).

database/src/main/resources/db/V2__Tag_System.sql
illuminator3: The other file is also named like that

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+                    .setFooter(event.getUser().getAsTag())
+                    .setDescription("Successfully edited tag '" + tagId + "'")
+                    .build())
+                .queue();
+        }, failure -> {
+            if (failure instanceof ErrorResponseException ex
+                    && ex.getErrorResponse() == ErrorResponse.UNKNOWN_MESSAGE) {
+                event.reply("This message doesn't exist").setEphemeral(true).queue();
+            } else {
+                throw new RuntimeException(failure);
```
Tais993, line 130: You can also just print the existing failure, throwing a new runtime exception is kinda useless?

.gitignore
```diff
@@ -143,4 +143,4 @@ gradle-app.setting
 .classpath
 
 # End of https://www.toptal.com/developers/gitignore/api/netbeans,intellij,java,gradle,eclipse
-application/db/
+/db/
```
Zabuzard, line 146: did the structure change? please double check that this actually still covers the db stuff with the most recent version of develop.

application/build.gradle
```diff
@@ -39,6 +39,7 @@ shadowJar {
 
 dependencies {
     implementation project(':database')
+    implementation project(':tag-system')
```
Zabuzard, line 42: why is it its own "project", we did not do this to the other commands as well, did we? whats the benefit? maybe others should do that as well?

application/src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
```diff
@@ -1,7 +1,6 @@
-package org.togetherjava.tjbot.commands.example;
+package org.togetherjava.tjbot.commands;
```
Zabuzard, line 1: This change doesnt seem related to the tag system and will possibly clash with the changes introduced in #125 .

application/src/main/java/org/togetherjava/tjbot/commands/Command.java
```diff
@@ -15,8 +15,7 @@
 
 /**
  * The class all commands have to abstract. <br>
- * Instead of implementing this interface, you can also extend
- * {@link org.togetherjava.tjbot.commands.example.AbstractCommand}.
+ * Instead of implementing this interface, you can also extend {@link AbstractCommand}.
```
Zabuzard, line 18: Clashes with #125

application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
```diff
@@ -30,8 +34,11 @@
     private final List<Command> commandList = new ArrayList<>();
     private final Map<String, Command> commandMap;
 
-    public CommandHandler() {
-        commandList.addAll(List.of(new ReloadCommand(this)));
+    public CommandHandler(Database database) {
+        TagSystem tagSystem = TagSystemInstanceHolder.getOrCreate(database);
```
Zabuzard, line 38: I would prefer it if we could add the whole tag command family with a single command to keep the changes to this file minimal.

`CommandHandler` should ideally just add the command family in one argument and thats it.

Would it be easy to refactor it in a way to keep changes to this file as minimal as possible?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,88 @@
...
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.ButtonStyle;
+import net.dv8tion.jda.api.interactions.components.selections.SelectOption;
+import net.dv8tion.jda.api.interactions.components.selections.SelectionMenu;
+import org.togetherjava.tjbot.commands.AbstractCommand;
+import org.togetherjava.tjbot.tag.TagSystem;
+
+import java.util.List;
+
+public class TagCommand extends AbstractCommand {
```
Zabuzard, line 20: Missing Javadoc, please explain the command, give some example discord usages and similar.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public void onSlashCommand(SlashCommandEvent event) {
+        OptionMapping idOption = event.getOption("id");
+
+        if (idOption != null) {
+            String tagId = idOption.getAsString();
+
+            TagUtility.replyTag(event, tagId, event.getUser().getAsTag(), tagSystem,
+                    event.getOption("raw") != null, event.getUser().getId());
+        } else {
```
Zabuzard, line 44: Prefer early-out. So `if (...) { ... return; } ...` and then the rest without a big `else` block.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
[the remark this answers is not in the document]
Zabuzard: `illegal` is kinda funny but i would vote to rename it to `not allowed` or similar. it just sounds a bit too hardcore for people who dont understand jokes.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (event.getUser().getId().equals(userId)) {
+            SelectOption option = event.getSelectedOptions().get(0);
+            String tagId = option.getLabel();
+
+            event.getMessage().delete().queue();
+
+            TagUtility.sendTag(event.getMessageChannel(), tagId, event.getUser().getAsTag(),
+                    tagSystem, false, generateComponentId(userId));
+        } else {
```
Zabuzard, line 84: prefer early out. reverse the condition and then have `if (!...) { ... return;}` and the rest of the main logic without a conditional and at the bottom.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
```diff
@@ -0,0 +1,88 @@
...
+            event.reply(":police_car: Selection menu theft is illegal").setEphemeral(true).queue();
```
Zabuzard, line 85: `illegal` -> `not allowed` (as before)

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+import net.dv8tion.jda.api.requests.ErrorResponse;
+import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.commands.AbstractCommand;
+import org.togetherjava.tjbot.tag.TagSystem;
+
+import java.awt.*;
+import java.time.LocalDateTime;
+import java.util.List;
+
+public class TagManageCommand extends AbstractCommand {
```
Zabuzard, line 23: missing javadoc (as earlier)

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+    private static final String TAG_ID = "Tag id";
+    private static final String CONTENT = "content";
+    private static final String TAG_MINUS_ID = "tag-id";
```
Zabuzard, line 26: the names for those are pretty bad imo. prefer a name that explains its usage instead of the plain content. the `MINUS` triggers me, that should be removed definitely.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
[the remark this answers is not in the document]
Zabuzard: Please put some explanation string into the `IllegalStateException`, makes it easier for people to debug. In particular if the bot crashes and all we have is the log file.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
Zabuzard: minor but `e` is preferred over `ex` (nice usage of pattern matching btw)

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
Zabuzard: agreed. not only that, `RuntimeException` also does not add anything useful to the table.

either use a more specific exception, add error details to it or just rethrow instead. `throw failure;`.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+            event
+                .replyEmbeds(new EmbedBuilder().setColor(Color.GREEN)
+                    .setTitle(SUCCESS)
+                    .setTimestamp(LocalDateTime.now())
+                    .setFooter(event.getUser().getAsTag())
+                    .setDescription("Successfully created tag '" + tagId + "'")
+                    .build())
+                .queue();
+        }, failure -> {
+            if (failure instanceof ErrorResponseException ex
```
Zabuzard, line 196: swallows other errors. check the previous message and apply the improvements here as well.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -0,0 +1,233 @@
...
+                        .setTimestamp(LocalDateTime.now())
+                        .setFooter(event.getUser().getAsTag())
+                        .setDescription("Successfully deleted tag '" + tagId + "'")
+                        .build())
+                    .setEphemeral(true)
+                    .queue();
+            } else if (event.getButton().getLabel().equals("Abort")) {
+                event.getMessage().delete().queue();
+            }
+        } else {
```
Zabuzard, line 229: prefer early out. reverse the condition and have the rest without a big if. `if (!...) { ... return } ...`.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -0,0 +1,68 @@
...
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.ButtonStyle;
+import org.togetherjava.tjbot.tag.TagSystem;
+
+import java.awt.*;
+import java.time.LocalDateTime;
+
+public class TagUtility {
+    private TagUtility() {}
```
Zabuzard, line 16: if u want, u can use the enum-pattern `public enum TagUtility { ; ...}`

just minor though, i actually prefer it ur way cause an enum could distract a reader for only the minor benefit of adding security against serialization and reflection (not really worth the deal).

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -0,0 +1,68 @@
...
+import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.ButtonStyle;
+import org.togetherjava.tjbot.tag.TagSystem;
+
+import java.awt.*;
+import java.time.LocalDateTime;
+
+public class TagUtility {
```
Zabuzard, line 15: missing javadoc. what does the class offer? give some overview please

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -0,0 +1,68 @@
...
+        return s.replaceAll("([^a-zA-Z0-9 \n\r])", "\\\\$1");
+    }
+
+    public static EmbedBuilder buildAllTagsEmbed(String user, TagSystem tagSystem) {
+        return new EmbedBuilder().setColor(Color.MAGENTA)
+            .setTimestamp(LocalDateTime.now())
+            .setFooter(user)
+            .setDescription(String.join(", ", tagSystem.retrieveIds()));
+    }
+}
```
Zabuzard, line 68: all methods are missing javadoc.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -0,0 +1,68 @@
...
+            event
+                .replyEmbeds(buildAllTagsEmbed(requestor, tagSystem)
+                    .setTitle("Could not find tag '" + tagId + "'")
+                    .build())
+                .setEphemeral(true)
+                .queue();
+        }
+    }
+
+    public static String escape(String s) {
```
Zabuzard, line 58: this does not seem to be a `TagSystem` specific utility. it might be helpful for all commands. elevate it into some sort of global utility place

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagsCommand.java
```diff
@@ -0,0 +1,46 @@
...
+import net.dv8tion.jda.api.events.interaction.ButtonClickEvent;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.ButtonStyle;
+import org.togetherjava.tjbot.commands.AbstractCommand;
+import org.togetherjava.tjbot.tag.TagSystem;
+
+import java.util.List;
+
+public class TagsCommand extends AbstractCommand {
```
Zabuzard, line 14: missing javadoc

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagsCommand.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void onButtonClick(ButtonClickEvent event, List<String> idArgs) {
+        String userId = idArgs.get(0);
+
+        if (event.getUser().getId().equals(userId)
+                || event.getMember().hasPermission(Permission.MESSAGE_MANAGE)) {
+            event.getMessage().delete().queue();
+        } else {
+            event.reply(":police_car: Button theft is illegal").setEphemeral(true).queue();
```
Zabuzard, line 43: prefer reversing the condition and early out.

database/src/main/java/org/togetherjava/tjbot/db/Database.java
```diff
@@ -46,10 +46,8 @@ public Database(String jdbcUrl) throws SQLException {
         SQLiteDataSource dataSource = new SQLiteDataSource(sqliteConfig);
         dataSource.setUrl(jdbcUrl);
 
-        Flyway flyway = Flyway.configure()
-            .dataSource(dataSource)
-            .locations("/org/togetherjava/tjbot/db")
-            .load();
+        Flyway flyway =
+                Flyway.configure().dataSource(dataSource).locations("classpath:/db/").load();
```
Zabuzard, line 50: is this an intended change? please double check

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
+package org.togetherjava.tjbot.tag;
+
+import org.jooq.Result;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.Tags;
+import org.togetherjava.tjbot.db.generated.tables.records.TagsRecord;
+
+import java.util.*;
+
+public class TagSystem {
```
Zabuzard, line 10: missing javadoc

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            return out;
+        });
+    }
+
+    public Set<String> retrieveIds() {
+        return retrieve().keySet();
+    }
+}
```
Zabuzard, line 66: all methods are missing javadoc

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
...
+    public Map<String, String> retrieve() {
+        return database.readTransaction(ctx -> {
+            Result<TagsRecord> result = ctx.selectFrom(Tags.TAGS).fetch();
+            Map<String, String> out = new HashMap<>();
+
+            for (TagsRecord tagsRecord : result) {
+                out.put(tagsRecord.getId(), tagsRecord.getContent());
+            }
+
+            return out;
```
Zabuzard, line 59: if u want, u could do this nicely with streams. im okay with both though

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,66 @@
...
+        });
+    }
+
+    public Set<String> retrieveIds() {
+        return retrieve().keySet();
```
Zabuzard, line 64: if applicable, prefer instead:
```suggestion
        return Collections.unmodifiableSet(retrieve().keySet());
```

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
```diff
@@ -0,0 +1,33 @@
+package org.togetherjava.tjbot.tag;
+
+import org.togetherjava.tjbot.db.Database;
+
+public class TagSystemInstanceHolder {
```
Zabuzard, line 5: missing javadoc

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
```diff
@@ -0,0 +1,33 @@
...
+    private TagSystemInstanceHolder() {}
+
+    private static volatile TagSystem instance;
+
+    public static TagSystem getOrCreate(Database database) {
+        if (instance == null) {
+            setInstance(new TagSystem(database));
```
Zabuzard, line 12: is lazy init worth the trouble of making this class safe? why not just have the instance always?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        return instance;
+    }
+
+    public static TagSystem getInstance() {
+        if (instance == null) {
+            throw new UnsupportedOperationException("Instance not yet set");
```
Zabuzard, line 20: this is the wrong exception imo, since the operation **is supported**, just not with the current state
```suggestion
            throw new IllegalStateException("Instance not yet set");
```

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        return instance;
+    }
+
+    public static void setInstance(TagSystem tagSystem) {
+        if (instance != null) {
+            throw new UnsupportedOperationException("Instance already set");
```
Zabuzard, line 28:
```suggestion
            throw new IllegalStateException("Instance already set");
```

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
Zabuzard: In particular, i dislike that the two methods below can only give runtime safety and can not prevent wrong calls at compile-time.

without lazy init, all of those trouble is gone.

Zabuzard (changes requested):

Zabuzard:
All classes that are not explicitly designed for inheritance should be marked `final`. I add this as general request, since it applies to almost all classes you added.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: I can't directly rethrow it here, since the 2nd argument is a consumer.

.gitignore
illuminator3: That's actually independent for everyone. I might as well just do `**/db/`

application/build.gradle
illuminator3: Yes, I was planning on doing a PR for the command system after this one. It's so that you don't have a big blob of packages in a single module; the "application" module should only contain the `Application` enum.

application/src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
illuminator3: No, it doesn't. The commits of that PR are also in this branch.

application/src/main/java/org/togetherjava/tjbot/commands/Command.java
illuminator3: (same as above)

application/src/main/java/org/togetherjava/tjbot/commands/CommandHandler.java
illuminator3: I don't understand what you mean with this, where else am I supposed to register the commands?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
Tais993: I don't understand why you're throwing the same exception within a RuntimeException?
Instead, just print the error and tell the user something went wrong?

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: sonarcloud was telling me to do those. I'll revert it.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: Yea, I forgot to do this here too.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
illuminator3: Put into `CommandUtility`

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagsCommand.java
illuminator3: done

database/src/main/java/org/togetherjava/tjbot/db/Database.java
illuminator3: Yea, it is. I talked with I Al Istannen about it.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: > is lazy init worth the trouble of making this class safe? why not just have the instance always?

Because otherwise there will be no way to retrieve the database instance at the beginning.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: done

Zabuzard:
I have some questions/proposals I would like to clarify:

- Those buttons under the messages are kinda cool but really big and bloaty. Can we make sure (if not already happened), they are only present for the tag-management-commands but not for the "view" command?
- Can we restrict tag management to roles? I would like to only allow it for `Moderator`, `Staff Helper`, `Top Helpers`. In the past we had a specific `Tag Manager` role for that. I would like to get rid of that though.

illuminator3:
> I have some questions/proposals I would like to clarify:
> 
> * Those buttons under the messages are kinda cool but really big and bloaty. Can we make sure (if not already happened), they are only present for the tag-management-commands but not for the "view" command?

This is pretty subjective. I think they're good and needed to delete accidental taggings.

> * Can we restrict tag management to roles? I would like to only allow it for `Moderator`, `Staff Helper`, `Top Helpers`. In the past we had a specific `Tag Manager` role for that. I would like to get rid of that though.

This is a pretty big request. This would be a good fit for another PR.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: Yea, I'm just printing the stacktrace of the exception now.

Zabuzard:
> 
> 
> > I have some questions/proposals I would like to clarify:
> > 
> > * Those buttons under the messages are kinda cool but really big and bloaty. Can we make sure (if not already happened), they are only present for the tag-management-commands but not for the "view" command?
> 
> This is pretty subjective. I think they're good and needed to delete accidental taggings.
> 
> > * Can we restrict tag management to roles? I would like to only allow it for `Moderator`, `Staff Helper`, `Top Helpers`. In the past we had a specific `Tag Manager` role for that. I would like to get rid of that though.
> 
> This is a pretty big request. This would be a good fit for another PR.

Okay. Can you clarify how your current PR works though? With respect to the tag-manage roles (can anyone edit it?) and with respect to the buttons under the "tag view" command?

illuminator3:
> > > I have some questions/proposals I would like to clarify:
> > > 
> > > * Those buttons under the messages are kinda cool but really big and bloaty. Can we make sure (if not already happened), they are only present for the tag-management-commands but not for the "view" command?
> > 
> > 
> > This is pretty subjective. I think they're good and needed to delete accidental taggings.
> > > * Can we restrict tag management to roles? I would like to only allow it for `Moderator`, `Staff Helper`, `Top Helpers`. In the past we had a specific `Tag Manager` role for that. I would like to get rid of that though.
> > 
> > 
> > This is a pretty big request. This would be a good fit for another PR.
> 
> Okay. Can you clarify how your current PR works though? With respect to the tag-manage roles (can anyone edit it?) and with respect to the buttons under the "tag view" command?

Currently, everyone with the MESSAGE_MANAGE permission can create, edit, delete tags and can also click the "Delete" button under a tagging even if it's not theirs.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagCommand.java
illuminator3: added

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: added

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
d3vel0per: Even now it isn't safe. Check for double checked locking if you want to do it this way.
I would still prefer something like an enum or may a static class holding the instance if you want to do it lazily

database/src/main/resources/db/V2__Tag_System.sql
d3vel0per: May be rename both the files instead.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
d3vel0per: Ignored 😢

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
d3vel0per: why?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: Yea, because I had no clue what you were talking about / suggesting.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: I'd rather get a NPE and fail-hard instead of returning something not useful when the input's clearly wrong.

Tais993: An optional and a NPE?
Why return an optional if you want a NPE

illuminator3: I'm not returning an optional.

Tais993: Didn't notice the `database.readTransaction` cause I was reading from the conversation tab

application/src/main/java/org/togetherjava/tjbot/commands/CommandUtility.java
```diff
@@ -0,0 +1,9 @@
+package org.togetherjava.tjbot.commands;
+
+public final class CommandUtility {
```
Zabuzard, line 3: javadoc

application/src/main/java/org/togetherjava/tjbot/commands/CommandUtility.java
```diff
@@ -0,0 +1,9 @@
...
+    private CommandUtility() {}
+
+    public static String escape(String s) {
```
Zabuzard, line 6: javadoc

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -40,21 +33,21 @@ public TagManageCommand(TagSystem tagSystem) {
...
                 new SubcommandData("createtag", "Create a tag")
-                    .addOption(OptionType.STRING, "id", TAG_ID, true)
-                    .addOption(OptionType.STRING, CONTENT, "Tag content", true),
+                    .addOption(OptionType.STRING, "id", "tag id", true)
+                    .addOption(OptionType.STRING, "content", "Tag content", true),
                 new SubcommandData("createidtag", "Create a tag using a message id")
-                    .addOption(OptionType.STRING, TAG_MINUS_ID, TAG_ID, true)
-                    .addOption(OptionType.NUMBER, MESSAGE_MINUS_ID, "Message id", true));
+                    .addOption(OptionType.STRING, "tag-id", "tag id", true)
+                    .addOption(OptionType.NUMBER, "message-id", "Message id", true));
```
Zabuzard: I think using constants is better than having it with plain `String`s. That way we at least have compile-time safety.
Maybe you misunderstood my last comment, it was really only about the names of those constants. Please revert.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -127,7 +120,7 @@ private void editIdTag(SlashCommandEvent event) {
                     && ex.getErrorResponse() == ErrorResponse.UNKNOWN_MESSAGE) {
                 event.reply("This message doesn't exist").setEphemeral(true).queue();
             } else {
-                throw new RuntimeException(failure);
+                failure.printStackTrace();
```
Zabuzard, line 123: Please dont print the stack trace. Either rethrow or use the logging system. `log.error(failure)` or something.
I would prefer a rethrow though, then we could react on it and handle it higher up in the logic - if needed.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -196,6 +189,8 @@ private void createIdTag(SlashCommandEvent event) {
             if (failure instanceof ErrorResponseException ex
                     && ex.getErrorResponse() == ErrorResponse.UNKNOWN_MESSAGE) {
                 event.reply("This message doesn't exist").setEphemeral(true).queue();
+            } else {
+                failure.printStackTrace();
```
Zabuzard, line 193: (see above about rethrow and logger)

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
```diff
@@ -204,30 +199,32 @@ private void createIdTag(SlashCommandEvent event) {
...
+                .replyEmbeds(new EmbedBuilder().setColor(Color.GREEN)
+                    .setTitle("Success")
+                    .setTimestamp(LocalDateTime.now())
+                    .setFooter(event.getUser().getAsTag())
+                    .setDescription("Successfully deleted tag '" + tagId + "'")
+                    .build())
+                .setEphemeral(true)
+                .queue();
+        } else if (event.getButton().getLabel().equals("Abort")) {
+            event.getMessage().delete().queue();
```
Zabuzard, line 227: You could apply the early-return thing to the `Abort` option as well, that way moving the big block of "main logic" outside of its `if`.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: > I would prefer a rethrow though

Rethrowing is not possible here, since it's a `Consumer`.

application/src/main/java/org/togetherjava/tjbot/commands/AbstractCommand.java
Zabuzard: Oh, I see. Then I should look at a diff between you and that branch instead of you and develop. But that also means that this PR has to be blocked until #125 is merged. Otherwise you will merge it together with your PR.

illuminator3: #125 has already been merged.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
Zabuzard: i  think u misunderstood me here. having fields is good. i just dont like the names u gave. sorry for being not precise enough.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
Zabuzard: Fair, saw this discussion just now (disregard my new comment on this topic please).

Still, use the logging system instead of `printStackTrace` instead, if applicable.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: What should the names of the constants be then?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
Zabuzard: Makes sense, thanks for explaining.

Zabuzard: Any reason why you need a singleton though - why not just a plain regular instance?

Zabuzard:
This seems to be based on #125 , I guess merging has to be blocked until former is merged first.

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: > Any reason why you need a singleton though - why not just a plain regular instance?

Everyone would need to pass the instance around and store it somewhere, possibly creating circular dependencies.

I-Al-Istannen: Why would there be cyclic references? You create the system in `main` and pass it around where needed?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
d3vel0per: Try something like this:

```suggestion
ctx.insertInto(Tags.TAGS, Tags.TAGS.ID, Tags.TAGS.CONTENT)
      .values(tag, content)
      .onDuplicateKeyUpdate()
      .set(Tags.TAGS.CONTENT, content)
      .execute();
```

application/src/main/java/org/togetherjava/tjbot/commands/CommandUtility.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/CommandUtility.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: Done by printing the exception to the logger

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: Done by printing the exception to the logger

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystemInstanceHolder.java
illuminator3: removed the singleton

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagsCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagManageCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
illuminator3: done

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
Zabuzard: Sorry but I do not get this. Why do you return `null` and _abuse a `Optional` just for its API here? That is kinda misleading and usually considered a nono.

If you consider it an error, throw an exception. If not, return an `Optional`. Prefer not using `null` as return value please.
And in particular, do not abuse `Optional` just for its "nice API". If you do not want to return an `Optional`, then use normal if-else please. Then again, why not just return `Optional` out of this method?

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Checks if a tag already exists in the database
+     *
+     * @param tag id tag to check
+     * @return true if it exists, false if not
+     * @author illuminator3
+     */
+    public boolean exists(String tag) {
+        return database.readTransaction(ctx -> {
+            return Optional
```
Zabuzard, line 32: Do not abuse `Optional` as if-else. Return `... != null` instead.

Zabuzard (changes requested):

tag-system/src/main/java/org/togetherjava/tjbot/tag/TagSystem.java
illuminator3: changed

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -51,11 +51,12 @@ public static MessageEmbed generateEmbed(String content, String requestor) {
...
      * @param isRaw if the tag should be displayed raw
      * @param componentId generated componentId based on the user id
+     * @throws IllegalArgumentException if the tag does not exist
      * @author illuminator3
      */
     public static void sendTag(MessageChannel channel, String tagId, String requestor,
-            TagSystem tagSystem, boolean isRaw, String componentId) {
-        String content = tagSystem.get(tagId);
+            TagSystem tagSystem, boolean isRaw, String componentId) throws IllegalArgumentException {
+        String content = tagSystem.get(tagId).orElseThrow(() -> new IllegalArgumentException("Tag doesn't exist"));
```
Zabuzard: Minor but useful improvement: Add the `tagId` to the exception message.

application/src/main/java/org/togetherjava/tjbot/commands/tag/TagUtility.java
```diff
@@ -74,12 +75,13 @@ public static void sendTag(MessageChannel channel, String tagId, String requesto
...
      * @param componentId generated componentId based on the user id
+     * @throws IllegalArgumentException if the tag does not exist
      * @author illuminator3
      */
     public static void replyTag(SlashCommandEvent event, String tagId, String requestor,
-            TagSystem tagSystem, boolean isRaw, String componentId) {
+            TagSystem tagSystem, boolean isRaw, String componentId) throws IllegalArgumentException {
         if (tagSystem.exists(tagId)) {
-            String content = tagSystem.get(tagId);
+            String content = tagSystem.get(tagId).orElseThrow(() -> new IllegalArgumentException("Tag doesn't exist"));
```
Zabuzard: Minor but useful improvement: Add the `tagId` to the exception message.

Zabuzard (changes requested):

Zabuzard:
You are failing some of the required checks, did you review whats going on there? :thinking:

Zabuzard (approved):

Zabuzard:
@illuminator3 Is the PR done, besides conflicts and task failures?

Zabuzard:
I am increasing the priority since the longer this is open, the harder it gets to merge. I might have a look later today, if I find time.

Zabuzard:
CI is complaining about the remaining `TODO`s but they should be dealed with in a different PR, as second iteration of the system.

Zabuzard:
Rebase on `develop`

Zabuzard:
**NotNull** found out that the `MarkdownSanitizer.escape(...)` does not correctly escape `//`. It leaves it as `//` instead of making `////`. We have to test whether the raw-command works with such input and whether its an actual issue in practice (thinking about code comments for example).

Zabuzard:
> 
> 
> **NotNull** found out that the `MarkdownSanitizer.escape(...)` does not correctly escape `\\`. It leaves it as `\\` instead of making `\\\\`. We have to test whether the raw-command works with such input and whether its an actual issue in practice (thinking about code comments for example).

Tested. It "fails" only if using the regular `create` and `edit` commands but works fine for `xxx-with-message`. Probably because discord supports in-line escaping already inside the slash command input.

In any case, since it works for `xxx-with-message`, this is not an issue for this, we can ignore the _problem_.

PR #138, opened 2021-09-20
Fix or mute sonarlint rule violations

I-Al-Istannen:
This is in preparation for #123 and fixes all current sonarlint violations *except* `TODO` which should likely be disabled in #123.


Zabuzard (approved):

marko-radosavljevic (approved):
It looks good. :relaxed: 

Maybe we should always add a comment with a link to the rule, and reasoning behind the decision to suppress it. To make maintenance easier.

I'm fine with this tho ^^

Issue #100, opened 2021-09-12
Config system


Zabuzard:
The bot should have a configuration file system instead of plain `main` arguments.

This would also greatly simplify the management of the two different bots (dev and master), by just giving them a different configuration setup (`test-bot = true`).

Initially, this config should contain the path to the DB, the bot token and a flag whether its the test-bot or not.

The community can decide on the data format for the config system, e.g.

- `Properties` 🎉 
- JSON 🚀 
- YAML 👀 
- Other (name in comments)

PR #142, opened 2021-09-22
Added config file support

Zabuzard:
### What is this?

Implements #100 , i.e. adding support for a configuration file. Replaces the previous `args` system.

### Technical details

The configuration file is using JSON (result of votes in #100 ), using **Jackson** to deserialize it.

`args` now expects either no or a single argument (the path to the config file), if none was given, it will assume `config.json` as path.

### Example

For testing it locally, create `TJ-Bot\application\config.json` (or if running from a jar, put it right next to the jar) with content like:
```json
{
    "token": "<your_token_here>",
    "databasePath": "db/database.db",
    "projectWebsite": "https://github.com/Together-Java/TJ-Bot",
    "discordGuildInvite": "https://discord.com/invite/XXFUXzK"
}
```
(the wiki will be updated with this information as soon as this is merged)

### Singleton?

I went for a **singleton** to make it easier to use `Config`. Personally, I would prefer passing it around explicitly though - but I am okay with making an exception for a configuration. If you prefer passing it around explicitly, please tell me, then I will refactor that aspect (be aware that the consequence is populating all your constructors and methods with possibly lots of config arguments then though).

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -5,36 +5,48 @@ import net.dv8tion.jda.api.JDABuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.togetherjava.tjbot.commands.CommandHandler;
+import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.db.Database;
 
 import javax.security.auth.login.LoginException;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.sql.SQLException;
 
-/***
+/**
  * Main class of the application. Use {@link #main(String[])} to start an instance of it.
  */
 public enum Application {
     ;
 
     private static final Logger logger = LoggerFactory.getLogger(Application.class);
+    private static final String DEFAULT_CONFIG_PATH = "config.json";
 
     /**
      * Starts the application.
      *
-     * @param args command line arguments - [the token of the bot, the path to the database]
+     * @param args command line arguments - [the path to the configuration file (optional, by
+     *        default "config.json")]
      */
     public static void main(final String[] args) {
-        if (args.length != 2) {
-            throw new IllegalArgumentException("Expected two arguments but " + args.length
-                    + " arguments were provided. The first argument must be the token of the bot"
-                    + " and the second the path to the database.");
+        if (args.length > 1) {
+            throw new IllegalArgumentException("Expected no or one argument but " + args.length
+                    + " arguments were provided. The first argument is the path to the configuration file. If no argument was provided, '"
+                    + DEFAULT_CONFIG_PATH + "' will be assumed.");
         }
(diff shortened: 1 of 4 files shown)


Zabuzard (commented):
@I-Al-Istannen Please prepare the VPS with the corresponding config files before this is merged, otherwise the bot crashes.

application/src/main/java/org/togetherjava/tjbot/Application.java
```diff
@@ -5,36 +5,48 @@
...
-     * @param args command line arguments - [the token of the bot, the path to the database]
+     * @param args command line arguments - [the path to the configuration file (optional, by
+     *        default "config.json")]
      */
     public static void main(final String[] args) {
-        if (args.length != 2) {
-            throw new IllegalArgumentException("Expected two arguments but " + args.length
-                    + " arguments were provided. The first argument must be the token of the bot"
-                    + " and the second the path to the database.");
+        if (args.length > 1) {
```
illuminator3, line 32: What if `args.length == 0`?

application/src/main/java/org/togetherjava/tjbot/Application.java
```diff
@@ -5,36 +5,48 @@
...
-            runBot(token, Path.of(databasePath));
+            Config.load(configPath);
+        } catch (IOException e) {
+            logger.error("Unable to load the configuration file from path '{}'",
+                    configPath.toAbsolutePath(), e);
+            return;
+        }
+
+        try {
+            var config = Config.getInstance();
```
illuminator3, line 48: I wouldn't use `var` here because the type is not obvious.

illuminator3 (changes requested):

application/src/main/java/org/togetherjava/tjbot/Application.java
Zabuzard: Then the code explicitly takes the default value (see the next lines of the code).

application/src/main/java/org/togetherjava/tjbot/Application.java
Zabuzard: Imo it is very obvious cause `Config.getInstance()`. I dont care though, I can change it.

application/src/main/java/org/togetherjava/tjbot/Application.java
illuminator3: Ah, yes. Change `arguments were` to `argument(s) was/were` though.

Zabuzard: Since it is inside the `if (args.length > 1)` it can only trigger for 2 or more arguments. Singular is not an option, hence the text seems to be correct english.

illuminator3 (approved):

I-Al-Istannen (approved):
I'd prefer passing it around but I can also live with the singleton here

marko-radosavljevic (approved):

PR #143, opened 2021-09-22
Add shutdown hook

Zabuzard:
### What is this?

Adding a shutdown hook to the code to properly react to shutdown of the system.

### Details

For now, we simply log a message to make reading the logs easier (in case of multiple starts and shutdowns). In the future it might trigger additional cleanup logic if needed 🤷 

log example

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -58,6 +58,8 @@ public enum Application {
                 .build();
             jda.awaitReady();
             logger.info("Bot is ready");
+
+            Runtime.getRuntime().addShutdownHook(new Thread(Application::onShutdown));
         } catch (LoginException e) {
             logger.error("Failed to login", e);
         } catch (InterruptedException e) {
@@ -66,5 +68,9 @@ public enum Application {
         } catch (SQLException e) {
             logger.error("Failed to create database", e);
         }
+    }
+
+    private static void onShutdown() {
+        logger.info("Bot has been stopped");
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -87,7 +87,7 @@ public enum Application {
     private static void onShutdown() {
         logger.info("Bot has been stopped");
     }
-  
+
     /**
      * Set's any System-properties before anything else is touched.
      */


application/src/main/java/org/togetherjava/tjbot/Application.java
```diff
@@ -67,4 +69,8 @@ public static void runBot(String token, Path databasePath) {
             logger.error("Failed to create database", e);
         }
     }
+
+    private static void onShutdown() {
```
illuminator3, line 73: I wouldn't make this static. Also, maybe close the database inside it?

illuminator3 (changes requested):

application/src/main/java/org/togetherjava/tjbot/Application.java
Zabuzard: It had to be static cause the `runBot` is also `static`. I agree to the database thing though, and this will basically lead to the class not being static anymore, so that issue is also solved 👍

Zabuzard: After some research, I figured out that the database is correctly shutting down its connections already automatically. See https://github.com/jOOQ/jOOQ/issues/10512

Zabuzard: We can close the currently anonymous `connection` though.

Zabuzard: Seems `jda.shutdown()` and `jda.shutdownNow()`  takes too long for a shutdown hook.

Zabuzard: I can not find any nice solution to this. It is getting more and more complex with no clean end in sight.

At this point, it is not worth the little benefit. I do not see anyone using `Application` outside of the context of `main` anyways. So we do not have to offer a graceful shutdown. So the forced kill from outside is everything we need. And resource-closing during a JVM shutdown neither makes sense (they are released due to the JVM process dying anyways) nor is there time for anything that takes longer than a few milliseconds.

Zabuzard: If you can agree to this, I would like to dismiss your change request.

marko-radosavljevic (approved):

application/src/main/java/org/togetherjava/tjbot/Application.java
Zabuzard: @illuminator3 ?

illuminator3 (dismissed):

illuminator3 (approved):

marko-radosavljevic (approved):

illuminator3 (approved):

PR #149, opened 2021-09-23
Bugfix for Docker-Environments, set Parallelism Flag

krankkkk:
@I-Al-Istannen found an _interesting_ Bug with the new JDK, when we are executing the Dev-Bot in the Docker-Container on the VPS it only has 1 Core and With 1 Core Java provides 0 Workers to its ForkJoinPool per default. JDA heavily relies on the ForkJoinPool and since the pool does no work anymore, JDA also never start's up fully and the Bot is semi-alive.

Now there is a check against this situation and if we would run into this behavior we set an System-Flag wich overrides the default behavior.


I-Al-Istannen (approved):

marko-radosavljevic (approved):
weird

Issue #103, opened 2021-09-12
[FEATURE] Add a way to access logs


Zabuzard:
We need a way to be able to read the logs of the two bots, without actually logging in to the VPS and reading the text files.

This could for example be a full business solution like Kibana or maybe also just another routine in the bot that sends all log messages to a dedicated Discord channel.

Also, it would be good to have a way to change the logging level hot (while the application is running) - or at least without logging in to the VPS and changing the config manually.

d3vel0per:
I see 3 options in my mind:
* Spring's ecosystem brings in support of the mentioned functionality and a lot more (one of the reasons I was more inclined to D4J). So we spend some time in integrating JDA with spring
* Use something like https://github.com/portainer/portainer which is more of a docker management UI but achieves our current goal and a few others like bringing down a bot if needed.
* Roll out our own implementation using a combination of Elastic-Logstash-Kibana.

Others can chip in

Tais993:
Or abuse the logback appender to send it to a channel :p

marko-radosavljevic:
@d3vel0per Any updates on this one?

Zabuzard:
I like channel logging and hot log level change per discord command.

Unfortunately, that does not help if the bot is failung during JDA setup (like it is now, after the Java 17 migration).

krankkkk:
The most simplistic implementation would be to store the logevents in the DB and query them on demand per  some command. That way you don't hit rate-Limits because you are constantly pushing to some Discord-Log-Channel and you also don't rely on full Business solutions.

d3vel0per:
> @d3vel0per Any updates on this one?

Not really, I was waiting for more ideas. I was evaluating a few cloud products but the group didn't have a consensus on those.
Streaming back to a channel sounds interesting with the mentioned already mentioned caveats (may be its not a part of the bot itself)
what @krankkkk suggested also sounds good but then we will have to build a lot of infra around it viz. archiving, well-structured logs, etc.
or we may host a full log analysis stack (but i think the setup itself will become more costly than the bot itself)

d3vel0per:
> @d3vel0per Any updates on this one?

Not really, I was waiting for more ideas. I was evaluating a few cloud products but the group didn't have a consensus on those.
Streaming back to a channel sounds interesting with the mentioned already mentioned caveats (may be its not a part of the bot itself)
what @krankkkk suggested also sounds good but then we will have to build a lot of infra around it viz. archiving, well-structured logs, etc.
[...]

PR #150, opened 2021-09-25
WIP: #103 Add a Way to see Logs

krankkkk:
In #103 there is a cry for a Way to access Log's i created a little Spring project + Vaadin-flow.

It provides an Discord-OAuth2 secured way of accessing the Log's in realtime, like `tail -f` would do. 
It's an pure Java project so everyone can contribute.

For now it's fully functioning with what's requested and I hope i can get some early feedback on the whole thing. 
The Code-Documentation isn't done (or started), but that comes after the features are all realized.

If you wan't to test it locally you need 2 things
1. A [Discord Application](https://discord.com/developers/applications) with the redirect URL: `http://localhost:8080/login/oauth2/code/github`
2. A `config.json` like this:

```json
{
  "clientName": "Discord",
  "clientId": "DISCORD_OAUTH2_CLIENTID",
  "clientSecret": "DISCORD_OAUTH2_SECRET",
  "rootUserName": "YOUR_DISCORD_USERNAME",
  "rootDiscordID": "YOUR_DISCORD_ID",
  "logPath": "application/logs/app.log"
}
```

Either put the `config.json` in the `logviewer` directory or anywhere you want and provide the Path to it as Start-Args


Zabuzard:
Can you give a quick mini explanation of what this is, how it works, maybe show a quick screenshot?

krankkkk:
Right probably should have done that, It's a Web-App that displays the log's and pushed changes in the log to each Browser.

For now it only watches the default `app.log` from the application, but also watching any additional log's would a pretty small change.
It uses Discord OAuth2 as Identity provider,  but uses it's own Role-Management and if you have permission to view can see this this:

Logs
(It's my local bot put in trace so event's come through quickly)

And the small User-Management:
User-Management

krankkkk:
I need some advice/input on how to handle the Docker Environment the Bot is currently in, there is no easy or straightforward way to handle reading the log's the bot is writing.

Either the Container of the Bot is accessible from the Container of the Logviewer, which would make it pointless to have 2 containers.
Or the Bot writes the Log outside of it's Container, which makes Standardization harder  and makes stuff like changing the loglevel also unnecessarily hard.
Best theoretical Way would be that both live in the same "space", but I did not find a way yet to start 2 apps from a single Container. 

The most practical choice would probably be to create an API on the Bot-side that transfers the Log's over, but I did not want to touch the Bot itself without some feedback

Zabuzard:
> 
> 
> Note: The following is only about the logs written to an actual File
> 
> I need some advice/input on how to handle the Docker Environment the Bot is currently in, there is no easy or straightforward way to handle reading the log's the bot is writing.
> 
> Either the Container of the Bot is accessible from the Container of the Logviewer, which would make it pointless to have 2 containers. Or the Bot writes the Log outside of it's Container, which makes Standardization harder and makes stuff like changing the loglevel also unnecessarily hard. Best theoretical Way would be that both live in the same "space", but I did not find a way yet to start 2 apps from a single Container.
> 
> The most practical choice would probably be to create an API on the Bot-side that transfers the Log's over, but I did not want to touch the Bot itself without some feedback

I am not 100% sure I get in which direction this is going. The way I see it, this project (as of today) basically consists of a spring-self-made log-viewer website application. It did not solve the problems of actually connecting to the logs of the bot (yet) - which is the critical main focus of this task.

Please correct me if I am wrong but I challenge the benefit of having a self-written application versus just using any existing industry standard solution, such as Kibana. Dont get me wrong, your application looks nice - but I feel like we are reinventing the wheel here in regards to the UI/application that actually processes the log.

The challenge and focus of this is primarily how to give users the possibility to actually connect to our logs in one way or another (i.e. creating something that can be viewed from outside but still access the logs from inside). For technical details on that, and to answer your questions on that topic, @I-Al-Istannen is probably your best contact.

I-Al-Istannen:
We could share a file by mounting the same volume in both containers. We could use an appender that just throws logs over TCP to the container, as the containers can easily share an internal network. Log4J provides quite a few appenders for this purpose (see [here](https://logging.apache.org/log4j/2.x/manual/appenders.html)), for example over HTTP. We could wrap that in an AsyncAppender and would probably be good to go.


I think the main benefit a self rolled version provides here is Discord O-Auth integration and maybe being a bit more lightweight. We don't really need more than seeing a few lines of output, I think.

This is definitely re-inventing the wheel while adding a few novel things on top, but that isn't necessarily a bad thing in this context, is it?

krankkkk:
I already have an HTTP appender that throws the current events to the logviewer, that's the whole reason for this project.

I'm more concerned about the Events which were written to file. While the viewer itself saves any event it sees, i thought it might be good to have a way to view the actual logfile too.

I'm gonna try a few things with the shared mointPoint for that

krankkkk:
I figured out a way with docker-compose, so this is open for reviews

application/src/main/resources/log4j2.xml
```diff
@@ -4,18 +4,27 @@
...
+                     filePattern="logs/app-%d{dd-MM-yyyy}.log.gz"
                      ignoreExceptions="false">
-            <PatternLayout pattern="%d{yyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+            <PatternLayout charset="UTF-8"
+                           pattern="%d{yyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
             <Policies>
                 <SizeBasedTriggeringPolicy size="10 MB"/>
             </Policies>
         </RollingFile>
+        <Http name="Http" url="http://${env:TJ_LOG_TARGET:-localhost}:8080/rest/api/logs">
```
Zabuzard, line 16: How does this thing behave if the bot is started locally and not in the context of the real container?

application/src/main/resources/log4j2.xml
```diff
@@ -4,18 +4,27 @@
...
+            <JsonLayout/>
+        </Http>
+        <Async name="AsyncHttp">
+            <AppenderRef ref="Http"/>
+        </Async>
     </Appenders>
     <Loggers>
-        <Root level="info">
+        <Root level="Debug">
```
Zabuzard, line 24: Defaulf level should stay `Info`. To change the level we should provide config files and a way to change it hot while it runs, for example with this project (if possible) - but that might be part of another PR.

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.9"
+services:
+  bot:
+    image: krankkkk/dev-tj:latest
```
Zabuzard, line 4: Why is your name here?

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    #image: togetherjava/tjbot:latest
+    volumes:
+      - "./application/config.json:/config.json"
+      - "./application/logs/:/logs"
+    links:
+      - logs
+    environment:
+      - TJ_LOG_TARGET=logs
+  logs:
+    image: krankkkk/dev-logviewer:latest
```
Zabuzard, line 14: Why is your name here?

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    #image: togetherjava/tjlogs:latest
+    ports:
+      - "8080:8080"
```
Zabuzard, line 17: `8080` might already be in use by another application, it is too commonly used. Prefer another port please.

logviewer/build.gradle
```diff
@@ -0,0 +1,67 @@
+plugins {
+    id 'com.google.cloud.tools.jib' version '3.1.4'
```
Zabuzard, line 2: Please refactor to use the same style used in the main project.

logviewer/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+    // https://mvnrepository.com/artifact/com.h2database/h2
+    runtimeOnly 'com.h2database:h2:1.4.200'
+
+    developmentOnly 'org.springframework.boot:spring-boot-starter-actuator:2.5.4'
+    developmentOnly 'org.springframework.boot:spring-boot-devtools:2.5.4'
+}
+
+tasks.jib.dependsOn(tasks.vaadinBuildFrontend)
+jib {
+    from.image = 'eclipse-temurin:17'
```
Zabuzard, line 46: Not 100% sure but did we define the version somewhere else in a more global context already? In that case, please re-use existing variables and dont write it again.

logviewer/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+    to {
+        image = 'togetherjava.duckdns.org:5001/togetherjava/tjlogs:' + System.getenv('BRANCH_NAME') ?: 'latest'
+        auth {
+            username = System.getenv('REGISTRY_USER') ?: ''
+            password = System.getenv('REGISTRY_PASSWORD') ?: ''
+        }
+    }
+    container{
+        setPorts(["8080"].asList())
```
Zabuzard, line 55: Port topic (see earlier)

logviewer/src/main/java/org/togetherjava/logwatcher/Application.java
```diff
@@ -0,0 +1,33 @@
...
+import com.vaadin.flow.theme.Theme;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
+import org.vaadin.artur.helpers.LaunchUtil;
+
+/**
+ * The entry point of the Spring Boot application.
+ */
+@SpringBootApplication
```
Zabuzard, line 16: Please rewrite the comment, it should explain what this application specifically is about and give a short intro to it.

logviewer/src/main/java/org/togetherjava/logwatcher/oauth/OAuth2LoginConfig.java
```diff
@@ -0,0 +1,93 @@
...
+        @Override
+        public void configure(WebSecurity web) throws Exception {
+            super.configure(web);
+            web.ignoring().antMatchers("/rest/api/**");
+        }
+
+        @Override
+        protected void configure(HttpSecurity http) throws Exception {
+            //@formatter:off
+            http.authorizeRequests().antMatchers("/rest/api/**").anonymous();
```
Zabuzard, line 80: Why did you turn off the formatter? How does it look with proper formatting?

logviewer/src/main/java/org/togetherjava/logwatcher/util/LogReader.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Read's the content of the given Logfile in the configured Logging path
+     *
+     * @param log Name of the Logfile
+     * @return The Content of the Log
+     */
+    public List<String> readLog(final Path log) {
+        try {
+            return Files.readAllLines(log, StandardCharsets.UTF_8);
```
Zabuzard, line 46: Charset UTF_8 is obsolete on NIO methods, leave it away.

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
```diff
@@ -0,0 +1,180 @@
...
+        nav.getElement().setAttribute("aria-labelledby", "views");
+
+        H3 views = new H3("Views");
+        views.addClassNames("flex", "h-m", "items-center", "mx-m", "my-0", "text-s",
+                "text-tertiary");
+        views.setId("views");
+
+        for (RouterLink link : createLinks()) {
+            nav.add(link);
+        }
```
Zabuzard, line 111: If you want, you can rewrite this as:
```suggestion
        createLinks().forEach(nav::add);
```

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
```diff
@@ -0,0 +1,180 @@
...
+            .map(MainLayout::createLink)
+            .toList();
+    }
+
+    private boolean checkAccess(MenuItemInfo menuItemInfo) {
+        final Class<? extends Component> view = menuItemInfo.view;
+        final AllowedRoles annotation = view.getAnnotation(AllowedRoles.class);
+
+        if (annotation == null) {
+            LogManager.getLogger(MainLayout.class)
```
Zabuzard, line 130: This seems to be a different logging system than the main application has. If possible, please use SLF4J as well to not confuse any readers and align the code bases with each other.

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
```diff
@@ -0,0 +1,180 @@
...
+            .orElse(false);
+    }
+
+    private Footer createFooter() {
+        Footer layout = new Footer();
+        layout.addClassNames("flex", "items-center", "my-s", "px-m", "py-xs");
+
+        Optional<User> maybeUser = authenticatedUser.get();
+        if (maybeUser.isPresent()) {
+            User user = maybeUser.get();
```
Zabuzard, line 148: Do not use `Optional#get` it will be deprecated soon. Instead:
```suggestion
            User user = maybeUser.orElseThrow();
```

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/LogsView.java
```diff
@@ -0,0 +1,125 @@
...
+        fillTextField(event.getValue());
+    }
+
+    /**
+     * Reload the Log and set it in the textField
+     *
+     * @param logFileName Name of the Logfile
+     */
+    private void fillTextField(final Path logFileName) {
+        this.text.setValue(String.join("\n", getLogEntries(logFileName)));
```
Zabuzard, line 92: Is `\n` the desired action here? Please rethink whether you might need `System.lineSeparator()` instead.

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    private void setInstantFormatter() {
+        final Grid<LogEvent> innerGrid = this.grid.getGrid();
+        final Optional<Grid.Column<LogEvent>> column =
+                Optional.ofNullable(innerGrid.getColumnByKey(LogEventsConstants.FIELD_INSTANT));
+        if (column.isEmpty())
+            return;
+
+        final Grid.Column<LogEvent> instant = column.get();
```
Zabuzard, line 90: Same as earlier:
```suggestion
        final Grid.Column<LogEvent> instant = column.orElseThrow();
```

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
```diff
@@ -0,0 +1,153 @@
...
+    private void removeHook() {
+        StreamWatcher.removeSubscription(this.uuid);
+    }
+
+    private void setInstantFormatter() {
+        final Grid<LogEvent> innerGrid = this.grid.getGrid();
+        final Optional<Grid.Column<LogEvent>> column =
+                Optional.ofNullable(innerGrid.getColumnByKey(LogEventsConstants.FIELD_INSTANT));
+        if (column.isEmpty())
+            return;
```
Zabuzard, line 88: Missing curly braces. I am surprised that this passes our checks and style settings.
```suggestion
        if (column.isEmpty()) {
            return;
        }
```

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
```diff
@@ -0,0 +1,153 @@
...
+        final EnhancedDialog dialog = new EnhancedDialog();
+        dialog.setHeader("Choose the Columns you want to see.");
+        final Set<String> columns = this.grid.getGrid()
+            .getColumns()
+            .stream()
+            .map(Grid.Column::getKey)
+            .collect(Collectors.toSet());
+
+        final List<Checkbox> checkBoxes = new ArrayList<>();
+        for (String s : Arrays.asList(LogEventsConstants.FIELD_INSTANT,
```
Zabuzard, line 119: Please rename `s` to something more meaningful.

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
```diff
@@ -0,0 +1,153 @@
...
+            .toArray(String[]::new);
+
+
+        this.grid.getGrid().setColumns(columns);
+        final boolean includeTime = checkboxes.stream()
+            .filter(AbstractField::getValue)
+            .anyMatch(c -> LogEventsConstants.FIELD_INSTANT.equals(c.getLabel()));
+
+        if (includeTime)
+            setInstantFormatter();
```
Zabuzard, line 146: Missing curlies:
```suggestion
        if (includeTime) {
            setInstantFormatter();
        }
```

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Removes the given User from the Repository
+     *
+     * @param user User to remove
+     */
+    private void onRemove(User user) {
+        try {
+            this.repo.delete(user);
+        } catch (RuntimeException e) {
```
Zabuzard, line 60: Is this actually on a top level entry point of the application? If not, you should not catch everything.

logviewer/src/main/resources/application.properties
```diff
@@ -0,0 +1,16 @@
+server.port=${PORT:8080}
```
Zabuzard, line 1: Port topic (see earlier).

Zabuzard (changes requested):
I am a bit concerned about the amount of new technology this introduces into the project. If any of them seems to be cheap to replace with what we already have, please prefer that. (I will not block meaningful additions though - just trying to align our tech a bit).

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
Zabuzard: It seems our current style does not force it and unfortunately there is no easy way to enable it. I would still propose to change it though.

application/src/main/resources/log4j2.xml
krankkkk: The second Part `:-localhost` is an default value, so if the ENV-Variable is not set, it tries to talk to `localhost`

docker-compose.yml
krankkkk: Slipped through, changed back to the `togetherjava`

docker-compose.yml
krankkkk: Slipped through, changed back to the `togetherjava`

application/src/main/resources/log4j2.xml
krankkkk: Changed Back

logviewer/build.gradle
krankkkk: It has not been globally defined yet

logviewer/build.gradle
krankkkk: I don't see any difference to the buildscript in application or the parent script

docker-compose.yml
krankkkk: Refractoring to `5050` then

logviewer/build.gradle
Zabuzard: Right, sorry for that. Perfect.

logviewer/src/main/java/org/togetherjava/logwatcher/oauth/OAuth2LoginConfig.java
krankkkk: turned it back on, while testing the lines were way longer and the formatter mad it unreadable

logviewer/src/main/java/org/togetherjava/logwatcher/util/LogReader.java
krankkkk: Removed it

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
krankkkk: Changed it to this

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
krankkkk: Haven't seen anything about depreciating Optional#get yet, but changed it

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
krankkkk: Changed it

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
krankkkk: Changed it

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
krankkkk: changed it

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
krankkkk: Changed it to the more specific `IllegalArgumentException`

logviewer/src/main/java/org/togetherjava/logwatcher/views/logs/StreamedView.java
krankkkk: changed it to `field`

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
krankkkk: It still uses SLF4J, I just talked to the Log4j2 API
Refractored it to SLF4J directly

logviewer/src/main/java/org/togetherjava/logwatcher/Application.java
krankkkk: Rewrote it

krankkkk:
> I am a bit concerned about the amount of new technology this introduces into the project. If any of them seems to be cheap to replace with what we already have, please prefer that. (I will not block meaningful additions though - just trying to align our tech a bit).

Aside from the Spring specifc Stuf, the only Difference I see to the Bot, would be the DB. I used H2 and the Bot uses SQLite. I could change that to match. Everything else matches, from logging, to JSON-provider or the Testing-Framwork is the same as from the Bot

I-Al-Istannen:
> Edit: Seem's like SQLite is unable to use foreign-Keys in the default Settings, Should i Still switch over?

The `Database` class enables that stuff for you automatically, if you use it. H2 is also not really designed to be a resilient database for production use, I corrupted quite a few of them by shutting down my application...

If you want to use SQLite you can't use Hibernate anyways, as it didn't really have any stable SQLite support last I checked. So you can migrate it to JOOQ at the same time and use the "database" module :)

krankkkk:
> > Edit: Seem's like SQLite is unable to use foreign-Keys in the default Settings, Should i Still switch over?
> 
> The `Database` class enables that stuff for you automatically, if you use it. H2 is also not really designed to be a resilient database for production use, I corrupted quite a few of them by shutting down my application...
> 
> If you want to use SQLite you can't use Hibernate anyways, as it didn't really have any stable SQLite support last I checked. So you can migrate it to JOOQ at the same time and use the "database" module :)

The thing is, I never actually "talk" to the Database directly, I hide the the Complexity completely behind JPA.
So I could either
A) Add SQLite with an custom Dialect, this way only the DB would change
or
B) Switch to another, more resilient DB like postgres
or
C) I could Remove JPA and Hibernate completely so i can swap over to jooq+SQLite and reimplement everything JPA did for me
(This could mean ALOT of boilerplate Code, since Spring internally uses JPA, same as the Security module)

Zabuzard:
> 
> 
> > > Edit: Seem's like SQLite is unable to use foreign-Keys in the default Settings, Should i Still switch over?
> > 
> > 
> > The `Database` class enables that stuff for you automatically, if you use it. H2 is also not really designed to be a resilient database for production use, I corrupted quite a few of them by shutting down my application...
> > If you want to use SQLite you can't use Hibernate anyways, as it didn't really have any stable SQLite support last I checked. So you can migrate it to JOOQ at the same time and use the "database" module :)
> 
> The thing is, I never actually "talk" to the Database directly, I hide the the Complexity completely behind JPA. So I could either A) Add SQLite with an custom Dialect, this way only the DB would change or B) Switch to another, more resilient DB like postgres or C) I could Remove JPA and Hibernate completely so i can swap over to jooq+SQLite and reimplement everything JPA did for me (This could mean ALOT of boilerplate Code, since Spring internally uses JPA, same as the Security module)

Yeah the bummer here is Spring. I would say stick with your choice as that makes the most sense with Spring. I guess I just was not prepared to introduce big-boy-Spring and all its consequences yet. But it is fair.

Zabuzard:
Could you not force-push and rewrite history during development? This makes it hard to review your code since I can not see the incremental changes since my last review.

I appreciate a clean history but usually people do that when ready for merging and not during reviews, thanks.

Zabuzard (approved):

I-Al-Istannen:
I mean, you do not have to use JPA with spring. You can just exclude Hibernate and do your own database stuff? I do that for a Spring project of mine. I am not comfortable with storing Data in H2 to be honest, it corrupted too many of my databases.

I can give that a go if you want.

krankkkk:
Removed JPA and Hibernate Completely.

That was an Pain in the a...., jooq does not play nicely with spring.

logviewer/src/main/resources/application.properties
```diff
@@ -0,0 +1,10 @@
+server.port=${PORT:5050}
+logging.level.org.atmosphere=warn
+logging.level.root=info
+logging.file.path=./logviewer/logs
+spring.mustache.check-template-location=false
+
+# To improve the performance during development.
+# For more information https://vaadin.com/docs/flow/spring/tutorial-spring-configuration.html#special-configuration-parameters
+# vaadin.whitelisted-packages= org/vaadin/example
+
```
Zabuzard, line 10: As a note to this config file, in the main application module we decided (per poll in GH Issues) to use JSON config files that are deserialized into POJOs using Jackson.

logviewer/src/main/java/org/togetherjava/logwatcher/watcher/StreamWatcher.java
```diff
@@ -0,0 +1,52 @@
+package org.togetherjava.logwatcher.watcher;
+
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class StreamWatcher {
+    private static final Map<UUID, Runnable> consumerMap = new ConcurrentHashMap<>(3);
```
Zabuzard, line 10: Why `3`? Use a constant with a good name instead.

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
```diff
@@ -0,0 +1,200 @@
...
+ * The User-Management View in the Browser
+ */
+
+@PageTitle("UserManagement")
+@Route(value = "user", layout = MainLayout.class)
+@AllowedRoles(roles = {Role.ADMIN})
+@PermitAll
+public class UserManagement extends VerticalLayout {
+
+    private final transient IUserRepository repo;
```
Zabuzard, line 43: No `I`, `A`, `E` prefixes please (to align with the rest of the code base).

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
```diff
@@ -0,0 +1,200 @@
...
+        add(grid);
+
+
+        GridContextMenu<UserDTO> menu = grid.addContextMenu();
+        menu.addItem("Add", this::onAdd);
+        menu.addItem("Edit", this::onEdit);
+        menu.addItem("Delete", this::onDelete);
+
+        grid.setColumns("discordID", "userName", "roles");
+        grid.setItems(DataProvider.fromCallbacks(this::onAll, query -> (int) onAll(query).count()));
```
Zabuzard, line 57: Is this `int` case safe? What if its too big? At least please add a comment explaining why this cant be the case. Otherwise the next reader will be confused about this.

logviewer/src/main/java/org/togetherjava/logwatcher/Application.java
```diff
@@ -0,0 +1,43 @@
...
+@PWA(name = "LogViewer", shortName = "Logs", offlineResources = {"images/logo.png"})
+@NpmPackage(value = "line-awesome", version = "1.3.0")
+@Push
+public class Application extends SpringBootServletInitializer implements AppShellConfigurator {
+
+    public static void main(String[] args) {
+        String pathToConfig = "./logviewer/config.json";
+        if (args.length > 0) {
+            pathToConfig = String.join(" ", args).trim();
+        }
```
Zabuzard, line 37: Prefer to just accept one optional args argument, to make it possible to give a different config path; instead of multiple that are joined. (Alignment with main module, might also be interesting for our containers).

logviewer/src/main/java/org/togetherjava/logwatcher/Application.java
```diff
@@ -0,0 +1,43 @@
...
+
+        System.setProperty("TJ_CONFIG_PATH", pathToConfig);
```
Zabuzard, line 39: Not sure where you use this but if it is possible to not use custom system configs, I would highly prefer it.

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.sql.SQLException;
+import java.util.concurrent.atomic.AtomicReference;
+
+@Configuration
+public class DatabaseProvider {
+
+    private static final AtomicReference<Database> AtomDb = new AtomicReference<>();
```
Zabuzard, line 18: Name conflicts style:
```suggestion
    private static final AtomicReference<Database> ATOM_DB = new AtomicReference<>();
```

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public DatabaseProvider() {
+        synchronized (AtomDb) {
```
Zabuzard, line 21: `synchronized` should ideally not be used on regular objects but instead on dedicated lock objects. Prefer creating a dedicated `new Object()` only for this purpose (private static final object lock idiom)

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,66 @@
...
+            return new Database("jdbc:sqlite:%s".formatted(dbPath.toAbsolutePath()));
+        } catch (final SQLException e) {
+            LoggerFactory.getLogger(DatabaseProvider.class)
+                .error("Exception while creating Database.", e);
+            throw new FatalBeanException("Could not create Database.", e);
+        }
+    }
+
+    private Path getDBPath() {
+        final Path dbPath = Path.of("./logviewer/db/db.db");
```
Zabuzard, line 45: this should use the configuration instead

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+        try {
+            if (Files.notExists(dbPath.getParent())) {
```
Zabuzard, line 48: that `if` is obsolete, isnt it? IIRC `createDirectories` only does something if it doesnt exist, no?

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,66 @@
...
+            LoggerFactory.getLogger(DatabaseProvider.class)
+                .error("Exception while creating Database-Path.", e);
+        }
+
+        return dbPath;
+    }
+
+
+    @Bean
+    public Database getDB() {
```
Zabuzard, line 61: Rename to `Db` instead of `DB`
```suggestion
    public Database getDb() {
```

logviewer/src/main/java/org/togetherjava/logwatcher/accesscontrol/Role.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static Role forID(final int id) {
+        return Arrays.stream(values())
+            .filter(r -> r.id == id)
+            .findAny()
+            .orElseThrow(() -> new IllegalArgumentException("Unknown RoleID: %d".formatted(id)));
+    }
+
+    public static Set<Role> getDisplayableRoles() {
+        return Set.of(USER, ADMIN);
```
Zabuzard, line 49: Use `EnumSet` instead (is faster and more compact in memory)
```suggestion
        return EnumSet.of(USER, ADMIN);
```

logviewer/src/main/java/org/togetherjava/logwatcher/entities/LogDTO.java
```diff
@@ -0,0 +1,131 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.time.Instant;
+
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class LogDTO {
```
Zabuzard, line 11: Personally, I do not like suffixing `DTO` or similar. If you can agree to that and its appicable, maybe rename all of them to just `Log`, `User`, ... (also alignment with main application module)

I do not have a strong opinion here though.

logviewer/src/main/java/org/togetherjava/logwatcher/logs/ILogRepository.java
```diff
@@ -0,0 +1,22 @@
+package org.togetherjava.logwatcher.logs;
+
+import org.togetherjava.tjbot.db.generated.tables.pojos.Logevents;
+
+import java.util.List;
+
+public interface ILogRepository {
```
Zabuzard, line 7: No `I` prefix (see earlier)

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
```diff
@@ -0,0 +1,108 @@
...
+import org.togetherjava.tjbot.db.generated.tables.records.UsersRecord;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static org.togetherjava.tjbot.db.generated.tables.Users.USERS;
+
+@Component
+@SuppressWarnings("java:S1602") // Curly Braces are necessary here
```
Zabuzard, line 18: Why, where? Is it really necessary everywhere in the class? It doesnt seem like to me.

Please check if you can just move the suppression to the actual statements instead of the whole class.

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
```diff
@@ -0,0 +1,172 @@
...
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * The main view is a top-level placeholder for other views.
+ */
+@PageTitle("Main")
+@SuppressWarnings({"java:S1192"})
```
Zabuzard, line 30: Whats this warning, why did u suppress it?

Zabuzard (changes requested):

logviewer/src/main/resources/application.properties
krankkkk: Those are the properties for Spring, that are not supposed to be configurable for an user of the viewer

krankkkk: think log4j2.xml

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
krankkkk: I mean I could also write it 9 times (each time I talk to the DB), but then I would just rather set the Ignore on the whole class

logviewer/src/main/java/org/togetherjava/logwatcher/views/MainLayout.java
krankkkk: That's the Comment below it

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
krankkkk: DataProvider expect's an Integer, and I don't think were are gonna display more an 2 Billion Log-Entries at once, but I add an comment

logviewer/src/main/java/org/togetherjava/logwatcher/watcher/StreamWatcher.java
krankkkk: 3 is just to have an small initial Size, since I don't think more than 3 people will watch the logs at once

logviewer/src/main/java/org/togetherjava/logwatcher/Application.java
krankkkk: Just wanted to catch Cases where Spaces where in the Path, but I change it

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
krankkkk: Not in this instance, since i need to make sure I have the Path in the Docker-Compose script, since the DB should survive an update of the Image

logviewer/src/main/resources/application.properties
Zabuzard: They are configurable for the guy setting up the application though. That is what I mean and that is also what I was referring to. If its not possible to change that easily, we have to stick with it. But if it is, we should align it.

logviewer/src/main/java/org/togetherjava/logwatcher/watcher/StreamWatcher.java
Zabuzard: Fair. `EXPECTED_CONCURRENT_LOG_WATCHERS = 3` might be helpful then.

logviewer/src/main/java/org/togetherjava/logwatcher/views/usermanagement/UserManagement.java
Zabuzard: Makes sense. A comment saying exactly what you just said will be helpful in that case, imo

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
Zabuzard: Then lets talk about those 8 cases. Why is it needed there? I do not see it. They are single statements, a curly is obsolete.

Zabuzard (approved):

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
krankkkk: Because of the Design of the Database-class it's an ambiguous match:
[Invalid](https://imgur.com/EPAT4s1)

Zabuzard: Interesting. Cant you add some cast to it to make it clear?

(lemme loop in @I-Al-Istannen for this, who designed the DB class)

Zabuzard (dismissed):

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
krankkkk: Yes cast's are possible, but readability suffers greatly from them imo, see:
[Casts](https://imgur.com/5fGVr8C.jpg)

I-Al-Istannen: Without the braces java can not correctly infer that you actually want to return nothing there and it picks up the function variant of the method as another possible match. Sonarlint isn't smart enough to figure out that the braces are actually needed here and complains anyways.

I probably said this earlier but the only real solution (besides fixing Sonar) would be to rename one of the methods in the database class, which should probably be alright to do. We can do that after the current PRs are merged.

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
I-Al-Istannen: Then use the configuration or command line and provide both in the compose script? That way you only have the path in one place.

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
krankkkk: > that you actually want to return nothing there

But I do return something

I-Al-Istannen: That's the same problem but in reverse, the statement still applies. If you are interested you can look at JLS §15.27.3 I think. A Method invocation is an `Expression Statement` and its value might be discarded. Following the rules in the mentioned JSL paragraph, the invocation could either be an expression that returns its value (matching the Function method) or an expression statement that discards its value (matching the consumer method).

Java can not tell the difference and so you are forced to disambiguate. Adding the block solves the issue as you then have a void-compatible block that is not value-compatible and therefore the compiler can figure out what you wanted. That Sonar doesn't perform the same analysis is a bug.

My previous comment still applies in its entirety.

logviewer/frontend/themes/myapp/views/hello-world-view.css
```diff
@@ -0,0 +1,4 @@
+.hello-world-view {
```
I-Al-Istannen, line 1: What's a "hello-wold-view"?

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import java.sql.SQLException;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+@Configuration
+public class DatabaseProvider {
+
+    private static final AtomicReference<Database> ATOM_DB = new AtomicReference<>();
+    private static final Lock latch = new ReentrantLock();
```
I-Al-Istannen, line 21: Why are those two variables static? Can't you just always create an instance in the constructor?
(And potentially mark the DatabaseProvider as a singleton so Spring is happy?)

logviewer/src/main/java/org/togetherjava/logwatcher/entities/InstantWrapper.java
```diff
@@ -0,0 +1,68 @@
+package org.togetherjava.logwatcher.entities;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.util.Date;
+
+public class InstantWrapper {
```
I-Al-Istannen, line 10: What is this for?

database/src/main/resources/db/V2__Logviewer.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE logevents
```
I-Al-Istannen, line 1: You currently added this to the same database the main application uses. Does the main application need this table and does this application need the users table? It might make more sense to separate them?

logviewer/frontend/themes/myapp/views/hello-world-view.css
krankkkk: Psst, no-one looks at the .css :P

logviewer/src/main/java/org/togetherjava/logwatcher/entities/InstantWrapper.java
krankkkk: When an Event comes from the Bot, I can't directly parse an Instant from the JSON. 
So I added an little Record Class that can hold the values for the Instant and convert it after it arrived

database/src/main/resources/db/V2__Logviewer.sql
krankkkk: I saw that the configuration from the Bot-DB was also in the Database project, so I thought it's supposed to be there.

The Bot doesn't need anything from the viewer but I don't know where I would separate them without also breaking the code-generation

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
krankkkk: I only looked at `@Singleton` Annotation and found nothing that would belong to Spring.

It's now marked with the appropriate Scope

logviewer/src/main/java/org/togetherjava/logwatcher/entities/InstantWrapper.java
I-Al-Istannen: I see. If you feel like it you can add a small javadoc comment but it's not really important.

database/src/main/resources/db/V2__Logviewer.sql
I-Al-Istannen: Yea, that's a bit of an oddity as a few modules share their database apparently (at least that seems to be the intention of whoever extracted the database) but others do not need to. I will have a look at splitting off the viewer db.

Zabuzard (approved):

database/src/main/resources/db/V2__Logviewer.sql
I-Al-Istannen: I have created https://github.com/Together-Java/TJ-Bot/commit/46cb59ed4e3f5d20067fb6d381c9f26c8307a491 for this. If this works for you and you think it's fine, feel free to cherry-pick it into this PR.

Zabuzard (approved):

logviewer/src/main/resources/application.properties
krankkkk: While it is possible to convert some of these to an json-based Property File, I think that would be a bad choice for these.
These are not supposed to be changed even by someone setting up the application (Except the logging stuff).

I.e. changing the Port here, would result that you need to also change it in the logging-xml of the bot, in the Discord-OAuth config and in the config.json were the configurable Properties are living in.

The mustache property disables an error/warning message from being displayed. 

Converting all (future) Properties to something Bean-Based is not something that makes sense.

All Properties that are supposed to be configurable for someone that is setting up the Viewer are already in the `config.json`

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
krankkkk: Can this be marked as resolved? The Curly Braces and the Suppression are necessary here.

database/src/main/resources/db/V2__Logviewer.sql
krankkkk: Work's for me 👍🏽

logviewer/src/main/java/org/togetherjava/logwatcher/users/UserRepositoryImpl.java
I-Al-Istannen: Yea, let's fix this in a future PR.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: "3.9"
```
I-Al-Istannen, line 1: I am not sure this should be in the root as it can not be used to start our programs. Maybe this is better suited for documentation or should be renamed to "example" or something?

The images do not exist in docker hub and your compose file will therefore fail. Doesn't block the merge for me but I'll probably change it later on.

I-Al-Istannen (approved):

application/src/main/resources/log4j2.xml
```diff
@@ -4,18 +4,27 @@
...
+            <PatternLayout charset="UTF-8"
+                           pattern="%d{yyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
             <Policies>
                 <SizeBasedTriggeringPolicy size="10 MB"/>
             </Policies>
         </RollingFile>
+        <Http name="Http" url="http://${env:TJ_LOG_TARGET:-localhost}:5050/rest/api/logs">
+            <JsonLayout/>
+        </Http>
+        <Async name="AsyncHttp">
```
d3vel0per, line 19: Just wondering how an http appender can be made async unlike disk IO you can flush data in a batch it still has to be over multiple http calls. If it doesnt work as expected the cost of accumulated http calls can slow down the operations of the bot. Also what happens if the http endpoint becomes unavailable?

krankkkk: If the Endpoint isn't available, an error get's logger to console but not to file.

@I-Al-Istannen wanted the Async Wrapper around the HTTP Appender

logviewer/src/main/java/org/togetherjava/logwatcher/DatabaseProvider.java
krankkkk: Implemented it on the Java/Json Side, but found no way for standardization to include the json-properties in any compose-scripts

logviewer/src/main/java/org/togetherjava/logwatcher/config/Config.java
```diff
@@ -0,0 +1,122 @@
...
+        this.rootDiscordID = jsonNode.get("rootDiscordID").asText();
+        this.logPath = jsonNode.get("logPath").asText();
+        this.redirectPath = jsonNode.get("redirectPath").asText();
+        this.databasePath = jsonNode.get("databasePath").asText();
+    }
+
+    private JsonNode getJsonNode() {
+        final String pathToConfig = Objects.requireNonNull(CONFIG_PATH.get(), "Path not Set");
+        try {
+            return new ObjectMapper().readTree(Path.of(pathToConfig).toFile());
```
Zabuzard, line 85: Prefer having the `ObjectMapper` pre-created, maybe as a field.

Zabuzard (changes requested):

application/src/main/resources/log4j2.xml
I-Al-Istannen: https://logging.apache.org/log4j/2.x/manual/appenders.html
It just spins up a new thread as directly logging to HTTP would make log calls very expensive. The Async appender allows us to specify what should happen when the queue fills up - block or discard messages.

logviewer/src/main/java/org/togetherjava/logwatcher/config/Config.java
krankkkk: Or Autoinjected by Spring? Would that also be fine?

Zabuzard: Did you use dependency injection elsewhere in the program already? If you did not use it much yet, I would prefer to not introduce it just for this mini occasion to keep complexity and _black magic_ low for a beginner.

If you already made heavy usage of it anyways, I am completely fine with having it again. I am the last who blocks people writing their code the way they love in a community project like this.

krankkkk: I pretty much used it in in every Class

Zabuzard:
> 
> 
> Can this be merged? Are there any Issues remaining?
> 
> If not I'm going to merge dev into this branch and resolve any merge conflicts

Please check sonarcloud to not introduce any code flaws. You also need a second review. After thats done, please align with @I-Al-Istannen on whatever needs to be done on the VPS then or whatever is left to do to get this live.

Afterwards, if possible, please create some mini tutorial on the wiki to teach us how to use and maintain it.

I guess thats it, thank you very much for this contribution! ❤️

Zabuzard:
Appears that only staff can click the button.

@krankkkk Your commit history is a bit dirty (merge instead of rebase; a lot of just "Implement feedback" instead of meaningful descriptions).

- Either you spend some time rewriting your commit history, making it "nice"
- or I just squash it

Pick one :)

PR #119, opened 2021-09-15
Fixed command system issues

Tais993:
Moved AbstractCommand to the Commands package instead of the example package, this was an accident.
Improved docs:
- Added docs to the AbstractCommand constructor
- Added multiple @see tags
- Elaborated on the effect of Command#isGuildOnly

And optimized imports

PR #169, opened 2021-10-08
Multithreading for command system

Zabuzard:
Long running commands must not block other commands or processes. To ensure that, commands and guild updates are now executed multithreaded.

Port from the stale #119 which did not survive the massive refactoring introduced in `develop`.


Zabuzard:
Rebased on `develop` (merge conflicts). Did not change any code. So reviews are not actually stale.

PR #176, opened 2021-10-11
Bot does not create necessary Directories for the DB

krankkkk:
Currently the main Bot does not create any necessary Directories to the Path specified in the  `config.json` resulting in not being able to start with just following the wiki.

This 3 Line change fixes this.


Zabuzard:
> 
> 
> Should I extract it into a method and add a unit test?

There is no need to test that the built-in method works. And testing File IO can only be done properly if you mock the filesystem (we should not do real file io in unit tests).

Id say its not worth the hassle.

PR #171, opened 2021-10-08
Migrate TeX command

java-coding-prodigy:
Partially implements #128 

Adds  the `/tex` command.

PR #197, opened 2021-10-16
Add external component ID storage

### LINKED ISSUE #155: Add string-external component ID storage
### What is this?

The component ID system that will be introduced by #154 (see `ComponentIDs`, used by `SlashCommandAdapter` and the fairly complex documentation in `SlashCommand`) is quite hard to use for users and has hard and inpractical limitations (mostly the 100 character limit). Also, our Jackson CSV representation seems to mess up some stuff like `;`.

In short, our system tries to store information relevant to a button/selection-menu event inside the component ID (a plain string that is send to JDA), but that has a lot of limitations.

It would be better to store the actual information inside the application (for example a `Map` or the `Database`) and send around an identifier in the component ID instead, for example a `UUID`-string.

### Overview

**The advantages are clear**, the UUID-string itself will never go beyond 100 characters and is easy to handle for everyone. We can store as much and complex information as we want, since it will be held within the application only and not send over JDA and the wire.

**The disadvantages are small** and easy to counter: If the information is stored in-memory, it will be lost upon restart. So buttons that have been created on older instances are stale and the bot can not work with the ID anymore. An easy way to counter that is to (optionally?) additionally offer a persisted memory such as the database to commands.

### User impact

For the end-user, the API will likely be very similar and not change much at all. Unfortunately, we also will not get rid of the problems with the old API - namely that users could just bypass the system, provide their own component ID and break the systems internal routing by that.

There is no way to make this "type-safe" at "compile-time", unfortunately.

Zabuzard:
### Motivation

Component IDs (as used by buttons and selection menus) have a maximal length of 100 characters and hence are not able to store arbitrary information. Component IDs are, among other things, used to carry information through an event.

For example the user who triggered a slash command or details to the query, which are then stored _"inside"_ a button to retrieve the associated information back when it is clicked. By that, it can be implemented for example that a button can only be clicked by the original message author.

However, it is sometimes required to store more than just 100 characters. For example #171 already hits this limit hard and is basically blocked by this.

### Overview

This is a bigger system internal change (no user API changes) which addresses theses issues. Closes #155 .

It does so by storing the actual payload outside of the component ID used by the buttons, in a database table (and in-memory map), associating it to a generated **UUID**, which is then used as actual component ID for JDA.

The user interface does not change, existing commands do not need to be edited. I.e. they can still just call `generateComponentId(...)` and it will use the new mechanisms internally instead.

### `ComponentIdStore`

The central point of this is a new class called `ComponentIdStore`, which is exposed via the interfaces:
* `ComponentIdGenerator` - to all commands during setup for id generation
* `ComponentIdParser` - to `CommandSystem` for the routing of events

The store is basically a 2-layer `Map<UUID, ComponentId>`:
* first layer: `Cache` (by Caffeine), to speedup lookups, covers probably about 95% of all queries or so
* second layer: a database table `component_ids`

When an user wants to create a component ID, the store will add the payload to both layers and associate it to a generated UUID. When an user wants to parse back the payload from the UUID, the store looks up the in-memory map first, and if not found, also the database.

#### Eviction

To prevent the database from just growing indefinitely over the years, the `ComponentIdStore` implements a LRU-mechanism on both, the in-memory map, as well as on the database table.

For latter, it runs a periodic **eviction-routine**, which will locate records that have not been used for a long time and delete them.

Each lookup of an UUID **heats** the record in both, the in-memory map and in the database table. That means, its `last_used` timestamp will be updated, making it not targeted for the next evictions.

Users are able to listen to eviction events, by registering themselves as listener using `ComponentIdStore#addComponentIdRemovedListener`. While not used as of today, this might be interesting in the future, for example to deactivate actions (such as buttons) associated to an expired component ID.

Component IDs can also be associated with a `Lifespan.PERMANENT` to prevent eviction all together. While this should not be used per default, it can be useful for actions that are not used often but should still be kept alive (for example global role-assignment reactions).

Eviction details can be configured, but by default they are set to:
- evict every **15 minutes**
- evict entries that have not been used for longer than **20 days**
- in-memory map has a max size of **1_000**

### Database details

The table is created as such:
```sql
CREATE TABLE component_ids
(
    uuid         TEXT      NOT NULL UNIQUE PRIMARY KEY,
    component_id TEXT      NOT NULL,
    last_used    TIMESTAMP NOT NULL,
    lifespan     TEXT      NOT NULL
)
```
Content looks for example like this (after popping `/reload` three times):
table records

There was an issue with **SQLite** and **jOOQ** who assumed `LocalDateTime` as corresponding equivalent to `TIMESTAMP`, which is bad. This PR changes the configuration to force `Instant` as type for `TIMESTAMP` (note that there is also `TIMESTAMP WITH ZONE`, which is associated to `OffsetDateTime`, which is good). This will apply to all users of our database in the project (as of now, this affects **logviewer** as well).

### User API

**The user facing API did not change**, but let me give some examples nonetheless. There are two options:
- extend `SlashCommandAdapter` (and benefit)
- implement `SlashCommand`

#### `SlashCommandAdapter`

The abstract class already offers a `generateComponentId(...)` method that can just be called as it is. For example:
```java
.addActionRow(
  Button.of(ButtonStyle.SUCCESS, generateComponentId(member.getId()), "Yes"),
  Button.of(ButtonStyle.DANGER, generateComponentId(member.getId()), "No")
)
```

#### `SlashCommand`

Using the interface directly, one has to provide additional details (which would otherwise be auto-provided by `SlashCommandAdapter`) and capture the `ComponentIdGenerator` from the newly added method
```java
void acceptComponentIdGenerator(@NotNull ComponentIdGenerator generator);
```
which is called by the system during setup automatically, giving a command the option to capture a generator for later usage, e.g.:
```java
@Override
public void acceptComponentIdGenerator(@NotNull ComponentIdGenerator generator) {
    this.generator = generator;
}
```

The generator can then be used easily, for example:
```java
.addActionRow(
  Button.of(
    ButtonStyle.SUCCESS,
    generator.generate(
      new ComponentId(getName(), Arrays.asList(member.getId()),
      Lifespan.REGULAR),
    "Yes"),
  Button.of(
    ButtonStyle.DANGER,
    generator.generate(
      new ComponentId(getName(), Arrays.asList(member.getId()),
      Lifespan.REGULAR),
    "No")
)
```

### Checklist

- [x] add component ID storage with in-memory map
- [x] add database layer
- [x] create eviction routine and heating
- [x] polish the API
- [x] javadoc
- [x] linter issues
- [x] fix logviewer to also use `Instant` after the jOOQ change
- [x] ensure proper thread safety for the ID store
- [x] migrate LRU in-memory cache to **Caffeine**
- [x] unit test adapter
~~- [ ] unit test store~~


application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
```diff
@@ -122,34 +124,23 @@
...
-     * An example would be {@code 13,ping,"hello,world"}.
-     * <p>
-     * Any provided extra arguments can then be picked up during the corresponding event (see
+     * The component ID has to be a UUID-string (see {@link java.util.UUID}), which is associated to
+     * a specific database entry, containing meta information about the command being executed. Such
+     * a database entry can be created and a UUID be obtained by using
+     * {@link ComponentIdGenerator#generate(ComponentId, Lifespan)}, as provided by the instance
+     * given to {@link #acceptComponentIdGenerator(ComponentIdGenerator)} during system setup. The
+     * required {@link ComponentId} instance accepts optional extra arguments, which, if provided,
+     * can be picked up during the corresponding event (see
```
marko-radosavljevic, line 139: Since this is a central part of the documentation, and everything else points to it, It's maybe worth expending on it a bit.

The whole system is pretty abstract and complex, with lots of advanced java concepts. So I'm thinking, how can we make it more digestible to a newcomer?

I think a short usage example of the system would help a lot. Also, there is a ton of good information written in this PR, but it's not in the code. So I guess the best way to go about it would be to add a wiki page (slightly modified PR description) that is referenced in the code.

application/src/test/java/org/togetherjava/tjbot/commands/SlashCommandAdapterTest.java
```diff
@@ -67,30 +63,26 @@ void getData() {
...
+        assertEquals(Integer.toString(elements.length), componentIdText[1]);
+        assertEquals(Lifespan.REGULAR.toString(), componentIdText[2]);
+
+        // Explicit lifespan
+        for (Lifespan lifespan : Lifespan.values()) {
+            componentIdText = adapter.generateComponentId(lifespan, elements).split(";");
+            assertEquals(3, componentIdText.length);
+            assertEquals(NAME, componentIdText[0]);
+            assertEquals(Integer.toString(elements.length), componentIdText[1]);
+            assertEquals(lifespan.toString(), componentIdText[2]);
```
marko-radosavljevic, line 85: This can probably be in a separate test, with a better name describing tested behaviour (maybe with a `@DisplayName` too).

But I see other methods are tested the same way, so It's fine i guess. It's general enough, just a bit bulky with lots of assertions. ^^

marko-radosavljevic (commented):
Thanks for providing such a detailed and nicely written PR. It made review much easier. :relaxed: :heart: 

Code is clean, flexible, and it doesn't change user API. Great work! ^^

Heatmanofurioso (dismissed):

application/src/test/java/org/togetherjava/tjbot/commands/SlashCommandAdapterTest.java
Zabuzard: Yeah. I feel like we havent yet "agreed" on some sort of guideline on how to structure tests. Right now we just went with one test-method per method to test and then used comments to separate blocks.

I would say we let this pass through but discuss that topic in a separate GH issue, align on something and then adjust all existing tests (since its not only this file which is affected)

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
Zabuzard: Good point. I added a link to a wiki page which I will write now.

Zabuzard:
Created [wiki/Component-IDs](https://github.com/Together-Java/TJ-Bot/wiki/Component-IDs) and [wiki/Component-ID-Store](https://github.com/Together-Java/TJ-Bot/wiki/Component-ID-Store) as per request by @marko-radosavljevic

application/src/test/java/org/togetherjava/tjbot/commands/SlashCommandAdapterTest.java
marko-radosavljevic: Yup, I agree, let's keep it consistent with the current testing practices and revise it later.

marko-radosavljevic (approved):
33622cc confused me, because I could've sworn synchronization was there when I reviewed it last time. Only then realized I'm reviewing the same code, it's just new commits that are force pushed. Maybe I'm a bit slow, but at least I'm still somewhat sane. :D

Looks good. Let's merge this, it's time. ^^

Tais993 (approved):

Issue #222, opened 2021-10-25
Add mod audit logs


Zabuzard:
For moderators, it is important to be able to easily audit all moderative actions:

- User was banned
- User was unbanned
- User was kicked
- User was muted
- User was unmuted
- User deleted a message of another user

The Discord Audit Log does not provide good filtering tools for that and hence is too verbose as an alternative. The channels `#mod_commands` and `#commands` originally have been intended to fulfill this role but mods and staff dont always do their actions via bot-commands but also with the Discord UI. And with the slash commands of this bot the lines will blur even further.

A better alternative is to just let the bot hook on all mentioned actions (regardless how they were triggered in the first place) and log them into a dedicated channel which mods then can easily read and follow.

As far as implementation goes, here is a list of useful JDA methods:

- `onGuildBan`, `onGuildUnban`
- `onGuildMemberRemove`
- `onGuildMemberRoleAdd`, `onGuildMemberRoleRemove`
- `onGuildMessageDelete`
- `retrieveAuditLogs`

PR #225, opened 2021-10-25
Add mod audit logs

Zabuzard:
### Overview

Implements #222 .

Logs moderative actions to a dedicated channel.

example

### Supported actions

The current implementation supports logging of

- ban
- unban
- kick
- user muted
- user unmuted
- message deleted by another user

### Config changes

This changes the config file and adds two new entries:
```json
"modAuditLogChannelPattern": "mod_audit_log",
"mutedRolePattern": "Muted"
```
The entries are REGEX patterns used to identify the corresponding channel and role, based on their name.

### Implementation

The system is a **routine** which is executed automatically on a specific schedule. Due to hard rate limits on viewing the audit log, right now it is executed only **three times** per day, on fixed times:
* 04:00 UTC
* 12:00 UTC
* 20:00 UTC

There is no proper "system support" for routines yet, hence there are a bunch of TODOs linking to the corresponding GH issue #235 .

The system maintains a database table where it remembers for each guild a timestamp when it checked the logs last time. Whenever it executes, it retrieves all audit log messages after that specific date and logs them to the dedicated channel.

### Limitations (Discussion)

I see 3 approaches to implement this and unfortunately neither is really optimal:

1. Use `ListenerAdapter` to react on the events exposed by JDA
2. Check the Discord Audit Log on a certain interval (`guild.retrieveAuditLogs()`)
3. Let the bot commands log actions themselve (maybe offer a central point for them to report to)

Combinations are of course also an option, if necessary.

Approach **2.** is the one currently implemented by this PR. Approach **1.** was implemented by this PR in the past.

#### extend `ListenerAdapter`

This works and is reliable. Unfortunately, the information that the API exposes is extremely limited and sparse. For example, an user-ban-event only exposes the guild and the banned user. It does not tell who banned the user.

But more ciritical is that neither **kick** nor **message deletion** can be retrieved.

There is no `onGuildMemberKicked` event, only `onGuildMemberRemove​` but this, unfortunately, includes when a member left on their own and there is no way to differentiate them.

For message deletion, there is only `onGuildMessageDelete` but this also includes when an user deleted their own messages (which is not an action that should be logged) and there is no way to differentiate them.

#### Check Discord Audit Log

The Discord Audit Log exposes a lot of useful information. Everything expect showing actual message content can be implemented by reading the log.

However, it seems that reading the audit log is heavily rate limited. The rate limit seems to be something around 5 reads per day (maybe less, but still). And unfortunately there is no way to "subscribe" to the event of new actions being added.

So the audit log is barely usable and could, if at all, only be used to log events much after they actually happened. Extensive logs that appear only once or twice per day might still be useful though.

#### Let commands log themselves

This is the most flexible approach and offers the most details, also not subject to any rate limiting.

However, we can only control our own commands. That means, it is not possible to log any actions executed via other bot commands (e.g. Dyno) or via the Discord UI.

Which makes this option barely useful.


Heatmanofurioso (dismissed):

application/src/main/resources/db/V3__Add_Mod_Audit_Log_System.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE mod_audit_log_guild_process
+(
+    guild_id                       BIGINT    NOT NULL PRIMARY KEY,
```
java-coding-prodigy, line 3: Why is this `BigInt`?

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
```diff
@@ -0,0 +1,371 @@
...
+    private Optional<TextChannel> getModAuditLogChannel(@NotNull Guild guild) {
+        // Check cache first, then get full list
+        return guild.getTextChannelCache()
+            .stream()
+            .filter(isAuditLogChannel)
+            .findAny()
+            .or(() -> guild.getTextChannels().stream().filter(isAuditLogChannel).findAny());
+    }
+
+    private enum Action {
```
java-coding-prodigy, line 347: What about `purge` and `prune`? Are they included in `MESSAGE_DELETION`?

application/src/main/resources/db/V3__Add_Mod_Audit_Log_System.sql
Zabuzard: thats the type in sqlite that corresponds to long in java

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
Zabuzard: yes, any sort of message deletion will trigger this

Zabuzard: prune is not included (but could be). i dont think we want to monitor that though. at least i would not be interested in prunes.

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
```diff
@@ -0,0 +1,371 @@
...
+            case UNBAN -> handleUnbanEntry(entry);
+            case KICK -> handleKickEntry(entry);
+            case MEMBER_ROLE_UPDATE -> handleRoleUpdateEntry(entry);
+            case MESSAGE_DELETE -> handleMessageDeleteEntry(entry);
+            default -> Optional.empty();
+        };
+        return maybeMessage.map(message -> message.flatMap(auditLogChannel::sendMessageEmbeds));
+    }
+
+    @SuppressWarnings("MethodMayBeStatic")
```
Tais993, line 275: I'd make the methods static myself, like IntelliJ is saying

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
```diff
@@ -0,0 +1,371 @@
...
+
+        private final String title;
+        private final String verb;
+
+        Action(@NotNull String title, @NotNull String verb) {
+            this.title = title;
+            this.verb = verb;
+        }
+
+        String getTitle() {
```
Tais993, line 363:
```suggestion
        @NotNull String getTitle() {
```

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
```diff
@@ -0,0 +1,371 @@
...
+            return title;
+        }
+
+        String getVerb() {
```
Tais993, line 367:
```suggestion
        @NotNull String getVerb() {
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
Zabuzard: I think this is a design flaw. As a consequence, all of those methods should be static but the role ones cant be static...
I think those methods are non-static by design but could be made static as of today, due to their current implementation.

I dont have a strong voice here though. I can make it static.

Tais993: If only a portion of the methods can become static, I'm fine with keeping them all instance because of consistency reasoning.

If they all can get static, I'd for sure make them all static.

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
Zabuzard: Fixed

application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
Zabuzard: Fixed

application/src/main/java/org/togetherjava/tjbot/Application.java
```diff
@@ -81,6 +82,10 @@ public static void runBot(String token, Path databasePath) {
             jda.awaitReady();
             logger.info("Bot is ready");
 
+            // TODO This should be moved into some proper command system instead (see GH issue #235
+            // which adds support for routines)
```
RealYusufIsmail, line 86: Is the todo needed?

Zabuzard: yes. otherwise people might forget about this or dont understand why this code is there in the way it is there right now (which is not really the best way to write that code).

marko-radosavljevic (approved):
Short one, but tricky to get right. ^^

All questions I had have been answered in the code, which is neat. A bit heavier lambdas/streams on moments, but that's alright. All in all, clean. :relaxed: 

Since this is an isolated and easily replaceable piece of code, and also important for moderation, feel free to merge it.

Tais993 (approved):

PR #245, opened 2021-11-04
Added .DS_STORE to gitignore (mac users)

Zabuzard:
`.DS_STORE` is a metafile usually seen on MAC. According to a quick google research, it should be put on gitignore

(we have one contributor using MAC already who spot this problem)


java-coding-prodigy:
Yusuf's commit had a comment there explaining this in brief, shouldn't we do the same?

marko-radosavljevic (approved):

illuminator3 (approved):

PR #246, opened 2021-11-04
Fix broken mockito bump (3.12.4 -> 4.0.0)

Zabuzard:
#239 Updated only one line but forgot to update this line as well

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -61,7 +61,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
 
     testImplementation 'org.mockito:mockito-core:4.0.0'
-    testRuntimeOnly 'org.mockito:mockito-core:3.12.4'
+    testRuntimeOnly 'org.mockito:mockito-core:4.0.0'
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
 }


Tais993 (approved):

Heatmanofurioso:
We should also take note of all breaking changes and deprecated APIs
https://github.com/mockito/mockito/releases

Tais993:
They've all been deprecated for some time already, so none of our code should break.
We'll find it out if something is deprecated now.

Zabuzard:
> 
> 
> We should also take note of all breaking changes and deprecated APIs https://github.com/mockito/mockito/releases

We already approved, tested and released this change in #239 . This PR is not about upgrading Mockito in general, it is about fixing the partial update that Dependabot shipped.

java-coding-prodigy:
We should have caught that error previously, but its alright, did not harm much. Imo we should have a comment there reminding us to upgrade it at both places.

application/build.gradle
```diff
@@ -61,7 +61,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
```
java-coding-prodigy, line 62:
```suggestion
//Upgrade versions at both places
```

java-coding-prodigy (changes requested):

marko-radosavljevic (approved):

application/build.gradle
marko-radosavljevic: That's kinda obvious, that if you want to update the dependency, you have to do it everywhere. It's not something unique with mockito.

Real solution is to be more careful while reviewing dependabot PRs. We had 2 approving reviews, and 3rd person merging it, it's obvious that our process failed.

I think comment is not necessary, but it's good that you are trying to find solutions for this process failure. ^^

java-coding-prodigy: > That's kinda obvious, that if you want to update the dependency, you have to do it everywhere. It's not something unique with mockito.

Well, only mockito and JUnit are mentioned at 2 places.

>Real solution is to be more careful while reviewing dependabot PRs. We had 2 approving reviews, and 3rd person merging it, it's obvious that our process failed.

Definitely agree, I was one of the approvers in fact.

>I think comment is not necessary, but it's good that you are trying to find solutions for this process failure. ^^

I too think it might be too much but we should do it to be safe. Thank you for the compliment :)

java-coding-prodigy: Others do comment your opinions.

Zabuzard: I get your point but I disagree. This "problem" does not only exist for mockito but for pretty much everything in this file.

```
implementation 'org.apache.logging.log4j:log4j-api:2.14.1'
implementation 'org.apache.logging.log4j:log4j-core:2.14.1'
implementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.14.1'
```
and
```
implementation 'org.scilab.forge:jlatexmath:1.0.7'
implementation 'org.scilab.forge:jlatexmath-font-greek:1.0.7'
implementation 'org.scilab.forge:jlatexmath-font-cyrillic:1.0.7'
```
and
```
implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.13.0'
implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
```
and
```
testImplementation 'org.mockito:mockito-core:4.0.0'
testRuntimeOnly 'org.mockito:mockito-core:4.0.0'
```
and
```
testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
```

So either we add version-variables or we do our job at reviewing better. In either case, I would solve this "problem" in a different GH issue/PR if at all.

java-coding-prodigy: I see, I thought it was only for those two but now I realize we need to have the comment everywhere, which is bad. I guess we don't do it then.

java-coding-prodigy (approved):

PR #251, opened 2021-11-08
UX tag system - Suggest tags if typo

### LINKED ISSUE #250: UX tag system - suggest tags if typo
When mistyping a tag name, such as `/tag instane` instead of `/tag instance`, the tag system should suggest the correct tag to the user for a better UX.

This could for example be based on edit distance.

Zabuzard:
### Overview

Implements and closes #250 .

If a user types a tag wrong, for example `/tag hello` instead of `/tag hallo` or similar, the system will now suggest the best matching alternative:

BplvCXF

This applies to all tag commands:
* `/tag id`
* `/tag-manage raw id`
* `/tag-manage delete id`
* `/tag-manage edit id`
* and more...

### Details

The current implementation is based on [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) and **has unit test coverage**.

The utility is exposed via a new `StringDistances` utility class, which also offers useful related methods for other commands in the future, for example an `autocomplete` method, which could be useful in a search-as-you-type system.

---
### Button idea

To enhance the UX further, I was thinking of adding some sort of `Yes`, `No` buttons, which the user could click instead of having to write the command again.

While this is a cool and also easy idea, I ran into one issue that makes this a bad UX. Namely, the error message and the buttons should be **ephemeral** but the final response, after clicking `Yes` would **not be ephemeral**. This can only be realized by sending a totally new message, and not by simply editing the original message. But then the command has no context for other users, i.e. `"Bob used /tag ..."` is then not displayed.

So this idea **basically fails** due to
* not being able to either change the ephemeral status of an already send message,
* or not being able to resend an existing message (with different attributes),
* or not being able to fake a slash command usage with a bot

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagSystem.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagSystem.java
@@ -5,6 +5,7 @@ import net.dv8tion.jda.api.interactions.Interaction;
 import net.dv8tion.jda.api.interactions.components.Button;
 import net.dv8tion.jda.api.interactions.components.ButtonStyle;
 import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.commands.utils.StringDistances;
 import org.togetherjava.tjbot.db.Database;
 import org.togetherjava.tjbot.db.generated.tables.Tags;
 import org.togetherjava.tjbot.db.generated.tables.records.TagsRecord;
@@ -57,12 +58,16 @@ public final class TagSystem {
      * @param event the event to send messages with
      * @return whether the given tag is unknown to the system
      */
-    boolean isUnknownTagAndHandle(@NotNull String id, @NotNull Interaction event) {
+    @SuppressWarnings("BooleanMethodNameMustStartWithQuestion")
+    boolean handleIsUnknownTag(@NotNull String id, @NotNull Interaction event) {
         if (hasTag(id)) {
             return false;
         }
-        // TODO Add fuzzy string matching suggestions (Levenshtein edit distance)
-        event.reply("Could not find any tag with id '%s'.".formatted(id))
+        String suggestionText = StringDistances.closestMatch(id, getAllIds())
+            .map(", did you perhaps mean '%s'?"::formatted)
+            .orElse(".");
+
+        event.reply("Could not find any tag with id '%s'%s".formatted(id, suggestionText))
             .setEphemeral(true)
             .queue();
         return true;
--- a/application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
@@ -1,0 +1,146 @@
+package org.togetherjava.tjbot.commands.utils;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.stream.IntStream;
+
+/**
+ * Utility class for computing string distances, for example the edit distance between two words.
+ */
+public enum StringDistances {
+    ;
+
+    /**
+     * Computes the candidate that matches the given query string best.
+     *
+     * It is given that, if the candidates contain the query literally, the query will also be the
+     * returned match. If the candidates do not contain the query literally, the best match will be
+     * determined. The measures for this are unspecified.
+     * 
+     * @param query the query string to find a match for
+     * @param candidates the set of candidates to select a match from
+     * @param <S> the type of the candidates
+     * @return the best matching candidate, or empty iff the candidates are empty
+     */
+    public static <S extends CharSequence> Optional<S> closestMatch(@NotNull CharSequence query,
+            @NotNull Collection<S> candidates) {
+        return candidates.stream()
+            .min(Comparator.comparingInt(candidate -> editDistance(query, candidate)));
+    }
+
+    /**
+     * Attempts to autocomplete the given prefix string by selecting the candidate that matches the
+     * prefix best.
+     *
+     * It is given that, if the candidates contain the query literally, the query will also be the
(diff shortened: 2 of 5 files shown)


java-coding-prodigy (commented):
The problem is that if the user enters some gibberish, it will give the closest tag. Not sure whether that would be good UX.

Zabuzard:
> 
> 
> The problem is that if the user enters some gibberish, it will give the closest tag. Not sure whether that would be good UX.

I could put in a threshold, for example to not suggest if the edit distance is bigger than 5. But to be honest, I do not really see an issue here.

Like, having suggestions (even if it is wrong) is still better than having no suggestions at all.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -0,0 +1,146 @@
...
+            for (int y = 1; y < columns; y++) {
+                // Take minimum of all candidates
+                int upperCandidate = table[x - 1][y] + 1;
+                int leftCandidate = table[x][y - 1] + 1;
+                int diagonalCandidate = table[x - 1][y - 1];
+                if (source.charAt(x - 1) != destination.charAt(y - 1)) {
+                    diagonalCandidate++;
+                }
+
+                int bestCandidate = IntStream.of(upperCandidate, leftCandidate, diagonalCandidate)
```
illuminator3, line 137: Inline this variable.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        for (int x = 0; x < rows; x++) {
+            table[x][0] = x;
+        }
+
+        // Process row by row, selecting diagonal candidates
+        for (int x = 1; x < rows; x++) {
+            for (int y = 1; y < columns; y++) {
+                // Take minimum of all candidates
+                int upperCandidate = table[x - 1][y] + 1;
```
illuminator3, line 130: Inline this variable.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -0,0 +1,146 @@
...
+                int leftCandidate = table[x][y - 1] + 1;
```
illuminator3, line 131: Inline this variable.

illuminator3 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: i made this on purpose for readability, since i can then pick a meaningful name. this algorithm is not trivial for someone who doesnt know it and the variable names help understanding it and "visualizing it" on the table.

can you agree to that?

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: i made this on purpose for readability, since i can then pick a meaningful name. this algorithm is not trivial for someone who doesnt know it and the variable names help understanding it and "visualizing it" on the table.

can you agree to that?

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: i made this on purpose for readability, since i can then pick a meaningful name. this algorithm is not trivial for someone who doesnt know it and the variable names help understanding it and "visualizing it" on the table.

can you agree to that?

marko-radosavljevic:
Hm, not having a button is a real bummer, and makes this feature a lot less useful.

Because if you do `?tag persno` instead of person, and you see no bot response, you know you misspelled it, and you just repeat the same thing. Bot telling you it's actually 'person' is obvious, if you already know that you need that specific tag. We also rarely have long and difficult to spell tags.

On the other hand, bot automatically replying with the closest match is an even worse UX. Is there nothing else we can do here?

Why can't we do something similar to what !doc bot does. It's not ephemeral, and you can pick, and it edits the original message? Or that would be even worse UX, because if the suggestions are not useful, you have to pick your mouse, delete it, then go back to keyboard to type the tag you actually want..

What I would like instead would be some form of auto-suggestions with option to autocomplete, and keyboard centric workflow. Don't know if it's possible tho. 

It's still a cute feature, and will probably be more useful in the future. :relaxed: :heart:

marko-radosavljevic (approved):
Haven't reviewed the algo, but the tests pass, and it's well documented. That's more than enough for me. If we catch any weird edge cases, we will add them to the tests. :relaxed:

java-coding-prodigy:
I just realized that we should probably make the system non case sensitive

Zabuzard:
> 
> 
> I just realized that we should probably make the system non case sensitive

Agreed but that would go into a different PR since that would require making the whole tag system lower-case-only.

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagSystem.java
```diff
@@ -57,12 +58,16 @@ static Button createDeleteButton(String componentId) {
...
-    boolean isUnknownTagAndHandle(@NotNull String id, @NotNull Interaction event) {
+    @SuppressWarnings("BooleanMethodNameMustStartWithQuestion")
+    boolean handleIsUnknownTag(@NotNull String id, @NotNull Interaction event) {
         if (hasTag(id)) {
             return false;
         }
-        // TODO Add fuzzy string matching suggestions (Levenshtein edit distance)
-        event.reply("Could not find any tag with id '%s'.".formatted(id))
+        String suggestionText = StringDistances.closestMatch(id, getAllIds())
+            .map(", did you perhaps mean '%s'?"::formatted)
```
Tais993, line 67: This just blew my mind ngl

Tais993 (approved):

PR #255, opened 2021-11-10
Update JDA version

RealYusufIsmail:
This update includes:

add support for min/max option value (#1863)

Implement user banner & banner color (#1736)

Updated magic numbers to use the Message#MAX_CONTENT_LENGTH constant (#1870)

 add BOT_HTTP_INTERACTIONS user flag (#1865)

Added whoami as command to the MessageListenerExample which sends id,… (#1866)

Attempt a fix for overriding messages

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -45,7 +45,7 @@ shadowJar {
 dependencies {
     implementation project(':database')
 
-    implementation 'net.dv8tion:JDA:4.3.0_339'
+    implementation 'net.dv8tion:JDA:4.3.0_346'
 
     implementation 'org.apache.logging.log4j:log4j-api:2.14.1'
     implementation 'org.apache.logging.log4j:log4j-core:2.14.1'


Zabuzard:
Any reason why we need this **now** instead of waiting till the next dependabot update at the end of the month? The only thing in the list you shared that could be interesting for us today would be

* add support for min/max option value (#1863)

and we already solved that problem with the choice-system thing

RealYusufIsmail:
I could use min/max for the user id instead of keeping as a string I can limit it to a certain number of value

Zabuzard:
> 
> 
> I could use min/max for the user id instead of keeping as a string I can limit it to a certain number of value

for what user id? doesnt sound reasonable to me.

RealYusufIsmail:
> > I could use min/max for the user id instead of keeping as a string I can limit it to a certain number of value
> 
> for what user id? doesnt sound reasonable to me.

I mean like `/warn userid`. To make sure the user does not enter a value bigger than a user id

Tais993:
> 
> 
> > > I could use min/max for the user id instead of keeping as a string I can limit it to a certain number of value
> > 
> > 
> > for what user id? doesnt sound reasonable to me.
> 
> I mean like `/warn userid`. To make sure the user does not enter a value bigger than a user id

You should use the OptionType#USER instead

Zabuzard:
If there is no real need for this, I say we close this and go with the next automatic update based on dependabot.

RealYusufIsmail:
Ok

Issue #247, opened 2021-11-05
VC activity command UX - Direct message towards audience


Zabuzard:
I think we could improve the vc activity command text. It is a bit random to other readers who lack context.

It would be better if it would say like **"user x wants to start activity y. feel free to join by clicking z, enjoy"**

Right now it's directed towards the message author but it should be directed to the actual audience, other people.

![akvKddi](https://user-images.githubusercontent.com/13614011/140027173-611ef6f9-e09f-4ee9-ab87-2868050b7dc1.png)

Zabuzard:
Related #242

PR #261, opened 2021-11-16
Changed VC-Activity message

JJeeff248:
This is in reference to #247 

I have changed the reply message to:
image

I'm not sure why it's displaying the link embed instead of the invite with buttons. (below)

image

I've had a look around and couldn't seem to find much information about controlling what gets displayed. I think it would look nicer with the buttons instead of the embedded link, I'm just unsure how to change it.

Another thing I played around with was mentioning the user. I couldn't decide if the bot should mention them or just use their name. I ended up going with mentioning them but am happy to change it if needed.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -213,8 +213,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
                 %s wants to start %s.
                 Feel free to join by clicking %s , enjoy!
                 If it says the activity ended, click on the URL instead.
-                 """.formatted(Objects.requireNonNull(event.getMember()).getEffectiveName(),
-                applicationName, invite.getUrl()));
+                 """.formatted(event.getUser().getAsTag(), applicationName, invite.getUrl()));
     }
 
     private static void handleErrors(@NotNull SlashCommandEvent event,
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -169,34 +169,45 @@ public final class VcActivityCommand extends SlashCommandAdapter {
 
 
         String applicationId;
+        String applicationName;
 
         if (applicationOption != null) {
-            applicationId = VC_APPLICATION_TO_ID.get(applicationOption.getAsString());
+            applicationName = applicationOption.getAsString();
+            applicationId = VC_APPLICATION_TO_ID.get(applicationName);
         } else {
             applicationId = idOption.getAsString();
+            applicationName = applicationId;
+
+            // Get the application name from the ID
+            for (var entry : VC_APPLICATION_TO_ID.entrySet()) {
+                if (applicationId.equals(entry.getValue())) {
+                    applicationName = entry.getKey();
+                    break;
+                }
+            }
         }
 
-        handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge);
+        handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge, applicationName);
     }
 
     private static void handleSubcommand(@NotNull SlashCommandEvent event,
             @NotNull VoiceChannel voiceChannel, @NotNull String applicationId,
-            @Nullable Integer maxUses, @Nullable Integer maxAge) {
+            @Nullable Integer maxUses, @Nullable Integer maxAge, @NotNull String applicationName) {
 
         voiceChannel.createInvite()
             .setTargetApplication(applicationId)
             .setMaxUses(maxUses)
             .setMaxAge(maxAge)
-            .flatMap(invite -> replyInvite(event, invite))
+            .flatMap(invite -> replyInvite(event, invite, applicationName))
             .queue(null, throwable -> handleErrors(event, throwable));
     }
 
     private static @NotNull ReplyAction replyInvite(@NotNull SlashCommandEvent event,
-            @NotNull Invite invite) {
+            @NotNull Invite invite, @NotNull String applicationName) {
         return event.reply("""
-                I wish you a lot of fun, here's the invite: %s
-                If it says the activity ended, click on the URL instead.
-                 """.formatted(invite.getUrl()));
+                %s wants to start %s.
+                Feel free to join by clicking %s, enjoy!
+                 """.formatted(event.getMember(), applicationName, invite.getUrl()));
     }
 
     private static void handleErrors(@NotNull SlashCommandEvent event,
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -185,7 +185,8 @@ public final class VcActivityCommand extends SlashCommandAdapter {
         handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge, applicationName);
     }
 
-    private static <K, V> @NotNull Optional<K> getKeyByValue(@NotNull Map<K, V> map, @NotNull V value) {
+    private static <K, V> @NotNull Optional<K> getKeyByValue(@NotNull Map<K, V> map,
+            @NotNull V value) {
         for (Map.Entry<K, V> entry : map.entrySet()) {
             if (value.equals(entry.getKey())) {
                 return Optional.of(entry.getKey());
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -23,6 +23,7 @@ import org.togetherjava.tjbot.commands.SlashCommandVisibility;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Implements the {@code vc-activity} command. Creates VC activities.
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -169,34 +169,45 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
         if (applicationOption != null) {
-            applicationId = VC_APPLICATION_TO_ID.get(applicationOption.getAsString());
+            applicationName = applicationOption.getAsString();
+            applicationId = VC_APPLICATION_TO_ID.get(applicationName);
         } else {
             applicationId = idOption.getAsString();
+            applicationName = applicationId;
+
+            // Get the application name from the ID
+            for (var entry : VC_APPLICATION_TO_ID.entrySet()) {
```
Zabuzard, line 182: NIT: I am okay with `var` but we havent used it in the project so far. I guess most people would discourage using it in this specific case (since the types of the entry are not obvious)

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -169,34 +169,45 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
+                if (applicationId.equals(entry.getValue())) {
+                    applicationName = entry.getKey();
+                    break;
```
Zabuzard, line 185: * What if its never found? What behavior do you want then?
* Might create a little helper method `private static Optional<K> <K, V> getKeyByValue(Map<K, V> map, V value)`, makes this more readable.

Zabuzard (changes requested):

Zabuzard:
Good stuff, thanks for your contribution! 😃

Tais993:
I don't like it's behaviour atm.
Mentioning the user feels more cluttered, I'd rather see their tag since they're already mentioned.

And if the user uses their own ID, seeing "wants to start 85728572" is bad UX.

Not sure what others think, just my opinion.

Zabuzard:
> 
> 
> I don't like it's behaviour atm. Mentioning the user feels more cluttered, I'd rather see their tag since they're already mentioned.
> 
> And if the user uses their own ID, seeing "wants to start 85728572" is bad UX.
> 
> Not sure what others think, just my opinion.

I agree. My opinion:
* Dont mention the user again, use their tag.
* Regarding custom IDs, maybe add a name field to the user input - or how should others know what game it is anyways 🤷

Tais993:
Just say "wants to start a activity" then
The name isn't always known either and still, this bot shouldn't be used for "professional" invites (imo)

I still prefer it if the bot doesn't send a heart welcoming invite myself.

If someone wants a "professional" invite, they can write to a message themselves with the name.

Tais993:
> * or how should others know what game it is anyways 🤷 

You didn't look at the screenshots or tried the command I see.

It sends a Discord invite with the name, look at the screenshots in this PR.

Zabuzard:
> 
> 
> > * or how should others know what game it is anyways 🤷
> 
> You didn't look at the screenshots or tried the command I see.
> 
> It sends a Discord invite with the name, look at the screenshots in this PR.

I have seen that but OP is saying that Discord doesnt display it like that, unlike expected... Or did I read that wrong?

Tais993:
Yes they were talking about something else, I'm going to take a look at that when I get at school.

But that'll be fixed, look at the second screenshot they send.

That's how Discord displays it

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -169,34 +169,45 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
 
     private static @NotNull ReplyAction replyInvite(@NotNull SlashCommandEvent event,
-            @NotNull Invite invite) {
+            @NotNull Invite invite, @NotNull String applicationName) {
         return event.reply("""
-                I wish you a lot of fun, here's the invite: %s
-                If it says the activity ended, click on the URL instead.
-                 """.formatted(invite.getUrl()));
+                %s wants to start %s.
+                Feel free to join by clicking %s, enjoy!
```
Tais993, line 209:
```suggestion
                Feel free to join by clicking %s , enjoy!
```
If you include the , in the URL, it'll indeed behave weird

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -169,34 +169,45 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
         String applicationId;
+        String applicationName;
 
         if (applicationOption != null) {
-            applicationId = VC_APPLICATION_TO_ID.get(applicationOption.getAsString());
+            applicationName = applicationOption.getAsString();
+            applicationId = VC_APPLICATION_TO_ID.get(applicationName);
         } else {
             applicationId = idOption.getAsString();
+            applicationName = applicationId;
```
Tais993, line 179:
```suggestion
                applicationName = "an activity"
```
Makes more sense than a random number

Tais993 (changes requested):
Id' still add a note so people understand they can use the URL if the invite doesn't work

Like I had before
> If it says the activity ended, click on the URL instead.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: Would replacing `var` with `Map.Entry<String, String>` be better here then? 
Or is there a better alternative?

java-coding-prodigy: I don't think there is any other option, without using the `forEach` method

JJeeff248:
> Mentioning the user feels more cluttered, I'd rather see their tag since they're already mentioned.

When you say see their tag would that be their effective name or is it something else?

This is the effective name: `event.getMember().getEffectiveName()`

Tais993:
I'm talking about their tag, User#getAsTag

Example is `Tijs#0002`, or `Zabuzard#0123`

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: > What if its never found? What behavior do you want then?

If the activity is not found the bot currently replies `Something went wrong :/` and logs a warning. Should it do something else?

Tais993: He's talking about when the name isn't found, not when the activity doesn't exist.

And when the name isn't found imo it should say 

`Tais#0002 wants to start an activity.` 
Instead of
`Tais#0002 wants to start 01258284129.`

JJeeff248: Oh yup ok, thanks :)

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: Changed in 107d309b978ad9aeb2783e0fb415b1ca750841e1

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: > What if its never found? What behavior do you want then?

Changed in 16db30d18f0dbe01b7948578f5f9ad451f41c605

> Might create a little helper method private static Optional<K> <K, V> getKeyByValue(Map<K, V> map, V value), makes this more readable.

Changed in 107d309b978ad9aeb2783e0fb415b1ca750841e1

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: Fixed in 16db30d18f0dbe01b7948578f5f9ad451f41c605

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: Changed in 16db30d18f0dbe01b7948578f5f9ad451f41c605

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -169,34 +170,50 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
             applicationId = idOption.getAsString();
+
+            applicationName =
+                    getKeyByValue(VC_APPLICATION_TO_ID, applicationId).orElse("an activity");
+        }
+
+        handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge, applicationName);
+    }
+
+    private static <K, V> Optional<K> getKeyByValue(@NotNull Map<K, V> map, @NotNull V value) {
```
Zabuzard, line 188: NotNull on return type as well :)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
JJeeff248: Fixed in 52d86d77b2f17896138b8bc16f30200de343664e

Zabuzard (dismissed):

Zabuzard:
Please dont forget to apply **Spotless**. Your code has some minor formatting issue right now, it seems.

Zabuzard (approved):

Tais993:
Don't really like the message it responds with.
With the old message we had `If it says the activity ended, click on the URL instead.`

I feel like we should add a specific note about Discord saying that the activity ended doesn't matter and the URL could be used.
While it does tell the users to press on the URL atm, it doesn't mention that Discord says the activity ended but that that can be ignored.

JJeeff248:
> With the old message we had If it says the activity ended, click on the URL instead.

That is still included in the current message. The full message at the moment is
```
user#tag wants to start an activity.
Feel free to join by clicking *url* , enjoy!
If it says the activity ended, click on the URL instead.
```
Is there a more specific message you think should be used instead?
I'm not too sure about including more in the message as it might get a bit long and seems unnecessary.

If you want it to specify that Discord says it has ended would `If Discord says the activity ended, click on the URL instead.` be better?

Tais993:
> That is still included in the current message. The full message at the moment is

In the screenshot it wasn't visisble, so didn't notice that.
I'm happy with it then

Tais993 (approved):

marko-radosavljevic:
Thanks for contributing @JJeeff248, much appreciated. :relaxed: :heart:

Merged.

PR #252, opened 2021-11-08
Migrate mute, unmute commands

Zabuzard:
### Overview

Partially implements #64 . Adds commands:
* `/mute user reason`
* `/unmute user reason`

muted_unmuted

The commands are fairly simple and follow the pattern introduced by #244 . That said, the branch is based on that branch and hence also blocked by its release.

---
### Notes

This PR slightly changes the initialization process of the bot by moving routines and listeners down into `Commands.java` (instead of `Application.java`). This is a temporary change that is unfortuantely required, since they need access to systems such as `TagSystem` or `ModAuditLogStore`, and also the `JDA` instance (which is now also passed down through `CommandSystem`).

All of this will be made nicer anyways with #235 and #236 , so I wouldnt bother too much for now.

---
### Bot permission change

The bot needs to listen to user-join events, which requires the privileged **Server Members Intent** to be enabled (already enabled it on our bots and added it to the config, since it doesnt harm - but we have to tell contributors in an announcement).

setting

---
### Checklist

Basic
* [x] Mute a user
* [x] Unmute a user

DM
* [x] Mute a user who has disabled DMs
* [x] Mute a user who has enabled DMs
* [x] Unmute a user who has disabled DMs
* [x] Unmute a user who has enabled DMs

Foreign users
* [x] Attempt to mute a user who is not in the guild
* [x] Attempt to unmute a user who is not in the guild

Logic
* [x] Attempt to mute an already muted user
* [x] Attempt to unmute an user who is not muted yet

Permissions
* [x] Mute with the bot not having role permissions
* [x] Unmute with the bot not having role permissions
* [x] Mute with the author not having the soft moderation role
* [x] Unmute with the author not having the soft moderation role
* [x] Mute with the bot being lower in the hierarchy than the target
* [x] Unmute with the bot being lower in the hierarchy than the target
* [x] Mute with the bot being lower in the hierarchy than the muted role
* [x] Unmute with the bot being lower in the hierarchy than the muted role
* [x] Mute with the author being lower in the hierarchy than the target
* [x] Unmute with the author being lower in the hierarchy than the target
* [x] Mute with the author being lower in the hierarchy than the muted role
* [x] Unmute with the author being lower in the hierarchy than the muted role

Rejoin
* [x] Mute user, then leave and rejoin, user should be muted again
* [x] Mute user, then unmute. Then leave and rejoin, user should not be muted
* [x] Mute user, then leave. Unmute and rejoin, user should not be muted
[...]

Issue #64, opened 2021-09-09
Migrate other moderation commands


Zabuzard:
Next to the main moderation system (#63 ) we also have to migrate other important moderation commands, such as the mute, whois and modlogs system.

For reference, this is how it works as of today:

* `?whois <user>` by Dyno can be used to get an overview of a specific user. For example when he joined and registered.
* ~~`?modlogs <user>`  by Dyno shows all moderation actions that have been taken against an user.~~ (implemented)
* ~~`>mute <user>` and `>unmute <user>` by Inori Chan to mute or unmute an user.~~ (implemented)
* ~~`?mute <user> <time> <message>` by Dyno to mute an user. `<time>` and `<message>` are optional.~~ (implemented)

Zabuzard:
Naturally blocked by #59

Zabuzard:
Only `whois` is missing still, need to elaborate on the design and use case though

Tais993:
We'll have to discuss whenever we still want the mute command.

[Discord just released time-outs](https://support.discord.com/hc/en-us/articles/4413305239191)
This is our equivalent of the mute command, difference is that Discord's version as a better UX and UI for both moderators and the people muted.

Only issue is, UI allows upto 1 week.
Bots can set upto 1 month, so we might still want a command to set it to a month.

Tais993:
> Only `whois` is missing still, need to elaborate on the design and use case though

Will make a prototype of the whois command ^^

Issue #63, opened 2021-09-09
Migrate ban/warn system


Zabuzard:
We have to migrate the ban and warning moderation system. The details are up to be decided by the community.

For reference, this is how it works as of today:

### Ban and warning system by Dyno

The commands `?ban` and `?warn` can be used for moderation.

#### Banning
The available command is

- `?ban <user> <time> <message>`

with

- `<user>` - mandatory - either being the name of the user, like `@Zabuzard` or his Discord account ID like `157994153806921728`
- `<time>` - optional - empty for a permanent ban, otherwise a period like `1h` or `1d`
- `<message>` - optional - a ban message/reason which is also written to the user in a DM by Dyno

Banning an user also deletes all of his recent messages.

#### Warning
The available command is

- `?warn <user> <message>`

with

- `<user>` - mandatory - either being the name of the user, like `@Zabuzard` or his Discord account ID like `157994153806921728`
- `<message>` - mandatory - the warning message/reason which is also written to the user in a DM by Dyno

Also, there is

- `?warnings <user>` - which will display an overview of all warnings an user received already

Zabuzard:
Naturally blocked by #59

java-coding-prodigy:
We could have a similar system with some changes:

- Change the prefix.
- Change the order of queries.
- Have an automatic message displayed by the bot telling the user why they were warned/muted/banned, depending on the message provided.
- We could also remove the `?tag interrupt` and display the message in the warning message mentioned above.

illuminator3:
> * Have an automatic message displayed by the bot telling the user why they were warned/muted/banned, depending on the message provided.
it already private messages the user

RealYusufIsmail:
I will be adding the warn command in a future pr.

Zabuzard:
As soon as #267 has been merged, this can be closed :)

Zabuzard:
Has been implemented and merged 🎉

PR #267, opened 2021-11-17
Migrate temporary ban/mute

Zabuzard:
### Overview

Dyno supports temporary bans and mutes, so should we. This adds support for such a feature. Part of #63 and #64 .

Therefore, a `duration` option is added to the existing `ban` and `mute` commands, which allows selection from a menu, such as:

choice menu

response

user DM

### Details

The feature is implemented by using the existing `ModerationActionsStore`, which already supports a field `expires_at`.

A new class `TemporaryModerationRoutine` is scheduled every 5 minutes. The routine basically selects all actions from the store where `expires_at` is behind `now` (minus edge cases, like already revoked, overwriten by another action, ...) and then revokes those actions by unbanning/unmuting the user.

The system is **designed modular** enough to easily be expanded with new temporary moderation actions (i.e. other than temp ban and temp mute). Therefore, the slim interface `RevocableModerationAction` is implemented and an implementation is added to a list in `TemporaryModerationRoutine`, the rest is automatic.

The system also gracefully supports long downtimes. Hence, it is not a problem if the actual revocation date of an action is missed, the system will revoke it on the next execution when the bot is online again. This also holds true if the bot is temporarily removed from a guild or, more importantly if an user is temporarily not a member of the guild anymore (think about a temporary mute, then user leaves and wants to rejoin after the expiration date, mute should not be applied anymore).

### Notes

Error handling, such as users who got deleted in the meantime, left the guild or the even the bot leaving the guild and similar, are handled solely by `logger` messages. There is no direct feedback in any guild channel (on purpose).

logger success

Revocation is picked up by the existing mod-audit-log feature:

audit log

As well as by the `/audit` command (ignore the timestamps, I didnt have the bot running for some time):

audit command

#### TODO

* [x] Wait for #252 
* [x] Add support for temp mutes
* [x] Fix remaining issues in the code
* [x] Javadoc, Linter

#### Testing

Base cases:
* [x] Perm ban
* [x] Perm mute
* [x] Temp ban
* [x] Temp unmute

Bot leaves guild
* [x] Temp ban/mute, then remove bot from guild and wait until revoke kicks in (can be mocked easily by editing the DB guildId value)

User does not exist
* [x] Temp ban/mute, then delete user account from Discord and wait until revoke kicks in (can be mocked easily by editing the DB userId value)

User is not a member
* [x] Temp mute user, kick user, wait until revoke kicks in

Multiple overlapping actions
* [x] Temp ban user, then unban, then temp ban again with different time (all within less than 5 minutes (the routine schedule))

Stale event
* [x] Temp ban user, then unban, temp unban should not be executed anymore
* [x] Temp mute user, then unmute, temp unmute should not be executed anymore

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/TemporaryModerationRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/TemporaryModerationRoutine.java
@@ -20,8 +20,8 @@ import java.util.stream.Collectors;
 // FIXME javadoc
 public final class TemporaryModerationRoutine {
     private static final Logger logger = LoggerFactory.getLogger(TemporaryModerationRoutine.class);
-    private static final Set<ModerationUtils.Action> REVOCABLE_ACTIONS =
-            EnumSet.of(ModerationUtils.Action.BAN, ModerationUtils.Action.MUTE);
+    private static final Set<ModerationAction> REVOCABLE_ACTIONS =
+            EnumSet.of(ModerationAction.BAN, ModerationAction.MUTE);
 
     private final ModerationActionsStore actionsStore;
     private final JDA jda;
@@ -96,9 +96,9 @@ public final class TemporaryModerationRoutine {
         // Do not revoke an action which was already revoked by another action issued afterwards
         // For example if a user was unbanned manually after being temp-banned,
         // but also if the system automatically revoked a temp-ban already itself
-        ModerationUtils.Action revokeActionType = switch (groupIdentifier.type) {
-            case BAN -> ModerationUtils.Action.UNBAN;
-            case MUTE -> ModerationUtils.Action.UNMUTE;
+        ModerationAction revokeActionType = switch (groupIdentifier.type) {
+            case BAN -> ModerationAction.UNBAN;
+            case MUTE -> ModerationAction.UNMUTE;
             default -> throw new AssertionError("Unsupported action type: " + groupIdentifier.type);
         };
         ActionRecord lastRevokeAction = actionsStore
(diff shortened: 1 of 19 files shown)


Zabuzard:
Blocked by #252

RealYusufIsmail:
Would it not be better waiting for the audit command?

Zabuzard:
> 
> 
> Would it not be better waiting for the audit command?

Why? I could ask you the same. I do not see any actual blocking dependency between the two. Only inconvenience and extra work introduced for both of them.

Zabuzard:
Blocked by #297

application/src/main/java/org/togetherjava/tjbot/commands/moderation/TemporaryModerationRoutine.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private void revokeAction(long userId, long guildId, @NotNull ModerationUtils.Action action) {
+        Guild guild = jda.getGuildById(guildId);
+        if (guild == null) {
+            logger.info(
+                    "Attempted to revoke a temporary moderation action but the bot is not connected to the guild '{}' anymore, skipping revoking.",
+                    guildId);
+            return;
+        }
+        if (action != ModerationUtils.Action.BAN) {
```
Heatmanofurioso, line 69: You could call this before retrieving the guild on line 62 `jda.getGuildById(guildId);`
If the action type is unsupported, it doesn't matter of the bot is connected or not to the guild, and we won't make an "unnecessary" call to the getGuildById

Heatmanofurioso (changes requested):

Zabuzard:
Yikes, got no time to resolve the conflict before holiday anymore. Will do it afterwards 🤷

Zabuzard:
Conflict resolved (no real code change)

Zabuzard:
Merging since 2 weeks without actual code changes and everything looks fine.

PR #268, opened 2021-11-17
Fix database methods are ambiguous without curlies

### LINKED ISSUE #201: Database methods are ambiguous without curlies
When using `database.read/write` or `database.read/writeTransaction` and using it with a one-liner lambda without curly braces, the call is **ambiguous**.

ambiguous example

We either have to explicitly cast or use curlies. But when adding curlies, we get a linter warning about having unecessary curlies. This ends up in polluting the code base with annoying warning suppressions `@SuppressWarnings("java:S1602")`, since this happens pretty much in majority of all database use cases.

The code base already has some of those suppressions in it and it keeps getting more.

suppression

----

A fix for this could be to rename the methods so that they are not overloads anymore. It is super unfortunate, but I think it would be better than having to go through this cumbersome annoyance each time using the database class.

Zabuzard:
### Overview

Fixes #201 .

**tl;dr,** the database interface lead to annoying situations, such as:

Its now fixed by renaming the methods slightly:
* the `void read` method got renamed to `readAndConsume`
* the non-void `T write` method got renamed to `writeAndProvide`

(analogously for the transaction methods)

This also fixes all currently existing instances of `java:S1602`, which have been **suppressed** so far.

### Details

I did not rename either both void or both non-void methods on purpose. Mostly to reduce the otherwise heavy merge conflicts. The `void read` method is the one that is generally not used, same for the `T write` method. Hence I only renamed the "rarely used" methods. Context-wise it also makes sense that `read(...)` actually returns a result while `write(...)` usually does not return a result.

### Merge conflicts

I made the changes in a way that the merge conflicts are **as minimal as possible**. In general, you should be able to just accept any change without any interferences (unless you changed the logic itself that this refactoring touches - unlikely though).

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/DatabaseCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/DatabaseCommand.java
@@ -85,12 +85,12 @@ public final class DatabaseCommand extends SlashCommandAdapter {
         // /db get hello
         String key = Objects.requireNonNull(event.getOption(KEY_OPTION)).getAsString();
         try {
-            Optional<String> value = database.read(context -> {
-                try (var select = context.selectFrom(Storage.STORAGE)) {
-                    return Optional.ofNullable(select.where(Storage.STORAGE.KEY.eq(key)).fetchOne())
-                        .map(StorageRecord::getValue);
-                }
-            });
+            Optional<String> value = database.read(
+                    context -> Optional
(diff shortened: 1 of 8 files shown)


RealYusufIsmail (approved):

java-coding-prodigy (approved):
All the unit tests pass, and you must have tested this manually too, LGTM

Tais993 (approved):

PR #271, opened 2021-11-19
Fixed bug with tag-raw not working with code snippets

### LINKED ISSUE #269: Tag raw doesnt work with code snippets
In some more complex setups, such as in the presence of code snippets and other, `/tag-manage raw` does not seem to work as expected.

As an example, use `/tag-manage raw id: bfs`:
bfs tag

It messes up the starting code snippet, the ending code snippets and also some of the bold-markers.

The raw content of this message, for testing purposes, is: https://pastebin.com/4tuEQ7Bz

Zabuzard:
### Bug

Fixes #269 .

Basically, `/tag-manage raw` did not escape triple backticks correctly.

issue

Resulting in crap like this:

tag

### Fix

The fix applied here is relatively simple, we still use the existing escaping utility but apply a patch on top where we replace its result with the proper result.

I slightly rewrote the code section to make it clearer that we now have a replace before and one after the escape utility. Otherwise its too easy to confuse with two replacements happening before, i.e.
```java
escape(text.replace(...)).replace(...)
// vs
escape(text.replace(...).replace(...))
```
I also adjusted the existing test cases and added a new unit test to cover this particular complex case as well.

### DIFF
--- a/application/src/test/java/org/togetherjava/tjbot/commands/utils/MessageUtilsTest.java
+++ b/application/src/test/java/org/togetherjava/tjbot/commands/utils/MessageUtilsTest.java
@@ -12,28 +12,29 @@ final class MessageUtilsTest {
     void escapeMarkdown() {
         List<TestCase> tests = List.of(new TestCase("empty", "", ""),
                 new TestCase("no markdown", "hello world", "hello world"),
-                new TestCase("basic markdown", "\\*\\*hello\\*\\* \\_world\\_",
-                        "**hello** _world_"),
+                new TestCase(
+                        "basic markdown", "\\*\\*hello\\*\\* \\_world\\_", "**hello** _world_"),
                 new TestCase("code block", """
-                        \\```java
+                        \\`\\`\\`java
                         int x = 5;
-                        \\```
+                        \\`\\`\\`
                         """, """
                         ```java
                         int x = 5;
                         ```
-                        """), new TestCase("escape simple", "hello\\\\\\\\world\\\\\\\\test",
+                        """),
+                new TestCase("escape simple", "hello\\\\\\\\world\\\\\\\\test",
                         "hello\\\\world\\\\test"),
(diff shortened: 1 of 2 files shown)


java-coding-prodigy:
Could you share a screenshot of the output after fixing the bug?

Zabuzard:
> Could you share a screenshot of the output after fixing the bug?

Sure:

after fix

Zabuzard:
That said, I spot another bug. Embeds seem to mess up leading spaces (they are part of the message though).

marko-radosavljevic (approved):
Tests failed to catch this, and if I reviewed this code (and tests) I wouldn't catch it either. Just something to note. ^^

application/src/test/java/org/togetherjava/tjbot/commands/utils/MessageUtilsTest.java
```diff
@@ -44,7 +45,118 @@ void escapeMarkdown() {
                         ```java
                         "Hello \\" World\\\\\\"" haha
                         ```
-                        """));
+                        """),
+                new TestCase("escape real example",
+                        """
+                                Graph traversal can be accomplished easily using \\*\\*BFS\\*\\* or \\*\\*DFS\\*\\*. The algorithms only differ in the order in which nodes are visited: https://i.imgur.com/n9WrkQG.png
```
illuminator3, line 51: Indent this like every other test case, please.

illuminator3 (dismissed):

application/src/test/java/org/togetherjava/tjbot/commands/utils/MessageUtilsTest.java
Zabuzard: Spotless likes to disagree with you 😆

Zabuzard:
Been a week since the last change. Got one approval by marko and one semi approval by illu (I guess). Merging.

PR #277, opened 2021-11-23
Preserving whitespace on tag-raw

### LINKED ISSUE #273: Tag-raw does not display leading spaces
### Introduction

Discord seems to not render leading spaces on a message (unless it is inside a formatted code block).

Input like

input

results in

rendered input

### Problem

This is not new and also not really a "bug". However, it becomes a problem with the `/tag-manage raw` command. See this:

code

The tag-raw command intentionally removes code blocks so that we can copy paste the raw text to edit it. But that leads to leading spaces in the code being stripped. So the copy-pasted code is fully left-trimmed.

This makes the tag-raw command barely usable in practice, very cumbersome.

### Ideas

**Dyno** attempts to fix this by putting the raw content into a **code block**, this fixes the left-trimming issue but introduces a much bigger problem... If everything is put into a code block, we cant have code blocks in the text anymore, since they would otherwise exit the surrounding code block otherwise... (leading to the mess of why dynos tag-raw was even worse UX for us the last years).

To be honest, I do not really have any nice solution right now...

Research lead to this [discussion](https://support.discord.com/hc/en-us/community/posts/360034461151-Keep-the-spaces-at-the-beginning-of-a-message) which suggests replacing leading spaces by whitespace-ish characters that look like space but are not, hence Discord does not touch it.

I guess that could fix our problem. Kinda dirty though.

### Proposal (from comments)

A simple and good fix would be to not try to hack our way through Discords rendering quirks and instead bypass the regular message system by sending file attachments.

So instead of `replyEmbed(...)` we basically could do `sendFile(...)` and send the content as file (for example with `.md` extension). That way, Discord will not mess up anything about the message content and users can get the original content easily.

This also does not interrupt the flow to create/edit tags afterwards, since it nicely synergises with `/tag edit-with-message` after posting the edited message content in Discord again.

As a note, this might make #269 obsolete (we can still keep it as general purpose helper though). In fact, `/tag raw` might not need any sort of escaping with this file-fix anymore.

Zabuzard:
Fixes and closes #273 .

Note that this is a rather dirty solution. Join in the discussion at the GH issue if you have better ideas.

---

See the GH issue for a detailed description of the problem.

---

The solution proposed by this is to replace all groups of 2 leading-whitespaces by braille-blanks. They look similar to whitespace but Discord does not recognize them and hence does not left-trim them.

To not mess up the actual messages when creating or editing, the fix is undone on tag-create and tag-edit.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagManageCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagManageCommand.java
@@ -22,6 +22,7 @@ import java.util.OptionalLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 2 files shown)


Zabuzard:
Closing this PR in favor of a better not-so-dirty fix, as proposed in #273 (txt files).

PR #282, opened 2021-11-26
Add top-helpers command

### LINKED ISSUE #65: Migrate TopHelper system

marko-radosavljevic:
## Overview

Implements and closes #65. Originally based on #117.

Adds the `/top-helpers` command that computes and displays the top helpers from the last month.

example

Also has a fallback message if there is no data in the time range yet:

no data

For now the command has no options but it was designed with optionally picking a time range (start and end) in mind. This can be done once Discord/JDA add date-pickers (currently in progress). A `TODO` in the class keeps track of this.

## Details

The code is divided into three classes:

* `TopHelpersCommand` - the slash command itself, will read the database, compute the message and display it
* `TopHelpersMessageListener` - the listener that receives all help messages and adjusts the database accordingly
* `TopHelpersPurgeMessagesRoutine` - the routine that deletes old messages from the database

### Slash command

The `TopHelpersCommand` class is pretty straightforward:

1. Check author permissions, needs "soft moderation role"
2. Compute the time range (last month, this is properly computed with `java.time` and not just last 30 days)
3. Fetch the top helpers from the database (some `groupBy` with `count` and `orderBy`, nothing fancy)
4. Extra logic if there is no data in the time range
5. Retrieve `Member` information about those users **async**
6. Error handler if it fails
7. Success handler that basically stringifies the user data using an external `AsciiTable` library

### Message listener

The `TopHelpersMessageListener` class has one simple jobs right now:

1. any incoming message will be added to the database table (just ids and timestamps, no content)

~~latter should in the near future be replaced with #346 , a `TODO` keeps track of it already.~~

Help messages are identified by channel name regex (as per `MessageReceiverAdapter`), the regex is:
```json
"helpChannelPattern": "([a-zA-Z_]+_)?help(_\\d+)?"
```

In the future, `FreeCommand` could be changed to utilize this config entry as well, instead of the channel ids.

### Message purging

The class `TopHelpersPurgeMessagesRoutine` is a simple `Routine` which is executed once every **4 hours** and deletes all top-helper message entries in the database that are **older than 90 days**.

### Database

The database table that supports all the magic is quite simple, its just:
```sql
CREATE TABLE help_channel_messages
(
    message_id BIGINT    NOT NULL PRIMARY KEY,
    guild_id   BIGINT    NOT NULL,
    sent_at    TIMESTAMP NOT NULL
)
```

## Remarks for release

The config has changed, the following has been added:
```java
"helpChannelPattern": "([a-zA-Z_]+_)?help(_\\d+)?"
```
(see the `config.json.template` file in the code diff)


Zabuzard:
Valid but parked, will be picked up by someone soonish (january maybe).

Zabuzard:
Will take this over but for the moment blocked by #346 (which itself is blocked by #345 ).

Zabuzard:
Revived it. It is working again now:

[example](https://i.imgur.com/XUadpOX.png)

Will polish it now.

Zabuzard (approved):

Zabuzard:
No changes since a week, merging.

Issue #217, opened 2021-10-22
Migrate the role reactions activity


borgrel:
**Is your feature request related to a problem? Please describe.**
This bot's brief is "to replace all bots currently used on together java"

However the role reaction has not been discussed. We need a way for new users to mark themselves with roles to identify their proficiency and the level of help they require.

**Describe the solution you'd like**
Migrate the role reaction bot

**Describe alternatives you've considered**
Make a role slash command

illuminator3:
We should use a system like TCCPP's: https://i.imgur.com/Eu7TMff.png

Zabuzard:
Thats indeed pretty cool, I like it.

PR #284, opened 2021-11-28
Migrated Role Reactions

JJeeff248:
This PR adds the `/role-select` command which allows Moderators to send a message where users can select their own roles much like we currently have in the demand_a_role channel on the server.

This addresses #217 and also implements the suggestion of using selection menus.

So far there are two sub-commands, `all` and `choose`. 
- `all` will list all the available roles for users to select from
- `choose` lets the mod select the roles to display

Both sub-commands allow for a title and description which get sent in an embedded message above the selection menu.

image

Things I'm wanting to improve/add:
- [ ] UX design. I was thinking a way to include emotes could be nice.
- [x] Allow users to deselect roles to remove them


Tais993:
Currently this shows who ran the command.
image

Thoughts on sending an ephemeral message to the user ".. went successfull"
And sending an actual message to the channel? This way it doesn't display the command.

JJeeff248:
> Thoughts on sending an ephemeral message to the user ".. went successfull"
> And sending an actual message to the channel?

I like it, I think this helps keep things looking clean.

Implemented in 9312729.

Now looks like this:
image

JJeeff248:
I realised I should've shown how the different commands look so have some screen shots below.

This is `/role-select all`

image

This is `/role-select choose`

image

And this is what happens when you select a role
image

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        if (ephemeral) {
+            event.replyEmbeds(makeEmbed(title, description))
+                .addActionRow(menu.build())
+                .setEphemeral(true)
+                .queue();
+        } else {
+            event.getChannel()
+                .sendMessageEmbeds(makeEmbed(title, description))
+                .setActionRow(menu.build())
+                .queue();
```
Tais993, line 117:
```suggestion
            event.replyEmbeds(makeEmbed(title, description))
                .addActionRow(menu.build())
                .setEphemeral(ephemeral)
                .queue();
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        if (title == null) {
+            title = "Select your roles:";
+        }
+
+        EmbedBuilder embedBuilder = new EmbedBuilder();
+        embedBuilder.setTitle(title);
+
+        if (description != null) {
+            embedBuilder.setDescription(description);
+        }
```
Tais993, line 178:
```suggestion
        embedBuilder.setDescription(description);
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+
+        return embedBuilder.build();
+    }
+
+    @Override
+    public void onSelectionMenu(@NotNull SelectionMenuEvent event, @NotNull List<String> args) {
+        Member member = Objects.requireNonNull(event.getMember(), "Member is null");
+
+        // Get the roles the bot can interact with
+        List<Role> guildRoles = Objects.requireNonNull(event.getGuild()).getRoles();
```
Tais993, line 190:
```suggestion
        List<Role> guildRoles = event.getGuild().getRoles();
```

All Objects#requireNonNull does is throw a NPE with less info.
So this is kinda useless since it'll throw a NPE either way.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        List<String> roleIds = guildRoles.stream().map(Role::getId).collect(Collectors.toList());
```
Tais993, line 191:
```suggestion
        List<String> roleIds = guildRoles.stream().map(Role::getId).toList();
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+
+        List<Role> selectedRoles = new ArrayList<>();
+        for (SelectOption role : Objects.requireNonNull(event.getSelectedOptions())) {
+            if (roleIds.contains(role.getValue())) {
+                selectedRoles.add(guildRoles.get(roleIds.indexOf(role.getValue())));
+            }
```
Tais993, line 197: 2 things.

I doubt you need the Objects#requireNonNull, it'll throw an NPE either way to my knowledge.

And the SelectOption is named "role" here, which is really misinforming the user since it's not a Role.
It's a SelectOption, thoughts on renaming it to a.e `selectedOption`?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        Member member = Objects.requireNonNull(event.getMember(), "Member is null");
+
+        // Get the roles the bot can interact with
+        List<Role> guildRoles = Objects.requireNonNull(event.getGuild()).getRoles();
+        List<String> roleIds = guildRoles.stream().map(Role::getId).collect(Collectors.toList());
+
+        List<Role> selectedRoles = new ArrayList<>();
+        for (SelectOption role : Objects.requireNonNull(event.getSelectedOptions())) {
+            if (roleIds.contains(role.getValue())) {
+                selectedRoles.add(guildRoles.get(roleIds.indexOf(role.getValue())));
```
Tais993, line 196: This looks really error prone, if a role gets added everything breaks.

Maybe make the title the role name, and the value it's ID?
Not sure how the UX would be like exactly, but it makes the backend way more safe.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        // True if the event option was 'choose'
+        if (event.getMessage().isEphemeral()) {
+
+            SelectionMenu.Builder menu = SelectionMenu.create(generateComponentId(member.getId()));
+            menu.setPlaceholder("Select your roles").setMaxValues(selectedRoles.size());
+
+            for (SelectOption roleOption : Objects.requireNonNull(event.getSelectedOptions())) {
+                Role role = guildRoles.get(roleIds.indexOf(roleOption.getValue()));
+                menu.addOption(role.getName(), role.getId());
+            }
```
Tais993, line 209: 2 things.

I doubt you need the Objects#requireNonNull, it'll throw an NPE either way to my knowledge.

And the SelectOption is named "role" here, which is really misinforming the user since it's not a Role.
It's a SelectOption, thoughts on renaming it to a.e selectedOption?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        List<SelectOption> menuOptions =
+                Objects.requireNonNull(event.getInteraction().getComponent()).getOptions();
+
+        // Remove deselected roles
+        for (SelectOption option : menuOptions) {
+            Role role = guildRoles.get(roleIds.indexOf(option.getValue()));
+            if (!selectedRoles.contains(role)) {
+                Objects.requireNonNull(event.getGuild()).removeRoleFromMember(member, role).queue();
```
Tais993, line 228: You already run Objects#requireNonNull earlier on the Guild.

Instead of adding "random" requireNonNull in the code, you should validate whatever you can at the start of the method in the first few lines.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+
+            return;
+        }
+
+        List<SelectOption> menuOptions =
+                Objects.requireNonNull(event.getInteraction().getComponent()).getOptions();
+
+        // Remove deselected roles
+        for (SelectOption option : menuOptions) {
+            Role role = guildRoles.get(roleIds.indexOf(option.getValue()));
```
Tais993, line 226: This looks really error prone, if a role gets added everything breaks.

Maybe make the title the role name, and the value it's ID?
Not sure how the UX would be like exactly, but it makes the backend way more safe.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,265 @@
...
+            if (!selectedRoles.contains(role)) {
+                Objects.requireNonNull(event.getGuild()).removeRoleFromMember(member, role).queue();
+            }
+        }
+
+        // Add the selected roles to the member
+        for (Role role : guildRoles) {
+            if (selectedRoles.contains(role)) {
+                Objects.requireNonNull(event.getGuild()).addRoleToMember(member, role).queue();
```
Tais993, line 235: You already run Objects#requireNonNull earlier on the Guild.

Instead of adding "random" requireNonNull in the code, you should validate whatever you can at the start of the method in the first few lines.

Tais993 (changes requested):
Regardless of my comments, you did a good job 👍

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Removing this would mean if the user used `/role-select all` the message would be sent as a reply to the command which we decided wasn't the best right?

image

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: > Maybe make the title the role name, and the value it's ID?

This is how it currently is? Here is where I add the roles to the menu:
```JAVA
for (Role role : roles) {
    if (role.getName().equals("@everyone"))
        continue;
    menu.addOption(role.getName(), role.getId());
}
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Changed in 134824ebe25f2b2528c70328c7ec981835a88ba5

JJeeff248: Changed in 4f91e805565ad19111a4fcedac9fe3db01fe2165

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Changed in 134824ebe25f2b2528c70328c7ec981835a88ba5

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Changed in 134824ebe25f2b2528c70328c7ec981835a88ba5

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Changed in 134824ebe25f2b2528c70328c7ec981835a88ba5

> Instead of adding "random" requireNonNull in the code, you should validate whatever you can at the start of the method in the first few lines.

Also, thanks for the tip, Intellij just recommends having requireNotNull in a lot of places but I get it's not necessary all the time. I'll be sure to check things like this in the future :)

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: I'm not sure I follow how it would break sorry. All it's doing is checking the selected roles against the menu options. If the is a menu option that isn't selected it removes the role. 

Currently each menu item is made using the role's name and id. Using something like `selectedOption.getValue()` returns the id of the selected role.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: Could you elaborate?
```java
.setEphemeral(ephemeral)
```
I still run this method.
This is just a lot shorter and easier readable by that.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: You'll have to explain me what the complicated
```java
selectedRoles.add(guildRoles.get(roleIds.indexOf(role.getValue())));
```
is

From what I understand, you can just run
```java
guild.getRoleById(selectedOption.getValue());
```
instead

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: I'll resolve this for now in favor of the comment above

Zabuzard:
@JJeeff248 Your still on it? Or should someone else take over? :)

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: Never mind!

I completely missed the fact that one isn't a reply, I'll resolve this

JJeeff248:
I'm still on it, just been a busy week :)

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
JJeeff248: Gonna be honest, I just overcomplicated it without realising. Using the shorter version makes a lot more sense and def helps clean up the code a bit more. Will change this in the next commit

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        EmbedBuilder embedBuilder = new EmbedBuilder();
+        embedBuilder.setTitle(title);
+
+        embedBuilder.setDescription(description);
+
+        embedBuilder.setColor(new Color(24, 221, 136));
+
+        return embedBuilder.build();
```
Tais993, line 180:
```suggestion
        return new EmbedBuilder()
                .setTitle(title)
                .setDescription(description)
                .setColor(new Color(24, 221, 136))
                .build();
```

Tais993 (commented):
Not a full review, but noticed it so

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,261 @@
...
+                guildRoles.subList(guildRoles.indexOf(highestBotRole) + 1, guildRoles.size()));
+
+        if (minValues != null) {
+            menu.setMinValues(minValues);
+        }
+
+        menu.setPlaceholder(placeHolder).setMaxValues(roles.size());
+
+        for (Role role : roles) {
+            if (role.getName().equals("@everyone") || role.getTags().getBotId() != null)
```
Heatmanofurioso, line 150: You should add brackets to this if
Although it's easy to understand, it is a bit error-prone if someone in the future tries to add conditions to it before the continue and doesn't notice the brackets

marko-radosavljevic: I was surprised that this doesn't get picked up by our static analyzers. So I activated this rule: [Control structures should use curly braces](https://sonarcloud.io/organizations/togetherjava/rules?open=java%3AS121&rule_key=java%3AS121). :+1:

It's a major issue, so it doesn't fail the build, but it will be visible in bot response and sonarcloud dashboard... speaking of bot, why bot didn't respond to this PR… :(

Heatmanofurioso: Same here.
Another reason for this change, just so `JJeeff248` understands, is consistency.
Since we use curly brackets everywhere, it's usually a good practice to be consistent in the way you structure your code

JJeeff248: All added, it makes sense to keep the code consistent, I'll keep it in mind for future :)

Fixed in b76a15fb4e674f7cef9645f756a25ff188181d53

JJeeff248:
**Where I'm at with this PR**
I am unsure how / if I should add the option for emotes to be included in the menu. I feel like if it were done *wrong* then it might clutter the menu but it could also add to it and make it a bit nicer to use. Looking for more opinions on this please.

I am also unable to test if role icons get displayed in the selection menu and am wanting to know,
1. If someone would be able to test it out?
2. If we would want this if they do/don't show up?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param menu the menu to add options to {@link SelectionMenu.Builder}
+     * @param placeHolder the placeholder for the menu {@link String}
+     * @param minValues the minimum number of selections. nullable {@link Integer}
+     */
+    private static void addMenuOptions(@NotNull SlashCommandEvent event,
+            @NotNull SelectionMenu.Builder menu, @NotNull String placeHolder,
+            @Nullable Integer minValues) {
+
+        Role highestBotRole =
+                Objects.requireNonNull(event.getGuild()).getSelfMember().getRoles().get(1);
```
Tais993, line 137: Wouldn't the highest role be 0?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+                        "The {} ({}) role has been removed but is still an option in the selection menu",
+                        selectedOption.getLabel(), selectedOption.getValue());
+                continue;
+            }
+
+            if (selectedRoles.contains(role)) {
+                guild.addRoleToMember(member, role).queue();
+            } else {
+                event.getGuild().removeRoleFromMember(member, role).queue();
+            }
```
Tais993, line 233: Instead, use Guild#modifyMemberRoles and add/remove all roles together, if possible

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+    private static @Nullable String handleOption(@Nullable OptionMapping option) {
+
+        if (option == null) {
+            return null;
+        }
+
+        if (option.getType() == OptionType.STRING) {
+            return option.getAsString();
+        } else if (option.getType() == OptionType.BOOLEAN) {
+            return option.getAsBoolean() ? "true" : "false";
```
Tais993, line 257:
```suggestion
            return option.getAsBoolean().toString();
```

Not sure, but I'd assume toStringing would also result in `true` or `false`

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+                .setActionRow(menu.build())
+                .queue();
+
+            event.reply("Message sent successfully!").setEphemeral(true).queue();
+
+            return;
+        }
+
+        List<SelectOption> menuOptions =
+                Objects.requireNonNull(event.getInteraction().getComponent()).getOptions();
```
Tais993, line 216:
```suggestion
                event.getInteraction().getComponent().getOptions();
```

Why throw a NPE when you can throw a NPE?

Exactly, if you remove the requireNonNull the code is a ton shorter, while it's behavior doesn't change.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+
+        Guild guild = Objects.requireNonNull(event.getGuild());
+
+        List<Role> selectedRoles = new ArrayList<>();
+        for (SelectOption selectedOption : Objects.requireNonNull(event.getSelectedOptions())) {
+            Role selectedRole = guild.getRoleById(selectedOption.getValue());
+            if (selectedRole != null && guild.getSelfMember().canInteract(selectedRole)) {
+                selectedRoles.add(selectedRole);
+            }
+        }
```
Tais993, line 192: Thoughts on using the Stream API instead?

```java
event.getSelectedOptions()
   .stream()
   .map(selectedOption -> guild.getRoleById(selectedOption.getValue())
   .filter(Objects::requireNonNull) 
   .filter(role -> guild.getSelfMember().canInteract(role))
   .toList();
``` 

Not sure how advanced you are with Java, so if you don't understand how this works, feel free to ask ^^

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,262 @@
...
+                guildRoles.subList(guildRoles.indexOf(highestBotRole) + 1, guildRoles.size()));
+
+        if (minValues != null) {
+            menu.setMinValues(minValues);
+        }
+
+        menu.setPlaceholder(placeHolder).setMaxValues(roles.size());
+
+        for (Role role : roles) {
+            if (role.getName().equals("@everyone") || role.getTags().getBotId() != null) {
```
Tais993, line 150:
```suggestion
            if (role.getIdLong() == event.getGuild().getIdLong() || role.getTags().getBotId() != null) {
```
You can easily create a role named `@everyone`
So instead, check or the role's ID is equal to the guild's ID.

Thankfully, the real `@everyone` role is always equal to the guild's ID


And to use Role#getTags you need to enable this in the cache, which I highly doubt is enabled.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993, line 137:
```suggestion
                event.getGuild().getSelfMember().getRoles().get(1);
```

Tais993 (changes requested):

Tais993:
> I am also unable to test if role icons get displayed in the selection menu and am wanting to know,

No they aren't shown in the selection menu, there's no way for Discord to know the values you gave are roles, thus it won't do anything with it.

>     1. If someone would be able to test it out?

See above ^^

>     2. If we would want this if they do/don't show up?

Can't change it, once selection menus with Roles come in Discord & JDA it probably becomes possible to show them.

Tais993:
@JJeeff248 do you've time this week or possibly next week to finish this PR, otherwise I (or maybe someone else) could finish it

JJeeff248:
Sorry I've been inactive. I wont have much free time to finish off this command for awhile unfortunately. I am happy for someone else to take over and finish it up :)

Tais993:
Will take over ^^

PR #303, opened 2021-12-05
Fixed Voice Activity Command crashing on high ages

### LINKED ISSUE #300: VC Activity command crashes
**Describe the bug**
The vc activity command crashes when the max-age exceeds about 1 million.

**Screenshots**
image

--- 

This could be fixed by, for example, implementing a max value for the max-age.

RealYusufIsmail:
fixed the issue were when age option was set to a huge number caused the command to crash. The max is 365 days and the min is 1 day now. 

This closes #300

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -38,8 +38,8 @@ import java.util.concurrent.TimeUnit;
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
-    private static final int MIN_AGE_DURATION = 0;
-    private static final long MAX_AGE_DURATION = TimeUnit.DAYS.toSeconds(7);
+    private static final int  MIN_AGE_DURATION_SECONDS = 0;
+    private static final long MAX_AGE_DURATION_SECONDS = TimeUnit.DAYS.toSeconds(7);
 
     private static final int MIN_USE_AMOUNT = 0;
     private static final int MAX_USE_AMOUNT = 100;
@@ -95,7 +95,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
                         .setRequiredRange(MIN_USE_AMOUNT, MAX_USE_AMOUNT),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                        .setRequiredRange(MIN_AGE_DURATION, MAX_AGE_DURATION));
+                        .setRequiredRange(MIN_AGE_DURATION_SECONDS, MIX_AGE_DURATION_SECONDS));
 
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -20,6 +20,7 @@ import org.slf4j.LoggerFactory;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.commands.SlashCommandVisibility;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -36,6 +37,8 @@ import java.util.Optional;
  */
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
+
+    private static final Duration DURATION = Duration.ofDays(7);
 
     private static final String APPLICATION_SUBCOMMAND = "application";
 
@@ -88,7 +91,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
                         .setRequiredRange(0, 100),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                        .setRequiredRange(0, 604800));
+                        .setRequiredRange(0, DURATION.getSeconds()));
 
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -95,7 +95,8 @@ public final class VcActivityCommand extends SlashCommandAdapter {
                         .setRequiredRange(MIN_USE_DURATION, MAX_USE_DURATION),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                        .setRequiredRange(MIN_AGE_DURATION.getSeconds(), MAX_AGE_DURATION.getSeconds()));
+                        .setRequiredRange(MIN_AGE_DURATION.getSeconds(),
+                                MAX_AGE_DURATION.getSeconds()));
 
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -87,7 +87,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
                     "The amount of times the invite can be used, default is infinity", false),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                    .setRequiredRange(1, 365));
+                        .setRequiredRange(1, 365));
 
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -45,7 +45,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
     private static final String MAX_USES_OPTION = "max-uses";
     private static final String MAX_AGE_OPTION = "max-age";
 
-    public static final String YOUTUBE_TOGETHER_NAME = "YouTube Together";
+    public static final String YOUTUBE_TOGETHER_NAME = "YouTube_Together";
     public static final String POKER_NAME = "Poker";
     public static final String BETRAYAL_IO_NAME = "Betrayal.io";
     public static final String FISHINGTON_IO_NAME = "Fishington.io";
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -85,7 +85,7 @@ public final class VcActivityCommand extends SlashCommandAdapter {
     private static final List<OptionData> inviteOptions = List.of(
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
                     "The amount of times the invite can be used, default is infinity", false)
-                    .setRequiredRange(0, 100),
+                        .setRequiredRange(0, 100),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
                         .setRequiredRange(0, 604800));
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -162,10 +162,18 @@ public final class VcActivityCommand extends SlashCommandAdapter {
             applicationName =
                     getKeyByValue(VC_APPLICATION_TO_ID, applicationId).orElse("an activity");
         }
+        OptionMapping maxUsesOption = event.getOption(MAX_USES_OPTION);
+        OptionMapping maxAgeOption = event.getOption(MAX_AGE_OPTION);
 
-        handleSubcommand(event, voiceChannel, applicationId,
-                event.getOption(MAX_USES_OPTION).getAsLong(),
-                event.getOption(MAX_AGE_OPTION).getAsLong(), applicationName);
+
+        if (maxUsesOption == null || maxAgeOption == null) {
+            handleSubcommand(event, voiceChannel, applicationId, 0, 86400, applicationName);
+        } else {
+            handleSubcommand(event, voiceChannel, applicationId, (int) maxUsesOption.getAsLong(),
+                    (int) maxAgeOption.getAsLong(), applicationName);
+        }
+
+
     }
 
     private static <K, V> @NotNull Optional<K> getKeyByValue(@NotNull Map<K, V> map,
@@ -181,12 +189,12 @@ public final class VcActivityCommand extends SlashCommandAdapter {
 
     private static void handleSubcommand(@NotNull SlashCommandEvent event,
             @NotNull VoiceChannel voiceChannel, @NotNull String applicationId,
-            @Nullable Long maxUses, @Nullable Long maxAge, @NotNull String applicationName) {
+            @Nullable int maxUses, @Nullable int maxAge, @NotNull String applicationName) {
 
         voiceChannel.createInvite()
             .setTargetApplication(applicationId)
-            .setMaxUses(Math.toIntExact(maxUses))
-            .setMaxAge(Math.toIntExact(maxAge))
+            .setMaxUses(maxUses)
+            .setMaxAge(maxAge)
             .flatMap(invite -> replyInvite(event, invite, applicationName))
             .queue(null, throwable -> handleErrors(event, throwable));
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
@@ -38,7 +38,11 @@ import java.util.Optional;
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
-    private static final Duration DURATION = Duration.ofDays(7);
+    private static final int MIN_AGE_DURATION = 0;
+    private static final Duration MAX_AGE_DURATION = Duration.ofDays(7);
+
+    private static final int MIN_USE_DURATION = 0;
+    private static final int MAX_USE_DURATION = 100;
 
     private static final String APPLICATION_SUBCOMMAND = "application";
 
@@ -88,10 +92,10 @@ public final class VcActivityCommand extends SlashCommandAdapter {
     private static final List<OptionData> inviteOptions = List.of(
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
                     "The amount of times the invite can be used, default is infinity", false)
-                        .setRequiredRange(0, 100),
+                        .setRequiredRange(MIN_USE_DURATION, MAX_USE_DURATION),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                        .setRequiredRange(0, DURATION.getSeconds()));
+                        .setRequiredRange(MIN_AGE_DURATION, MAX_AGE_DURATION.getSeconds()));
 
 
     /**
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -86,7 +86,8 @@
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
                     "The amount of times the invite can be used, default is infinity", false),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
-                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false));
+                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
+                        .setRequiredRange(1, 365));
```
Zabuzard, line 90: 1 to 365 seconds is definitely not correct.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed to 604800 and that is correct since it says in the guild.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -84,9 +84,11 @@
 
     private static final List<OptionData> inviteOptions = List.of(
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
-                    "The amount of times the invite can be used, default is infinity", false),
+                    "The amount of times the invite can be used, default is infinity", false)
+                        .setRequiredRange(0, 100),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
-                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false));
+                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
+                        .setRequiredRange(0, 604800));
```
Zabuzard, line 91: This number is an unreadable mess. Use the `java.time` API to make it clear what this is.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed it to use java.time.Duration

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -121,10 +126,10 @@ public VcActivityCommand() {
     public void onSlashCommand(@NotNull SlashCommandEvent event) {
         Member member = Objects.requireNonNull(event.getMember(), "member is null");
         GuildVoiceState voiceState = Objects.requireNonNull(member.getVoiceState(),
-                "Voicestates aren't being cached, check the JDABuilder");
+                "Voice states aren't being cached, check the JDABuilder");
```
Tais993, line 129: Not for this PR and disagree

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -121,10 +126,10 @@ public VcActivityCommand() {
...
 
         if (!voiceState.inVoiceChannel()) {
-            event.reply("You need to be in a voicechannel to run this command!")
+            event.reply("You need to be in a voice channel to run this command!")
```
Tais993, line 132: Not for this PR

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -37,6 +38,8 @@
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
+    private static final Duration DURATION = Duration.ofDays(7);
```
Tais993, line 41: What duration?

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -84,9 +87,11 @@
 
     private static final List<OptionData> inviteOptions = List.of(
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
-                    "The amount of times the invite can be used, default is infinity", false),
+                    "The amount of times the invite can be used, default is infinity", false)
+                        .setRequiredRange(0, 100),
```
Tais993, line 91: Use constants instead of magic numbers

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -84,9 +87,11 @@
...
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
-                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false));
+                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
+                        .setRequiredRange(0, DURATION.getSeconds()));
```
Tais993, line 94: Use constants instead of magic numbers

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed to MAX_USE_DURATION

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed to MIN_USE_AMOUNT and MAX_USE_AMOUNT

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed to MIN_AGE_DURATION and MAX_AGE_DURATION

Tais993 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -95,7 +95,7 @@
                         .setRequiredRange(MIN_USE_DURATION, MAX_USE_DURATION),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
                     "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
-                        .setRequiredRange(MIN_AGE_DURATION, MAX_AGE_DURATION.getSeconds()));
+                        .setRequiredRange(MIN_AGE_DURATION.getSeconds(), MAX_AGE_DURATION.getSeconds()));
```
Zabuzard: Just do `0, TimeUnit.DAYS.toSeconds(7)` instead.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -38,7 +38,7 @@
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
-    private static final int MIN_AGE_DURATION = 0;
+    private static final Duration MIN_AGE_DURATION = Duration.ofSeconds(0);
     private static final Duration MAX_AGE_DURATION = Duration.ofDays(7);
```
Zabuzard: Remove those obsolete fields-

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
illuminator3: Tijs told him to add them.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
Zabuzard: Imo mini fixes like typos can be done in another PR as well. Other than that I would agree though.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
Tais993: ^^^
To keep consistency, and against magic numbers.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Changed it now.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -37,6 +38,12 @@
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
+    private static final int MIN_AGE_DURATION = 0;
+    private static final long MAX_AGE_DURATION = TimeUnit.DAYS.toSeconds(7);
```
Zabuzard, line 42: The constants should have the unit in their name to prevent confusion. `MIN/MAX_AGE_DURATION_SECONDS`

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -84,9 +91,11 @@
 
     private static final List<OptionData> inviteOptions = List.of(
             new OptionData(OptionType.STRING, MAX_USES_OPTION,
-                    "The amount of times the invite can be used, default is infinity", false),
+                    "The amount of times the invite can be used, default is infinity", false)
+                        .setRequiredRange(MIN_USE_AMOUNT, MAX_USE_AMOUNT),
             new OptionData(OptionType.INTEGER, MAX_AGE_OPTION,
-                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false));
+                    "Max age in seconds. Set this to 0 to never expire, default is 1 day", false)
+                        .setRequiredRange(MIN_AGE_DURATION, MAX_AGE_DURATION));
```
Zabuzard, line 98: Since you are using min and max now, is the `handleIntegerTypeOption` still needed? If not, remove. (@Tais993 please confirm, thanks)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
Tais993: The logic checking or it's a valid Integer isn't needed, but the null checks and such still are.

Zabuzard: Okay, then lets just adjust it a bit and all cool, I guess.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Fixed

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -38,8 +38,8 @@
 public final class VcActivityCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(VcActivityCommand.class);
 
-    private static final int MIN_AGE_DURATION = 0;
-    private static final long MAX_AGE_DURATION = TimeUnit.DAYS.toSeconds(7);
+    private static final int  MIN_AGE_DURATION_SECONDS = 0;
```
Zabuzard: You have two spaces now `int__MIN`, also shows me that you pushed without applying spotless - dont.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: Fixed

Zabuzard: Its not. You changed completely different parts of the code.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
RealYusufIsmail: I ran spotless. I will change it manually

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -157,22 +153,22 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
+        OptionalInt maxAge;
 
         // the user already received the error in the handleIntegerTypeOption method
         // it still throws to tell us to return this method and stop the proceeding code
         try {
-            maxAge = handleIntegerTypeOption(event, maxAgeOption);
+            maxAge = getIntegerOptionIfPresent(maxAgeOption);
         } catch (IllegalArgumentException ignore) {
             return;
         }
```
Zabuzard, line 174: The try-catch is completely obsolete, this exception can not be thrown here...

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -191,7 +187,8 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
                     getKeyByValue(VC_APPLICATION_TO_ID, applicationId).orElse("an activity");
         }
 
-        handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge, applicationName);
+        handleSubcommand(event, voiceChannel, applicationId, maxUses.getAsInt(), maxAge.getAsInt(),
```
Zabuzard, line 190: And what if the optional is not present? Then this `.getAsInt()` call crashes! You clearly havent tested this. This is a critical bug.

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -235,48 +232,18 @@ private static void handleErrors(@NotNull SlashCommandEvent event,
...
      * <p>
      * <p/>
      *
-     * @param event the {@link SlashCommandEvent}
-     * @param optionMapping the {@link OptionMapping}
-     * @return nullable {@link Integer}
-     * @throws java.lang.IllegalArgumentException if the option's value is - outside of
-     *         {@link Integer#MAX_VALUE} - negative
+     * @param optionMapping the {@link OptionMapping} to validate
      */
```
Zabuzard, line 241: The Javadoc is in a bad state right now, the text is hard to understand and not fitting. Also, what is this `<p><p/>` mess? And why do you take in the `event`, it is not even used!

Rewrite:
```java
* Interprets the option as integer, if present.
*
* @param option the option to interpret or null if not present
* @return the integer contained in the option, if present
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -235,48 +232,18 @@ private static void handleErrors(@NotNull SlashCommandEvent event,
...
-                .setEphemeral(true)
-                .queue();
-            throw new IllegalArgumentException(optionMapping.getName() + " can't be negative");
-        }
-
-
-        return optionValue;
+    private static OptionalInt getIntegerOptionIfPresent(@Nullable OptionMapping optionMapping) {
+        return Optional.ofNullable(optionMapping)
+            .stream()
```
Zabuzard, line 244: NIT: I dislike the `.stream()` hack. Its a dirty abuse of the API. I know its very common on StackOverflow though... your choice.

```java
return optionMapping == null
  ? OptionalInt.empty();
  : OptionalInt.of(Math.toIntExact(optionMapping.getAsLong()));
```

Also, gets rid of the unecessary boxing conversion from `long` to `Long` that `map(OptionMapping::getAsLong)` introduces...

Zabuzard (changes requested):

Tais993 (changes requested):
As I explained in the original PR, and in the discord last night.

Using optional is useless, null is a valid value, using null makes it a ton easier.

If we set the max-age to null, it'll use the default value.
Otherwise we've to add an if check and such, which just add code without any purpose

Zabuzard:
> As I explained in the original PR, and in the discord last night.
> 
> Using optional is useless, null is a valid value, using null makes it a ton easier.
> 
> If we set the max-age to null, it'll use the default value. Otherwise we've to add an if check and such, which just add code without any purpose

Are you talking about the bigger context, i.e. where the `OptionalInt` will be used again later? I.e. the setter where its passed to?

Not sure why you think you have to add ifs, you can just write `setFoo(value.orElse(null))`. But sure, I havent thought about the use-site yet, only about the creation-site.

That said, I could argue that methods should be designed in an isolated fashion and not with how their call-site will use them in mind - and then `OptionalInt` is cleaner. But its a private method anyways, so its only purpose is to assist the surrounding code as best as it can, making `@Nullable Integer` totally valid again.

**tl;dr**: I dont care, I accept both variants now.

Tais993:
You can just use orElse yeah, but adds code, which I don't like since it doesn't improve anything.

I'd go for Integer, the author of this PR may decide.

RealYusufIsmail:
Will continue next week I have had a tests this week including one mock and 2 normal test tomorrow

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -187,8 +163,9 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
                     getKeyByValue(VC_APPLICATION_TO_ID, applicationId).orElse("an activity");
         }
 
-        handleSubcommand(event, voiceChannel, applicationId, maxUses.getAsInt(), maxAge.getAsInt(),
-                applicationName);
+        handleSubcommand(event, voiceChannel, applicationId,
+                event.getOption(MAX_USES_OPTION).getAsLong(),
+                event.getOption(MAX_AGE_OPTION).getAsLong(), applicationName);
```
Tais993: This code doesn't even work, besides that, still extremely disagree to usage of Optional.

Optional exists to "wrap" around null, so handling becomes "easier" (subjective)
So why for gods sake, use Optional if it's literally a value that we encourage. (in this context)

This means you need to use `optional.orElse(null)` which honestly, is really damn useless 
Adds code, for no reason. Not for readability, maintainability, just, because you can?

Completely disagree with that.

Also talked about why we shouldn't use Optional on your [other PR](https://github.com/Together-Java/TJ-Bot/pull/303#pullrequestreview-824980517)

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -204,12 +181,12 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
             @NotNull VoiceChannel voiceChannel, @NotNull String applicationId,
-            @Nullable Integer maxUses, @Nullable Integer maxAge, @NotNull String applicationName) {
+            @Nullable Long maxUses, @Nullable Long maxAge, @NotNull String applicationName) {
 
         voiceChannel.createInvite()
             .setTargetApplication(applicationId)
-            .setMaxUses(maxUses)
-            .setMaxAge(maxAge)
+            .setMaxUses(Math.toIntExact(maxUses))
+            .setMaxAge(Math.toIntExact(maxAge))
```
Tais993: Why use a long if we want an int?
Currently the flow is like:

Int -> long -> int

Why not just keep it an int?

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/VcActivityCommand.java
```diff
@@ -182,7 +163,16 @@ public void onSlashCommand(@NotNull SlashCommandEvent event) {
...
-        handleSubcommand(event, voiceChannel, applicationId, maxUses, maxAge, applicationName);
+        OptionMapping maxUsesOption = event.getOption(MAX_USES_OPTION);
+        OptionMapping maxAgeOption = event.getOption(MAX_AGE_OPTION);
+
+        if (maxUsesOption == null || maxAgeOption == null) {
+            handleSubcommand(event, voiceChannel, applicationId, 0, 86400, applicationName);
+        } else {
+            handleSubcommand(event, voiceChannel, applicationId,
+                    getOptionAsInteger(maxUsesOption).getAsInt(),
+                    getOptionAsInteger(maxAgeOption).getAsInt(), applicationName);
```
krankkkk, line 174: - The If doesn't really make sense, if just one of the 2 is null you also ignore the other
 -  Whats the point of the OptionalInt? In else they will always exist.
 - Why not squash the if/else into an `getOptionalAsInteger(...).orElse(defaultValue)`?

krankkkk: Also, magic numbers what does 0 / 86400 mean in this context?

Zabuzard:
Ownership of this is transferred.

Zabuzard:
The current state of this PR is not very helpful, its probably better to do it from scratch again with all the learnings earned in the discussions here. Closing for now.

PR #249, opened 2021-11-06
Migrate /warn command

RealYusufIsmail:
# Overview

This adds the `/warn user reason` command, which lets moderators warn users. Warnings can then be later retrieved using `/audit` (#312 ).

As usual, the command will try to DM the target to inform them about the action, have the usual permission checks (heavy moderation role in this case) and similar. Nothing really special to see here.

Part of #63 .

- [x] The command saves the warns
- [x] Users without ban perms can not use the warn command

PR #312, opened 2021-12-10
Add /audit command

### LINKED ISSUE #301: Create an Audit Command
This will used to store informations such as bans, kicks and more. This will be useful since a mod can check why the user was banned or kicked for example

This command will be done by doing `/audit @user` for example on which the bot will show there actions

Zabuzard:
# Overview

This adds the `/audit user` command, which is basically a migration of Dynos `?modlogs` and `?warnings` commands. It displays all moderative actions that have been taken against a user:

empty case

and

base use case

Closes #301 .

The primary use case for this is to review **warnings** (shipped by #249 ).

The actions are currently displayed **chronologically ascending**. And there is a short summary in the beginning. The "description" text in the embeds is the reason of the action (sorry, I was too lazy to type something nice during testing 😆 ).

# Notes

Only actions known to the `ModerationActionsStore` can be retrieved and displayed. For now, this means that only actions issued by our own commands are shown (`/ban`, `/kick`). In the future however, we could also think about adding a routine/listener that populates our store with actions done through the Discord UI or other bots.

The cool thing about how `/audit` is designed now (being solely based on the **store** (database table)) is that it is fully **forward compatible** with commands added in the future. In particular `/mute`, `/unmute` (#252 ) and `/warn` (#249 ), as well as **temporary actions** like temporary bans or mutes (#267 ).

In order to use the command, one needs the `heavyModerationRole` (**Moderator**) and the bot, as well as the user need enough permission to **interact** with the target user.

# Future ideas

* The summary in the beginning can also be used to give the **warning count** when #249 is added.
* I was thinking about adding different colors for the embeds depending on type (red for BAN, green for MUTE, yellow for warning, ...), but then I would need to edit that `Action` enum. Possible, but I first wanted to hear your opinions about this.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/ModerationUtils.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/ModerationUtils.java
@@ -25,7 +25,7 @@ enum ModerationUtils {
      * {@link Guild#ban(User, int, String)}.
      */
     private static final int REASON_MAX_LENGTH = 512;
-    private static final Color AMBIENT_COLOR = Color.decode("#895FE8");
+    static final Color AMBIENT_COLOR = Color.decode("#895FE8");
 
     /**
      * Checks whether the given reason is valid. If not, it will handle the situation and respond to
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Commands.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Commands.java
@@ -6,10 +6,7 @@ import org.togetherjava.tjbot.commands.basic.PingCommand;
 import org.togetherjava.tjbot.commands.basic.VcActivityCommand;
 import org.togetherjava.tjbot.commands.free.FreeCommand;
 import org.togetherjava.tjbot.commands.mathcommands.TeXCommand;
-import org.togetherjava.tjbot.commands.moderation.BanCommand;
-import org.togetherjava.tjbot.commands.moderation.KickCommand;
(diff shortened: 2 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/moderation/ModerationUtils.java
```diff
@@ -25,7 +25,7 @@
      * {@link Guild#ban(User, int, String)}.
      */
     private static final int REASON_MAX_LENGTH = 512;
-    private static final Color AMBIENT_COLOR = Color.decode("#895FE8");
+    static final Color AMBIENT_COLOR = Color.decode("#895FE8");
```
RealYusufIsmail, line 28: What is the point of having it inModerationUtils. What if you want to use that colour for a non moderation command. If you want to keep it their maybe making it public would be better.

RealYusufIsmail (changes requested):

Heatmanofurioso (approved):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ModerationUtils.java
Zabuzard: I do not understand what your suggestion, sorry.

The color was already in `ModerationUtils` beforehand. I just changed it from `private` to package-visible. So that I can use it in `/audit` to streamline the color used by moderation commands...

RealYusufIsmail: oh ok

RealYusufIsmail (approved):

PR #318, opened 2021-12-10
Change free command message

illuminator3:
This will change
```
Thank you for asking a question in an available channel.
When a helper who can answer this question reads it they will help you.
Please be patient.

__Please do not post your question in other channels__
```
to
```
Thank you for asking in an available channel. A helper will be with you shortly!
Please be patient.

__**Do not post your question in other channels.**__
```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
@@ -12,7 +12,7 @@ enum UserStrings {
             Thank you for asking in an available channel. A helper will be with you shortly!
             Please be patient.
 
-            __**Do not post your question in other channels.**__
+            __Do not post your question in other channels.__
             """),
     MARK_AS_FREE("""
             This channel is now available for a question to be asked.
--- a/application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
@@ -10,7 +10,6 @@ import org.jetbrains.annotations.Nullable;
 enum UserStrings {
     NEW_QUESTION("""
             Thank you for asking in an available channel. A helper will be with you shortly!
-            Please be patient.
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
```diff
@@ -9,11 +9,10 @@
  */
 enum UserStrings {
     NEW_QUESTION("""
-            Thank you for asking a question in an available channel.
-            When a helper who can answer this question reads it they will help you.
+            Thank you for asking in an available channel. A helper will be with you shortly!
             Please be patient.
 
-            __Please do not post your question in other channels__
+            __**Do not post your question in other channels.**__
```
Zabuzard, line 15: I think the **bold** is a bit too much and takes away attention from the core message.

```suggestion
            __Do not post your question in other channels.__
```

Zabuzard (changes requested):
I like the change 👍

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
```diff
@@ -9,11 +9,10 @@
  */
 enum UserStrings {
     NEW_QUESTION("""
-            Thank you for asking a question in an available channel.
-            When a helper who can answer this question reads it they will help you.
+            Thank you for asking in an available channel. A helper will be with you shortly!
             Please be patient.
```
marko-radosavljevic, line 13: I feel like 'Please be patient' is redundant.

We have:
> A helper will be with you shortly!

and,

> Do not post your question in other channels.

I think our message is clear. Without it, it will be shorter and cleaner, and maybe less aggressive when we don't repeat the same thing 10 times. What do you think?

marko-radosavljevic (commented):
I like the changes, I think the response is overall more concise and clearer. :relaxed:

marko-radosavljevic (approved):

Tais993 (approved):

PR #324, opened 2021-12-14
Spring + Log4j2 Update on Logviewer

krankkkk:
This time really set log4j on 2.16+ (2.16 since 2.15 seems still to be vulnerable https://logging.apache.org/log4j/2.x/changes-report.html#a2.16.0)
Updated Spring to 2.6.1

### DIFF
--- a/logviewer/build.gradle
+++ b/logviewer/build.gradle
@@ -40,13 +40,13 @@ dependencies {
         version {
             require '2.16.0'
         }
-        because 'Log4Shell happend'
+        because 'Log4Shell happened'
     }
     runtimeOnly('org.apache.logging.log4j:log4j-core') {
         version {
             require '2.16.0'
         }
-        because 'Log4Shell happend'
+        because 'Log4Shell happened'
     }
 
     runtimeOnly 'org.apache.logging.log4j:log4j-jul:2.16.0'
--- a/logviewer/build.gradle
+++ b/logviewer/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id "com.google.cloud.tools.jib" version "3.1.4"
-    id "org.springframework.boot" version "2.5.5"
+    id "org.springframework.boot" version "2.6.1"
     id "io.spring.dependency-management" version "1.0.11.RELEASE"
     id "com.vaadin" version "21.0.2"
     id 'java'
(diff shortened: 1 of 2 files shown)


logviewer/build.gradle
```diff
@@ -36,35 +36,47 @@ jooq {
...
+    }
+    runtimeOnly('org.apache.logging.log4j:log4j-core') {
+        version {
+            require '2.16.0'
+        }
+        because 'Log4Shell happend'
+    }
+
+    runtimeOnly 'org.apache.logging.log4j:log4j-jul:2.16.0'
+    runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl:2.16.0'
```
Zabuzard, line 53: * Why the different style and why this "because" thing?
* If we already update log4j, might also update it in the base application and not just in logviewer? (or is it already 2.16 there?)
* Small typo, it is "happened" not "happend"

Zabuzard (changes requested):

logviewer/build.gradle
krankkkk: Different Style is necessary to force gradle to never use a version lower than 2.16.0, even if transitive dependencies would like a lower version.
"because" Is a reason so that if someone look the dependency tree, it will show this reason why it's set at 2.16+

Can include the 2.16 from the base application in this PR if you want.

Typo removed

logviewer/build.gradle
```diff
@@ -36,35 +36,47 @@ jooq {
...
+            require '2.16.0'
+        }
+        because 'Log4Shell happened'
+    }
+    runtimeOnly('org.apache.logging.log4j:log4j-core') {
+        version {
+            require '2.16.0'
+        }
+        because 'Log4Shell happened'
+    }
```
RealYusufIsmail, line 50:
```suggestion
    implementation('org.apache.logging.log4j:log4j-api') {
        version {
            require '2.16.0'
        }
        because 'Log4Shell happened'
    }
    
    runtimeOnly('org.apache.logging.log4j:log4j-core') {
        version {
            require '2.16.0'
        }
        because 'Log4Shell happened'
    }
```

RealYusufIsmail (changes requested):

logviewer/build.gradle
RealYusufIsmail: Add gap to make the readability easer if you want.

logviewer/build.gradle
Zabuzard: > Can include the 2.16 from the base application in this PR if you want.

Yes, please. thanks

logviewer/build.gradle
Zabuzard: I say no, its better to group by topics (log4j) and have empty lines when the topic changes.

Zabuzard (approved):

RealYusufIsmail (approved):

Issue #311, opened 2021-12-10
Create /modmail command


Zabuzard:
The `/modmail` command is used to write a message to _a moderator_. The advantages over a simple DM are:

* the user does not need to know which users of a server are moderators
* the user does not need to befriend a moderator in case they have disabled non-friend DMs

The requirements for this are:

* the message must not be visible to other members
* the message is either visible to a single or all moderators

Optionally:

* the command should also be usable if not part of the guild anymore, for example after getting banned (needs careful design though to not be abused - rate limiting etc)

---

One cheap and low hanging possibility could be to write the message into the `#mod_audit_log` channel by using `ModAuditLogWriter` introduced in #296 . To make the message invisible, just use `ephemeral`.

borgrel:
Making it not visible is easy, its an interaction so the bot reply just have to be epheral and poof, invisible command

However: there are plenty of people who will push and act badly UNTIL a mod is called and then they quickly backoff
there should be a visible and an invisible way to call mods?

(one to make a complaint and another to call for assistance)

Zabuzard:
> there should be a visible and an invisible way to call mods?

The way I see it, if someone wants quick visible mod attention they just ping `@Moderator help, there is a scammer`, while `/modmail` is for when you want to contact a moderator privately, for example if you want to complain about the behavior of another user or if you have been banned unjustified.

borgrel:
so the first global slash command use case we have

talentedasian:
I'd like to work on this if no one has started working on this command. Also, would the command be allowed to used practically everywhere (within the guild)  and not just on a particular channel such as `bots`?

Zabuzard:
> I'd like to work on this if no one has started working on this command. Also, would the command be allowed to be used practically everywhere (within the guild) and not just on a particular channel such as `bots`?

Yes, as all responses are **ephemeral**, it could be popped anywhere, no restrictions. But you do not have to do anything to control that, just a regular slash command handler would be enough, like any of the other slash commands we already have, `/ping` for example.

Zabuzard:
@talentedasian Why did you add "blocked"? By what is this issue blocked?

talentedasian:
The `ModAuditLogWriter` from #296 isn't merged yet

marko-radosavljevic:
@Zabuzard

>One cheap and low hanging possibility could be to write the message into the #mod_audit_log channel by using ModAuditLogWriter introduced in #296 . To make the message invisible, just use ephemeral.

[...]

PR #329, opened 2021-12-15
/modmail slash command

talentedasian:

## Pull-request

- [x] I have a relating [issue][new_issue].

### Changes

- [x] Existing code
- [x] New feature

Closes Issue: #311

## Description

This is a draft of my PR on the `/modmail` command. The commit body contains
most of my work but I also wanted to address one more thing. I'm not entirely
sure if I should use the `findMemberWithRoles` on `Guild` class by JDA because:

* Well first I don't really understand the warning that it says ["You MUST NOT use blocking operations such as Task.get()! The response handling happens on the event thread by default."](https://ci.dv8tion.net/job/JDA/javadoc/net/dv8tion/jda/api/entities/Guild.html#findMembersWithRoles(net.dv8tion.jda.api.entities.Role...))
* Second, I could use just the `getMemberWithRoles` although it only takes the mods that are in thecache.
* If i decide to hardcode the mods to be chosen by the user who invoked the  command then I could remove the newly added config `guildId`. This was at the  bottom of my list; I don't even know all the mods in the server. 
* I could use buttons, but I prefer selection menus. Buttons hurt my eyes!!


Tais993:
There's a big issue in this code, it retrieves every single member of the server each time you call `findMemberWithRoles`
So this should only be used once preferably at all time, maybe even never depending on how you solve it.


So instead, retrieve once and create an selection menu in advance.

This means you find all the members with the role, create a selection menu based on the members found.
And you always re-use this selection menu.

To improve UX for the moderators, add a command only moderators can use to reload the list

talentedasian:
Right now, the private message does get sent and the buttons getting disabled as well but the interaction somehow fails.

talentedasian:
Most of the work now is done, I'll now put this up as ready for review.

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
```diff
@@ -0,0 +1,138 @@
...
+                    an unknown error has occurred.
+                    """);
+
+        event.reply("The moderator you chose is not on the list of moderators on the guild")
+            .setEphemeral(true)
+            .queue();
+    }
+
+    /**
+     * Creates a list of options containing the moderators for use in the modmail slash command.
```
talentedasian, line 123: Forgot to change the javadoc for this. I'll just do it tomorrow.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReloadModMailCommand.java
```diff
@@ -0,0 +1,45 @@
...
+        super("reloadmod", "Reloads the list of moderators in the modmail selection menu",
+                SlashCommandVisibility.GUILD);
+
+        this.jda = Objects.requireNonNull(jda);
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull SlashCommandEvent event) {
+        Guild guild = Objects.requireNonNull(jda.getGuildById(config.getGuildId()),
+                "A Guild is required to use this command. Perhaps the bot isn't on the guild yet");
```
Zabuzard, line 34: you should be able to do `event.getGuild()`, no?

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReloadModMailCommand.java
talentedasian: I could but I didn't due to some pedantic reason. If the command was called on Guild A, but the `guildId` was Guild B, then it would have inconsistent results.

talentedasian: Thinking about that, I think I could perhaps do `event.getGuild()` on the `ModMailCommand` as well.

Tais993: I do think you should SlashCommandEvent#getGuild ^^

If you run the command in Guild B, right now it runs it like it's in Guild A, which is really weird imo

The bot is only in 1 guild, and it'll never join another one

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
```diff
@@ -0,0 +1,138 @@
...
+
+public class ModmailCommand extends SlashCommandAdapter {
+    private static final Logger logger = LoggerFactory.getLogger(ModmailCommand.class);
+
+    private static final String COMMAND_NAME = "modmail";
+    private static final String TARGET_OPTION = "message";
+    private static final Config config = Config.getInstance();
+
+    static final List<SelectOption> mods = new ArrayList<>();
+    static final Map<String, User> modsMap = new HashMap<>();
```
Tais993, line 29: It shouldn't be done this way, keep the variables private, use the onReady event to load the options/mods.

So keep it within this class, make the variables private and don't make them static.

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull SlashCommandEvent event) {
+        String memberId = event.getUser().getId();
+
+        // checks if selection menu already contains the moderators
+        if (mods.size() == 1) {
+            loadMenuOptions();
+        }
```
Tais993, line 58: Remove this in favor of the onReady event, see above

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
```diff
@@ -0,0 +1,138 @@
...
+     * <p>
+     * This method also stores the moderators on a map for later use. The map's values are always
+     * and should be exactly the same with the previous results.
+     */
+    private void loadMenuOptions() {
+        Guild guild = Objects.requireNonNull(jda.getGuildById(config.getGuildId()),
+                "A Guild is required to use this command. Perhaps the bot isn't on the guild yet");
+
+        ModMailUtil.loadMenuOptions(guild);
+    }
```
Tais993, line 136: See review above, remove this in favor of the onReady event

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public void onSelectionMenu(@NotNull SelectionMenuEvent event, @NotNull List<String> args) {
+        String message = args.get(1);
+        // Ignore if another user clicked the button which is only possible when used within the
+        // guild.
+        String userId = args.get(0);
+        if (event.isFromGuild()
+                && !userId.equals(Objects.requireNonNull(event.getMember()).getId())) {
```
Tais993, line 79:
```suggestion
                && !userId.equals(event.getMember().getId())) {
```

You're effectively doing

```java
if (member == null) {
    member.toString();
}
```
Here, which is, useless

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModMailUtil.java
```diff
@@ -0,0 +1,79 @@
...
+     * <p/>
+     * The previous result stored by this method will be overwritten if there was any.
+     */
+    public static void loadMenuOptions(@NotNull Guild guild) {
+        clearMods();
+
+        Role modRole = getModRole(guild)
+            .orElseThrow(() -> new IllegalStateException("No moderator role found"));
+
+        guild.findMembersWithRoles(modRole).get().stream().forEach(mod -> {
```
Tais993, line 50: Why the get and the stream?

Just use onSuccess?

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/modmail/ModmailCommand.java
Tais993, line 29: Another note, NEVER store Discord entities.

They can be invalidated (garbage collected) which will result in really weird errors.

Instead, store their ID (or username in this case)

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/Commands.java
```diff
@@ -70,6 +71,10 @@
         commands.add(new MuteCommand(actionsStore));
         commands.add(new UnmuteCommand(actionsStore));
 
+        ModmailCommand modmailCommand = new ModmailCommand();
+        commands.add(modmailCommand);
+        commands.add(modmailCommand.new ReloadModmailCommand());
```
illuminator3, line 76: This will not compile.

illuminator3 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/Commands.java
talentedasian: Really? I'm sure I've tested and ran it on my machine numerous times.

talentedasian: I also ran gradle `build` so I'm really sure it works. Do you mind checking it again?

Tais993: It doesn't compile no, I'm not sure what you're doing there `modmailCommand.new ReloadModmailCommand()`

talentedasian: The `ReloadModmailCommand` is an inner class since it's the only solution I could think of to reload the mods in the `/modmail` command. Thus, I am [instantiating](https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html) the `ReloadModmailCommand` from the `Modmail` object.

Tais993: Ah that explains

talentedasian: I am genuinely confused why it doesn't compile. Do you mind helping me out on this one?

Tais993: Well, it compiles like you said, it's just done in a reallyyyyy weird way.
To my knowledge, you can remove `modmailCommand.` from `modmailCommand.new ReloadModmailCommand()`
So please do that, if you can't, refer to the class instead of an instance?

illuminator3: You can't. That's the required syntax for instantiating a non-static inner class. It's really weird.

Tais993:
@talentedasian do you've the time to work on this PR (this week, next week)? Otherwise I'll take this PR over so it can be merged

talentedasian:
I do but you can take over this PR if you want.

Tais993:
If you've the time, you can handle it, it takes some time setting up and you deserve most of the credits \:p

Zabuzard:
> If you've the time, you can handle it, it takes some time setting up and you deserve most of the credits :p

If you take over, you can also copy their branch and hence basically keep their commits. Or alternatively manually add them as author to a commit of urs (for credits).

Tais993:
I'll work in the branch in that case, your branch will be copied so you'll be credited ^^

Tais993:
This PR will be closed in favor of a new PR (will be linked)
 
 In advance, thanks for your help talentedasian! Your branch will be the base of the new PR

PR #341, opened 2022-01-12
Added event listener support for (slash) commands

### LINKED ISSUE #236: Add Message Listener support
The command system, as of now, only supports SlashCommands. However, we slowly get more and more commands with a need to listen for any incoming message.

The command system should offer a generic way to register message listeners, which should also be combinable with SlashCommands and other things (for example if using the interfaces).

The scope of this issue is only to add message listening support, not other events (such as user banned, message deleted, ...), just message arrived. Maybe also message edit events.

Examples for such commands are the `!bytecode` command (#207 ) and free-command system (#221 ).

### LINKED ISSUE #237: Add arbitrary JDA event support
The command system, as of now, only supports SlashCommands. However, we slowly get more and more commands with a need to listen for any sort of events.

While we do favor offering specific solutions (SlashCommands, MessageListener, ...), we should also offer a catch-them-all solution where commands can just listen to all incoming JDA events. It should be well integrated with existing solutions, i.e. it should be possible to create a slash command that also receives all other events (for example if using the interfaces).

There is no need to put a lot of effort into the arbitrary event support, since it is only meant as backup solution in case we do not offer more concrete and specific support. Commands that need this are for example the mod audit system (#225 )

Tais993:
SlashCommand received onEvent(GenericEvent), SlashCommandAdapter received an event method for each event (as example, onMessageReceived(MessageReceivedEvent)

 
Closes #237


application/src/main/java/org/togetherjava/tjbot/commands/system/CommandSystem.java
```diff
@@ -99,19 +101,21 @@ public CommandSystem(@NotNull JDA jda, @NotNull Database database) {
...
         // itself
         logger.debug("Command system is now ready");
 
         // Propagate the onReady event to all commands
         // NOTE 'registerReloadCommands' will not be finished running, this does not wait for it
-        nameToSlashCommands.values()
-            .forEach(command -> COMMAND_SERVICE.execute(() -> command.onReady(event)));
+        nameToSlashCommands.values().forEach(command -> {
+            COMMAND_SERVICE.execute(() -> command.onReady(event));
+        });
```
Tais993, line 118:
```suggestion
        nameToSlashCommands.values()
            .forEach(command -> COMMAND_SERVICE.execute(() -> command.onReady(event)));
```

Just a guess to what sonarcloud is talking about

Zabuzard:
I dont like this design. Goes away from what I had in mind. I think it would be better to keep slash commands and general event listeners separated. Otherwise, eventually, we end up with a behemoth of a class that can do everything instead of separating our concerns.

Maybe something up for discussion. Code-wise, looks fine though.

Zabuzard:
Actually, let me propose a PR shortly which adds message receiver and generic event receiver support the way I visioned it.

Tais993:
See new PR and issue

PR #342, opened 2022-01-14
Removes title for all AuditLog

### LINKED ISSUE #279: Remove title from mod audit log
The mod audit log messages have a **title**, but it interrupts the read-flow, since its displayed in between the author and the message. As in, it reads as _"John Message Deletion deleted messages from Bob"_ instead of _"John deleted messages from Bob"_.

example

To increase UX, I would simply **remove the title all together**. Alternatively, we could also move the title part into the **footer**, up to be decided by who implements this.

Tais993:
Removes the title from all auditlogs.

1 thing we'll need to discuss is whenever we also want to remove the `title` field from the Action enum, since those might be useful in the future? Not sure, opinions appreciated

Closes #279


application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
```diff
@@ -363,4 +362,4 @@ String getVerb() {
             return verb;
         }
     }
-}
```
Zabuzard, line 366: dont we have a consistent line ending setting? the never ending story of line ending diffs 😆

Zabuzard (approved):

Heatmanofurioso (approved):

Issue #235, opened 2021-10-30
Add Routine support


Zabuzard:
More and more commands need to execute code on a scheduled routine (for example every x minutes or once per day or once at a specific date).

The command system should offer a generic way to all commands (also, not just SlashCommands) to schedule their actions on a central managed scheduler.

The interface should be easy and convenient to use for all registered commands. It should at least offer ways to:

- ~~execute code at date/time (send birthday greetings to zabu at 8th november)~~
- ~~execute code after a specific waiting time (expire this message in 2 days)~~
- repeatedly execute code (check the audit logs every 5 hours)
- ~~optionally persist a schedule, so that it is not lost if the bot is shutdown (/remind me at ... command)~~

The exact needs are not yet known but commands wanting to use such mechanisms are slowly comming (mod audit log, top helper, free command, ...)

----

Edit: Removed any sort of persisted-one-shot-tasks from the scope of this issue (see comments for why).

Zabuzard:
> This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.

This is actually still relevant, we need it eventually.

Zabuzard:
I feel like this has to be split into two separate topics:

1. support for executing something on a schedule (the actual routine), e.g. every 5 mins (also, doesnt need persistance)
2. add a service that is accessible to all features, allowing them to schedule one-shot tasks at a given date/time or with a delay, which is persisted in case the bot shuts down.

Zabuzard:
Regarding the second part, we decided that there is no pleasant way to offer a generic scheduler solution for persisted-one-shot-tasks.

The problem is that its pretty much impossible to persist a generic `Runnable` (the task to persist) in the database in a non-fragile way. One possibility was to persist something like
```java
@FunctionalInterface
public interface Task extends Serializable {
  void run(JDA jda);
}
```
(with Java serialization or JSON or similar)
which at least has the benefit that it can inject a fresh JDA, while the user can choose what exactly to serialize and what not. However, it is still too fragile for practical use. For example, suppose a `/remind-me what:"happy birthday" when:2 years`, when this is executed in 2 years, chances are high that JDA (or another component in the code base) was updated in a way that makes the old serialized code invalid (for example, because the Discord end-point to post messages has changed).
Alternative approaches, such as simply storing the name of a static method to execute (so that only the location of the code but not the code itself is serialized), are not powerful enough and too limiting.

[...]

PR #345, opened 2022-01-17
Add Message and Event receivers

Zabuzard:
### Overview

This implements and closes long awaited #236 and #237 by adding two new interfaces `MessageReceiver` (+ `MessageReceiverAdapter`) and `EventReceiver` to the system.

Additionally, some refactoring had to be done to keep a good flow. First of all, `SlashCommand`, `MessageReceiver` and `EventReceiver` (and soon also `Routine`, see #235 ) are summarized under a marker interface called `Feature`. This marks all features that our bot core system supports.

Consequently, the previous starting class `Commands` has been renamed to `Features` and is now the origin of all features, not only commands (we already moved other stuff there in the past to prepare for this).

While at it, the existing class `RejoinMuteListener` has been adapted and now implements `EventReceiver` instead.

### Details

System-wise, `Features#createFeatures` returns a list of all features to `BotCore` and it will iterate and filter it based on the type (a sealed interface would be awesome here to not miss any future type).

For `SlashCommands`, nothing changed. `EventReceivers` are directly forwarded to `jda.addEventListener`, since that interface extends JDAs `EventListener`. `MessageReceivers` subscribe to channels they are interested in by a REGEX pattern matching channel names, `BotCore` takes care of the message event forwarding by iterating a simple `Map<Pattern, MessageReceiver>`.

### Examples

Quick example for a `MessageReceiver`, via its `MessageReceiverAdapter` helper:
```java
class Foo extends MessageReceiverAdapter {
  public Foo() {
    super(Pattern.compile("lobby"));
  }
  
  @Override
  void onMessageReceived(GuildMessageReceivedEvent event) {
    event.getMessage()
      .addReaction("🐸")
      .queue();
  }
}
```
a quick example for a `EventListener`:
```java
class Foo implements EventReceiver {
  @Override
  public void onEvent(GenericEvent event) {
    if (event instanceof GuildMessageReceivedEvent messageEvent) {
      System.out.println("Message was sent!");
    }
  }
}
```
and in both cases a simple addition in `Features.java`:
```java
features.add(new Foo());
```

### Remarks

When merging this, we have to go through our Wiki and do some minor adjusments caused by the renaming of `Commands` (now `Features`).

This PR competes against #341, which is a different design-approach to this problem, in which `SlashCommands` get the event listening capabilities directly instead of creating separate flows for them. (Which I am not a fan of, since this leads to creation of a behemoth class that can do everything instead of separating capabilities and concerns - SRP.)


borgrel:
While this adds great functionality, it defeats the original design of the bot?

The bot has a SINGLE command listener and manually validates / splits all incoming slash commands to drastically reduce overhead (by only having the correct command respond to the event instead of letting each command manually filter out invalid commands they dont need to respond to)

The message system should work the same way? currently every message listener responds to all message events and have to manually filter the messages they are not interested in, instead of only receiving messages they do want.

application/src/main/java/org/togetherjava/tjbot/commands/Features.java
```diff
@@ -0,0 +1,82 @@
...
+        // NOTE The system can add special system relevant commands also by itself,
+        // hence this list may not necessarily represent the full list of all commands actually
+        // available.
+        Collection<Feature> features = new ArrayList<>();
+
+        // Routines
+        // TODO This should be moved into some proper command system instead (see GH issue #235
+        // which adds support for routines)
+        new ModAuditLogRoutine(jda, database).start();
+        new TemporaryModerationRoutine(jda, actionsStore).start();
```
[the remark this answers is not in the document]
Zabuzard: The routine support is added shortly after this PR, #346 , solving the valid issue you are mentioning.

So for the context of this PR, I would say we keep this for now.

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
```diff
@@ -84,16 +84,16 @@
...
      *
      * @return the command data of this command
      */
     @NotNull
     CommandData getData();
 
     /**
-     * Triggered by the command system after system startup is complete. This can be used for
+     * Triggered by the core system after system startup is complete. This can be used for
      * initialisation actions that cannot occur during construction.
```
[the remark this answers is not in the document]
Zabuzard: Agreed. So you are suggestion to remove this `onReady` method and instead let current users (probably only the help system) use the alternatives that are now available?

Zabuzard:
> While this adds great functionality, it defeats the original design of the bot?

I am not sure I get your concern. I am in no way saying that people should use the new interfaces when all they want to do is to create a slash command. In particular, you should rarely have to use `EventReceiver` and instead use the alternatives `SlashCommand` and `MessageReceiver` if possible.

But it is also clear that there are a number of commands that do need more capabilities than just being a slash command (your help system is a good example for this). Hence why we need to introduce those features as well.
People can of course also combine things, `Foo implements SlashCommand, MessageReceiver`, if needed.

Or is your concern more about the fact that we have `jda.addEventListener(...)` on all the message/event receivers instead of letting them all go into `BotCore` and routing them ourselves? In that case, I was indeed also concidering this but at the end I couldn't see any benefit to this (other than introducing possible bugs). After all, why should we reinvent the existing routing logic inside JDA with our own if we have no benefit. Maybe you see benefit that I overlooked. For the slash commands there was clearly benefit since we had a need to decorate our slash command events with extra information.
The good thing is though that the detail of who routes the message/event receiver events is hidden inside `BotCore`, so it could be changed at any time if needed without impacting any user.

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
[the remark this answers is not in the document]
borgrel: no, the onready still has use / purpose

for example routine support could use it to log:
'INFO routine system loaded: 2 routines rolled back from invalid termination, 23 routines restarted, 5 old routines deleted'

we just need to remove the statement in the javadoc saying: 'use this to start things the bot doesnt support yet'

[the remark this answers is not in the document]
Zabuzard: Routines are not slash commands though. Routines will use `Routine` (#346 ). So it should then rather be moved into that interface then.

But I get your point and agree, the Javadoc has to be adjusted.

borgrel:
> Or is your concern more about the fact that we have `jda.addEventListener(...)` on all the message/event receivers instead of letting them all go into `BotCore` and routing them ourselves?

This

> The good thing is though that the detail of who routes the message/event receiver events is hidden inside `BotCore`, so it could be changed at any time if needed without impacting any user.

True enough, but not if we want to add internal routing


The fact is that few of our message listeners will want / need global message listening. for example the free system will only need to listen on the help channels. Illum's suggestion of a dad bot would only need to listen on lobby and geek speak. if the bot gets functionality that blocks ppl from talking in snippets (by deleting any posts after the first by the same user in a 24hr period) would only need to listen to messages in the snippets channel.

The way it currently is every single message in every single channel would be sent to every single listener and then every single listener has to filter out the 99% they dont need individually so assuming 12 listeners each message will be filtered 12 times and ignored 11 times, instead of being filtered once and only sent to the one listener that is interested in it.

The routing system is already designed and filters slash commands once to send them to the relevant command handler instead of just letting every command manually register its own listener and personally filter out the unneeded commands .... the routing should be expanded re-used to reduce overhead in message listeners in the same way

Zabuzard:
> The fact is that few of our message listeners will want / need global message listening. for example the free system will only need to listen on the help channels. Illum's suggestion of a dad bot would only need to listen on lobby and geek speak. if the bot gets functionality that blocks ppl from talking in snippets (by deleting any posts after the first by the same user in a 24hr period) it would only need to listen to messages in the snippets channel.

Okay, I think I get your point now. You are suggessting to not only add a general `MessageReceiver` but a more specific way instead where people will also tell you which channels they are interested in, to only receive some messages instead.
So basically you are saying that majority (if not all) message receivers are only interested in a small portion of channels (or maybe even other filters, such as by author) and we should take care of that filtering to make it easier for users and also get rid of the code duplication that occurs when everyone has to filter themselves.

And if someone really has an interest in all channels, they could always also still fall back to `EventReceiver` which is seen as the unfiltered cumbersome fallback option.

I like that idea and will consider it, thanks.

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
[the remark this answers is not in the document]
Zabuzard: I removed `onReady` now. The free-system now implements `EventReceiver` (instead of `EventListener`), which also gets a `ReadyEvent` handed over. And now that the `BotCore` adds it to `jda.addEventListener(...)`, it also gets triggered early enough. Seems more elegant that way.

application/src/main/java/org/togetherjava/tjbot/commands/MessageReceiver.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface MessageReceiver extends Feature {
+    /**
+     * Triggered by the core system whenever a new message was sent in a text channel of a guild the
+     * bot has been added to.
+     *
+     * @param event the event that triggered this, containing information about the corresponding
+     *        message that was sent
+     */
+    void onMessageSent(@NotNull GuildMessageReceivedEvent event);
```
borgrel, line 26: Shouldnt this be `onMessageReceived`?

`onMessageSent` is confusing

borgrel (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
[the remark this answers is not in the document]
borgrel: it looks like that will work.

technically what should happen is the JDA should be passed in on construction and the onReady code called in the constructor and then there is no need for generic events or the onready event (it was just being used as a hook to access the JDA without having to rewrite all of the code in slashadapter because commands didnt have access to JDA, Feature does.)

because the slashadapter and the messageadapter covers all functionality.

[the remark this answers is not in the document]
borgrel: but then again ...... maybe it should be left in `onReady`?
the Features construction is not threaded so heavy code there will block the bot's startup
and whether onReady is called by the JDA or by our bots router, it is threaded there.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -268,4 +281,22 @@ private static void onComponentIdRemoved(ComponentId componentId) {
...
+
+        @Override
+        public void onGuildMessageReceived(@NotNull GuildMessageReceivedEvent event) {
+            messageReceiver.onMessageSent(event);
+        }
+
+        @Override
+        public void onGuildMessageUpdate(@NotNull GuildMessageUpdateEvent event) {
+            messageReceiver.onMessageUpdated(event);
+        }
```
Tais993, line 300: I'd personally we say, we implement EventListener and do instanceof checks ourselves.

Currently it goes from onEvent -> onGuildMessageUpdate -> onMessageUpdated, I don't think we need the ListenerAdapter

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: Sounds reasonable.

application/src/main/java/org/togetherjava/tjbot/commands/SlashCommand.java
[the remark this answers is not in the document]
Zabuzard: i feel like once this is merged, the free command can be simplified and made nicer but i leave that up to u for another PR. it would be too much for this PR to adjust /free beyond "just making it work", i think.

application/src/main/java/org/togetherjava/tjbot/commands/MessageReceiver.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: done

Zabuzard:
`MessageReceiver` now subscribe to channels they are interested in by their name and `BotCore` takes care of the forwarding (as requested by @borgrel ).

In a future PR, `/free` could now be simplified to basically implementing `MessageReceiver` and using that channel-name-subscription system, probably simplifying its own logic quite a bit (and getting rid of the `EventReceiver` part).

Tais993 (dismissed):
It's good enough for now, it's blocking some other PR's (including #350)

We can change the subscription system for the Message events later one if required.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeCommand.java
```diff
@@ -285,9 +282,14 @@ private void checkBusyStatusAllChannels(@NotNull JDA jda) {
...
+                                     // the second case is unreachable; but it passes without
+                                     // pattern-matching. Probably a bug in SonarLint with Java 17.
     @Override
     public void onEvent(@NotNull GenericEvent event) {
-        if (event instanceof GuildMessageReceivedEvent guildEvent) {
+        if (event instanceof ReadyEvent readyEvent) {
+            onReady(readyEvent);
+        } else if (event instanceof GuildMessageReceivedEvent guildEvent) {
             if (guildEvent.isWebhookMessage() || guildEvent.getAuthor().isBot()) {
                 return;
```
java-coding-prodigy, line 294: I think we should put the whole `GuildMessageRevievedEvent` handling in a separate method. Why don't we just implement `MessageReciever` and have that block in the `onGuildMessageRecieved` method

application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
```diff
@@ -23,8 +23,8 @@
             Command not ready please try again in a minute.
             """),
     NOT_MONITORED_ERROR("This channel is not being monitored for free/busy status. If you"
-            + "believe this channel should be part of the free/busy status system, please discuss it"
-            + "with a moderator"),
+            + " believe this channel should be part of the free/busy status system, please discuss it"
```
illuminator3, line 26: Maybe change this to `, please consult a moderator`?

application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
```diff
@@ -271,9 +271,10 @@ public void updateStatusFor(@NotNull Guild guild) {
         }
         long channelId = guildIdToStatusChannel.get(guild.getIdLong());
         TextChannel channel = guild.getTextChannelById(channelId);
-        if (channel == null)
+        if (channel == null) {
             throw new IllegalStateException("Status channel %d does not exist in guild %s"
                 .formatted(channelId, guild.getName()));
+        }
```
illuminator3, line 277: New lines please.

illuminator3 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeCommand.java
Zabuzard: I agree, but this should not be done in this PR but in a dedicated PR dealing with the free-command. borgrel is already awaitng this to get merged to do exactly those changes.

For this PR, i only did like the bare minimum to integrate the existing code so that it still works as before. Any further change would likely be a bigger refactoring and also needs more testing.

java-coding-prodigy: I realized that from our discussion in the discord.

application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
Zabuzard: Done, please check.

application/src/main/java/org/togetherjava/tjbot/commands/free/UserStrings.java
Zabuzard: Done, please check.

application/src/main/java/org/togetherjava/tjbot/commands/MessageReceiver.java
```diff
@@ -0,0 +1,51 @@
...
+     * Retrieves the pattern matching the names of channels of which this receiver is interested in
+     * receiving sent messages from. Called by the core system once during the startup in order to
+     * register the receiver accordingly.
+     * <p>
+     * Changes on the pattern returned by this method afterwards will not be picked up.
+     *
+     * @return the pattern matching the names of relevant channels
+     */
+    @NotNull
+    Pattern getChannelNamePattern();
```
java-coding-prodigy, line 32:
```suggestion
    Predicate<AbstractChannel> getChannelPredicate();
```

application/src/main/java/org/togetherjava/tjbot/commands/MessageReceiverAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+     *        from matching channels will be received
+     */
+    protected MessageReceiverAdapter(@NotNull Pattern channelNamePattern) {
+        this.channelNamePattern = channelNamePattern;
+    }
+
+    @Override
+    public final @NotNull Pattern getChannelNamePattern() {
+        return channelNamePattern;
+    }
```
java-coding-prodigy, line 34:
```suggestion
    public final @NotNull Predicate<AbstractChannel> getChannelPredicate() {
       return channel -> channelNamePattern.matcher(channel.getName()).matches();
  }
```

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -29,40 +31,60 @@
...
-public final class CommandSystem extends ListenerAdapter implements SlashCommandProvider {
-    private static final Logger logger = LoggerFactory.getLogger(CommandSystem.class);
+public final class BotCore extends ListenerAdapter implements SlashCommandProvider {
+    private static final Logger logger = LoggerFactory.getLogger(BotCore.class);
     private static final String RELOAD_COMMAND = "reload";
     private static final ExecutorService COMMAND_SERVICE = Executors.newCachedThreadPool();
     private final Map<String, SlashCommand> nameToSlashCommands;
     private final ComponentIdParser componentIdParser;
     private final ComponentIdStore componentIdStore;
+    private final Map<Pattern, MessageReceiver> channelNameToMessageReceiver = new HashMap<>();
```
java-coding-prodigy, line 57:
```suggestion
    private final Map<MessageReciever, Predicate<AbstractChannel>> channelNameToMessageReceiver = new HashMap<>();
```

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -29,40 +31,60 @@
...
-            .stream()
+    public BotCore(@NotNull JDA jda, @NotNull Database database) {
+        Collection<Feature> features = Features.createFeatures(jda, database);
+
+        // Message receivers
+        features.stream()
+            .filter(MessageReceiver.class::isInstance)
+            .map(MessageReceiver.class::cast)
+            .forEach(messageReceiver -> channelNameToMessageReceiver
+                .put(messageReceiver.getChannelNamePattern(), messageReceiver));
```
java-coding-prodigy, line 76:
```suggestion
                .put(messageReceiver, messageReciever.getChannelPredicate()));
```

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -106,12 +128,30 @@ public void onReady(@NotNull ReadyEvent event) {
...
+
+    private @NotNull Stream<MessageReceiver> getMessageReceiversSubscribedTo(
+            @NotNull AbstractChannel channel) {
+        String channelName = channel.getName();
+        return channelNameToMessageReceiver.entrySet()
+            .stream()
+            .filter(patternAndReceiver -> patternAndReceiver.getKey()
+                .matcher(channelName)
+                .matches())
+            .map(Map.Entry::getValue);
```
java-coding-prodigy, line 154:
```suggestion
            .filter(patternAndReceiver -> patternAndReceiver.getValue().test(channel))
            .map(Map.Entry::getKey);
```

java-coding-prodigy (dismissed):
I have put my suggestions for the refactoring of our `MessageReciever` interface. However, you still have to change the imports, variable names, and documentation.

Tais993 (approved):

Zabuzard:
7 days without real code changes + approval -> merging

Zabuzard:
1 week without real code changes + 1 approval -> merging

PR #346, opened 2022-01-18
Add routine support

### LINKED ISSUE #235: Add Routine support

Zabuzard:
### Overview

Implements and closes #235. Adds an interface `Routine implements Feature`, which allows classes to be executed on a pre-defined fixed schedule by the bots core system (i.e. every 5 minutes).

Also migrated the two existing routines:
* `TemporaryModerationRoutine` (for temp bans, mutes, ...)
* `ModAuditLogRoutine`

### Example

A quick example routine would be
```java
class Foo implements Routine {
  @Override
  public void run(JDA jda) {
    System.out.println("hello");
  }

  @Override
  public Schedule createSchedule() {
    return new Routine.Schedule(Routine.ScheduleMode.FIXED_RATE, 0, 5, TimeUnit.SECONDS);
  }
}
```
with an addition in `Features.java`:
```java
features.add(new Foo());
```
The system will now basically hand over the routine to a `ScheduledExecutorService` and execute its `run(JDA)` method on the schedule defined by its `createSchedule()`.

### Remarks

This is based on (and hence also blocked by) #345 .

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
@@ -51,10 +51,8 @@ public enum Features {
         Collection<Feature> features = new ArrayList<>();
 
         // Routines
-        // TODO This should be moved into some proper command system instead (see GH issue #235
-        // which adds support for routines)
-        new ModAuditLogRoutine(jda, database).start();
-        new TemporaryModerationRoutine(jda, actionsStore).start();
+        features.add(new ModAuditLogRoutine(database));
+        features.add(new TemporaryModerationRoutine(jda, actionsStore));
 
         // Message receivers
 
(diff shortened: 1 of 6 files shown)


Zabuzard:
No changes since a week, merging

PR #347, opened 2022-01-18
Fixed bug with tag-raw not displaying leading spaces

### LINKED ISSUE #273: Tag-raw does not display leading spaces

Zabuzard:
### Overview

Implements and closes #273 . In a nutshell, the tag-raw command cant display leading zeroes, which is super annoying when working with code, for example:

Indents are missing

This is a Discord limitation/feature and cant easily be bypassed or turned off.

The PR fixes the issue by going away from sending escaped embeds and instead sending a plain attachment. Discord doesnt mess with the attachments, so the content is displayed plain as it is.

attachment

Which not only fixes the bug, but argueably also looks better.


Zabuzard:
@Tais993 I dont like the `event.reply("").addFile(...)`, a better way to add a file without that `reply("")` that I havent seen?

Tais993:
> @Tais993 I dont like the `event.reply("").addFile(...)`, a better way to add a file without that `reply("")` that I havent seen?

Unfortunately, there is not

Tais993 (approved):
iOS and Android do not have a preview for files, so they'll receive the raw file instead.
Since most (if not all) tags will be viewed, and edited on desktop this solution is fine to me.

Heatmanofurioso (approved):

PR #348, opened 2022-01-18
Add config template

### LINKED ISSUE #234: Add default `config.json` to git
Right now, users are expected to read our tutorial (Wiki) and copy the example `config.json` given there into their code base for the project to work.

This is easy to miss for users. And hard to maintain for us (it got outdated several times already).

Instead, the code should either generate it on-the-fly (using code or build scripts) or we simply add the `config.json` to **git**. This was not done originally, since local changes should not be tracked by git on user end - users should not push in changes to the file accidentally.

A solution to this might actually be:
https://stackoverflow.com/questions/9794931/keep-file-in-a-git-repo-but-dont-track-changes

Zabuzard:
### Overview

Closes #234. This adds a `config.json.template` file, containing reasonable default values and some placeholders for the configuration file of the bot.

This is a solution to the problem of maintaining the `config.json` that we all have to use in a transparent way without git-tracking the real `config.json` and accidentally commiting our real bot tokens when doing changes.

That approach seems to be the most idiomatic way currently used (see for example [https://stackoverflow.com/questions/9794931/keep-file-in-a-git-repo-but-dont-track-changes](https://stackoverflow.com/questions/9794931/keep-file-in-a-git-repo-but-dont-track-changes)).

This also gets rid of the problems we had with forgetting to maintain the default config in our Wiki (we can now simply refer to this template instead) and it eases the flow for people when rebasing `develop` when it introduces changes to the config.

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -1,0 +1,19 @@
+{
+    "token": "<put_your_token_here>",
+    "databasePath": "local-database.db",
+    "projectWebsite": "https://github.com/Together-Java/TJ-Bot",
+    "discordGuildInvite": "https://discord.com/invite/XXFUXzK",
+    "modAuditLogChannelPattern": "mod_audit_log",
+    "mutedRolePattern": "Muted",
+    "heavyModerationRolePattern": "Moderator",
+    "softModerationRolePattern": "Moderator|Staff Assistant",
+    "tagManageRolePattern": "Moderator|Staff Assistant|Top Helpers .+",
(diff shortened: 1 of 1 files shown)


Tais993 (approved):
Looks wrapped up to me

https://user-images.githubusercontent.com/49957334/150323151-ad85a11e-6edc-491a-99b9-e081bfd6dbca.mp4

!

Heatmanofurioso (approved):

PR #351, opened 2022-01-23
Reaction roles using a SelectionMenu

Tais993:
Implementation of the reaction roles, this is based of the original PR by JJeff248, see #284 

At the moment of writing
- [x] implement emotes/emojis support based of the Role's icon
- [x] Look for other possible improvements
- [x] Take pictures for how the GUI looks
- [x] replace nonce cause it's unsupported for interactions
- [x] Fix history

UX:

# Slash-command

When running the choose or all command, you get asked for what the embed's title and description has to be.
image

# Choose sub-command

When running the choose command, it displays the embed you gave, and it allows you to select roles. 
image

After selecting the roles, it sends a message with a SelectionMenu for the roles.
image

# Role selection by user

After this, the UX is the same for choosing the roles yourself or just picking all roles.

Once you select the roles, it sends a message saying your roles have been "updated"
image

# Notes

- Emoji roleicon support has been added, emotes support unfortunately can't be added (to my knowledge). Can't test due to not having a level 2 server

- For in the future when modals / forms get released, we might want to make it a modal so we can add a description and such. But that won't be this PR *most likely*


application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,320 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(RoleSelectCommand.class);
+
+    private static final String ALL_OPTION = "all";
+    private static final String CHOOSE_OPTION = "choose";
+
+    private static final String TITLE_OPTION = "title";
+    private static final String DESCRIPTION_OPTION = "description";
+
+    private static final Color embedColor = new Color(24, 221, 136);
```
java-coding-prodigy, line 52:
```suggestion
    private static final Color embedColor = Color.decode("#18DD88");
```

java-coding-prodigy (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
java-coding-prodigy: We should use `Color#decode` so we can easily change the hex when needed and also for consistency

Tais993: I'm missing the point

java-coding-prodigy: Point is we should be using hex instead of RGB, at least imo.

Tais993: And why?

java-coding-prodigy: Because it is easier to edit in the future and other things too.

Tais993: What other things? And I don't see how it's easier, just use IntelliJ. 
"But what if you don't use an IDE", simple, because it's RGB you can use your brain to guess what the color is.

I doubt many people here will be able to guess what color something is by hex.


Unless better arguments come in, or people agree with you I won't change this.

Tais993:
Can't test due to DB issues, so back to a draft for now

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Zabuzard: i think both are bad for humans to read and understand. everyone will either use a plugin to understand the color or just google it anyways... too much brain power needed to interpret it 😆 

that said, other commands in the code base used the hex code so far. personally, i dont mind at all and am fine with both.

a point could be made about intellij plugins showing the color and i believe intellij cant do that with the rgb but only with the hex? i might be totally wrong here though.

Tais993: Totally wrong there
image

Full support for everything
Can even change the hex in there

Tais993:
Added screenshots for UX ^^

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
java-coding-prodigy: > Full support for everything
Can even change the hex in there

That happens even if we use hex

Zabuzard: In that case I do not have any argument and officially **dont care**.

Heatmanofurioso: I agree with Zabu.
If we heavily used "standard" colors, we could use a lib of sorts to get the color from an enum "ex: `IndexedColors.LIGHT_BLUE.getIndex()`", but that not being the case, both solutions have the same readability, and IDE's can easily display them

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.Objects;
+
+
+/**
+ * Implements the {@code roleSelect} command.
+ *
+ * <p>
+ * Allows users to select their roles without using reactions, instead it uses selection menus where
+ * you can select multiple roles. <br />
+ * Note: the bot can only use roles below its highest one
```
Heatmanofurioso, line 39: This could be slightly rephrased to be a bit more easy to understand.
Example: "Note: The bot can only use roles with a permission level below it's highest one"

Not sure if the example given is a suitable substitution 😆

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+     * @param member the member to update the roles of
+     * @param guild what guild to update the roles in
+     * @param additionRoles the roles to add
+     * @param removalRoles the roles to remove
+     */
+    private static void handleRoleModifications(@NotNull final Interaction event,
+            final Member member, final @NotNull Guild guild, final Collection<Role> additionRoles,
+            final Collection<Role> removalRoles) {
+        guild.modifyMemberRoles(member, additionRoles, removalRoles)
+            .flatMap(empty -> event.reply("Updated your roles!").setEphemeral(true))
```
Heatmanofurioso, line 292: Just to follow a bit of the wording sense that's used in most other messages.
Could this be changed to something like "Your roles have been updated!" ?

Heatmanofurioso (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993, line 292:
```suggestion
            .flatMap(empty -> event.reply("Your roles have been updated!").setEphemeral(true))
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: While I get the point, don't like the way that phrases it either.

It can only use roles with a lower position, not less permissions,

Heatmanofurioso: You're right. I miss-worded that really badly, but the same logic applies, if you agree.
Something like: `Note: the bot can only use roles with a position below its highest one`
Just adding the "with a position" makes it a bit easier to understand, I believe

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Heatmanofurioso: Sounds great!

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993:
```suggestion
* Note: the bot can only use roles with a position below its highest one
```
I fully agree!

Heatmanofurioso (approved):

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: Can this be resolved?

java-coding-prodigy: I reckon so

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        if (!member.hasPermission(Permission.MANAGE_ROLES)) {
+            event.reply("You dont have the right permissions to use this command")
+                .setEphemeral(true)
+                .queue();
+            return;
+        }
+
+        Member selfMember = Objects.requireNonNull(event.getGuild()).getSelfMember();
+        if (!selfMember.hasPermission(Permission.MANAGE_ROLES)) {
+            event.reply("The bot needs the manage role permissions").setEphemeral(true).queue();
```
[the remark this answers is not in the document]
Tais993: It already does that?
This command can only be ran by moderators (or well, people with the MANAGE_ROLES permission)

Tais993:
This PR has been stale for 10 days, could anyone review please?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * Implements the {@code roleSelect} command.
+ *
+ * <p>
+ * Allows users to select their roles without using reactions, instead it uses selection menus where
+ * you can select multiple roles. <br />
+ * Note: the bot can only use roles with a position below its highest one
+ */
+public class RoleSelectCommand extends SlashCommandAdapter {
```
Zabuzard, line 41: final

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+    private static final Color AMBIENT_COLOR = new Color(24, 221, 136, 255);
+
+    private static final List<OptionData> messageOptions = List.of(
+            new OptionData(OptionType.STRING, TITLE_OPTION, "The title for the message", false),
+            new OptionData(OptionType.STRING, DESCRIPTION_OPTION, "A description for the message",
+                    false));
+
+
+    /**
+     * Construct an instance
```
Zabuzard, line 60: dot

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+            event.reply("You dont have the right permissions to use this command")
+                .setEphemeral(true)
+                .queue();
+            return;
+        }
+
+        Member selfMember = Objects.requireNonNull(event.getGuild()).getSelfMember();
+        if (!selfMember.hasPermission(Permission.MANAGE_ROLES)) {
+            event.reply("The bot needs the manage role permissions").setEphemeral(true).queue();
+            logger.warn("The bot needs the manage role permissions");
```
Zabuzard, line 90: You can flag an `error` for that. Its a wrong bot-setup.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+            return;
+        }
+
+        SelectionMenu.Builder menu = SelectionMenu.create(generateComponentId(member.getId()));
```
Zabuzard, line 94: you have to choose `Lifetime.PERMANENT` in `generateComponentId`, otherwise the component id might get stale after a few weeks/months. (see the overload of the generate method)

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        boolean ephemeral = false;
+
+        if (Objects.equals(event.getSubcommandName(), CHOOSE_OPTION)) {
```
Zabuzard, line 97: NIT: Swap order and use `.equals`

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+
+        Member selfMember = Objects.requireNonNull(event.getGuild()).getSelfMember();
+        if (!selfMember.hasPermission(Permission.MANAGE_ROLES)) {
+            event.reply("The bot needs the manage role permissions").setEphemeral(true).queue();
+            logger.warn("The bot needs the manage role permissions");
+            return;
+        }
+
+        SelectionMenu.Builder menu = SelectionMenu.create(generateComponentId(member.getId()));
+        boolean ephemeral = false;
```
Zabuzard, line 95: NIT: rename to `isEphemeral`?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+
+        MessageBuilder messageBuilder = new MessageBuilder(makeEmbed(title, description))
+            .setActionRows(ActionRow.of(menu.build()));
+
+        if (ephemeral) {
+            event.reply(messageBuilder.build()).setEphemeral(true).queue();
+        } else {
+            event.getChannel().sendMessage(messageBuilder.build()).queue();
+
+            event.reply("Message sent successfully!").setEphemeral(true).queue();
```
Zabuzard, line 119: I might be missing something but why this split? Dont I see the message directly in the same channel or are there 2 different channels?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+     * @param event the {@link SlashCommandEvent}
+     * @param menu the menu to add options to {@link SelectionMenu.Builder}
+     * @param placeHolder the placeholder for the menu {@link String}
+     * @param minValues the minimum number of selections. nullable {@link Integer}
+     */
+    private static void addMenuOptions(@NotNull final Interaction event,
+            @NotNull final SelectionMenu.Builder menu, @NotNull final String placeHolder,
+            @Nullable final Integer minValues) {
+
+
```
Zabuzard, line 136: Two newlines is one too much

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        Guild guild = Objects.requireNonNull(event.getGuild(), "The given guild cannot be null");
+
+        Role highestBotRole = guild.getSelfMember().getRoles().get(0);
```
Zabuzard, line 139: This seems cheezy, is the role ordering on `getRoles` required by its javadoc?

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        List<Role> guildRoles = guild.getRoles();
+
+        Collection<Role> roles = new ArrayList<>(
+                guildRoles.subList(guildRoles.indexOf(highestBotRole) + 1, guildRoles.size()));
```
Zabuzard, line 143: This logic might be weird to someone not familiar with it, I would recommend moving it into a helper method like `getRolesBotCanUse` or similar (maybe u have a better name)

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+
+        if (null != minValues) {
+            menu.setMinValues(minValues);
+        }
+
+        menu.setPlaceholder(placeHolder)
+            .setMaxValues(roles.size())
+            .addOptions(roles.stream()
+                .filter(role -> !role.isPublicRole())
```
Zabuzard, line 152: NIT: `Predicate.not(Role::isPublicRole)`, ur choice

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+                .filter(role -> !role.getTags().isBot())
+                .map(role -> {
+                    RoleIcon roleIcon = role.getIcon();
+
+                    if (null == roleIcon || !roleIcon.isEmoji()) {
+                        return SelectOption.of(role.getName(), role.getId());
+                    } else {
+                        return SelectOption.of(role.getName(), role.getId())
+                            .withEmoji((Emoji.fromUnicode(roleIcon.getEmoji())));
+                    }
```
Zabuzard, line 162: Could be moved into a helper method.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+                })
+                .toList());
+    }
+
+    /**
+     * Creates an embedded message to send with the selection menu
+     *
+     * <p>
+     * </p>
```
Zabuzard, line 171: This seems like a copy pasta mistake.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+                    } else {
+                        return SelectOption.of(role.getName(), role.getId())
+                            .withEmoji((Emoji.fromUnicode(roleIcon.getEmoji())));
+                    }
+                })
+                .toList());
+    }
+
+    /**
+     * Creates an embedded message to send with the selection menu
```
Zabuzard, line 168: dot

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+
+    @Override
+    public void onSelectionMenu(@NotNull final SelectionMenuEvent event,
+            @NotNull final List<String> args) {
+
+        Guild guild = Objects.requireNonNull(event.getGuild(), "The given guild cannot be null");
+        List<SelectOption> selectedOptions = Objects.requireNonNull(event.getSelectedOptions(),
+                "The given selectedOptions cannot be null");
+
+
```
Zabuzard, line 196: Two newlines is one too much

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        List<Role> selectedRoles = selectedOptions.stream()
+            .map(SelectOption::getValue)
+            .map(guild::getRoleById)
+            .filter(Objects::nonNull)
+            .filter(role -> guild.getSelfMember().canInteract(role))
+            .toList();
+
+
```
Zabuzard, line 204: Two newlines is one too much

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        if (event.getMessage().isEphemeral()) {
+            handleNewRoleBuilderSelection(event, selectedRoles);
+        } else {
+            handleRoleSelection(event, selectedRoles, guild);
+        }
+    }
+
+    /**
+     * Handles selection of a {@link SelectionMenuEvent}
```
Zabuzard, line 213: dot

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+            }
+
+            return role;
+        }).filter(Objects::nonNull).forEach(role -> {
+            if (selectedRoles.contains(role)) {
+                rolesToAdd.add(role);
+            } else {
+                rolesToRemove.add(role);
+            }
+        });
```
Zabuzard, line 238: A bit complex to read, maybe helper methods could help, in particular for the `map` call.

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
```diff
@@ -0,0 +1,319 @@
...
+        guild.modifyMemberRoles(member, additionRoles, removalRoles)
+            .flatMap(empty -> event.reply("Your roles have been updated!").setEphemeral(true))
+            .queue();
+    }
+
+    /**
+     * This gets the OptionMapping and returns the value as a string if there is one
+     *
+     * <p>
+     * </p>
```
Zabuzard, line 300: Copy pasta issue

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: Highest role is at 0, lowest is at the lowest index. Makes sense

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: Or Java adds `Role::!isPublicRole` :(

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Zabuzard: Yeah. But is this an undocumented side-effect or actually mandated by this `getRoles` method - that its sorted correctly?

Tais993: It's documented

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: You're missing a dot in your message :p

Zabuzard: so true 😆

application/src/main/java/org/togetherjava/tjbot/commands/basic/RoleSelectCommand.java
Tais993: The reply is for the slash-command, I don't want the reaction-role message to say that it's from a slash-command.

Thus I send the role-selection in a message beforehand.

Zabuzard (approved):

Heatmanofurioso (approved):

PR #352, opened 2022-01-24
whois command

Tais993:
## Open to suggestions for properties to add to the command

# Implementation

I decided to go for the description instead of fields, fields are useful but fill your screen easily.
When adding a lot of them, it becomes harder to read and (imho) ugly.

The embed color goes as follows:
Member color -> profile color -> default (black)

So if it's an user, that has a custom banner it'll result in black.
But if it's a member, with a member color it'll show that color instead.

# Current to-do's:
- [x] Look for more things to show
- [x] Find a way to fix code duplications (it doesn't feel perfect, but it's alright)

# Command signature
`/whois User:user (boolean:show_server_specific_info)`

# Member info
If the member is boosting the guild, it'll also display since when they started boosting.
If the member isn't in VC, it won't show the `in voicechannel` option

image

# User info
If the user has any flags (other than nitro and Hypesquad), it shows them

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
@@ -1,23 +1,20 @@
 package org.togetherjava.tjbot.commands.moderation;
 
 import net.dv8tion.jda.api.EmbedBuilder;
-import net.dv8tion.jda.api.entities.GuildVoiceState;
-import net.dv8tion.jda.api.entities.Member;
-import net.dv8tion.jda.api.entities.Role;
-import net.dv8tion.jda.api.entities.User;
-import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.entities.*;
+import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
+import net.dv8tion.jda.api.interactions.Interaction;
+import net.dv8tion.jda.api.interactions.callbacks.IReplyCallback;
 import net.dv8tion.jda.api.interactions.commands.OptionMapping;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
-import net.dv8tion.jda.api.interactions.components.Button;
-import net.dv8tion.jda.api.interactions.components.ButtonStyle;
-import net.dv8tion.jda.api.requests.restaction.interactions.ReplyAction;
+import net.dv8tion.jda.api.requests.restaction.interactions.ReplyCallbackAction;
 import org.jetbrains.annotations.NotNull;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.commands.SlashCommandVisibility;
 import org.togetherjava.tjbot.commands.utils.DiscordClientAction;
 
 import javax.annotation.CheckReturnValue;
-import java.awt.*;
+import java.awt.Color;
 import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
@@ -52,7 +49,7 @@ public final class WhoIsCommand extends SlashCommandAdapter {
     }
 
     @Override
-    public void onSlashCommand(@NotNull final SlashCommandEvent event) {
+    public void onSlashCommand(@NotNull final SlashCommandInteractionEvent event) {
         OptionMapping userOption = Objects.requireNonNull(event.getOption(USER_OPTION),
                 "The given user option cannot be null");
         OptionMapping showServerSpecificInfoOption = event.getOption(SHOW_SERVER_INFO_OPTION);
@@ -73,63 +70,56 @@ public final class WhoIsCommand extends SlashCommandAdapter {
     }
 
     @CheckReturnValue
-    private static @NotNull ReplyAction handleWhoIsUser(final @NotNull SlashCommandEvent event,
+    private static @NotNull ReplyCallbackAction handleWhoIsUser(final @NotNull IReplyCallback event,
             final @NotNull User user, final @NotNull User.Profile profile) {
-
-        StringBuilder descriptionBuilder =
-                new StringBuilder().append(userIdentificationToStringItem(user))
-                    .append("\n**Is bot:** ")
-                    .append(user.isBot())
-                    .append(userFlagsToStringItem(user.getFlags()))
-                    .append("\n**Registration date:** ")
-                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()));
+        String description = userIdentificationToStringItem(user) + "\n**Is bot:** " + user.isBot()
+                + userFlagsToStringItem(user.getFlags()) + "\n**Registration date:** "
+                + DATE_TIME_FORMAT.format(user.getTimeCreated());
 
         EmbedBuilder embedBuilder =
                 generateEmbedBuilder(event, user, profile, profile.getAccentColor()).setAuthor(
                         user.getName(), user.getEffectiveAvatarUrl(), user.getEffectiveAvatarUrl())
-                    .setDescription(descriptionBuilder);
+                    .setDescription(description);
 
-        return event.replyEmbeds(embedBuilder.build())
-            .addActionRow(Button.of(ButtonStyle.LINK, "discord://-/users/" + user.getId(),
-                    "Click to see profile"));
+        return sendEmbedWithProfileAction(event, embedBuilder.build(), user.getId());
     }
 
     @CheckReturnValue
-    private static @NotNull ReplyAction handleWhoIsMember(final @NotNull SlashCommandEvent event,
-            final @NotNull Member member, final @NotNull User.Profile profile) {
+    private static @NotNull ReplyCallbackAction handleWhoIsMember(
+            final @NotNull IReplyCallback event, final @NotNull Member member,
+            final @NotNull User.Profile profile) {
         User user = member.getUser();
 
         Color memberColor = member.getColor();
         Color effectiveColor = (null == memberColor) ? profile.getAccentColor() : memberColor;
 
-        StringBuilder descriptionBuilder =
-                new StringBuilder().append(userIdentificationToStringItem(user))
-                    .append(voiceStateToStringItem(member))
-                    .append("\n**Is bot:** ")
-                    .append(user.isBot())
-                    .append(possibleBoosterToStringItem(member))
-                    .append(userFlagsToStringItem(user.getFlags()))
-                    .append("\n**Join date:** ")
-                    .append(DATE_TIME_FORMAT.format(member.getTimeJoined()))
-                    .append("\n**Registration date:** ")
-                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()))
-                    .append("\n**Roles:** ")
-                    .append(formatRoles(member));
+        String description = userIdentificationToStringItem(user) + voiceStateToStringItem(member)
+                + "\n**Is bot:** " + user.isBot() + possibleBoosterToStringItem(member)
+                + userFlagsToStringItem(user.getFlags()) + "\n**Join date:** "
+                + DATE_TIME_FORMAT.format(member.getTimeJoined()) + "\n**Registration date:** "
+                + DATE_TIME_FORMAT.format(user.getTimeCreated()) + "\n**Roles:** "
+                + formatRoles(member);
 
         EmbedBuilder embedBuilder = generateEmbedBuilder(event, user, profile, effectiveColor)
             .setAuthor(member.getEffectiveName(), member.getEffectiveAvatarUrl(),
                     member.getEffectiveAvatarUrl())
-            .setDescription(descriptionBuilder);
+            .setDescription(description);
 
-        return event.replyEmbeds(embedBuilder.build())
-            .addActionRow(DiscordClientAction.General.USER.asLinkButton("Click to see profile!",
-                    user.getId()));
+        return sendEmbedWithProfileAction(event, embedBuilder.build(), user.getId());
+    }
+
+    private static @NotNull ReplyCallbackAction sendEmbedWithProfileAction(
+            final @NotNull IReplyCallback event, @NotNull MessageEmbed embed,
+            @NotNull String userId) {
+        return event.replyEmbeds(embed)
+            .addActionRow(
+                    DiscordClientAction.General.USER.asLinkButton("Click to see profile!", userId));
     }
 
     private static @NotNull String voiceStateToStringItem(@NotNull final Member member) {
         GuildVoiceState voiceState = Objects.requireNonNull(member.getVoiceState(),
                 "The given voiceState cannot be null");
-        if (voiceState.inVoiceChannel()) {
+        if (voiceState.inAudioChannel()) {
             return "\n**In voicechannel:** " + (voiceState.getChannel().getAsMention());
         } else {
             return "";
@@ -140,15 +130,15 @@ public final class WhoIsCommand extends SlashCommandAdapter {
     /**
      * Generates whois embed based on the given parameters.
      *
-     * @param event the {@link SlashCommandEvent}
+     * @param event the {@link SlashCommandInteractionEvent}
      * @param user the {@link User} getting whois'd
      * @param profile the {@link net.dv8tion.jda.api.entities.User.Profile} of the whois'd user
      * @param effectiveColor the {@link Color} that the embed will become
      * @return the generated {@link EmbedBuilder}
      */
-    private static @NotNull EmbedBuilder generateEmbedBuilder(
-            @NotNull final SlashCommandEvent event, @NotNull final User user,
-            final @NotNull User.Profile profile, final Color effectiveColor) {
+    private static @NotNull EmbedBuilder generateEmbedBuilder(@NotNull final Interaction event,
+            @NotNull final User user, final @NotNull User.Profile profile,
+            final Color effectiveColor) {
 
         EmbedBuilder embedBuilder = new EmbedBuilder().setThumbnail(user.getEffectiveAvatarUrl())
             .setColor(effectiveColor)
@@ -183,14 +173,11 @@ public final class WhoIsCommand extends SlashCommandAdapter {
      * Handles the user's identifying properties (such as ID, tag)
      *
      * @param user the {@link User} to take the identifiers from
-     * @return user readable {@link StringBuilder}
+     * @return user readable {@link String}
      */
-    private static @NotNull StringBuilder userIdentificationToStringItem(final @NotNull User user) {
-        return new StringBuilder("**Mention:** ").append(user.getAsMention())
-            .append("\n**Tag:** ")
-            .append(user.getAsTag())
-            .append("\n**ID:** ")
-            .append(user.getId());
+    private static @NotNull String userIdentificationToStringItem(final @NotNull User user) {
+        return "**Mention:** " + user.getAsMention() + "\n**Tag:** " + user.getAsTag()
+                + "\n**ID:** " + user.getId();
     }
 
     /**
@@ -213,13 +200,13 @@ public final class WhoIsCommand extends SlashCommandAdapter {
     private static @NotNull StringBuilder userFlagsToStringItem(
             final @NotNull Collection<User.UserFlag> flags) {
         String formattedFlags = formatUserFlags(flags);
+        StringBuilder result = hypeSquadToStringItem(flags);
 
-        if (formattedFlags.isBlank()) {
-            return hypeSquadToStringItem(flags);
-        } else {
-            return hypeSquadToStringItem(flags).append("\n**Flags:** ")
-                .append(formatUserFlags(flags));
+        if (!formattedFlags.isBlank()) {
+            result.append("\n**Flags:** ").append(formattedFlags);
         }
+
+        return result;
     }
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
@@ -1,0 +1,263 @@
+package org.togetherjava.tjbot.commands.moderation;
+
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.entities.GuildVoiceState;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.User;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.commands.OptionMapping;
+import net.dv8tion.jda.api.interactions.commands.OptionType;
+import net.dv8tion.jda.api.interactions.components.Button;
+import net.dv8tion.jda.api.interactions.components.ButtonStyle;
+import net.dv8tion.jda.api.requests.restaction.interactions.ReplyAction;
+import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.SlashCommandVisibility;
+import org.togetherjava.tjbot.commands.utils.DiscordClientAction;
+
+import javax.annotation.CheckReturnValue;
+import java.awt.*;
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * This command allows you to look up user (or member) info.
+ */
+@SuppressWarnings("ClassWithoutLogger")
+public final class WhoIsCommand extends SlashCommandAdapter {
+    private static final String USER_OPTION = "user";
+    private static final String SHOW_SERVER_INFO_OPTION = "show_server_specific_info";
+
+    private static final String USER_PROFILE_PICTURE_SIZE = "4096";
+
+    // Sun, December 11, 2016, 13:36:30
+    private static final DateTimeFormatter DATE_TIME_FORMAT =
+            DateTimeFormatter.ofPattern("E, MMMM d, u, HH:mm:ss");
+
+    /**
+     * Creates an instance.
+     */
+    public WhoIsCommand() {
+        super("whois", "Provides info about the given user", SlashCommandVisibility.GUILD);
+
+        getData().addOption(OptionType.USER, USER_OPTION, "the user to look up", true)
+            .addOption(OptionType.BOOLEAN, SHOW_SERVER_INFO_OPTION,
+                    "Whenever to show info that is specific to this server, such as their roles. This is true by default.",
+                    false);
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull final SlashCommandEvent event) {
+        OptionMapping userOption = Objects.requireNonNull(event.getOption(USER_OPTION),
+                "The given user option cannot be null");
+        OptionMapping showServerSpecificInfoOption = event.getOption(SHOW_SERVER_INFO_OPTION);
+
+        User user = userOption.getAsUser();
+        Member member = userOption.getAsMember();
+
+        boolean showServerSpecificInfo = null != member && (null == showServerSpecificInfoOption
+                || showServerSpecificInfoOption.getAsBoolean());
+
+        user.retrieveProfile().flatMap((User.Profile profile) -> {
+            if (showServerSpecificInfo) {
+                return handleWhoIsMember(event, member, profile);
+            } else {
+                return handleWhoIsUser(event, user, profile);
+            }
+        }).queue();
+    }
+
+    @CheckReturnValue
+    private static @NotNull ReplyAction handleWhoIsUser(final @NotNull SlashCommandEvent event,
+            final @NotNull User user, final @NotNull User.Profile profile) {
+
+        StringBuilder descriptionBuilder =
+                new StringBuilder().append(userIdentificationToStringItem(user))
+                    .append("\n**Is bot:** ")
+                    .append(user.isBot())
+                    .append(userFlagsToStringItem(user.getFlags()))
+                    .append("\n**Registration date:** ")
+                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()));
+
+        EmbedBuilder embedBuilder =
+                generateEmbedBuilder(event, user, profile, profile.getAccentColor()).setAuthor(
+                        user.getName(), user.getEffectiveAvatarUrl(), user.getEffectiveAvatarUrl())
+                    .setDescription(descriptionBuilder);
+
+        return event.replyEmbeds(embedBuilder.build())
+            .addActionRow(Button.of(ButtonStyle.LINK, "discord://-/users/" + user.getId(),
+                    "Click to see profile"));
+    }
+
+    @CheckReturnValue
+    private static @NotNull ReplyAction handleWhoIsMember(final @NotNull SlashCommandEvent event,
+            final @NotNull Member member, final @NotNull User.Profile profile) {
+        User user = member.getUser();
+
+        Color memberColor = member.getColor();
+        Color effectiveColor = (null == memberColor) ? profile.getAccentColor() : memberColor;
+
+        StringBuilder descriptionBuilder =
+                new StringBuilder().append(userIdentificationToStringItem(user))
+                    .append(voiceStateToStringItem(member))
+                    .append("\n**Is bot:** ")
+                    .append(user.isBot())
+                    .append(possibleBoosterToStringItem(member))
+                    .append(userFlagsToStringItem(user.getFlags()))
+                    .append("\n**Join date:** ")
+                    .append(DATE_TIME_FORMAT.format(member.getTimeJoined()))
+                    .append("\n**Registration date:** ")
+                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()))
+                    .append("\n**Roles:** ")
+                    .append(formatRoles(member));
+
+        EmbedBuilder embedBuilder = generateEmbedBuilder(event, user, profile, effectiveColor)
+            .setAuthor(member.getEffectiveName(), member.getEffectiveAvatarUrl(),
+                    member.getEffectiveAvatarUrl())
+            .setDescription(descriptionBuilder);
+
+        return event.replyEmbeds(embedBuilder.build())
+            .addActionRow(DiscordClientAction.General.USER.asLinkButton("Click to see profile!",
+                    user.getId()));
+    }
+
+    private static @NotNull String voiceStateToStringItem(@NotNull final Member member) {
+        GuildVoiceState voiceState = Objects.requireNonNull(member.getVoiceState(),
+                "The given voiceState cannot be null");
+        if (voiceState.inVoiceChannel()) {
+            return "\n**In voicechannel:** " + (voiceState.getChannel().getAsMention());
+        } else {
+            return "";
+        }
+    }
+
+
+    /**
+     * Generates whois embed based on the given parameters.
+     *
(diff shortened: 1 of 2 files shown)


Zabuzard:
A regular user is probably not aware of the difference between user and member and will not understand what to expected from `hide_member_info`. Maybe rename it to `hide_server_specific_info` or similar. Something that the user will naturally understand better.

I would also turn around the logic, `show_server_info` instead of `hide`, is easier to understand for most people (I believe). (with default being `true` i.e. `show`)

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.SlashCommandVisibility;
+
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class WhoIsCommand extends SlashCommandAdapter {
```
Zabuzard, line 24: (i know this is just a draft, im putting it here anyways while im here)
* Missing javadoc
* should be final

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(WhoIsCommand.class);
+
+    private static final String USER_OPTION = "user";
+    private static final String HIDE_MEMBER_INFO = "hide_member_info";
+    private static final DateTimeFormatter DATE_TIME_FORMAT =
+            DateTimeFormatter.ofPattern("E, MMMM d, u, HH:mm:ss");
```
Zabuzard, line 30: Personally, I think it is always helpful to have a comment on such things to show a quick example of how the date/time would look like with this. So just sth like `// Example: 2022, April 5, 4, 19:20:31` or whatever it is.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public WhoIsCommand() {
+        super("whois", "who is the given user?", SlashCommandVisibility.GUILD);
```
Zabuzard, line 33: the description is funny but kinda not as helpful to a confused reader as it could be. maybe rewrite `"provides detailed information about the user"`

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+
+        getData().addOption(OptionType.USER, USER_OPTION, "the user to look up", true);
+        getData().addOption(OptionType.BOOLEAN, HIDE_MEMBER_INFO,
+                "whenever to show the member info of this user", false);
```
Zabuzard, line 37: you could cascade the call if u want. `getData().addOption().addOption()`

Zabuzard, line 37: `"whether to hide information about this user that is specific to this server, such as when they joined"`

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+            OptionMapping hideMemberInfoOption = event.getOption(HIDE_MEMBER_INFO);
+
+            if (null != hideMemberInfoOption && hideMemberInfoOption.getAsBoolean()) {
+                handleWhoIsUser(event, user);
+            } else {
+                handleWhoIsMember(event, member, user);
+            }
+        } else {
+            handleWhoIsUser(event, user);
+        }
```
Zabuzard, line 58: The if-else is a bit hard to read/understand. It is easy to see that its about 3 cases but its not clear on first sight what those 3 cases are about, i.e.
* user is not part of guild
* user is part of guild and we want to show member info
* user is part of guild and we want to hide member info

Maybe introduce some well named `boolean` variables or helper methods to make this clearer.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+        User user = userOption.getAsUser();
+        Member member = userOption.getAsMember();
+
+        if (null != member) {
+            OptionMapping hideMemberInfoOption = event.getOption(HIDE_MEMBER_INFO);
+
+            if (null != hideMemberInfoOption && hideMemberInfoOption.getAsBoolean()) {
+                handleWhoIsUser(event, user);
+            } else {
+                handleWhoIsMember(event, member, user);
```
Zabuzard, line 54: These method overloads are kinda confusing and definitely not obvious by just reading the signature. I.e., it is not obvious that one shows member extra info and the other not, just by passing in different arguments.

I think it is worth renaming them and only passing the relevant parameter (user can be retrieved from member anyways).

Also, both methods basically build an embed and the sending section is duplicated. It might be better to change them from standalone-handle methods to methods that just compute and return the embed and then doing the sending bit here again.

Maybe we can also append the member specific details (based on the flag) at the end of the embed build by the version that hides them.

So in a nutshell we would aim for sth like
```java
EmbedBuilder embed;
if (isUserMember) {
  embed = createResponseNotMember(user);
} else {
  embed = createResponseMember(member);
  if (isShowDetails) {
    appendDetails(embed);
  }
}

event.replyEmbeds(embed.build()).queue();

File: application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java [hunk 7]
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    /*
+     * .addField("Mention:", user.getAsMention(), false) .addField("Tag:", user.getAsTag(), false)
+     * .addField("ID", user.getId(), false) .addField("Join date",
+     * DATE_TIME_FORMAT.format(member.getTimeJoined()), true) .addField("Registration date",
+     * DATE_TIME_FORMAT.format(user.getTimeCreated()), true) .addField("Roles", formatRoles(member),
+     * false)
+     */
```
Zabuzard, line 131: (Dont forget to delete)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Thing I struggled with was readability, I'd want join date and registration date to be after each other.
But other info might come after that too, so keeping it in a sensible order while also splitting things up into 2 methods, doesn't seem possible.

Take a look at the examples, `is bot` and `is booster` are after each other too as example ^^

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Yeah removed it locally, didn't push yet

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Introduced variable

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Edited

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Cascaded 👍

Tais993: edited

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Still regards this, to me it's clear.

the handleWhoIsUser, handles the who is command for the user.
the  handleWhoIsMember, handles the who is command for the member.

Not sure how else I can "improve" it, the name should be enough

Tais993:
Now I get it, sonarcloud is crying about the logger variable.

Tais993:
Been 7 days, but due to the merge a new commit came (without changes)

Can this be merged in the new few days?

Tais993:
This PR has been stale for 12 days, could anyone review please?

Zabuzard:
I wasnt aware its ready, sorry.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * You ever wanted to stalk someone on Discord? Well no need to worry anymore! I got you covered,
+ * this command allows you to look up user (or member) info.
+ */
```
Zabuzard, line 30: This is quite funny but Javadoc is not the place for fun (unfortunately, lol).

Please rewrite more formally 😬

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+@SuppressWarnings("ClassWithoutLogger")
+public final class WhoIsCommand extends SlashCommandAdapter {
+    private static final String USER_OPTION = "user";
+    private static final String SHOW_SERVER_SPECIFIC_INFO = "show_server_specific_info";
```
Zabuzard, line 34: The constant name should hint what it is, i.e. an option or a subcommand or something else. Maybe `SHOW_SERVER_INFO_OPTION`

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Creates an instance of who is command
+     */
+    public WhoIsCommand() {
+        super("whois", "Provides info about the given user", SlashCommandVisibility.GUILD);
+
+        getData().addOption(OptionType.USER, USER_OPTION, "the user to look up", true)
+            .addOption(OptionType.BOOLEAN, SHOW_SERVER_SPECIFIC_INFO,
+                    "whenever to show info that is specific to this server, such as ", false);
```
Zabuzard, line 48: "such as ..." missing the remaining text?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+    @Override
+    public void onSlashCommand(@NotNull final SlashCommandEvent event) {
+        OptionMapping userOption = Objects.requireNonNull(event.getOption(USER_OPTION),
+                "The given user option cannot be null");
+        OptionMapping showServerSpecificInfoOption = event.getOption(SHOW_SERVER_SPECIFIC_INFO);
+
+        User user = userOption.getAsUser();
+        Member member = userOption.getAsMember();
+
+
```
Zabuzard, line 60: Two newlines is one newline too much. Afaik sonar also flags that later on.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+        boolean showServerSpecificInfo = null != member && (null == showServerSpecificInfoOption
+                || showServerSpecificInfoOption.getAsBoolean());
```
Zabuzard, line 62: Am I misreading this or do we show server specific info if its a member but the option has not been set? I.e. is the option `true` "by default"?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+    private static @NotNull ReplyAction handleWhoIsUser(final @NotNull SlashCommandEvent event,
+            final @NotNull User user, final @NotNull User.Profile profile) {
+
+        StringBuilder descriptionBuilder =
+                new StringBuilder().append(handleUserIdentification(user))
+                    .append("\n**Is bot:** ")
+                    .append(user.isBot())
+                    .append(handleUserFlags(user.getFlags()))
+                    .append("\n**Registration date:** ")
+                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()));
```
Zabuzard, line 83: NIT: For such small strings, string concat is totally okay as well. Elaborate if it would be more readable to change it to a formatted triple-string like:
```java
"""
foo: %s
bar: %s
baz: %s
""".formatted(foo, bar, baz)
```

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+
+        EmbedBuilder embedBuilder =
+                generateEmbedBuilder(event, user, profile, profile.getAccentColor()).setAuthor(
+                        user.getName(), user.getEffectiveAvatarUrl(), user.getEffectiveAvatarUrl())
+                    .setDescription(descriptionBuilder);
+
+        return event.replyEmbeds(embedBuilder.build())
+            .addActionRow(Button.of(ButtonStyle.LINK, "discord://-/users/" + user.getId(),
```
Zabuzard, line 91: This discord-URI usage is new to me, probably also new to others. It might be worth to move that up into some more general helper class with helper methods. Some sort of `DiscordActionBuilder` or `DiscordActions` or similar maybe, with a `createShowProfileAction(id)` and stuff like that 🤷

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+        EmbedBuilder embedBuilder = generateEmbedBuilder(event, user, profile, effectiveColor)
+            .setAuthor(member.getEffectiveName(), member.getEffectiveAvatarUrl(),
+                    member.getEffectiveAvatarUrl())
+            .setDescription(descriptionBuilder);
+
+
+        return event.replyEmbeds(embedBuilder.build())
+            .addActionRow(Button.of(ButtonStyle.LINK, "discord://-/users/" + user.getId(),
+                    "Click to see profile"));
+    }
```
Zabuzard, line 127: Hard to see during review but is this method a 1:1 copy of the other method? Only with different strings? The embed and button part look almost identical.

Maybe we can extract some stuff into helper methods or get rid of the duplication by other means. Not a big deal though, im okay with this for now.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+
+    private static @NotNull String handleVoiceState(@NotNull final Member member) {
```
Zabuzard, line 129: The name seems a bit confusing to me. It seems to be more of a "voice state to string" method. Maybe you can find a better name.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param user the {@link User} getting whois'd
+     * @param profile the {@link net.dv8tion.jda.api.entities.User.Profile} of the whois'd user
+     * @param effectiveColor the {@link Color} that the embed will become
+     * @return the generated {@link EmbedBuilder}
+     */
+    private static @NotNull EmbedBuilder generateEmbedBuilder(
+            @NotNull final SlashCommandEvent event, @NotNull final User user,
+            final @NotNull User.Profile profile, final Color effectiveColor) {
+
+
```
Zabuzard, line 153: Two newlines is one too much

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+        EmbedBuilder embedBuilder = new EmbedBuilder().setThumbnail(user.getEffectiveAvatarUrl())
+            .setColor(effectiveColor)
+            .setFooter("Requested by " + event.getUser().getAsTag(),
+                    event.getMember().getEffectiveAvatarUrl())
+            .setTimestamp(Instant.now());
+
+        if (null != profile.getBannerId()) {
+            embedBuilder.setImage(profile.getBannerUrl() + "?size=4096");
```
Zabuzard, line 161: size should be moved into a constant

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param user the {@link User} to take the identifiers from
+     * @return user readable {@link StringBuilder}
+     */
+    private static @NotNull StringBuilder handleUserIdentification(final @NotNull User user) {
+        return new StringBuilder("**Mention:** ").append(user.getAsMention())
+            .append("\n**Tag:** ")
+            .append(user.getAsTag())
+            .append("\n**ID:** ")
+            .append(user.getId());
+    }
```
Zabuzard, line 195: same issue with the name i also had with the voice state. a "handle" method is, for me, a method that actually does `event.queue()` and similar stuff, i.e. without a real return value.

this is more like a "format/tostring the current state of property x".

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Formats Hypesquad and the flags
+     *
+     * @param flags the {@link Collection} of {@link net.dv8tion.jda.api.entities.User.UserFlag}
+     *        (recommend {@link java.util.EnumSet}
+     * @return user readable {@link StringBuilder}
+     */
+    private static @NotNull StringBuilder handleUserFlags(
```
Zabuzard, line 214: name (see earlier)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.SlashCommandVisibility;
+
+import javax.annotation.CheckReturnValue;
+import java.awt.*;
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.Objects;
```
java-coding-prodigy, line 24: test

java-coding-prodigy, line 24: test

java-coding-prodigy, line 24: test

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
java-coding-prodigy: Maybe write that in a normal `//` comment instead of a Javadoc comment?

java-coding-prodigy: Maybe write that in a normal `//` comment instead of a Javadoc comment?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
java-coding-prodigy: I think the triple strings might be better here.

marko-radosavljevic:
Hm, I like it, but the default is a bit cluttered. Do we really need fields 'is bot', 'is booster', 'hypersquad'.. etc on each /whois we do?

Maybe it makes more sense to have 2 commands, one concise and clear, so you can extract the most important info quickly, and another with more detailed info.

Short one would be without banner/background image too, and that way it would be super space efficient, you can even use it in lobby, and it wouldn't be that spammy, since it's only few fields.

Tais993:
You'll have to elaborate, what important info exactly?

marko-radosavljevic:
Basically just this: profile picture, ID, join/registration date. At least that's the only thing I use whois for.

Personally, I would like to know user activity on the server as well, past month and overall. That would be the most useful field for me. Since I often have to determine if the member is positively contributing in any way, or it's just a spambot. This would probaby go under more detailed view tho.

Tais993:
> I would like to know user activity on the server as well, past month and overall

What does this mean?

> Since I often have to determine if the member is positively contributing in any way, or it's just a spambot. 

Same for this, how can a bot determine this?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: My bad, didn't even remember writing this :p

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Will add that it's true by default to the option's description

marko-radosavljevic:
> > I would like to know user activity on the server as well, past month and overall
> 
> What does this mean?

Right, apologies for not being clear enough. Number of messages basically.

> > Since I often have to determine if the member is positively contributing in any way, or it's just a spambot.
> 
> Same for this, how can a bot determine this?

I would determine that, I just need bot to dig up some info for me.

So use case goes like this:

1. Member reports getting an advertisement in dms, or discord invite link.
2. I'm not familiar with the person being reported, so I decide to gather basic info by doing /whois. 
3. I require answers to a few questions to determine if the person is bot or a positively contributing member:
    a) Has this member ever typed anything on the server?
    b) Has this member joined very recently?
    c) Account age?
4. Check message history if I'm unsure.
5. Make a moderation action based on all this. Action is wildly different if it's a spambot harassing every member on the server with malicious intent, or just a positively contributing member sending one bad dm.

But again, this is how and why I use` whois`, it's quite possible that people use it for different things. I think it makes sense to have member activity on `whois` anyway, I would argue it's one of the most useful stats for a community member.

Tais993:
> Number of messages basically.

Not possible, we could assume the number the tophelpers command retrieves.
But that's only last month, and it doesn't feel right. All the info it displays is based off Discord, then we're adding custom elements. Not sure what others think about it.

> a) Has this member ever typed anything on the server?

Not possible

> b) Has this member joined very recently?
> c) Account age?

It seems the only info you want is

- PFP
- ID 
- join date
- account age

Now the question becomes, how do you want me to implement this?
Most sensible implementation would be

`/whois basic User:user`
`/whois advanced User:user (boolean:show_server_specific_info)`

But I personally don't really like it, it's becoming too much of an hassle to request data.
I'll let others decide what they want.

Could you put your comment in #64 so it can be discussed further over there?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: So
```java
"""
**Is bot:** %s
%s
**Registration date:** %s
""".formatted(user.isBot(), handleUserFlags(user.getFlags()), DATE_TIME_FORMAT.format(user.getTimeCreated());
```

This has no benefit to readability, it even makes it worse due to the random `%s` imho

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: I'd love to add that, but preferably in another PR in the future.

There's a ton of Discord URL's (which I'd love to wrap, maybe people can do cool things with it)
But I've to check how many work, and to what extent.

Just adding that class for method feels wrong to me, I'd rather just create a PR in the future with all of the actions

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Struggled a ton with this.

Issue is simple, to create a user friendly order I put certain flags in a certain order. (join date before registration date a.e)
So if we compare:

```diff
               new StringBuilder().append(handleUserIdentification(user))
+                    .append(handleVoiceState(member))
                    .append("\n**Is bot:** ")
                    .append(user.isBot())
+                    .append(handlePossibleBooster(member))
                    .append(handleUserFlags(user.getFlags()))
+                    .append("\n**Join date:** ")
+                    .append(DATE_TIME_FORMAT.format(member.getTimeJoined()))
                    .append("\n**Registration date:** ")
                    .append(DATE_TIME_FORMAT.format(user.getTimeCreated()))
+                    .append("\n**Roles:** ")
+                    .append(formatRoles(member));
```

You notice that the differences are everywhere, and I can't keep readable order (unless doing weird af trickery with Map's/Lists) while making it maintainable.

So this means, when adding a field it has to be done twice unfortunately.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Format feels wrong to me when comparing to normal usages of the format method.
`userIdentificationToString` is acceptable, but still feels weird?

Do you've any other suggestions? Otherwise I'll go with `userIdentificationToString`

marko-radosavljevic:
> It seems the only info you want is
> 
>     * PFP
> 
>     * ID
> 
>     * join date
> 
>     * account age

Yup, that sounds good to me, since that's the best I can get. ^^

I assume most people need those as well? I don't even know what `hypesquad` is. Bots are easy to recognize, and somewhat rare use case. Same with boosters.

Also, I think UI-wise, `is bot: false` is too nerdy, programming jargon, don't think most people think this way.

> Now the question becomes, how do you want me to implement this? Most sensible implementation would be
> 
> `/whois basic User:user` `/whois advanced User:user (boolean:show_server_specific_info)`
> 
> But I personally don't really like it, it's becoming too much of an hassle to request data. I'll let others decide what they want.

Yeah, I don't like it either, I think it's bad UI/UX. We are also used to just `whois`, so there is a bit of conditioning.

Maybe two separate commands `/whois` and `/whois-advanced` is better? That way you have easy to use `/whois`, but also can use the more advanced one if you need it. It clutters the command space a bit, but it's not terrible.

Another option is to just trim the current implementation.

> Could you put your comment in #64 so it can be discussed further over there?

Eh, it seems it's just an umbrella issue tracking migration from dyno, also few months old. Don't think we will have fruitful community-wide discussion in there.

* * * *
I'm not very opinionated on this, and also not sure what the wider community needs are. So, ultimately, your choice on the implementation. :relaxed:

Tais993:
> Also, I think UI-wise, is bot: false is too nerdy, programming jargon, don't think most people think this way

I mean, we're a programming Discord.
If it wasn't a programming Discord I would've created a method that converted it to yes/no

And Marko, you're a nerd, I'm a nerd, don't try to hide that :p

Tais993:
Can anyone else give me their thoughts on Marko's comment(s)?

Zabuzard:
> Can anyone else give me their thoughts on Marko's comment(s)?

I think that somone who uses `/whois` wants to see all info available. The more the better. Why would I ever want to use a `/whois-show-minimal` when there is a `/whois-show-everything`. With that reasoning, I would say that everything should be showed with no option to turn it off.

I might have a biased view on this though, since my use case for this is pretty limited.

Heatmanofurioso:
I agree with Zabu on this one
I wouldn't separate it in two commands. It's a bit redundant, considering the probable usual usage of the command

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: See #391

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: I know that you keep saying this, but still I don't believe it, why can't sonar just... say that?

Sonar should be the one speaking, not you?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Added

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Zabuzard: Yeah. I think it only flags this after pushing. Not sure why though.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Zabuzard: Really? For me its easier to read cause I can focus on the structure of the message.
With your original code its easier to detect whether the correct args are given but the message itself is less obvious to read.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Zabuzard: I guess you could get rid of some of the duplication though, no? Not talking about the embed itself now, you convinced me regarding this. But what about the code shortly before and after the embed building? I.e. maybe just move the embed-building into two helper methods and keep the rest in a single method?

But yeah, if thats also not easily possible, its fine.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: I personally don't like it because it adds these `%s` without indication of what they are.
So this means you've to read the arguments, the second argument to be specific, handleUserflags?
Current code clearly shows what happens I think personally.

To show what I mean with your example, the perfect world:
```java
"""
foo: %s
bar: %s
baz: %s
""".formatted(foo, bar, baz)
``` 
Reality:
```java
"""
foo: %s
%s
baz: %s
""".formatted(foo, bar, baz)
```
has a weird %s, and you don't know what it is, you've to read the arguments to figure out what one it is.
So first you've to count the amount of `%s`, then you've to find where we format, and go to that `...` argument.
To read the name/method/whatever, to understand the message structure

Zabuzard: Thats fair. Its a pitty that there are no "named placeholders" or similar. Feel free to resolve this conversation.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Zabuzard: Why not add a `sendEmbedWithUserProfile` or similar, where u move this duplicated code into.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: If they existed, I would've been using them!

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Since you spoiled me with the correct name, sure will create it.
Originally I wasn't sure how to name it, but that method sounds good to me, so sure will create it.

Heatmanofurioso:
@Tais993 Can you give Sonar some love?

Tais993:
Closed to delete the branch temporary

Tais993:
@Heatmanofurioso, @Zabuzard changes have been made, and Spotless has received their not so deserving attention

Zabuzard (changes requested):
See last open conversation

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
Tais993: Yes it was

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+            .addActionRow(DiscordClientAction.General.USER.asLinkButton("Click to see profile!",
+                    user.getId()));
+    }
+
+    private static @NotNull String voiceStateToStringItem(@NotNull final Member member) {
+        GuildVoiceState voiceState = Objects.requireNonNull(member.getVoiceState(),
+                "The given voiceState cannot be null");
+        if (voiceState.inVoiceChannel()) {
+            return "\n**In voicechannel:** " + (voiceState.getChannel().getAsMention());
+        } else {
```
illuminator3, line 134: Not needed else.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+     *
+     * @param member the {@link Member} to take the booster properties from
+     * @return user readable {@link String}
+     */
+    private static @NotNull String possibleBoosterToStringItem(final @NotNull Member member) {
+        OffsetDateTime timeBoosted = member.getTimeBoosted();
+        if (null != timeBoosted) {
+            return "\n**Is booster:** true \n**Boosting since:** "
+                    + DATE_TIME_FORMAT.format(timeBoosted);
+        } else {
```
illuminator3, line 177: Not needed else.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/WhoIsCommand.java
```diff
@@ -0,0 +1,263 @@
...
+     *        (recommend {@link java.util.EnumSet}
+     * @return user readable {@link StringBuilder}
+     */
+    private static @NotNull StringBuilder userFlagsToStringItem(
+            final @NotNull Collection<User.UserFlag> flags) {
+        String formattedFlags = formatUserFlags(flags);
+
+        if (formattedFlags.isBlank()) {
+            return hypeSquadToStringItem(flags);
+        } else {
```
illuminator3, line 219: Not needed else.

illuminator3 (commented):
In my opinion, it would look cleaner without the elses (what even is the plural of else?). Albeit, it's not required, and thus this review is not requesting changes.

Zabuzard:
Taking over. Ready to review.

Zabuzard (approved):

Zabuzard:
Tested, works:

test

illuminator3 (approved):

PR #355, opened 2022-01-25
Add unit tests for tag-system

### LINKED ISSUE #213: Add unit tests to the tag commands
We recently merged command unit testing utilities.

That being done, we should now add unit tests for the tag system (`tags`, `tag`, `tag-manage`) to increase our test coverage.

Zabuzard:
### Overview

Implements and closes #213 . Adds unit tests to the tag system, i.e.
* `TagSystem`,
* `TagCommand`,
* `TagsCommand` and
* `TagManageCommand`.

The tests themselves should be rather self-explanatory.

Majority of changes in this PR are not logic changes but come from the fact that we made `Config` mockable, and hence had to get rid of the **singleton** - which bleeds into almost all commands.

### Features

While at it, we had to implement multiple more general features from which other unit tests will also benefit in the future:

#### Database

We added a method `Database.createMemoryDatabase(...)` to conveniently use an in-memory database for testing. An usage can be seen in `TagsCommandTest`, where it is called `@Before` each test.

#### Mockito

The whole mockito test suite, in particular revolving around `JdaTester` has been enhanced and generally supports more calls now.

And we also added some bigger features to it:

##### Button events

In order to test button events, we added a `JdaTester#createButtonClickEvent()` method that returns a builder that can create mocks of the event, similar to how `JdaTester#createSlashCommandEvent(...)` works.

The API of this builder is very similar to JDAs `Method` interface and similar:
```java
// Default message with a delete button
jdaTester.createButtonClickEvent()
  .actionRows(ActionRow.of(Button.of(ButtonStyle.DANGER, "1", "Delete"))
  .buildWithSingleButton();
```
or
```java
// More complex message with a user who clicked the button that is not the message author and multiple buttons
Button clickedButton = Button.of(ButtonStyle.PRIMARY, "1", "Next");

jdaTester.createButtonClickEvent()
  .message(new MessageBuilder()
    .setContent("See the following entry")
    .setEmbeds(
      new EmbedBuilder()
        .setDescription("John")
        .build())
  .build())
  .userWhoClicked(jdaTester.createMemberSpy(5))
  .actionRows(
    ActionRow.of(Button.of(ButtonStyle.PRIMARY, "1", "PREVIOUS"),
    clickedButton)
  .build(clickedButton);
```

##### Command spies

There is now a method `JdaTester#spySlashCommand(command)` that makes it possible to work with a command that is spied upon and additionally is setup properly. For example, it works with a mocked component ID generator.

##### Member spies

We also added a method `JdaTester#createMemberSpy(userId)` that is able to create valid member instances which are spied upon. Useful for testing things that require members. For example buttons that have been clicked by a user different to the message author.

##### USER option for slash command events

We enhanced the `SlashCommandEventBuilder` and added overloads `option(String, User)` and `option(String, Member)`.

So now it also supports options with types `USER`, not just `String`. For example:
```java
Member member = jdaTester.createMemberSpy(1);

var event = jdaTester.createSlashCommandEvent(command)
  .option("id", "foo");
  .option("reply-to", member)
  .build();
```

To make this work, the implementation had to be changed a bit. Therefore, other types can now also easily be added in the future.

##### Setting the user who triggered a slash command event

The `SlashCommandEventBuilder` now has a method `userWhoTriggered(user)` that can be used to set the member who triggered the event. Useful for mocking, for example to adjust permissions of a user who wants to use the ban command.

##### `RestAction` mocking

`JdaTester` now has two methods to create **success** and **failure** mocks of `RestAction`s. This can be very useful when testing stuff that involves calls such as `TextChannel#retrieveMessageById` or similar. Example:
```java
var jdaTester = new JdaTester();

var message = new MessageBuilder("Hello World!").build();
var action = jdaTester.createSucceededActionMock(message);

doReturn(action).when(jdaTester.getTextChannelSpy()).retrieveMessageById("1");
```
or also
```java
var jdaTester = new JdaTester();

var reason = new FooException();
var action = jdaTester.createFailedActionMock(reason);

doReturn(action).when(jdaTester.getTextChannelSpy()).retrieveMessageById("1");
```

### Checklist

Unit tests
* [x] TagSystem
* [x] TagsCommand
* [x] TagCommand
* [x] TagManageCommand - raw
* [x] TagManageCommand - create
* [x] TagManageCommand - create-with-message
* [x] TagManageCommand - edit
* [x] TagManageCommand - edit-with-message
* [x] TagManageCommand - delete

Mocking
* [x] spySlashCommand
* [x] createButtonClickEvent
* [x] add `USER` as option to `SlashCommandEventBuilder`
* [x] `userWhoTriggered(user)` for `SlashCommandEventBuilder`
* [x] `Config.getInstance()` mocking
* [x] `RestAction<T>` mocking (success & failure)
* [x] Discord API exception mocking

### DIFF
--- a/application/src/test/java/org/togetherjava/tjbot/commands/basic/PingCommandTest.java
+++ b/application/src/test/java/org/togetherjava/tjbot/commands/basic/PingCommandTest.java
@@ -1,22 +1,39 @@
 package org.togetherjava.tjbot.commands.basic;
 
 import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.togetherjava.tjbot.commands.SlashCommand;
 import org.togetherjava.tjbot.jda.JdaTester;
 
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 final class PingCommandTest {
-    @Test
-    void pingCommand() {
-        SlashCommand command = new PingCommand();
-        JdaTester jdaTester = new JdaTester();
+    private JdaTester jdaTester;
+    private SlashCommand command;
 
+    private @NotNull SlashCommandEvent triggerSlashCommand() {
         SlashCommandEvent event = jdaTester.createSlashCommandEvent(command).build();
         command.onSlashCommand(event);
+        return event;
+    }
 
-        verify(event, times(1)).reply("Pong!");
+    @BeforeEach
+    void setUp() {
+        jdaTester = new JdaTester();
+        command = jdaTester.spySlashCommand(new PingCommand());
+    }
+
+    @Test
+    @DisplayName("'/ping' responds with pong")
+    void pingRespondsWithPong() {
+        // GIVEN
+        // WHEN using '/ping'
+        SlashCommandEvent event = triggerSlashCommand();
+
+        // THEN the bot replies with pong
+        verify(event).reply("Pong!");
     }
 }
--- a/application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
+++ b/application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
@@ -1,0 +1,413 @@
+package org.togetherjava.tjbot.commands.tags;
+
+import net.dv8tion.jda.api.MessageBuilder;
+import net.dv8tion.jda.api.Permission;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.requests.ErrorResponse;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.togetherjava.tjbot.commands.SlashCommand;
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.Tags;
+import org.togetherjava.tjbot.jda.JdaTester;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.AdditionalMatchers.aryEq;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+final class TagManageCommandTest {
+    private TagSystem system;
+    private JdaTester jdaTester;
+    private SlashCommand command;
+    private Member moderator;
+
+    private static @NotNull MessageEmbed getResponse(@NotNull SlashCommandEvent event) {
+        ArgumentCaptor<MessageEmbed> responseCaptor = ArgumentCaptor.forClass(MessageEmbed.class);
+        verify(event).replyEmbeds(responseCaptor.capture());
+        return responseCaptor.getValue();
+    }
+
+    @BeforeEach
+    void setUp() {
+        Config config = mock(Config.class);
+        String moderatorRoleName = "Moderator";
+        when(config.getTagManageRolePattern()).thenReturn(moderatorRoleName);
+
+        Database database = Database.createMemoryDatabase(Tags.TAGS);
+        system = spy(new TagSystem(database));
+        jdaTester = new JdaTester();
+        command = new TagManageCommand(system, config);
+
+        moderator = jdaTester.createMemberSpy(1);
+        Role moderatorRole = mock(Role.class);
+        doReturn(true).when(moderator).hasPermission(any(Permission.class));
+        doReturn(List.of(moderatorRole)).when(moderator).getRoles();
+        when(moderatorRole.getName()).thenReturn(moderatorRoleName);
+    }
+
+    private @NotNull SlashCommandEvent triggerRawCommand(@NotNull String tagId) {
+        return triggerRawCommandWithUser(tagId, moderator);
+    }
+
+    private @NotNull SlashCommandEvent triggerRawCommandWithUser(@NotNull String tagId,
+            @NotNull Member user) {
+        SlashCommandEvent event = jdaTester.createSlashCommandEvent(command)
+            .setSubcommand(TagManageCommand.Subcommand.RAW.getName())
+            .setOption(TagManageCommand.ID_OPTION, tagId)
+            .setUserWhoTriggered(user)
+            .build();
+
+        command.onSlashCommand(event);
+        return event;
+    }
+
+    private @NotNull SlashCommandEvent triggerCreateCommand(@NotNull String tagId,
+            @NotNull String content) {
+        return triggerTagContentCommand(TagManageCommand.Subcommand.CREATE, tagId, content);
+    }
+
+    private @NotNull SlashCommandEvent triggerEditCommand(@NotNull String tagId,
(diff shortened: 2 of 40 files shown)


application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
```diff
@@ -0,0 +1,393 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.AdditionalMatchers.aryEq;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+final class TagManageCommandTest {
```
Heatmanofurioso, line 30: Add a DisplayName to all tests in this class

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagSystemTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.togetherjava.tjbot.db.generated.tables.Tags;
+import org.togetherjava.tjbot.jda.JdaTester;
+
+import java.util.Optional;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+final class TagSystemTest {
```
Heatmanofurioso, line 16: Add a DisplayName to all tests in this class

application/src/test/java/org/togetherjava/tjbot/commands/basic/PingCommandTest.java
```diff
@@ -5,7 +5,6 @@
 import org.togetherjava.tjbot.commands.SlashCommand;
 import org.togetherjava.tjbot.jda.JdaTester;
 
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 final class PingCommandTest {
```
Heatmanofurioso, line 10: Add a DisplayName to the test

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagsCommandTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.togetherjava.tjbot.db.generated.tables.Tags;
+import org.togetherjava.tjbot.jda.JdaTester;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.Mockito.*;
+
+final class TagsCommandTest {
```
Heatmanofurioso, line 25: Add a DisplayName to all tests in this class

Heatmanofurioso (changes requested):

Heatmanofurioso:
I usually follow a convention on the Display Names of tests, where I describe my GIVEN, WHEN, THEN in the test name itself.

Would you think it's a good idea to adapt to that convention @Zabuzard ? Probably starting on this PR "hence me asking it here"

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
Zabuzard: What if the test-method-name is already descriptive enough? I.e. if the DisplayName would just be a 1:1 copy of it?

I get that it makes sense to use it if u need more room to explain it, but thats not always the case, is it?

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagSystemTest.java
Tais993: This class has clear methods names, I'm confident I know what they do. I'd still rather read a sentence than the methods name though.

application/src/test/java/org/togetherjava/tjbot/jda/ButtonClickEventBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+ *     .build())
+ *   .userWhoClicked(jdaTester.createMemberSpy(5))
+ *   .actionRows(
+ *     ActionRow.of(Button.of(ButtonStyle.PRIMARY, "1", "Previous"),
+ *     clickedButton)
+ *   .build(clickedButton);
+ * }
+ * </pre>
+ */
+public final class ButtonClickEventBuilder {
```
Tais993, line 70: It's not a record, these methods are setters? I don't understand where the naming is from

They should be named "set" so it's clear what they do, currently I'd think they all *add* to the existing value.
But they set, not add, they replace the old value.

Tais993 (changes requested):

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
Heatmanofurioso: As Tais said. The intention of the DisplayName is for a more human-readable label.

IDEs show it when running tests, coverage reports also use them as titles, and your build logs also use them as a title "I don't remember if this last one has to be toggled on or not"

Zabuzard: Thats fair. I tried to experiment a bit with the different conventions I know. So you see some more basic tests that have just the name of the method they test and then there are also more complex tests where each test only tests one specific scenario, using the `GIVEN, WHEN, THEN` syntax.

I agree on the `@DisplayName` bit, will add it.

application/src/test/java/org/togetherjava/tjbot/jda/ButtonClickEventBuilder.java
Zabuzard: Kinda depends on which builder pattern you are following. There are builder patterns that use `setXXX` and some that just use `xxx` (Intellijs built-in builder-generator for example). Records also use just `xxx`.

Tais993: Records aren't builders, that's a huge difference.

And I just feel like we've to be clear about it's behaviour, either add it in docs or make it clear in the method's name

Zabuzard: As said, this is not just records. It is actually a common pattern. Lombdok and Intellijs default settings create builders like that. Maybe you are not aware of this convention or dont like it, thats fair - but it is common.

At the same time, `setXXX` and `withXXX` are also common 🤷 

Personally, I do not care that much and am willing to change it if it helps you. Just want to make sure that its clear that all 3 are commonly used patterns for builders.

Zabuzard: Renamed it to `setXXX` now.

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagsCommandTest.java
Zabuzard: Done.

application/src/test/java/org/togetherjava/tjbot/commands/basic/PingCommandTest.java
Zabuzard: done

application/src/test/java/org/togetherjava/tjbot/commands/tags/TagManageCommandTest.java
```diff
@@ -0,0 +1,413 @@
...
+            .retrieveMessageById(id);
+    }
+
+    private void failOnRetrieveMessage(@NotNull String messageId, @NotNull Throwable failure) {
+        doReturn(jdaTester.createFailedActionMock(failure)).when(jdaTester.getTextChannelSpy())
+            .retrieveMessageById(messageId);
+    }
+
+    @Test
+    @DisplayName("Users without a specific role can not use '/tag-manage'")
```
Tais993, line 150:
```suggestion
    @DisplayName("Users without the required role can not use '/tag-manage'")
```

Zabuzard:
> If possible, could you make it so this PR becomes 2 commits.
> 
> 1 updating JDATester, and ButtonClickEventBuilder etc 1 adding the tests
> 
> This makes imho more sense

I beautified the commit history (no code changes). Cheers

Tais993 (approved):

Heatmanofurioso (approved):

PR #359, opened 2022-02-01
Fixed broken encoding for unit tests

Zabuzard:
### Overview

Apparently the previous setup didnt work and unit tests have been executed on the wrong encoding.

Resulting in minor issues, such as unicode usage inside tests not working as expected. For example the following test fails:
```java
assertEquals("\u2022 foo", "• foo");
```
> expected: <• foo> but was: <â€¢ foo>
> Expected :• foo
> Actual   :â€¢ foo

This fixes the issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,8 +51,7 @@ subprojects {
         }
     }
 
-
-    compileJava {
+    def compileTasks = {
         options.encoding = 'UTF-8'
 
         // Makes spotlessApply task run on every compile/build.
@@ -63,16 +62,8 @@ subprojects {
         targetCompatibility = JavaVersion.VERSION_17
     }
 
-    compileTestJava {
-        options.encoding = 'UTF-8'
-
-        // Makes spotlessApply task run on every compile/build.
-        dependsOn 'spotlessApply'
-
-        // Nails the Java-Version of every Subproject
-        sourceCompatibility = JavaVersion.VERSION_17
-        targetCompatibility = JavaVersion.VERSION_17
-    }
+    compileJava(compileTasks)
+    compileTestJava(compileTasks)
 
     spotless {
         java {


build.gradle
```diff
@@ -63,6 +63,17 @@ subprojects {
...
+    compileTestJava {
+        options.encoding = 'UTF-8'
+
+        // Makes spotlessApply task run on every compile/build.
+        dependsOn 'spotlessApply'
+
+        // Nails the Java-Version of every Subproject
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
+    }
```
Zabuzard, line 75: This is a direct copy of `compileJava`. Can we get rid of the duplication somehow (helper method or similar)?

@illuminator3 any idea?

illuminator3: `compileTestJava` and `compileJava` both want a closure. You can just make a variable for it and then assign both to it in a single expression:
```groovy
def javaCompileThings = {
        options.encoding = 'UTF-8'

        // Makes spotlessApply task run on every compile/build.
        dependsOn 'spotlessApply'

        // Nails the Java-Version of every Subproject
        sourceCompatibility = JavaVersion.VERSION_17
        targetCompatibility = JavaVersion.VERSION_17
}

compileTestJava = compileJava = javaCompileThings
```I'm not sure as of right now if the above works, in case it doesn't, try the following:
```gradle
compileTestJava(javaCompileThings)
compileJava(javaCompileThings)
```

Zabuzard: Cool stuff, ill try

Zabuzard: Former didnt work:
> Cannot assign 'Closure' to 'JavaCompile' 

But latter did, thanks

Heatmanofurioso (approved):

Tais993 (approved):

java-coding-prodigy (approved):

PR #361, opened 2022-02-05
Feature/jda v5

Tais993:
### This PR is still a draft, JDA v5 is still in alpha

I'll keep updating this PR, so when JDA v5 releases it can be as smooth as possible, and as fast as possible.
Since JDA v5 will add support for context menus, threads, autocomplete and a lot more this is really important for further development. A big example is our help-system, we can't switch to threads if JDA doesn't support it.
So once v5 releases we can jump on that as fast as possible.

For more info on what exactly is changing during JDA v5, join [JDA's Discord ](https://discord.gg/jda), and read the `jda5-changes` channel. I'll update this when a wiki article is created


Tais993:
ah fuck

PR #366, opened 2022-02-06
Added logger field

### LINKED ISSUE #365: Generation of logger inside methods
**Describe the bug**
`DatabaseProvider` generates new loggers inside its methods

**Expected behavior**
The logger should be put into a private static field called `logger`

**Additional context**
Occurences:
1) line 37
```java
catch (final SQLException e) {
            LoggerFactory.getLogger(DatabaseProvider.class)
                .error("Exception while creating Database.", e);
            throw new FatalBeanException("Could not create Database.", e);
        }
```
2) line 49
```java
catch (final IOException e) {
            LoggerFactory.getLogger(DatabaseProvider.class)
                .error("Exception while creating Database-Path.", e);
        }
```

java-coding-prodigy:
Fixes #365

### DIFF
--- a/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
+++ b/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
@@ -35,6 +35,7 @@ import org.vaadin.artur.helpers.LaunchUtil;
 @Push
 public class Application extends SpringBootServletInitializer implements AppShellConfigurator {
 
+    @SuppressWarnings({"java:S2387"})
     private static final Logger logger = LoggerFactory.getLogger(Application.class);
 
     public static void main(String[] args) {
--- a/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
+++ b/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
@@ -35,12 +35,11 @@ import org.vaadin.artur.helpers.LaunchUtil;
 @Push
 public class Application extends SpringBootServletInitializer implements AppShellConfigurator {
 
-    @SuppressWarnings({"java:S2387"})
-    private static final Logger logger = LoggerFactory.getLogger(Application.class);
+    private static final Logger applicationLogger = LoggerFactory.getLogger(Application.class);
 
     public static void main(String[] args) {
         if (args.length > 1) {
-            logger
+            applicationLogger
                 .error("Usage: Provide a single Argument, containing the Path to the Config-File");
             System.exit(1);
         }
--- a/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
+++ b/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
@@ -5,6 +5,7 @@ import com.vaadin.flow.component.page.AppShellConfigurator;
 import com.vaadin.flow.component.page.Push;
 import com.vaadin.flow.server.PWA;
 import com.vaadin.flow.theme.Theme;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
@@ -34,9 +35,11 @@ import org.vaadin.artur.helpers.LaunchUtil;
 @Push
 public class Application extends SpringBootServletInitializer implements AppShellConfigurator {
 
+    private static final Logger logger = LoggerFactory.getLogger(Application.class);
+
     public static void main(String[] args) {
         if (args.length > 1) {
-            LoggerFactory.getLogger(Application.class)
+            logger
                 .error("Usage: Provide a single Argument, containing the Path to the Config-File");
             System.exit(1);
         }
--- a/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/DatabaseProvider.java
+++ b/logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/DatabaseProvider.java
@@ -1,5 +1,6 @@
 package org.togetherjava.tjbot.logwatcher;
 
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.FatalBeanException;
 import org.springframework.beans.factory.config.BeanDefinition;
@@ -21,6 +22,8 @@ public class DatabaseProvider {
     private final Database db;
     private final Config config;
 
+    private static final Logger logger = LoggerFactory.getLogger(DatabaseProvider.class);
+
     public DatabaseProvider(final Config config) {
         this.config = config;
         this.db = createDB();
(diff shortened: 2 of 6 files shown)


Tais993 (approved):

krankkkk:
As said in the ticket, no reason for the field. So I would close this

marko-radosavljevic (changes requested):
I like consistency, so the changes make sense to me. Please check if there is any other place in the codebase where similar thing occurs, so we can align.

I will temporarily put a hold on this PR, because code owner (@krankkkk) disagrees with the changes. ^^

java-coding-prodigy:
> Please check if there is any other place in the codebase where similar thing occurs, so we can align.

If I find something will I add that to this PR itself, or another one?

Tais993:
This PR is fine

java-coding-prodigy:
SonarLint is unhappy that `SpringBootServletInitializer`(super class of `Application` in logviewer) has a protected field with the name `logger`. What do we do about this?

Tais993:
~~Time to remove sonarlint since it's giving me mental issues~~

All loggers should be private, so change that

java-coding-prodigy:
> All loggers should be private, so change that

Um, I cannot do that, it is a class from Spring that is causing this issue. Note that the logger there is of class `org.apache.commons.logging.Log` and not an SLF4J logger

Tais993:
Rename our logger to log if?

Just Sonarlint being stupid again

java-coding-prodigy:
> Just Sonarlint being stupid again

I *could* suppress that warning, imo that would be better.

Tais993:
Then I'd do that

logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
```diff
@@ -34,9 +35,12 @@
 @Push
 public class Application extends SpringBootServletInitializer implements AppShellConfigurator {
 
+    @SuppressWarnings({"java:S2387"})
```
Zabuzard, line 38: What is this suppress doing here? I dont like it and would remove it. But if you want to keep it, put a comment explaining what it is about.

Zabuzard (changes requested):

logviewer/src/main/java/org/togetherjava/tjbot/logwatcher/Application.java
java-coding-prodigy: The `SpringBootServletInitializer` class has a protected field called `logger` and sonar lint is unhappy that the field names are same.

Zabuzard: then dont ignore the warning but rename the field lol.

Zabuzard: or, if springs logger is basically the logger already, **reuse it** instead of creating a new logger.

java-coding-prodigy: The springs logger is from the apache logging framework, also renaming this field would lose consistency which is against the whole idea of this PR.

Tais993: > or, if springs logger is basically the logger already, **reuse it** instead of creating a new logger.

It's our code, not spring's code. So I 100% disagree

If we have an error, and the error is in our code, it should be logged in OUR code, not Spring's

Tais993: > then dont ignore the warning but rename the field lol.

Inconsistency, no

Zabuzard: > It's our code, not spring's code. So I 100% disagree
It kinda depends on the purpose of this field. If this is a protected field meant to be used by implementing classes and if it follows our existing logging configuration already, then it might make sense.

Other than that I agree with you and we should rename our field to get rid of the name clash and confusion.

Zabuzard: > > then dont ignore the warning but rename the field lol.
> 
> Inconsistency, no

Consistency is not an excuse to write bad code. Hiding **visible** fields is a bad idea, causes bugs and confusion.

Zabuzard: > Technically we could reuse that logger, but it is from the apache loggings framework, I hope that is ok.

What does its javadoc say? Its protected/public on purpose. So I would assume that they explain when and how to use it in their Javadoc.

java-coding-prodigy: > What does its javadoc say?

There is no javadoc

```java
public abstract class SpringBootServletInitializer implements WebApplicationInitializer {

	protected Log logger; // Don't initialize early

	private boolean registerErrorPageFilter = true;
	...
}
```

java-coding-prodigy: Wait, wtf have we been thinking 🤦 It is a non-static field, we cannot use it in the `main` method.

Zabuzard: > Wait, wtf have we been thinking 🤦 It is a non-static field, we cannot use it in the main method.

We are still hiding it, which we should not. We are extending this class after all. Imagine you would have a super class
```java
class Animal {
  protected String name;
  ...
}
```
then you also wouldnt start doing
```java
class Dog extends Animal {
  private static final String name = "Foo";
  ...
}
```
but call it differently to make clear what differentiates it from the `name` that you have inside the application already.

Otherwise people are deeply confused when for example using **inside a non-static Dog method** `name`, the call is ambigious then to a reader.

java-coding-prodigy: We never call `logger` in a non-static context though. I think it should be fine.

Zabuzard: @java-coding-prodigy Just call it `applicationLogger` please - gets rid of the name hiding, makes its purpose clear (or rather who owns it) and everyone is happy.

Zabuzard: @java-coding-prodigy could u finish this quickly when u have time? :) or should someone take over?

java-coding-prodigy: Really sorry, I had quite a lot of work to do(I still am busy, but can do this). Working on it rn.

Zabuzard: No worries at all. Take your time.

java-coding-prodigy: Done 👍

Zabuzard (approved):

PR #354, opened 2022-01-25
Pushing the remider command, which works while the bot is online.

IslamSakrak:

PR #373, opened 2022-02-09
Adding /remind command

### LINKED ISSUE #325: Add a reminder command
Used to do remind people to do something in at a certain time or in a certain number of days, hour, etc 

Can be useful to stop people from forgetting things.

The user will be able to type `/remindme time what to remind(optional)`

Zabuzard:
### Overview

Originally from #354 , taken over. Implements and closes #325 .

This adds the `/remind` command which can be used to automatically send reminders to oneself at a future date.

Reminders are sent in the channel the command has been used in (or via DM if something goes wrong).

example

### Details

Technically, reminders are simple saved in a database table, where an automatic routine will pick them up at time, send them and delete them from the table.

#### Database table

The table is quite straightforward. It has some metadata, data to be able to locate the channel to post to and the actual content and time to fire the reminder.

```sql
CREATE TABLE pending_reminders
(
    id         INTEGER   NOT NULL PRIMARY KEY AUTOINCREMENT,
    created_at TIMESTAMP NOT NULL,
    guild_id   BIGINT    NOT NULL,
    remind_at  TIMESTAMP NOT NULL,
    content    TEXT      NOT NULL
)
```

#### `RemindCommand` UX

The user can enter a time period as `amount` and `unit`, which determine when the reminder will be send.
* `amount` - an integer between `1` and `1_000`
* `unit` - dropdown of `"minutes", "hours", "days", "weeks", "months", "years"`

After using the command, it verifies that the time is within reasonable limits (max 3 years) and if the user has not reached the limit of pending reminders per user (100). If not, an error message is send.

limit example

max pending reminders

Those restrictions are mostly to prevent the database and hence our disk space from overflowing when users are spamming around.

#### `RemindRoutine`

The routine is quite trivial. It implements `Routine`, with a schedule of **30 seconds**. In each run, it gets hands on all pending reminders that are due, sends them and deletes them from the table afterwards.

To send them, it locates the corresponding guild and channel, retrieves the original author and builds a slim embed.

##### Error cases

If the guild and channel could be found, but the author couldnt be retrieved anymore, it sends the message with `Unknown User`.

user not found

If the guild or channel cant be found, the bot attempts to send it via DM to the author instead.

DM send

If sending the DM failed as well (user deleted, user disabled DMs, ...), the bot sends a `WARN` log message:

> WARN RemindRoutine - Failed to send a reminder (id '19'), skipping it. This can be due to a network issue, but also happen if the bot disconnected from the target guild and the user has disabled DMs or has been deleted.

### Future plans

The command would greatly benefit from JDA/Discord adding a date/time selector. It could then be split into `/remind in` and `/remind at`. A `// TODO` keeps track of this.

Unfortunately, it is not really possible to add `/remind at` (eventhough the use case is very reasonable) as of today without having hands on the users timezone.

### Checklist

* [x] Store reminders
* [x] Send & Delete reminders
* [x] Mention the original user
* [x] Limit period to 3 years
* [x] Limit reminders per user
* [x] Send user DM if unknown guild/channel


Heatmanofurioso (dismissed):

Tais993:
Silently ignoring the reminder shouldn't happen in my honest opinion.

If the channel/guild can't be found we should (try to) DM the user instead.

Zabuzard:
> If the channel/guild can't be found we should (try to) DM the user instead.

Good idea. Will do that.

Zabuzard (commented):
Send user DM if reminder cant be sent due to unknown guild/channel.

Zabuzard:
> If the channel/guild can't be found we should (try to) DM the user instead.

Implemented.

Heatmanofurioso (dismissed):

Tais993 (changes requested):
I'm not happy with how you ended up handling it.

This ReminderRoute class had me confused for too long, these onErrorMap's into null are unreadable and it resulted in a mess of ternary and logic which should be separated. 

Instead, don't map into null resulting in unreadable ternary mess, handle the error separately.

Right now  you just throw a IllegalStateException, after adding a lot of logic which makes it severely more complicated, when this can also just be done by handling the error when queueing

I'm sorry, but this isn't gonna cut it for me.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+                "Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue,"
+                        + " but also happen if the bot disconnected from the target guild and the"
+                        + " user has disabled DMs or has been deleted.",
+                id));
+    }
+
+    private static @NotNull MessageEmbed createReminderEmbed(@NotNull CharSequence content,
+            @NotNull TemporalAccessor createdAt, @Nullable User author) {
+        String authorName = author == null ? "Unknown user" : author.getAsTag();
+        String authorIconUrl = author == null ? null : author.getAvatarUrl();
```
Tais993, line 123: This method only runs if ReminderRoute#isUndeliverable is false, so why is User nullable?

And use User#getEffectiveAvatarUrl, this one doesn't work when no avatar has been set

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+    private static RestAction<ReminderRoute> computeReminderRoute(@NotNull JDA jda, long guildId,
+            long channelId, long authorId) {
+        // If guild and channel can still be found, send there
+        Guild guild = jda.getGuildById(guildId);
+        if (guild != null) {
+            TextChannel channel = guild.getTextChannelById(channelId);
+            if (channel != null) {
+                return createGuildReminderRoute(jda, authorId, channel);
+            }
+        }
```
Tais993, line 76:
```suggestion
        TextChannel channel = jda.getTextChannelById(channelId);
        if (channel != null) {
            return createGuildReminderRoute(jda, authorId, channel);
        }
```

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static @NotNull RestAction<ReminderRoute> createDmReminderRoute(@NotNull JDA jda,
+            long authorId) {
+        return jda.openPrivateChannelById(authorId)
+            .onErrorMap(error -> null)
+            .map(channel -> channel == null ? new ReminderRoute(null, null, null)
+                    : new ReminderRoute(channel, channel.getUser(),
+                            "(Sending your reminder directly, because I was unable to"
+                                    + " locate the original channel you wanted it to be send to)"));
```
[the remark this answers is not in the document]
Tais993: Convo on Discord

Tais993, line 97:
```suggestion
        return jda.openPrivateChannelById(authorId)
                .map(channel -> new ReminderRoute(channel, channel.getUser(),
                        "(Sending your reminder directly, because I was unable to"
                                + " locate the original channel you wanted it to be send to)"));
```

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static void sendReminderViaRoute(@NotNull RestAction<ReminderRoute> routeAction,
+            long id, @NotNull CharSequence content, @NotNull TemporalAccessor createdAt) {
+        routeAction.flatMap(route -> {
+            if (route.isUndeliverable()) {
+                throw new IllegalStateException("Route is not deliverable");
+            }
```
Tais993, line 105: ReminderRoute#isUndeliverable is always false after the suggestion above, thus remove the method and it's check

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+
+            MessageEmbed embed = createReminderEmbed(content, createdAt, route.target());
+            if (route.description() == null) {
+                return route.channel().sendMessageEmbeds(embed);
+            }
+            return route.channel().sendMessage(route.description()).setEmbeds(embed);
+        }).queue(message -> {
+        }, failure -> logger.warn(
```
Tais993, line 113:
```suggestion
            return route.channel().sendMessage(route.description()).setEmbeds(embed);
            
        }).queue(message -> {
        }, failure -> logger.warn(
```
Because spotless puts the success and failure consumer on the same line, it becomes harder to read.
Preferably add an empty line to indicate the end of the lambda above

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+            .setTimestamp(createdAt)
+            .setColor(AMBIENT_COLOR)
+            .build();
+    }
+
+    private record ReminderRoute(@Nullable MessageChannel channel, @Nullable User target,
+            @Nullable String description) {
+        boolean isUndeliverable() {
+            return channel == null && target == null;
+        }
```
Tais993, line 137:
```suggestion
```
See above

Tais993 (changes requested):
After convo and some clearing up in the Discord, new suggestions have come. The old ones have been resolved/deleted by me.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: Oh, the channel ids are unique across guilds as well? Cool, thanks.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,139 @@
...
+        routeAction.flatMap(route -> {
+            if (route.isUndeliverable()) {
+                throw new IllegalStateException("Route is not deliverable");
+            }
+
+            MessageEmbed embed = createReminderEmbed(content, createdAt, route.target());
+            if (route.description() == null) {
+                return route.channel().sendMessageEmbeds(embed);
+            }
+            return route.channel().sendMessage(route.description()).setEmbeds(embed);
```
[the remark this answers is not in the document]
Zabuzard: `setContent(null)` doesnt work, unfortunately. I remember that I tried it but it didnt work.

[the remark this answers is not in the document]
Tais993: It's nullable, sounds to me like you did something else.

Did you try `sendMessage(null).addEmbeds();`
That could be not null, it actually makes sense to be not null.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,137 @@
...
+            .sendMessageEmbeds(createReminderEmbed(content, createdAt, route.target()))
+            .content(route.description());
+
+        Consumer<Throwable> logFailure = failure -> logger.warn(
+                "Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue,"
+                        + " but also happen if the bot disconnected from the target guild and the"
+                        + " user has disabled DMs or has been deleted.",
+                id);
+
+        routeAction.flatMap(sendMessage).queue(doNothing(), logFailure);
```
Tais993, line 113: I understand your intention, but I don't really like it personally?

It removes, the normal order you have when reading.

I rather have them as inline lambda's tbh

Tais993 (commented):
This time I actually understood it.

There's 2 notes I'd like to give, one is my review.

Second one, everything is split up with methods.
I understand why, but it kinda doesn't feel like there's 1 central place to look at to understand the code. I'd personally inline the computeReminderRoute.
It's maybe split up, too much? I don't know, hopefully anyone else has an opinion on this matter.
@Heatmanofurioso, @marko-radosavljevic?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private static <T> @NotNull Consumer<T> doNothing() {
+        return a -> {
+        };
+    }
+
+    private record ReminderRoute(@NotNull MessageChannel channel, @Nullable User target,
+            @Nullable String description) {
+    }
```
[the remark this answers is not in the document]
Zabuzard: sounds good

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,137 @@
...
+        // Otherwise, attempt to DM the user directly
+        return createDmReminderRoute(jda, authorId);
+    }
+
+    private static @NotNull RestAction<ReminderRoute> createGuildReminderRoute(@NotNull JDA jda,
+            long authorId, @NotNull TextChannel channel) {
+        return jda.retrieveUserById(authorId)
+            .onErrorMap(error -> null)
+            .map(author -> new ReminderRoute(channel, author,
+                    author == null ? null : author.getAsMention()));
```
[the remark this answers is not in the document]
Zabuzard: sounds good

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        Instant now = Instant.now();
+        database.write(context -> context.selectFrom(PENDING_REMINDERS)
+            .where(PENDING_REMINDERS.REMIND_AT.lessOrEqual(now))
+            .stream()
+            .forEach(pendingReminder -> {
+                sendReminder(jda, pendingReminder.getId(), pendingReminder.getChannelId(),
+                        pendingReminder.getAuthorId(), pendingReminder.getContent(),
+                        pendingReminder.getCreatedAt());
```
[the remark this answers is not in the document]
Zabuzard: Imo that would clutter the surrounding context. Readers of `runRoutine` shouldnt have to care in this step how the reminder is send.

It is unfortunate though that the list of arguments is so long. Maybe I can simplify that part instead.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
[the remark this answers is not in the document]
Zabuzard: decided for factories instead.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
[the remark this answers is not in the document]
Zabuzard: decided for factories instead.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
```diff
@@ -0,0 +1,152 @@
...
+        // TODO As soon as JDA offers date/time selector input, this should also offer
+        // "/remind at" next to "/remind in" and use subcommands then
+        OptionData whenAmount = new OptionData(OptionType.INTEGER, WHEN_AMOUNT_OPTION,
+                "when to remind you, the amount of the time period (e.g. [5] weeks)", true)
+                    .setRequiredRange(MIN_WHEN_AMOUNT, MAX_WHEN_AMOUNT);
+        OptionData whenUnit = new OptionData(OptionType.STRING, WHEN_UNIT_OPTION,
+                "when to remind you, the unit of the time period (e.g. 5 [weeks])", true);
+        WHEN_UNITS.forEach(unit -> whenUnit.addChoice(unit, unit));
+
+        getData().addOptions(whenAmount, whenUnit)
```
Tais993, line 67: I'd personally do unit before amount. People might get confused and think it's the amount in weeks (because of the example)

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
Zabuzard: I get that. But it reads kinda weird then `"remind in weeks 5 content huhu"` vs `"remind in 5 weeks content huhu"`.

But yeah, I guess its better like that if we split the input into two sections.

Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,142 @@
...
+    private static void sendReminderViaRoute(@NotNull RestAction<ReminderRoute> routeAction,
+            long id, @NotNull CharSequence content, @NotNull TemporalAccessor createdAt) {
+        Function<ReminderRoute, MessageAction> sendMessage = route -> route.channel
+            .sendMessageEmbeds(createReminderEmbed(content, createdAt, route.target()))
+            .content(route.description());
+
+        Consumer<Throwable> logFailure = failure -> logger.warn(
+                "Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue,"
+                        + " but also happen if the bot disconnected from the target guild and the"
+                        + " user has disabled DMs or has been deleted.",
```
Tais993, line 105: Textblock?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private record ReminderRoute(@NotNull MessageChannel channel, @Nullable User target,
+            @Nullable String description) {
+        static ReminderRoute toPublic(@NotNull TextChannel channel, @Nullable User target) {
+            return new ReminderRoute(channel, target,
+                    target == null ? null : target.getAsMention());
+        }
+
+        static ReminderRoute toPrivate(@NotNull PrivateChannel channel, @NotNull User target) {
```
Tais993, line 136: In the future this might require some handling since the PrivateChannel#getUser method will become nullable.
This is more of a note for in the future, could add a TODO with a note so we don't forget that?

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: No, cause then I would have to add newlines and it shouldnt have newlines.

Zabuzard: Unless I use `\`.
```java
"""foo \
bar \
baz"""
```
what do u think?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: in which version? If we add such a TODO, we need to explicitly link the "when" somehow. otherwise its causing too much confusion when its not coming in the next few weeks

Tais993: JDA 5, no "when" since there's no ETA

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Tais993: > No, cause then I would have to add newlines and it shouldnt have newlines.

Not sure what I'd do then?

Zabuzard: My point is, I can write it like that:
```java
"""Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue, \
but also happen if the bot disconnected from the target guild and the \
user has disabled DMs or has been deleted."""
```
It uses the `\` syntax to get into the next line without actually having a newline in the text.

Some people might be familiar with this syntax (in particular console people), some not.

If you are okay with that, I will add it. I think its pretty much as-good/as-bad as using `+` 😆

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: yeah but now i can at least give it a name. thanks

Zabuzard: what does it mean when `getUser` on private channel is `null` in JDA 5? maybe i can implement a safe logic directly now already, which would work for both versions.

Tais993: It's nullable, the exact causation I do not remember, something with Discord changing iirc. 
A solution is to retrieve the user first, and use `User#openPrivateChannel`.
But that's an extra API call we might not even need, so whenever you add that is your decision.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Tais993: I know, I'm just not sure what I'd do.

I'm fine with either, so you can stick to this.

Zabuzard: I like it. I think ill try it out. Thanks.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: Okay, thanks for the info. I would be interested in the cause of it being nullable though. If you can open a private channel, the user to that channel should exist lol. Unless they change the meaning of private channel maybe.

Tais993: After doing some quick research, I can explain why, and this shouldn't matter so a comment isn't required.

To elaborate on this:

The MESSAGE_CREATE event gives information of the person that sent the message
The REACTION_ADD event **only** gives the ID of the person that adds the reaction

When you open a private channel, the user and such get retrieved, the info is there.


The issue starts if your bot becomes big, you'll start to use multiple shards. (shard = JDA instance, allows you to run tjhe bot on separate processes / machines)

**All** DM events get send to shard 0, if you open DM's in shard 1, shard 0 **doesn't** have any info about the DM's.
We then get the event, but the event only gives information of the person that triggers it. 

This means the person the DM's are from is unknown.



### Don't worry!
Because our bot is < 3000 (out of my head, could be wrong) guilds this doesn't matter, if our bot was using multiple shards this would be an issue.

So you can ignore this, luckily

Zabuzard: Makes sense. Thank you for investigating! :)

marko-radosavljevic (approved):
RemindRoutine is much more readable than the last time i glanced over it.  Routing logic is still somewhat convoluted tho, possibly extracting it to a new class is something that can be done, since it's not really routine's responsibility to figure out correct channels/users.

Will approve since it's not really worth 10h of combined development effort to go back and forth on naming and such, when this is not an integral feature to our core software. :relaxed:

marko-radosavljevic:
Oh yeah, how feasible would it be to add a few test cases for this?

Zabuzard:
> Oh yeah, how feasible would it be to add a few test cases for this?

Not for now. We need scheduler/clock mocking for this. Can be done and I also wanted to do this at a later point in time (other stuff like temp ban or mod audit log routine need this as well).

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
```diff
@@ -0,0 +1,152 @@
...
+ * Pending reminders are processed and send by {@link RemindRoutine}.
+ */
+public final class RemindCommand extends SlashCommandAdapter {
+    private static final String COMMAND_NAME = "remind";
+    private static final String WHEN_AMOUNT_OPTION = "when-amount";
+    private static final String WHEN_UNIT_OPTION = "when-unit";
+    private static final String CONTENT_OPTION = "content";
+
+    private static final int MIN_WHEN_AMOUNT = 1;
+    private static final int MAX_WHEN_AMOUNT = 1_000;
```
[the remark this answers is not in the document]
Zabuzard: I tried to name it according to what it logically represents in the code. Probably also had a tunnel vision with the split of `remind-in` and `remind-at` at which point I had multiple variables that required a consistent naming that clearly differentiates both from each other (the time period to remind "in"; and the time moment to remind "at").

But with the current code I totally agree. `TIME` is clearer and currently there is only one time (remind-in) anyways.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
```diff
@@ -0,0 +1,143 @@
...
+            long authorId, @NotNull TextChannel channel) {
+        return jda.retrieveUserById(authorId)
+            .onErrorMap(error -> null)
+            .map(author -> ReminderRoute.toPublic(channel, author));
+    }
+
+    private static @NotNull RestAction<ReminderRoute> createDmReminderRoute(@NotNull JDA jda,
+            long authorId) {
+        return jda.openPrivateChannelById(authorId)
+            .map(channel -> ReminderRoute.toPrivate(channel, channel.getUser()));
```
Tais993, line 93: Since we're just using PrivateChannel#getUser here, we can also just create a ReminderRouter#toPrivate(PrivateChannel) method? Allows us to use a method reference, and I love method references \:p

Tais993 (changes requested):
Last one I promise :p

application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
Zabuzard: Good spot

Zabuzard: Done. Please recheck, thanks.

Tais993 (approved):

marko-radosavljevic (approved):
I'm happy with the naming now. 
image

PR #384, opened 2022-02-16
Added /note command

### LINKED ISSUE #349: Create /note command
### Overview

The `/note` command would pretty much be a 1:1 copy of `/warn`, but without the "DM the user"-bit.

So UX-wise, we are talking about a command like `/note user:@John reason:Kinda toxic sometimes`.

Hence, it should also result in an entry in the DB and being shown in the `/audit` command.

Technically, this is a no-brainer. However, since its basically a 1:1 copy of `WarnCommand`, there is potential to do some refactoring and get rid of code duplication. Might be kinda tricky in practice though, maybe not worth it.

### Motivation

Sometimes mods handle a semi-critical situation informally in chat and dont want to issue a full warning that also notifies the user formally again via DM.

It is still useful to be able to just write down the incident to not forget about it in the future when judging about future situations - a note command could be useful for this.

Zabuzard:
### Overview

Implements and closes #349. This adds the `/note` command, which allows moderators **and Staff Assistants** to write a note about a given user, which is then accessible via `/audit`.

It is basically like `/warn`, but without noticing the target user.

note example

Here is an example of how the `/audit` looks like:

audit summary example
audit note example

### Details

There are the usual edge cases that are handled similar to other moderation commands:

* target user is too powerful for bot to interact with
* target user is too powerful for command user to interact with
* comamnd user is lacking the required role (soft moderation)

edge case example

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
@@ -1,0 +1,106 @@
+package org.togetherjava.tjbot.commands.moderation;
+
+import net.dv8tion.jda.api.entities.*;
+import net.dv8tion.jda.api.events.interaction.SlashCommandEvent;
+import net.dv8tion.jda.api.interactions.Interaction;
+import net.dv8tion.jda.api.interactions.commands.OptionMapping;
+import net.dv8tion.jda.api.interactions.commands.OptionType;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.SlashCommandVisibility;
+import org.togetherjava.tjbot.config.Config;
+
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+
+/**
+ * This command allows users to write notes about others. Notes are persisted and can be retrieved
+ * using {@link AuditCommand}, like other moderative actions.
+ * <p>
+ * The command fails if the user triggering it is lacking permissions to either write a note about
+ * other users or to write a note about the specific given user (for example a moderator attempting
+ * to write a note about an admin).
+ */
+public final class NoteCommand extends SlashCommandAdapter {
+    private static final Logger logger = LoggerFactory.getLogger(NoteCommand.class);
+    private static final String USER_OPTION = "user";
+    private static final String CONTENT_OPTION = "content";
+    private static final String ACTION_VERB = "write a note about";
+    private final ModerationActionsStore actionsStore;
+    private final Predicate<String> hasRequiredRole;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param actionsStore used to store actions issued by this command
+     * @param config the config to use for this
+     */
+    public NoteCommand(@NotNull ModerationActionsStore actionsStore, @NotNull Config config) {
+        super("note", "Writes a note about the given user", SlashCommandVisibility.GUILD);
+
+        getData()
+            .addOption(OptionType.USER, USER_OPTION, "The user who you want to write a note about",
+                    true)
+            .addOption(OptionType.STRING, CONTENT_OPTION,
+                    "The content of the note you want to write", true);
+
+        hasRequiredRole = Pattern.compile(config.getSoftModerationRolePattern()).asMatchPredicate();
+        this.actionsStore = Objects.requireNonNull(actionsStore);
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull SlashCommandEvent event) {
+        OptionMapping targetOption = event.getOption(USER_OPTION);
+        Member author = event.getMember();
+        Guild guild = event.getGuild();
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
```diff
@@ -0,0 +1,106 @@
...
+    private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
+            @Nullable Member target, CharSequence content, @NotNull Interaction event) {
+        if (target != null && !ModerationUtils.handleCanInteractWithTarget(ACTION_VERB, bot, author,
+                target, event)) {
+            return false;
+        }
+        if (!ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event)) {
+            return false;
+        }
+        return ModerationUtils.handleReason(content, event);
```
Tais993, line 81:
```suggestion
        if (target != null && !ModerationUtils.handleCanInteractWithTarget(ACTION_VERB, bot, author,
                target, event)) {
            return false;
        }
        
        if (!ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event)) {
            return false;
        }
        
        return ModerationUtils.handleReason(content, event);
```

Yes it's me again, like always with the styling :p

application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
```diff
@@ -0,0 +1,106 @@
...
+
+        actionsStore.addAction(guild.getIdLong(), author.getIdLong(), target.getIdLong(),
+                ModerationAction.NOTE, null, content);
+    }
+
+    private static void sendFeedback(@NotNull User target, @NotNull Member author,
+            @NotNull String noteContent, @NotNull Interaction event) {
+        MessageEmbed feedback = ModerationUtils.createActionResponse(author.getUser(),
+                ModerationAction.NOTE, target, null, noteContent);
+        event.replyEmbeds(feedback).queue();
```
Tais993, line 104:
```suggestion
        MessageEmbed feedback = ModerationUtils.createActionResponse(author.getUser(),
                ModerationAction.NOTE, target, null, noteContent);
                
        event.replyEmbeds(feedback).queue();
```

To elaborate on the logic.

```java
storeNote(target, author, content, guild);
sendFeedback(target, author, content, event);
```

Here both method calls are 1 line, and it's clear.
In this case, one method call takes 2 lines, and it's harder to read because of that.

So having that gap, is nicer for readability imo, it makes it clear where the separation is.

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
```diff
@@ -0,0 +1,106 @@
...
+        if (!handleChecks(guild.getSelfMember(), author, targetOption.getAsMember(), content,
+                event)) {
+            return;
+        }
+
+        sendNote(targetOption.getAsUser(), author, content, guild, event);
+    }
+
+    @SuppressWarnings("BooleanMethodNameMustStartWithQuestion")
+    private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
```
Heatmanofurioso, line 72: Instead of suppressing the warning here.
Why not rename the method to something like `doesHandleChecks`?

Naming on boolean methods does make it more easy to understand their purpose without documentation, or reading the signature

Heatmanofurioso (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/NoteCommand.java
Zabuzard: @Heatmanofurioso the problem is that the name is inaccurate then. The `boolean` return type indicates whether the user passed all the checks (or was it the opposite? 😬) and whether action has been taken (such as printing an error message).

So accurate would be sth like `doesPassChecksAndHandleIfNot` but thats kinda awkward.

That said, we kinda established this `handleXXX` thing as a (probably not-so-good) pattern already. `handle` means "check something and if it doesnt hold, also handle the situation already and send an error message to the user". All the moderation commands use this pattern everywhere already.

In hindsight it would probably have been better to use custom exceptions and move the "handle" part to top-level in the `onSlashCommand` with a try-catch. But it is what it is...

Heatmanofurioso: I agree with that reasoning. Thanks. Really does make sense for it to be a PR and update everything together, so things are consistent.
Resolving this conversation then

Heatmanofurioso (approved):

Tais993 (approved):

Zabuzard:
Merged since Heatman previously approved and nothing really changed since then.

PR #389, opened 2022-02-18
Removal of ALL subcommand on RoleSelectCommand

### LINKED ISSUE #381: Remove "/role-select all"
Recently, we merged #351 and it is pretty cool. However, it has two subcommands:

* `/role-select all` - users can pick any of the existing roles on the server
* `/role-select choose` - users can only pick roles that the mod pre-selected

I do not see a valid use case for `/role-select all`. Why would you want people to give the possibility to choose from ALL roles (including staff assistant, moderator, ... and other dangerous roles).

In fact, it only takes a careless mod to accidentally run `/role-select all`, hit enter and oops people could already give themselves `Moderator`, `Staff Assistant` and other "protected" roles (and you wouldnt even notice).

I propose to remove `/role-select all` and lift `/role-select choose` to just `/role-select`.

Tais993:
This removes the implementation of the ALL subcommand of the option, while keeping the CHOOSE subcommand 

Closes #381


marko-radosavljevic (approved):

marko-radosavljevic (commented):
Missing newline at the end of the file. (Also spotless complaining, probably about the same thing)

Zabuzard (dismissed):

marko-radosavljevic (approved):

Zabuzard (approved):

PR #394, opened 2022-02-21
Detecting and blocking scam

### LINKED ISSUE #390: Block scam links

There is too much scam links (nitro, steam, etc) posted on the discord, and it's taking a lot of moderation time, while annoying and harassing the whole server. Sadly, people still fall for these phishing attempts, so it's also a security issue we should address.

Block the scam links with our bot, and possibly temporarily remove offending members. (if false-positives are rare enough)

**Describe alternatives you've considered**

To waste time indefinitely until discord does something about it.

**Additional context**

Phishing link can look like this (list will be edited):

```
https://disceord.gift/EF2Zx27xxxx
```

On the same IP, these domains can be found:

- dlscors.gift
- dlscorb.gift
- disczrd.gift
- discorcl-nitro-gift.com
- discoerd.gift
- discerd.gift
- disceord.gift
- disccord.gift

How it looks in chat:
discord-nitro-scam-example

Text phrases usually found in link previews:

`You've been gifted a subscription!`
`Discord has gifted you Nitro for 1 month`

I can host a full website example, or just html page if needed.

Zabuzard:
### Overview

Implements and closes #390.

This adds:
* utility to detect scam messages `ScamDetector`
* a message receiver taking action against scam messages `ScamBlocker`
* a wrapper around a new database table `ScamHistoryStore`
* a cleanup routine `ScamHistoryPurgeRoutine` to purge the store

It automatically attempts to detect scam messages, such as nitro scam and then takes actions agains, ranging from just logging the issue to deleting the message and quarantining the user.

The system is highly configurable and supports several modes:
* `OFF`
* `ONLY_LOG`
* `APPROVE_FIRST`
* `AUTO_DELETE_BUT_APPROVE_QUARANTINE`
* `AUTO_DELETE_AND_QUARANTINE`

We will probably start with a more manual mode, such as `APPROVE_FIRST` and then increase it incrementally depending on how well it goes.

### `ScamBlocker`

The blocker is highly configurable and has different actions depending on its mode.

#### OFF

The blocker is deactivated and does not scan for scam at all.

#### ONLY_LOG

The blocker detects scam but only logs a `WARN` level message, no further action taken:

scam log

#### APPROVE_FIRST

Detected scam will be sent to moderators for review. Any action has to be approved explicitly first.

approve dialog

If lacking the soft moderation role:

lacking perms

#### AUTO_DELETE_BUT_APPROVE_QUARANTINE

Detected scam will automatically be deleted. A moderator will be informed for review. They can then decide whether the user should be put into quarantine.

dialog

#### AUTO_DELETE_AND_QUARANTINE

The blocker will automatically delete any detected scam and put the user into quarantine.

auto delete quarantine report

### `ScamDetector`

The scam detector analyzes strings for scam. Its heuristic is highly configurable.

In general, it searches the content for keywords, such as `nitro` and `@everyone`.

Additionally, it searches for an URL and analyzes the host of the URL. It supports a white- and blacklist for the URL hosts. Also, it checks whether the URL contains an infix that is similar to a suspicious keyword, such as `discord` (for example `www.foo_disc0rd_bar.com`).

Based on those, two rules that determine scam are defined:
1. `contains("nitro")` and `contains("@everyone")` and `hasUrl`
2. `contains("nitro")` and `hasSuspiciousUrl`

In both rules, the presence of `nitro` in the message is mandatory.

The following would be detected as a scam message by the first rule:
example scam first rule

And the following would be matched by the second rule:
example scam second rule

### `ScamHistoryStore` and Co.

The store is mainly a wrapper around the new database table:
```sql
CREATE TABLE scam_history
(
    id           INTEGER   NOT NULL PRIMARY KEY AUTOINCREMENT,
    sent_at      TIMESTAMP NOT NULL,
    guild_id     BIGINT    NOT NULL,
    channel_id   BIGINT    NOT NULL,
    message_id   BIGINT    NOT NULL,
    author_id    BIGINT    NOT NULL,
    content_hash TEXT      NOT NULL,
    is_deleted   BOOLEAN   NOT NULL
)
```

Entries are purged after **14 days** by `ScamHistoryPurgeRoutine`.

The main purpose of the store is to detect scam duplicates for a graceful handling of multi-spam. In practice, users dont just post a single scam message, but instead they spam multiple channels with the same message. However, we only want to report it once to the mods for decision-making and take action, such as quarantining, also only once. Therefore, we track each scam in the store and can easily find duplicate scam messages.

In detail, when a scam message is detected, we will check whether there are recent duplicates (**15 minutes**) and if so, we will just not take any further action. For modes that issue an immediate deletion, we will silently delete the message though. For other modes, the action will be taken on all scam duplicates after clicking the button on the scam report.

### Config

The config has changed. See `config.json.template`. Here is a good default configuration:
```json
"scamBlocker": {
     "mode": "AUTO_DELETE_BUT_APPROVE_QUARANTINE",
     "reportChannelPattern": "commands",
     "hostWhitelist": ["discord.com", "discord.gg", "discord.media", "discordapp.com", "discordapp.net", "discordstatus.com"],
     "hostBlacklist": ["bit.ly"],
     "suspiciousHostKeywords": ["discord", "nitro", "premium"],
     "isHostSimilarToKeywordDistanceThreshold": 2
 }
```

### Checklist

General:
* [x] Scam detection
* [x] Config
* [x] Add quarantine role (like mute) (#398)
* [x] Persist quarantine role (after rejoin, like mute) (#398)
* [x] Unit tests for the `ScamDetector`

Modes:
* [x] OFF mode
* [x] ONLY_LOG mode
* [x] APPROVE_FIRST mode
* [x] AUTO_DELETE_BUT_APPROVE_QUARANTINE mode
* [x] AUTO_DELETE_AND_QUARANTINE mode

Features:
* [x] Button action routing for non-slashcommands (component IDs)
* [x] Detect and gracefully handle multi-messages from the same user
* [x] Approval buttons only clickable by soft-mod role
* [x] Use SHA hash instead of message content in the database table


Zabuzard:
Waiting for #398

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,364 @@
...
+        };
+
+        guild.retrieveMemberById(args.authorId)
+            .queue(onRetrieveAuthorSuccess, onRetrieveAuthorFailure);
+    }
+
+
+    private record ComponentIdArguments(@NotNull ScamBlockerConfig.Mode mode, long guildId,
+            long channelId, long messageId, long authorId, String contentHash) {
+        static ComponentIdArguments fromList(@NotNull List<String> args) {
```
Tais993, line 348:
```suggestion
            long channelId, long messageId, long authorId, String contentHash) {
            
        static ComponentIdArguments fromList(@NotNull List<String> args) {
```
I thought this was the constructor, got really confused lol

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamDetector.java
```diff
@@ -0,0 +1,105 @@
...
+            return true;
+        }
+        return results.containsNitroKeyword && results.hasSuspiciousUrl;
+    }
+
+    private void analyzeToken(@NotNull String token, @NotNull AnalyseResults results) {
+        if ("@everyone".equalsIgnoreCase(token)) {
+            results.pingsEveryone = true;
+        }
+        if ("nitro".equalsIgnoreCase(token)) {
```
Tais993, line 36:
```suggestion
        }
        
        if ("nitro".equalsIgnoreCase(token)) {
```
Make sure that the if statements are vaccinated, preferably even with a booster.
Otherwise keep the distance

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamDetector.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private void analyzeUrl(@NotNull String url, @NotNull AnalyseResults results) {
+        String host;
+        try {
+            host = URI.create(url).getHost();
+        } catch (IllegalArgumentException e) {
+            // Invalid urls are not scam
+            return;
+        }
+        if (host == null) {
```
Tais993, line 53:
```suggestion
        }
        
        if (host == null) {
```
Same for try-catch and if statements.
Especially since the try-catch is more vulnerable to the Omicron variant

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -55,7 +55,8 @@
     private static final ScheduledExecutorService ROUTINE_SERVICE =
             Executors.newScheduledThreadPool(5);
     private final Config config;
-    private final Map<String, SlashCommand> nameToSlashCommands;
+    private final Map<String, UserInteractor> nameToInteractor;
```
Tais993, line 58: Seems really vulnerable, if we add context-commands with the same name as a slash-command this won't work

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,364 @@
...
+                        """
+                    .formatted(guild.getName());
+
+        jda.openPrivateChannelById(userId)
+            .flatMap(channel -> channel.sendMessage(dmMessage))
+            .queue();
+    }
+
+    private Optional<TextChannel> getReportChannel(@NotNull Guild guild) {
+        // Check cache first, then get full list
```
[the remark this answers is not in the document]
Zabuzard: pretty sure i copied that from somewhere. so that other location is still faulty then as well.

that said, afaik retrieve checks the cache beforehand first anyways, no? iirc i read that in the doc of other commands, such as retrieveUser.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: each slash command is also a userinteractor at the same time. so `nameToSlashCommands` is a subset of `nameToInteractor`. At first I hoped I can get away with just one of the maps, but then I needed access to slash commands specifically.

I can probably still get around it by doing `(SlashCommand) nameToInteractor.get(name)` and that way only have one map.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: It does, but the point I'm making is that it's getTextChannels not retrieveTextChannels.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,364 @@
...
+        return guild.getTextChannelCache()
+            .stream()
+            .filter(isReportChannel)
+            .findAny()
+            .or(() -> guild.getTextChannels().stream().filter(isReportChannel).findAny());
+    }
+
+    private List<ActionRow> createConfirmDialog(@NotNull GuildMessageReceivedEvent event) {
```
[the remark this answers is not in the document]
Tais993: Use addActionRow instead of addActionRows

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Tais993: Issue is that there will be multiple commands with the same name in the future

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: > It does, but the point I'm making is that it's getTextChannels not retrieveTextChannels.

I understand. My point is, I can just do `retrieveFoo()` and only that, since it will internally first check the cache anyways, no?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: @Tais993 That doesnt exist for the situation im in here.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: Why that? That shouldnt exist. That would mess with the system and as a solution we would need to add prefixes like "slashcommand;foo;..." to differentiate it from "whatevercommand;foo;...". But why do we need to support that?

Tais993: So in case we want to have 2 commands with the same name, you want us to just name them differently? Why shouldn't it exist? It makes sense for the user's UX, 2 commands with the same behaviour, with the same name.
The system isn't made for it indeed, I have to create an issue for.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: Still don't think a method should return a List, when it only has 1 element.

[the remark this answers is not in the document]
Tais993: Regardless, there's always a non-collection overload.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: Correc

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: You mean like being able to trigger, for example `/ban` also via user context menu? I see, makes sense. Yes, the system is not prepared for that yet. It is not very difficult though, prefixes will do the trick.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: Checked the code and we both overlooked something in this discussion. The code is correct and cant be improved imo:

screen

Basically, we dont have the channels ID or name, we have a predicate matching the name. So we need to check the list of all channels against the predicate for any-match. but `getTextChannels()` (no cache) will always create a full new list, it says itself that one should first check the cache:

screen2

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: Checked the code. The reason why its a collection is bc the result is handled around a few times before it is actually used.

Methods like `takeActionAutoDeleteButApproveQuarantine` call `createConfirmDialog` and pass the list to `reportScamMessage`, which then does `new MessageBuilder()....setActionRows(actionRows)...`. But the method can also be called with an empty list if no dialog is desired.

I made the List a Collection now though.

The only other solution I see is to change it to varargs mixed with arrays. Imo thats uglier, in particular since I need an empty array for no-dialog then. Not so nice to create.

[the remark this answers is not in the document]
Tais993: I still want createConfirmDialog to return an ActionRow, the methods that call it can make it a list.
createConfirmDialog doesn't need a List, it's useless

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: I didn't overlook anything, you are the one overlooking, still after me explaining to you a few times.

`getTextChannels = new List(getTextChannelCache())` (not exactly, but 90% the same)

getTextChannels **doesn't retrieve**, it would be called *retrieveTextChannels* then.

use getTextChannelCache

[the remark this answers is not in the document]
Zabuzard: I see, thats fair. Now it clicked.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: But what if a confirm dialog might consist of multiple rows? Kinda an implementation detail, isnt it? Why should the method enforce that its only a single row.

I dont have a strong preference here, I can make it a single `@Nullable ActionRow` as well. But then I probably have to do an if-else to either call the `addActionRow` or not, depending on whether its null (unless JDA handles null gracefully here, have to check the code).

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: And I can't find were you copied it from, don't see any class call getTextChannels even. So maybe misremembering or copied it from an outdated PR? Not sure

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Tais993: > But what if a confirm dialog might consist of multiple rows? Kinda an implementation detail, isnt it? Why should the method enforce that its only a single row.

If that happens in the future, I guess we have to change our code. 
Since the dialog is only 1 row, it's easy for the calling methods to add their own rows too, otherwise they need to do weird checks, or move rows. So regarding future compatibility, I'd argue that 1 actionrow is better.

Now the dialog just creates the row, and that is all that it needs to do.

>  I can make it a single @Nullable ActionRow as wel

Why would it be nullable randomly? It's never null? 
And if it's null, JDA would throw too if you give it a Collection with a null entry.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: Cant find it either anymore. Anyways, fixed.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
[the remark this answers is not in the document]
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -205,7 +206,7 @@ private void quarantineAuthor(@NotNull Guild guild, @NotNull Member author,
     }
 
     private void reportScamMessage(@NotNull GuildMessageReceivedEvent event,
-            @NotNull String reportTitle, @NotNull Collection<? extends ActionRow> actionRows) {
+            @NotNull String reportTitle, @Nullable ActionRow confirmDialog) {
```
Tais993, line 209: We misunderstood each-other, this should be a collection.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -249,19 +251,18 @@ If you think this was a mistake (for example, your account was hacked, but you g
             .queue();
     }
 
-    private Optional<TextChannel> getReportChannel(@NotNull Guild guild) {
+    private @NotNull Optional<TextChannel> getReportChannel(@NotNull Guild guild) {
         return guild.getTextChannelCache().stream().filter(isReportChannel).findAny();
     }
 
-    private Collection<ActionRow> createConfirmDialog(@NotNull GuildMessageReceivedEvent event) {
+    private @NotNull ActionRow createConfirmDialog(@NotNull GuildMessageReceivedEvent event) {
```
Tais993, line 258: This should return an ActionRow

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,363 @@
...
+
+    private void takeActionLogOnly(@NotNull GuildMessageReceivedEvent event) {
+        addScamToHistory(event);
+        logScamMessage(event);
+    }
+
+    private void takeActionApproveFirst(@NotNull GuildMessageReceivedEvent event) {
+        addScamToHistory(event);
+        logScamMessage(event);
+        reportScamMessage(event, "Is this scam?", createConfirmDialog(event));
```
Tais993, line 156:
```suggestion
        reportScamMessage(event, "Is this scam?", List.of(createConfirmDialog(event)));
```
And the calling methods create a List of it.

Zabuzard: That defeats the purpose of the helper method. Its supposed to return exactly whats needed for the reportScamMessage call, to simplify the call-site. So I would rather rename it to `createActionRowsForConfirmDialog` to prevent what you are suggesting.

Zabuzard: 🤷

Zabuzard: That said, why should the `reportScamMessage` take a list? It could as well just either take a single confirm dialog or nothing. It is domain-specific, doesnt need to be abstract.

Tais993: Don't like that either since it creates an action row, not action rows.
But w/e, you can undo the changes if you want.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,363 @@
...
+                    return;
+                }
+
+                if (errorResponseException.getErrorResponse() == ErrorResponse.UNKNOWN_MEMBER) {
+                    logger.debug(
+                            "Attempted to handle scam, but user '{}' is not a member of the guild anymore.",
+                            args.authorId);
+                }
+            }
+        };
```
Tais993, line 336: You can use JDA's ErrorHandler instead
```suggestion
        Consumer<Throwable> onRetrieveAuthorFailure = new ErrorHandler()
             .handle(ErrorResponse.UNKNOWN_USER, 
                 (unknownUserResponse) -> logger.debug("Attempted to handle scam, but user '{}' does not exist anymore.",
                            args.authorId)
             .handle(ErrorResponse.UNKNOWN_MEMBER, 
                 (unknownMemberResponse) ->  logger.debug(
                            "Attempted to handle scam, but user '{}' is not a member of the guild anymore.",
                            args.authorId);
```

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -128,18 +130,24 @@ public BotCore(@NotNull JDA jda, @NotNull Database database, @NotNull Config con
...
+            .map(SlashCommand.class::cast)
+            .toList();
     }
 
     @Override
     public @NotNull Optional<SlashCommand> getSlashCommand(@NotNull String name) {
-        return Optional.ofNullable(nameToSlashCommands.get(name));
+        return Optional.ofNullable(nameToInteractor.get(name))
+            .filter(SlashCommand.class::isInstance)
+            .map(SlashCommand.class::cast);
```
Tais993, line 150: I honestly want a second map for SlashCommands, this is really under performant.

Tais993 (changes requested):
Some classes don't have documentation, can you take a look at that?
Not saying everything needs documentation, more of a question to you to take a look at that so classes that do need it receive it.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: This is new to me, cool.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: Why? Whats the issue that you see? I dont like maintaining a map and a second map that is just a subview of it.

Zabuzard:
> Some classes don't have documentation, can you take a look at that?

@Tais993 Do you remember which? I thought I added it everywhere.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Tais993: The first map you loop through all items, filter and map.
The second, you still filter and map.

I personally find that a waste of performance, there's nothing to "maintain"
The map gets created on start-up, after that it doesn't anymore.

But now I just see less performance, and more code, and in return?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamDetector.java
```diff
@@ -0,0 +1,107 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.commands.utils.StringDistances;
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.config.ScamBlockerConfig;
+
+import java.net.URI;
+import java.util.regex.Pattern;
+
+public class ScamDetector {
```
Tais993, line 11: doc

Tais993 (changes requested):
I see, it's only 1 class.

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Zabuzard: The first can be solved by creating a restricted view of the set (that just blends out the stuff with the wrong type), if you really care that much. The map is small though and the method is only used once by reload - so its not really a big deal anyways.

The second one is not an iteration at all, the filter and map call are on Optional, so the second one is perfectly fine.

application/src/main/java/org/togetherjava/tjbot/commands/UserInteractor.java
```diff
@@ -0,0 +1,84 @@
...
+ * {@link SlashCommand}.
+ * <p>
+ * An interactor must have a unique name and can react to button clicks and selection menu actions.
+ */
+public interface UserInteractor extends Feature {
+
+    /**
+     * Gets the name of the interactor.
+     * <p>
+     * Requirements for this are documented in {@link CommandData#CommandData(String, String)}.
```
Tais993, line 23: Looks like broken copy paste

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,356 @@
...
+        MessageEmbed embed =
+                new EmbedBuilder().setDescription(event.getMessage().getContentStripped())
+                    .setTitle(reportTitle)
+                    .setAuthor(author.getAsTag(), null, author.getAvatarUrl())
+                    .setTimestamp(event.getMessage().getTimeCreated())
+                    .setColor(AMBIENT_COLOR)
+                    .setFooter(author.getId())
+                    .build();
+        Message message =
+                new MessageBuilder().setEmbeds(embed).setActionRows(confirmDialog).build();
```
Tais993, line 229: Don't really see why you would want this?
Using the return value of sendMessage would be nicer I'd say?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    private @NotNull ActionRow createConfirmDialog(@NotNull GuildMessageReceivedEvent event) {
+        ComponentIdArguments args = new ComponentIdArguments(mode, event.getGuild().getIdLong(),
+                event.getChannel().getIdLong(), event.getMessageIdLong(),
+                event.getAuthor().getIdLong(),
+                ScamHistoryStore.hashMessageContent(event.getMessage()));
+
+        return ActionRow.of(Button.of(ButtonStyle.SUCCESS, generateComponentId(args), "Yes"),
+                Button.of(ButtonStyle.DANGER, generateComponentId(args), "No"));
```
Tais993, line 265:
```suggestion
        return ActionRow.of(Button.success(generateComponentId(args), "Yes"),
                Button.danger(generateComponentId(args), "No"));
```

Tais993 (changes requested):
That's why I review multiple times :p

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: cool

Tais993: You & your personality yes

application/src/main/java/org/togetherjava/tjbot/commands/UserInteractor.java
Zabuzard: which part is broken? looks good to me, no? I guess ur referring to the link? Well, unfortunately we dont have another place where we documented the name requirements :D

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: i dont follow, what do u mean by that?

application/src/main/java/org/togetherjava/tjbot/commands/UserInteractor.java
Tais993: UserInteractor doesn't sound like something which is Discord command specific
Those requirements only exist for commands that get sent to Discord (slash-commands, context-commands etc)

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
```diff
@@ -0,0 +1,356 @@
...
+                    .setTitle(reportTitle)
+                    .setAuthor(author.getAsTag(), null, author.getAvatarUrl())
+                    .setTimestamp(event.getMessage().getTimeCreated())
+                    .setColor(AMBIENT_COLOR)
+                    .setFooter(author.getId())
+                    .build();
+        Message message =
+                new MessageBuilder().setEmbeds(embed).setActionRows(confirmDialog).build();
+
+        reportChannel.orElseThrow().sendMessage(message).queue();
```
Tais993, line 231:
```suggestion
        reportChannel.orElseThrow().sendMessage(message)
            .setEmbeds(embed)
            .setActionRows(confirmDialog)
            .queue();
```

This @Zabuzard

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Tais993: See below

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: this is actually all over in the project. nice

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: its not possible. we have a special situation here.

i want to send a message with no message content, but emebeds and action rows.

problem is, `sendMessage("")` and `sendMessage(null)` both dont work. and `sendEmbeds(...)` doesnt take action rows (does it?)

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: fixed

application/src/main/java/org/togetherjava/tjbot/commands/UserInteractor.java
Zabuzard: yeah. but i think its better if we have the same requirements on all of them. after all, slash commands implement UserInteractor and use the same getName method.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Tais993: `sendMessageEmbeds(embed)`

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Tais993: Keep forgetting it myself too sssh

application/src/main/java/org/togetherjava/tjbot/commands/moderation/scam/ScamBlocker.java
Zabuzard: fair 😆

Zabuzard: Actually. There is an issue. This `setActionRows` only allows a non-nullable action row (or non-empty action row lists).

The other one allows a nullable row (or empty lists).

So if we still want to stick with that design, I have to create an if and break the fluent-chain:
```java
var foo = reportChannel.orElseThrow().sendMessageEmbeds(embed);
if (confirmDialog != null) {
  foo = foo.setActionRows(confirmDialog);
}
foo.queue();
```
at which point I prefer the previous solution.

Tais993: In that case nvm since that's way too hard to reaad.

Tais993 (approved):

PR #395, opened 2022-02-22
Skipping audit logs against bot message deletions

### LINKED ISSUE #393: Dont audit-log message deletions against bots
### Overview

Our `mod_audit_log` channel often contains message deletions against our bot, such as:

example

There is nothing wrong with this per-se, but it generates unecessary noise and the point of the channel is to be noise-free.

So I am suggesting that the audit-logger routine should simply skip message deletions against bots. (there is always discords own audit log if a true full history of all actions has to be checked).

Other actions against bots should be kept, such as bans.

Zabuzard:
### Overview

Implements and closes #393.

This adds a little check that skips writing audit log messages for message deletions against bots. Cause they generate noise that doesnt need to be reviewed and we want the channel to be noise-free. I.e. logs like this are gone now:

message deletion against bot entry

### Details

The code changes are not trivial but quite straightforward. Unfortunately, the setup didnt allow a simple `if-else`, cause at the point where we know the target user, we are already in an async chain of `RestAction`.

I had to bubble up the target-user information one level from the more generic `handleAction` to the place where we have the context of `MESSAGE_DELETION`, i.e. `handleMessageDeleteEntry`. Therefore, I decided to create a little record that basically holds the `User target` and the `MessageEmbed` that `handleAction` used to create. I was able to enhance this further by letting the record actually generate the embed, moving the code from one place into the other.

Now that the embed and target is known at `handleMessageDeleteEntry`, we can check for `target.isBot()` and abort the computation. Unfortunately, I was not able to find any way to gracefully stop/cancel/fail a `RestAction` other than actually throwing (there is `addCheck`, but it has no hands on the value from the `RestAction`, so it cant be used for this) (@Tais993 any better ideas?).

I found a exception in the JDK that seems fitting for this purpose and reused it, `CancellationException`.

### Checklist

* [x] Implement
* [x] Absorb this exception, it shouldnt lead to logs and noise

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/routines/ModAuditLogRoutine.java
@@ -25,9 +25,8 @@ import java.awt.*;
 import java.time.*;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAccessor;
-import java.util.*;
 import java.util.List;
-import java.util.concurrent.CancellationException;
+import java.util.*;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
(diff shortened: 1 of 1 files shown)


Tais993:
Nope, you have the correct solution.
Another solution is to use CompletaebleFuture's, but I'm not a fan of that since they'll over-complicate the code a lot (imho)

A second note, you're throwing, this means that exception gets throw and gets logged.
We could set a default failure consumer(RestAction#setDefaultConsumer), which logs CancellationExceptions at warning instead of error.

Zabuzard:
> Nope, you have the correct solution. Another solution is to use CompletaebleFuture's, but I'm not a fan of that since they'll over-complicate the code a lot (imho)
> 
> A second note, you're throwing, this means that exception gets throw and gets logged. We could set a default failure consumer(RestAction#setDefaultConsumer), which logs CancellationExceptions at warning instead of error.

Good point. Ill have a look at that. Its unfortunately quite tricky to test this due to the audit log rate limitting. Maybe I can mock it quickly for testing it out.

Tais993 (approved):

Heatmanofurioso (approved):

Issue #386, opened 2022-02-17
/tex should support inline TeX


Zabuzard:
### Overview

The `/tex` command should support inline TeX, for example using the idiomatic $-notation:
```java
/tex latex: Hello, look at this $\frac{x}{2}$, nice stuff!
```
Right now, the command interprets the whole content as TeX and renders it accordingly:
![example](https://i.imgur.com/TMgYhPl.png)

It should instead be rendered as:
![desired](https://i.imgur.com/F07NARI.png)

### Implementation hint

This could be solved by identifying the `$...$` areas. If not present, simply use the existing flow. But if present, wrap the part to the left and right of it inside `\text{...}` blocks.

So `A$B$C` would be turned into `\text{A}B\text{C}`.

Be careful though, that there could be multiple `$`-sections.
* For example `See $\frac{x}{2}$ and also $\frac{y}{2}$`.
* Also the edge cases that it either starts with a section `$\frac{x}{2}$ nice`
* or ends with it `See $\frac{x}{2}$`.

The previous example `Hello, look at this $\frac{x}{2}$, nice stuff!` would then be turned into
```tex
\text{Hello, look at this }\frac{x}{2}\text{, nice stuff!}
```
and send like that to the TeX-engine.

IslamSakrak:
Can I be assigned to this I want to give it a try?

IslamSakrak:
Could you elaborate a bit on this?

`/tex latex: Hello, look at this $\frac{x}{2}$, nice stuff!`

the second image it looks like just that the /`tex `contains `$` it makes the whole /`tex `a regular text \`text{}`

While in the example 

`So A$B$C would be turned into \text{A}B\text{C}.`

it looks like it makes just specific parts into a text. 

So could you just clear it up for me please?

Zabuzard:
Taking over

PR #397, opened 2022-02-23
Adding Tex Inline support for the issue #386

IslamSakrak:
How it looks like 
image

The inputs
`A$B$C`
`$Hello$, look at this $\frac{x}{2}$, nice stuff!`
The Outputs
`\text{A}B\text{C}`
`Hello\text{, look at this }\frac{x}{2}\text{, nice stuff!}`

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
@@ -6,6 +6,7 @@ import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.components.Button;
 import net.dv8tion.jda.api.interactions.components.ButtonStyle;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.scilab.forge.jlatexmath.ParseException;
 import org.scilab.forge.jlatexmath.TeXConstants;
 import org.scilab.forge.jlatexmath.TeXFormula;
@@ -21,6 +22,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.List;
 import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Implementation of a tex command which takes a string and renders an image corresponding to the
@@ -34,6 +37,11 @@ import java.util.Objects;
 public class TeXCommand extends SlashCommandAdapter {
 
     private static final String LATEX_OPTION = "latex";
+    // Matches regions between two dollars, like '$foo$'.
+    private static final String MATH_REGION = "(\\$[^$]+\\$)";
+    private static final String TEXT_REGION = "([^$]+)";
+    private static final Pattern INLINE_LATEX_REPLACEMENT =
+            Pattern.compile(MATH_REGION + "|" + TEXT_REGION);
     private static final String RENDERING_ERROR = "There was an error generating the image";
     private static final float DEFAULT_IMAGE_SIZE = 40F;
     private static final Color BACKGROUND_COLOR = Color.decode("#36393F");
@@ -44,8 +52,7 @@ public class TeXCommand extends SlashCommandAdapter {
      * Creates a new Instance.
      */
     public TeXCommand() {
-        super("tex",
-                "This command accepts a latex expression and generates an image corresponding to it.",
+        super("tex", "Renders LaTeX, also supports inline $-regions like 'see this $frac[x}{2}$'.",
                 SlashCommandVisibility.GUILD);
         getData().addOption(OptionType.STRING, LATEX_OPTION,
                 "The latex which is rendered as an image", true);
@@ -57,6 +64,9 @@ public class TeXCommand extends SlashCommandAdapter {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            if (latex.contains("$")) {
+                latex = convertInlineLatexToFull(latex);
+            }
             formula = new TeXFormula(latex);
         } catch (ParseException e) {
             event.reply("That is an invalid latex: " + e.getMessage()).setEphemeral(true).queue();
@@ -90,6 +100,35 @@ public class TeXCommand extends SlashCommandAdapter {
             .editOriginal(renderedTextImageStream.toByteArray(), "tex.png")
             .setActionRow(Button.of(ButtonStyle.DANGER, generateComponentId(userID), "Delete"))
             .queue();
+    }
+
+    /**
+     *
+     * Converts inline latex like: {@code hello $\frac{x}{2}$ world} to full latex
+     * {@code \text{hello}\frac{x}{2}\text{ world}}.
+     *
+     */
+    @NotNull
+    private String convertInlineLatexToFull(@NotNull String latex) {
+        if (isInvalidInlineFormat(latex)) {
+            throw new ParseException("One dollar sign is invalid, wrap whole expression with it. ");
+        }
+        Matcher matcher = INLINE_LATEX_REPLACEMENT.matcher(latex);
+        StringBuilder sb = new StringBuilder();
+        while (matcher.find()) {
+            boolean isInsideMathRegion = matcher.group(1) != null;
+            if (isInsideMathRegion) {
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+
+        return sb.toString();
+    }
+
+    private boolean isInvalidInlineFormat(String latex) {
+        return latex.chars().filter(ch -> ch == '$').count() % 2 == 1;
     }
 
     @Override
--- a/application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
@@ -6,7 +6,6 @@ import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.components.Button;
 import net.dv8tion.jda.api.interactions.components.ButtonStyle;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 import org.scilab.forge.jlatexmath.ParseException;
 import org.scilab.forge.jlatexmath.TeXConstants;
 import org.scilab.forge.jlatexmath.TeXFormula;
@@ -63,6 +62,7 @@ public class TeXCommand extends SlashCommandAdapter {
         String latex = Objects.requireNonNull(event.getOption(LATEX_OPTION)).getAsString();
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
+
         try {
             if (latex.contains("$")) {
                 latex = convertInlineLatexToFull(latex);
@@ -72,49 +72,76 @@ public class TeXCommand extends SlashCommandAdapter {
             event.reply("That is an invalid latex: " + e.getMessage()).setEphemeral(true).queue();
             return;
         }
+
         event.deferReply().queue();
-        Image image = formula.createBufferedImage(TeXConstants.STYLE_DISPLAY, DEFAULT_IMAGE_SIZE,
-                FOREGROUND_COLOR, BACKGROUND_COLOR);
-        if (image.getWidth(null) == -1 || image.getHeight(null) == -1) {
-            event.getHook().setEphemeral(true).editOriginal(RENDERING_ERROR).queue();
-            logger.warn(
-                    "Unable to render latex, image does not have an accessible width or height. Formula was {}",
-                    latex);
-            return;
-        }
-        BufferedImage renderedTextImage = new BufferedImage(image.getWidth(null),
-                image.getHeight(null), BufferedImage.TYPE_4BYTE_ABGR);
-        renderedTextImage.getGraphics().drawImage(image, 0, 0, null);
-        ByteArrayOutputStream renderedTextImageStream = new ByteArrayOutputStream();
 
         try {
-            ImageIO.write(renderedTextImage, "png", renderedTextImageStream);
+            Image image = renderImage(formula);
+            sendImage(event, userID, image);
         } catch (IOException e) {
-            event.getHook().setEphemeral(true).editOriginal(RENDERING_ERROR).queue();
+            event.getHook().editOriginal(RENDERING_ERROR).queue();
             logger.warn(
                     "Unable to render latex, could not convert the image into an attachable form. Formula was {}",
                     latex, e);
-            return;
+
+        } catch (IllegalStateException e) {
+            event.getHook().editOriginal(RENDERING_ERROR).queue();
+
+            logger.warn(
+                    "Unable to render latex, image does not have an accessible width or height. Formula was {}",
+                    latex, e);
         }
+    }
+
+    private @NotNull Image renderImage(@NotNull TeXFormula formula) {
+        Image image = formula.createBufferedImage(TeXConstants.STYLE_DISPLAY, DEFAULT_IMAGE_SIZE,
+                FOREGROUND_COLOR, BACKGROUND_COLOR);
+
+        if (image.getWidth(null) == -1 || image.getHeight(null) == -1) {
+            throw new IllegalStateException("Image has no height or width");
+        }
+        return image;
+    }
+
+    private void sendImage(@NotNull SlashCommandEvent event, @NotNull String userID,
+            @NotNull Image image) throws IOException {
+
+        ByteArrayOutputStream renderedTextImageStream = getRenderedTextImageStream(image);
         event.getHook()
             .editOriginal(renderedTextImageStream.toByteArray(), "tex.png")
             .setActionRow(Button.of(ButtonStyle.DANGER, generateComponentId(userID), "Delete"))
             .queue();
     }
 
+    @NotNull
+    private ByteArrayOutputStream getRenderedTextImageStream(@NotNull Image image)
+            throws IOException {
+
+        BufferedImage renderedTextImage = new BufferedImage(image.getWidth(null),
+                image.getHeight(null), BufferedImage.TYPE_4BYTE_ABGR);
+
+        renderedTextImage.getGraphics().drawImage(image, 0, 0, null);
+        ByteArrayOutputStream renderedTextImageStream = new ByteArrayOutputStream();
+
+        ImageIO.write(renderedTextImage, "png", renderedTextImageStream);
+
+        return renderedTextImageStream;
+    }
+
     /**
-     *
      * Converts inline latex like: {@code hello $\frac{x}{2}$ world} to full latex
      * {@code \text{hello}\frac{x}{2}\text{ world}}.
-     *
      */
     @NotNull
     private String convertInlineLatexToFull(@NotNull String latex) {
         if (isInvalidInlineFormat(latex)) {
-            throw new ParseException("One dollar sign is invalid, wrap whole expression with it. ");
+            throw new ParseException(
+                    "The amount of $-symbols must be divisible by two. Did you forget to close an expression? ");
         }
+
         Matcher matcher = INLINE_LATEX_REPLACEMENT.matcher(latex);
         StringBuilder sb = new StringBuilder();
+
         while (matcher.find()) {
             boolean isInsideMathRegion = matcher.group(1) != null;
             if (isInsideMathRegion) {


application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +59,19 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            String regex = "(\\$\\S+\\$)|([a-zA-Z!., ]+)";
+            Pattern pattern = Pattern.compile(regex);
```
Zabuzard, line 63: Right now you constantly recompile this pattern. Instead, put it into a field (constant):
```java
private static final Pattern GIVE_ME_A_NICE_NAME = Pattern.compile("...");
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +59,19 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+            Matcher matcher = pattern.matcher(latex);
+            StringBuilder sb = new StringBuilder();
+            while (matcher.find()) {
+                if (matcher.group(1) != null) {
+                    sb.append(matcher.group(1).replaceAll("\\$", ""));
+                } else {
+                    sb.append("\\text{").append(matcher.group(2)).append("}");
+                }
+            }
+            latex = sb.toString();
```
Zabuzard, line 73: This should be moved into a helper method with a reasonable name.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +59,19 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            String regex = "(\\$\\S+\\$)|([a-zA-Z!., ]+)";
+            Pattern pattern = Pattern.compile(regex);
+            Matcher matcher = pattern.matcher(latex);
+            StringBuilder sb = new StringBuilder();
+            while (matcher.find()) {
+                if (matcher.group(1) != null) {
+                    sb.append(matcher.group(1).replaceAll("\\$", ""));
```
Zabuzard, line 68: Use `.replace("$", "")`, not `replaceAll` (which is for regex, that u dont need here)

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +59,19 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            String regex = "(\\$\\S+\\$)|([a-zA-Z!., ]+)";
+            Pattern pattern = Pattern.compile(regex);
+            Matcher matcher = pattern.matcher(latex);
+            StringBuilder sb = new StringBuilder();
+            while (matcher.find()) {
+                if (matcher.group(1) != null) {
```
Zabuzard, line 67: The logical affect that this condition has is unclear. Introduce a boolean with a nice name instead.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +59,19 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            String regex = "(\\$\\S+\\$)|([a-zA-Z!., ]+)";
```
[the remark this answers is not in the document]
IslamSakrak: It is linked with your comment https://github.com/Together-Java/TJ-Bot/pull/397#discussion_r812719622

There are 2 brackets which from my understanding represent 2 groups, one group matches `$...$`, which his the first bracket, the | represents or, and the second bracket ([...]+) represents the second match group. `matcher.group(2)`. 

So all together if the first match is not null, it means it has found a word with a dollar sign, then it does replace it with an empty space, while in the second match it appends `\text{ + [the words that match] + }`. 

This is all it if I am not mistaken, if I am someone please let me know.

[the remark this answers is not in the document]
Zabuzard: So first group is "i found something inside dollars" and second group is "i found something that is not a dollar"? then the second group should also be written like that `([^\\$]+)`. or similar

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +97,25 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+        latex = sb.toString();
+        return latex;
+    }
+
+    private boolean hasMatchedDollarSign(Matcher matcher) {
```
Tais993, line 115:
```suggestion
    private boolean hasMatchedDollarSign(@NotNull Matcher matcher) {
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +97,25 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
             .queue();
     }
 
+    @NotNull
+    private String getLatexToInline(String latex) {
```
Tais993, line 101: `@NotNull` or `@Nullable`

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +97,25 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+        StringBuilder sb = new StringBuilder();
+        while (matcher.find()) {
+            if (hasMatchedDollarSign(matcher)) {
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+        latex = sb.toString();
+        return latex;
```
Tais993, line 112:
```suggestion
        return sb.toString();;
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -36,6 +36,7 @@
 public class TeXCommand extends SlashCommandAdapter {
 
     private static final String LATEX_OPTION = "latex";
+    private static final Pattern LATEX_PATTERN = Pattern.compile("(\\$\\S+\\$)|([a-zA-Z!., ]+)");
```
Zabuzard, line 39: `LATEX_PATTERN` is not really a good name. It should describe more what it is actually about. It is your inline latex replacement pattern.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -107,6 +97,25 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
             .queue();
     }
 
+    @NotNull
+    private String getLatexToInline(String latex) {
```
Zabuzard, line 101: I think that name could need some improval. More like `convert` instead of `get`. Also, its the opposite direction. inline-latex gets translated into full-latex.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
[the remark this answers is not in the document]
Zabuzard: Its nice that you explained it to me now, but the explanation should be obvious from the code. So that people who read the code, also understand it.

So either do some renaming, split things apart and give them variable names, ... or as last resort use comments.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -107,6 +97,25 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+        latex = sb.toString();
+        return latex;
+    }
+
+    private boolean hasMatchedDollarSign(Matcher matcher) {
```
Zabuzard, line 115: No need for an extra method. Just create a simple boolean before the `if`. So `boolean hasMatchedDollarSign = matcher.group(1) != null);` is enough.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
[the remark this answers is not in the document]
Zabuzard: That said, I propose a different approach to the regex. While your approach is probably the most efficient, I think it is one of the less readable ways.

What you are doing now is: walk through the text, find any math-region or text-region and then do action on them.

Instead, I would just do: walk through the text, find any text-region and replace with `\text{...}`. Then, after that is fully done, simply do `.replace("$", "")`, done.

I am not convinced that you actually need to match math-regions in the first place. So your first group should be obsolete, I think.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: I would also rename it to `isInsideMathRegion`. Cause you essentially have either math-region or text-region.

Its probably easier to understand the code if its looked at like that. In fact, the regex could also be changed to `(%s|%s).formatted(MATH_REGION, TEXT_REGION)` for more readability.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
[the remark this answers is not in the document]
IslamSakrak: I changed it to this, would this be fine? I split it so it is more readable this way from this suggestion? https://github.com/Together-Java/TJ-Bot/pull/397#discussion_r813623023

```java
    private static final String MATH_REGION = "(\\$\\S+\\$)";
    private static final String TEXT_REGION = "([a-zA-Z!., ]+)";
    private static final Pattern INLINE_LATEX_REPLACEMENT = Pattern.compile(MATH_REGION + "|" + TEXT_REGION);
```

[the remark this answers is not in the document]
Tais993: Yes! I love that you split it up, the variable names make it very descriptive now.

I'd still add a quick comment:
```java
    // everything between 2 $'s
    private static final String MATH_REGION = "(\\$\\S+\\$)";
    // ..
    private static final String TEXT_REGION = "([a-zA-Z!., ]+)";
    private static final Pattern INLINE_LATEX_REPLACEMENT = Pattern.compile(MATH_REGION + "|" + TEXT_REGION);
    ```
    
    Just to make sure

File: application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java (hunk 6 above)
Comment (IslamSakrak) [reply application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java]: I have removed the method completely

File: application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java [hunk 12]
```diff
@@ -34,6 +36,11 @@
 public class TeXCommand extends SlashCommandAdapter {
 
     private static final String LATEX_OPTION = "latex";
+    // everything between 2 $'s
```
Zabuzard, line 39: I think an example would help as well.
```suggestion
    // Matches regions between two dollars, like '$foo$'.
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -34,6 +36,11 @@
...
+    private static final String MATH_REGION = "(\\$\\S+\\$)";
```
Zabuzard, line 40: `\\S+` is wrong. That way it wouldnt support whitespaces but the tex could as well contain whitespaces. You should accept anything in here. So `.+`.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,6 +64,7 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
         try {
+            latex = convertInlineLatexToFull(latex);
```
Zabuzard, line 67: What if the message doesnt even contain any inline latex? The input should be considered a full latex already then, i.e. ur method should not be executed in that case. A comment might also help to explain whats going on.

I.e.
```java
if (latex.contains("$")) {
  latex = convertInlineLatexToFull(latex);
}
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +100,21 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+        while (matcher.find()) {
+            boolean isInsideMathRegion = matcher.group(1) != null;
+            if (isInsideMathRegion) {
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+        return sb.toString();
+    }
```
Zabuzard, line 116: A quick comment at the beginning of the method, or as Javadoc could help. Sth along the lines of:
```java
// Converts inline latex like 'hello $\frac{x}{2}$ world'
// to full latex '\text{hello }\frac{x}{2}\text{ world}'.
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +100,21 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
             .queue();
     }
 
+    @NotNull
+    private String convertInlineLatexToFull(String latex) {
```
Zabuzard, line 104: method parameter is missing `@NotNull`

Zabuzard (changes requested):

Zabuzard (changes requested):
You should also check if you can adjust the description of the command so that users know that they can now also use inline-latex.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: Yes but it also shouldnt. Like, the `/tex` command should still be able to be used as it was originally intended. So like
`/tex \frac{x}{2}`, only with latex and no actual text.

So the command has to support both:
* just latex, but without `$...$`
* mixed text with inline latex, i.e. with `$...$`

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -99,9 +101,13 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
             .setActionRow(Button.of(ButtonStyle.DANGER, generateComponentId(userID), "Delete"))
             .queue();
     }
-
+    /**
+     * Converts inline latex like:
+     * 'hello $\frac{x}{2}$ world' to full latex '\text{hello }\frac{x}{2}\text{ world}'.
```
Zabuzard, line 106: Btw, you can do `{@code ...}` in javadoc. Might make the two examples more readable if u wrap it around them.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -52,7 +52,7 @@
      */
     public TeXCommand() {
         super("tex",
-                "This command accepts a latex expression and generates an image corresponding to it.",
+                "This command accepts a regular and an inline latex expression and generates an image corresponding to it.",
```
Zabuzard, line 55: Not your fault, but I honestly dont like this description. It feels like it is more written for a developer instead of an actual average Discord user.

My proposal:
```java
"Renders LaTeX, also supports inline $-regions like 'see this $\frac[x}{2\$'."
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Tais993: Okay but honestly, it's not the average Discord user.

How is the average Discord user going to use a LaTeX command if they don't know what LaTeX is?

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: I meant it more like
```java
Converts inline latex like:
{@code hello $\frac{x}{2}$ world}
to full latex
{@code \text{hello}\frac{x}{2}\text{ world}}.
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: Thats fair. But the description is really rather the opposite of compact. And it doesnt explain what an inline-latex is.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: Now its correct. Could you please double check whether Intellij actually renders the Javadoc correctly on mousehover? Maybe we have to escape the `{...}` of `frac` and `text` for it to not freak out.

Just tell me whether it works, then I will close this.

Zabuzard: (shown in discord, looks good).

Zabuzard (dismissed):
Well done 👍

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +101,27 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+        StringBuilder sb = new StringBuilder();
+        while (matcher.find()) {
+            boolean isInsideMathRegion = matcher.group(1) != null;
+            if (isInsideMathRegion) {
+                sb.append(matcher.group(1).replace("$", ""));
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+        return sb.toString();
```
Tais993, line 122:
```suggestion
        }
        
        return sb.toString();
```

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -92,6 +101,27 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+     *
+     * Converts inline latex like: {@code hello $\frac{x}{2}$ world} to full latex
+     * {@code \text{hello}\frac{x}{2}\text{ world}}.
+     *
+     */
+    @NotNull
+    private String convertInlineLatexToFull(@NotNull String latex) {
+        Matcher matcher = INLINE_LATEX_REPLACEMENT.matcher(latex);
+        StringBuilder sb = new StringBuilder();
+        while (matcher.find()) {
```
Tais993, line 114:
```suggestion
        StringBuilder sb = new StringBuilder();
        
        while (matcher.find()) {
```

Tais993 (commented):
subjective, so just think about it

Tais993 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -109,9 +110,11 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
      */
     @NotNull
     private String convertInlineLatexToFull(@NotNull String latex) {
+        if (isInvalidInlineFormat(latex)) {
+            throw new ParseException("One dollar sign is invalid, wrap whole expression with it. ");
```
Zabuzard, line 137: The error message is not accurate. For example, if you give 3 dollars, it also triggers. But the message only says "one dollar".

> The amount of $-symbols must be divisible by two. Did you forget to close an expression?

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -84,14 +98,62 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
             logger.warn(
                     "Unable to render latex, could not convert the image into an attachable form. Formula was {}",
                     latex, e);
-            return;
+            return null;
```
Zabuzard, line 101: dont use null to indicate absence. use Optional instead.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,21 +64,28 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
             return;
-        }
+        sendImage(event, userID, renderedTextImageStream);
+    }
+
+    @Nullable
+    private ByteArrayOutputStream getRenderedTextImageStream(@NotNull SlashCommandEvent event,
+            String latex, Image image) {
+        if (image == null)
+            return null;
```
Zabuzard, line 88: dont use null to indicate absence. use Optional instead.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,21 +64,28 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+        ByteArrayOutputStream renderedTextImageStream =
+                getRenderedTextImageStream(event, latex, image);
+        if (renderedTextImageStream == null)
             return;
-        }
+        sendImage(event, userID, renderedTextImageStream);
+    }
+
+    @Nullable
+    private ByteArrayOutputStream getRenderedTextImageStream(@NotNull SlashCommandEvent event,
```
Zabuzard, line 85: I dont like the name. You are using `get` as if it would be an innocant getter - but it is not. It is actually doing some heavy stuff.

The method converts an `Image` into `ByteArrayOutputStream`. So maybe `convertImageToRaw` would be more appropriate.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,21 +64,28 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+            String latex, Image image) {
```
Zabuzard, line 86: notnull/nullable on the params

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -84,14 +98,62 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
                     "Unable to render latex, could not convert the image into an attachable form. Formula was {}",
                     latex, e);
-            return;
+            return null;
+        }
+        return renderedTextImageStream;
+    }
+
+    @Nullable
+    private Image getImage(@NotNull SlashCommandEvent event, String latex, TeXFormula formula) {
```
Zabuzard, line 107: I dont like the name of this. It is not a getter. It computes some heavy stuff.

It takes the `latex` and gives out an `Image` for it. I would call it `renderImage` instead. Or `renderLatexAsImage`.

Zabuzard: While at it, please change the order of the methods to the order they appear in the code. So:
1. getImage
2. getRenderedTextImageStream
3. sendImage

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -57,21 +64,28 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
-                FOREGROUND_COLOR, BACKGROUND_COLOR);
-        if (image.getWidth(null) == -1 || image.getHeight(null) == -1) {
-            event.getHook().setEphemeral(true).editOriginal(RENDERING_ERROR).queue();
-            logger.warn(
-                    "Unable to render latex, image does not have an accessible width or height. Formula was {}",
-                    latex);
+        Image image = getImage(event, latex, formula);
+        ByteArrayOutputStream renderedTextImageStream =
+                getRenderedTextImageStream(event, latex, image);
+        if (renderedTextImageStream == null)
```
Zabuzard, line 79: I would probably move this "image to raw" part into `sendImage`.

Name-wise a method called `sendImage` should take an `Image` and not some raw data stream. (so `sendImage` should then basically call `getRenderedTextImageStream` internally)

Zabuzard (changes requested):

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -56,42 +62,102 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+            } else {
+                sb.append("\\text{").append(matcher.group(2)).append("}");
+            }
+        }
+
+        return sb.toString();
+    }
+
+    private boolean isInvalidInlineFormat(String latex) {
+        return latex.chars().filter(ch -> ch == '$').count() % 2 == 1;
```
Heatmanofurioso, line 158: Same here.
We should escape the possible escaped `\$` characters

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -56,42 +62,102 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
...
+                    "The amount of $-symbols must be divisible by two. Did you forget to close an expression? ");
+        }
+
+        Matcher matcher = INLINE_LATEX_REPLACEMENT.matcher(latex);
+        StringBuilder sb = new StringBuilder();
+
+        while (matcher.find()) {
+            boolean isInsideMathRegion = matcher.group(1) != null;
+            if (isInsideMathRegion) {
+                sb.append(matcher.group(1).replace("$", ""));
```
Heatmanofurioso, line 148: We should watch for escaped dollars in text `\$`

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
```diff
@@ -56,42 +62,102 @@ public void onSlashCommand(@NotNull final SlashCommandEvent event) {
         String latex = Objects.requireNonNull(event.getOption(LATEX_OPTION)).getAsString();
         String userID = (Objects.requireNonNull(event.getMember()).getId());
         TeXFormula formula;
+
         try {
+            if (latex.contains("$")) {
```
Heatmanofurioso, line 67: We should watch out for escaped dollar signs here

Heatmanofurioso (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: We discussed this and agreed that we can ignore it for now.

So unless you have a cheap and easy suggestion to adjust the code to support it, we can also go without it for now.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: We discussed this and agreed that we can ignore it for now.

So unless you have a cheap and easy suggestion to adjust the code to support it, we can also go without it for now.

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: We discussed this and agreed that we can ignore it for now.

So unless you have a cheap and easy suggestion to adjust the code to support it, we can also go without it for now.

Heatmanofurioso: Sorry about that.
Didn't know it was discussed.

But, wouldn't `(?<!\\)\$` be a good regex for it? It would only match unescaped dollar signs

Zabuzard: > Didn't know it was discussed.

No worries, not your fault. It was probably in Discord and not here.

Heatmanofurioso: I believe it does.
It's a negative look behind regex.

We could also add a really simple unit test to test it too.
Not all tests need to be integration tests 😆

Heatmanofurioso: image

Zabuzard: > We could also add a really simple unit test to test it too.

Yes, totally correct. We still have a separate GH issue open for adding tests to `/tex`, so I didnt want to bloat up this PR with more tasks than necessary (poor @IslamSakrak )

Zabuzard: I actually think that, as soon as this is merged, someone should tackle the unit tests for `/tex`. It would benefit a lot from it.

Heatmanofurioso: You're correct.
I'm gonna mention this point on that issue, so it doesn't fall into the forgotten realm when those tests get addressed

IslamSakrak: I wouldn't mind testing this also, I have written some tests already. Also, first time writing tests so might help me get better at it

IslamSakrak:
61 conversations personal best 😭

application/src/main/java/org/togetherjava/tjbot/commands/mathcommands/TeXCommand.java
Zabuzard: If you want, feel free to either create a second PR or also do it directly in this.

Zabuzard:
rip. @IslamSakrak

PR #410, opened 2022-03-08
Colorblind support for help system emojis

### LINKED ISSUE #409: Replace red/green circles from help system (colorblind accessibility)
## Overview

Our help systems status message uses 🔴 and 🟢 , which are hard to distinguish for people with red/green colorblindness. We should replace them with icons that also add a different shape.

status message

The relevant file is `ChannelStatusType.java`.

Zabuzard:
## Overview

Implements and closes #409.

As suggested by someone in #tj_suggestions who is red/green colorblind, the existing emojis for the help system status messages are barely distinguishable (🟢 🔴).

This changes the emojis to something that also has a different shape, while still having nice color coding for people who are not colorblind (✅ ❌).

free
busy

Looks for example like this for a colorblind person (deuteranopia):

colorblind

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelStatusType.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelStatusType.java
@@ -3,8 +3,8 @@ package org.togetherjava.tjbot.commands.free;
 import org.jetbrains.annotations.NotNull;
 
 enum ChannelStatusType {
-    FREE("free", ":green_circle:"),
-    BUSY("busy", ":red_circle:");
+    FREE("free", ":white_check_mark:"),
+    BUSY("busy", ":x:");
 
     private final String description;
     private final String emoji;


Tais993 (approved):

MaiTheLord:
as borgrel said here, https://github.com/Together-Java/TJ-Bot/issues/409#issuecomment-1065920526

> as soon as u change it from colours to symbols you get the problem that symbols need to be interpretted:
does a tick mean the channel is occupied or does it mean that its available? plenty of ppl will think it means the one way and plenty will think it means the other

this could be rather confusing. I would suggest to add a legend to the embed body to make it clearer, maybe something like this:
```
✅ - free
❌ - used
```

Zabuzard:
@MaiTheLord This is already explained in the message in the channel in our real server actually

screen

MaiTheLord:
I see. looks pretty good then!

Zabuzard:
@MaiTheLord If you like it, feel free to leave an APPROVE in the code review, then I can merge it :)

MaiTheLord (approved):

PR #412, opened 2022-03-14
Bugfix - Handling blocked users gracefully when reacting to their messages

### LINKED ISSUE #406: Autoreact on suggestions doesnt handle blocked users gracefully
## Overview

If someone blocks our TJ-Bot, it can not react on their suggestions. Right now, the code throws exceptions in that case:

log

## Proposed action

While we probably can not really work around that easily with the current system, we should at least handle this situation gracefully and not crash.

MaiTheLord:
warped the line that reacts to the message in a `try catch` block. see file changes.
resolves #406.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -62,6 +62,8 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
         }, exception -> {
             if (exception instanceof ErrorResponseException responseException
                     && responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
+                // User blocked the bot, hence the bot can not add reactions to their messages.
+                // Nothing we can do here.
                 return;
             }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -60,8 +60,8 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
             return message.addReaction(fallbackUnicodeEmote);
         }).queue(ignored -> {
         }, exception -> {
-            if (exception instanceof ErrorResponseException responseException &&
-                responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
+            if (exception instanceof ErrorResponseException responseException
+                    && responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
                 return;
             }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -53,18 +53,20 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
 
     private static void reactWith(@NotNull String emoteName, @NotNull String fallbackUnicodeEmote,
             @NotNull Guild guild, @NotNull Message message) {
-        try {
-            getEmoteByName(emoteName, guild).map(message::addReaction).orElseGet(() -> {
-                logger.warn(
-                        "Unable to vote on a suggestion with the configured emote ('{}'), using fallback instead.",
-                        emoteName);
-                return message.addReaction(fallbackUnicodeEmote);
-            }).complete();
-        } catch (ErrorResponseException exception) {
-            if (exception.getErrorResponse() != ErrorResponse.REACTION_BLOCKED) {
-                throw exception;
+        getEmoteByName(emoteName, guild).map(message::addReaction).orElseGet(() -> {
+            logger.warn(
+                    "Unable to vote on a suggestion with the configured emote ('{}'), using fallback instead.",
+                    emoteName);
+            return message.addReaction(fallbackUnicodeEmote);
+        }).queue(ignored -> {
+        }, exception -> {
+            if (exception instanceof ErrorResponseException && ((ErrorResponseException) exception)
+                .getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
+                return;
             }
-        }
+
+            logger.error(exception.getMessage(), exception);
+        });
     }
 
     private static @NotNull Optional<Emote> getEmoteByName(@NotNull String name,
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -60,8 +60,8 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
             return message.addReaction(fallbackUnicodeEmote);
         }).queue(ignored -> {
         }, exception -> {
-            if (exception instanceof ErrorResponseException && ((ErrorResponseException) exception)
-                .getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
+            if (exception instanceof ErrorResponseException responseException &&
+                responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
                 return;
             }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -65,7 +65,7 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
                 return;
             }
 
-            logger.error(exception.getMessage(), exception);
+            logger.error("Attempted to react to a suggestion, but failed", exception);
         });
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -4,6 +4,8 @@ import net.dv8tion.jda.api.entities.Emote;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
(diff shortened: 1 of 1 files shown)


MaiTheLord:
> Why complete

can't catch `queue()`, afaik.

MaiTheLord:
iirc, `queue()` throws exceptions on the executing thread, thus you can't catch them on the calling thread.

MaiTheLord:
```java
.queue(ignored -> {},
exception -> {
    if (exception instanceof ErrorResponseException && ((ErrorResponseException) exception).getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
        return;
    }
    exception.printStackTrace();
});
```
is this better?

Zabuzard:
u have to use our logging system, not e.printStackTrace. so at least `logger.error("...", e)`

Zabuzard:
> The best thing to do is to throw it, probably, but how do I do it from inside the function? 🤔

Then you would throw from within the JDA context where the lambda is executed, which is detached from our flow - not good.

If you stick with `queue`, you should log the error using the logger. If you stick with `complete`, try-catch is fine. That means that the call is blocking though and not async. Which means that you could potentially block one of our bot-core-system threads for a long time. We only have a few of these, so if there are many reactions going on and they, for whatever reason, take long to complete, that can block the whole bot and stuff like `/tag foo` can not be executed anymore and will die.

tldr, dont block our threads.

MaiTheLord:
that should do it.

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -56,7 +58,15 @@ private static void reactWith(@NotNull String emoteName, @NotNull String fallbac
...
             return message.addReaction(fallbackUnicodeEmote);
-        }).queue();
+        }).queue(ignored -> {
+        }, exception -> {
+            if (exception instanceof ErrorResponseException && ((ErrorResponseException) exception)
+                .getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
+                return;
+            }
+
+            logger.error(exception.getMessage(), exception);
```
Tais993, line 68: Just log the exception, it'll print the exception message twice now

Zabuzard: to add to what tais says, the loggers `logger.foo(..., exception)` already appends the full error message, stack trace whatsoever to the log. the first part is just ur text of whats going on and to provide context. like `logger.error("Attempted to react to a suggestion, but failed for an unknown reason", exception);`

Zabuzard: leads me into thinking that u actually didnt try out the code to see how it looks, but just coded and pushed it right of the bat. u shouldnt do that, trigger ur new code to verify it.

MaiTheLord: I am pretty sure it won't. how do you suggest I log it?

Zabuzard: > I am pretty sure it won't.

I am pretty sure it does. Right now it will say

> Foo bar
> FooException: Foo bar
> at ...
> at ...

So the message is twice. Just completely unecessary. Instead, add context.

> how do you suggest I log it?

Just like I showed above. So
```java
logger.error("Attempted to react to a suggestion, but failed for an unknown reason", exception);
```

Zabuzard: Dont assume. Try it out and you will see yourself.

MaiTheLord: I just tested it by removing the `if` statement completely, blocking the bot and sending a message, and I got this:
https://gist.github.com/MaiTheLord/90c678e4948d733c7a4edcb7a48dd715
it sent it twice as it tries to send two embeds.

MaiTheLord:
```
22:01:21.247 [ForkJoinPool.commonPool-worker-1] ERROR org.togetherjava.tjbot.commands.basic.SuggestionsUpDownVoter - 90001: Reaction blocked
net.dv8tion.jda.api.exceptions.ErrorResponseException: 90001: Reaction blocked
```

MaiTheLord: When it logs an exception the stack trace shows this.

image

is that supposed to happen? it doesn't happen if I don't use any error handler for the `queue()`.

MaiTheLord:
When it logs an exception the stack trace shows this.

image

is that supposed to happen? it doesn't happen if I don't use any error handler for the `queue()`.

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
MaiTheLord: this solves it, I guess. is that good?
```java
logger.error("", exception);
```

MaiTheLord: > leads me into thinking that u actually didnt try out the code to see how it looks, but just coded and pushed it right of the bat. u shouldnt do that, trigger ur new code to verify it.

actually I did test it. I just didn't pay much attention to the 2 first line of the stacktrace, unfortunately,

MaiTheLord: > ```java
> logger.error("Attempted to react to a suggestion, but failed for an unknown reason", exception);
> ```

I wouldn't say "unknown reason", as the reason is stated in the stacktrace. maybe just
```java
logger.error("Attempted to react to a suggestion, but failed", exception);
```

Zabuzard: sounds good

Tais993:
I don't know or it should, please give the full stacktrace

MaiTheLord:
Here ya go
https://gist.github.com/MaiTheLord/90c678e4948d733c7a4edcb7a48dd715

Zabuzard:
Sounds like you logged the error now. Didnt you want to ignore it and only log unknown failures? Whats the code that produced this?

MaiTheLord:
I commented the part that ignores the error, for the sake of seeing a stacktrace, as I couldn't think of any other way to make it fail.

MaiTheLord:
I found another way to make it fail (by revoking the bots permissions).
https://gist.github.com/MaiTheLord/99335c4206d4f81da055df4fbb5e6deb

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
MaiTheLord: can I resolve this?

Zabuzard:
Not sure what your question or confusion is, to be honest. This is the expected behavior - it logs the unknown error.

MaiTheLord:
I wasn't sure about the `caused by ContextException`.

when I look at it now, seems like its JDA's way to handle async exceptions.

Tais993:
> When it logs an exception the stack trace shows this.

ah yeah, that's just the exception, scroll up and you see the cause.

>>> I found another way to make it fail.
https://gist.github.com/MaiTheLord/99335c4206d4f81da055df4fbb5e6deb
produced by latest code, https://github.com/Together-Java/TJ-Bot/commit/8d554a3ede924c5db838bb5219f86cc92b5008f4, and revoking the Add Reactions permission

There's a ton of ways to make it fail, but we shouldn't do anything with this.
It logs the error, the bot owner can add permissions, not our issue.
Just keep it like this

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -56,7 +58,15 @@ private static void reactWith(@NotNull String emoteName, @NotNull String fallbac
                     "Unable to vote on a suggestion with the configured emote ('{}'), using fallback instead.",
                     emoteName);
             return message.addReaction(fallbackUnicodeEmote);
-        }).queue();
+        }).queue(ignored -> {
+        }, exception -> {
+            if (exception instanceof ErrorResponseException && ((ErrorResponseException) exception)
+                .getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
```
Tais993, line 64:
```suggestion
            if (exception instanceof ErrorResponseException responseException &&
                responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
```

Tais993 (changes requested):

Tais993 (approved):

MaiTheLord:
Zabuzard, can you review please?

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -56,7 +58,15 @@ private static void reactWith(@NotNull String emoteName, @NotNull String fallbac
                     "Unable to vote on a suggestion with the configured emote ('{}'), using fallback instead.",
                     emoteName);
             return message.addReaction(fallbackUnicodeEmote);
-        }).queue();
+        }).queue(ignored -> {
+        }, exception -> {
+            if (exception instanceof ErrorResponseException responseException
+                    && responseException.getErrorResponse() == ErrorResponse.REACTION_BLOCKED) {
```
Zabuzard, line 64: Maybe add a quick comment to explain this use case:
```java
// User blocked the bot, hence the bot can not add reactions to their messages. Nothing we can do here.
return;

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
MaiTheLord: sure.

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Zabuzard: yes

Tais993 (approved):

Zabuzard (approved):

PR #417, opened 2022-03-16
Removed code duplication from role based moderation actions.

MaiTheLord:
Introduced abstract class `RevocableRoleBasedAction` to remove that duplication.


application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
```diff
@@ -12,21 +12,14 @@
  */
 interface RevocableModerationAction {
     /**
-     * Classification of a revocation failure.
+     * The name of the action, that will be logged in case of a failed revocation.
+     *
+     * <p>
+     * Naming convention examples: {@code "ban"}, {@code "mute"}, {@code "quarantine"}
```
Zabuzard, line 18: The example would be better if it gives an example of a full sentence, to give the context in which the word might be used.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
```diff
@@ -12,21 +12,14 @@
  */
 interface RevocableModerationAction {
     /**
-     * Classification of a revocation failure.
+     * The name of the action, that will be logged in case of a failed revocation.
```
Zabuzard, line 15: Either you call it `actionName` but then also give a more generic and less detailed description of its use case - or you call it `failedRecovationActionName` and only then this javadoc description is okay (I would go for former).

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryBanAction.java
```diff
@@ -33,30 +32,4 @@
...
-                        targetId);
-                return FailureIdentification.KNOWN;
-            }
-
-            if (errorResponseException.getErrorResponse() == ErrorResponse.MISSING_PERMISSIONS) {
-                logger.warn("Attempted to revoke a temporary ban but the bot lacks permission.");
-                return FailureIdentification.KNOWN;
-            }
-        }
-        return FailureIdentification.UNKNOWN;
```
Zabuzard, line 60: For BAN the logic changes now, that ur using the merged logic. Is the difference in behavior acceptable or not?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
```diff
@@ -138,16 +140,34 @@ private void revokeAction(@NotNull RevocationGroupIdentifier groupIdentifier) {
...
+                case UNKNOWN_ROLE -> logger.warn(
+                        "Attempted to revoke a temporary {} but the {} role can not be found.",
+                        actionName, actionName);
+                case MISSING_PERMISSIONS -> logger.warn(
+                        "Attempted to revoke a temporary {} but the bot lacks permission.",
+                        actionName);
+                case UNKNOWN_BAN -> logger.debug(
+                        "Attempted to revoke a temporary {} but the user '{}' is not banned anymore.",
+                        actionName, targetId);
+                default -> logUnknownFailure.run();
```
Zabuzard, line 166: This is a design failure. Now you have the "parent class" requiring knowledge about its implementation. I.e. `Animal` having specific code for `Dog` and `Cat`. As in, this class now assumes its implementations are BAN and role-based actions. Not good.

If you want to have role-based actions, then they have to be in a specific class or method that deals only with role based actions, and then be given/picked by the subclass. Not the parent.

I am voting for going back to the idea of adding a utility class, which the implementations will then pick.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
```diff
@@ -138,16 +140,34 @@ private void revokeAction(@NotNull RevocationGroupIdentifier groupIdentifier) {
...
-            return;
-        }
-
-        logger.warn(
+    private void handleFailure(@NotNull Throwable failure, long targetId,
+            @NotNull String actionName) {
+        Runnable logUnknownFailure = () -> logger.warn(
                 "Attempted to revoke a temporary moderation action for user '{}' but something unexpected went wrong.",
-                groupIdentifier.targetId, failure);
+                targetId, failure);
```
Tais993, line 147: I personally find this random variable to decrease readability, it's such a small lambda too.
Just put it in the else statement?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
```diff
@@ -12,21 +12,14 @@
...
-         */
-        KNOWN,
-        /**
-         * The failure is unknown and has not been handled. Hence, further error handling should be
-         * continued.
-         */
-        UNKNOWN
-    }
+    @NotNull
+    String actionName();
```
Tais993, line 22: You can use the constructor instead, that's shorter in almost all cases

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
Tais993: Should've reloaded before my review, I do agree with Zabu.
Original intention was an utility method, and I would prefer that over this

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
Zabuzard: (its an interface though :D)

Tais993: My bad, alcohol is bad kids

MaiTheLord:
Tried using an abstract implementation. see the change.
Also, changed `handleFailure()` logic, to only forward an `ErrorResponse` to the implementation's handler, as this handler is supposed to handle errors thrown by a rest action.

MaiTheLord:
I'll appreciate a review.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link RevocableModerationAction} abstract implementation for a role based moderation action.
+ * <br />
+ * <br />
+ * Implements {@link RevocableModerationAction#handleRevokeFailure(ErrorResponse, long)} to handle
+ * possible errors for a role-based revocation action.
```
Zabuzard, line 13: A detail that shouldnt be part of the javadoc since it might only be true "today" but not tomorrow.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
```diff
@@ -0,0 +1,50 @@
+package org.togetherjava.tjbot.commands.moderation.temp;
+
+import net.dv8tion.jda.api.requests.ErrorResponse;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link RevocableModerationAction} abstract implementation for a role based moderation action.
```
Zabuzard, line 9: That javadoc is obsolete and doesnt add anything to the reader, not helpful. It is basically identical to the first line of code `abstract class RevocableRoleBasedAction implements RevocableModerationAction`.

Instead, you should take the opportunity and explain the use case, what this is.

> Role based moderation actions that can be revoked, for example a {@link ...Mute...} or {@link ...Quarantine...} which are applied implicitly purely by the presence of a role.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(RevocableRoleBasedAction.class);
+
+    private final String actionName;
+
+    /**
+     * Creates an instance of {@link RevocableRoleBasedAction} with the given action name.
+     *
+     * @param actionName the name of the moderation action
+     */
+    RevocableRoleBasedAction(String actionName) {
```
Zabuzard, line 25: not null

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
```diff
@@ -11,23 +12,6 @@
...
-         * Acknowledges that the failure is known and has been handled. Hence, further error
-         * handling should not be continued.
-         */
-        KNOWN,
-        /**
-         * The failure is unknown and has not been handled. Hence, further error handling should be
-         * continued.
-         */
-        UNKNOWN
-    }
```
Zabuzard, line 29: Why did you remove this? Please undo this. Now you reverted to "boolean" and the code is now worse, you landed in the so called "boolean-trap". Returning an enum is by far more readable to an user than a boolean. Please use the enum instead again, it was put there on purpose to avoid the boolean-trap.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
```diff
@@ -62,11 +46,9 @@
...
-     * @param failure the failure to handle
+     * @param errorResponse the error that occurred
      * @param targetId the id of the user who is targeted by the revocation
-     * @return a classification of the failure, decides whether the surrounding flow will continue
-     *         to handle the error further or not
+     * @return whether it should log an unknown failure message or not
      */
-    @NotNull
-    FailureIdentification handleRevokeFailure(@NotNull Throwable failure, long targetId);
+    boolean handleRevokeFailure(@NotNull ErrorResponse errorResponse, long targetId);
```
Zabuzard, line 53: I dont like that you changed `Throwable` to `ErrorResponse`. Design-wise, you are taking away capabilities from the system here. For example, if there is a temporary moderation action that can throw a FooException during revocation, it can not handle it now anymore. It can only handle Discord-API responses now, but not custom issues from our code.

I would suggest you revert back this part and use `Throwable failure` again and move the `if (failure instanceof ErrorResponse errorResponse) { ... }` back into the implementations (the abstract class and the temp-ban).

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
MaiTheLord: Doesn't JDA's `queue()` only throw `ErrorResponse`s?

Zabuzard: No. Anything can happen when it executes your code. For example, what if you have a `.map(...)` which throws some exception? It will be exposed here as well.

The exception you get here is not only the actual Discord/JDA exceptions but everything that could go wrong during the whole execution stage, including your transformations.

MaiTheLord: Maybe we should have this handle all throwables, so that the code will be valid in the future?

Zabuzard: And thats precisely what I proposed above. Just undo your changes on this part, they did not make the code better - but worse.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableModerationAction.java
MaiTheLord: undid it.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
MaiTheLord: done.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
MaiTheLord: changed it.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
MaiTheLord: removed it.

MaiTheLord:
I'll appreciate a review.

Tais993:
Could you please stop asking for reviews, use GitHub's GUI to request a review.
After 10 days or so feel free to post a comment, but I have things to do in my life, I can't constantly review on Discord, 1 day isn't a lot of time and makes me more interested in ignoring than reviewing.

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
```diff
@@ -0,0 +1,61 @@
...
+                logger.warn("Attempted to revoke a temporary {} but the {} role can not be found.",
+                        actionName, actionName);
+                return FailureIdentification.KNOWN;
+            }
+
+            if (errorResponseException.getErrorResponse() == ErrorResponse.MISSING_PERMISSIONS) {
+                logger.warn("Attempted to revoke a temporary {} but the bot lacks permission.",
+                        actionName);
+                return FailureIdentification.KNOWN;
+            }
```
Tais993, line 57: Switch expression/statement?

```java
return switch (errorResponseException.getErrorResponse()) {
    UNKNOWN_USER -> {
        logger.debug(
                "Attempted to revoke a temporary {} but user '{}' does not exist anymore.",
                actionName, targetId);
        yield FailureIdentification.KNOWN;
    }
}
```
And then for all choices + default case

Or maybe just a switch statement without yield, whatever you prefer

Tais993 (changes requested):

Tais993 (approved):

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
```diff
@@ -0,0 +1,55 @@
...
+                        actionName, targetId);
+                case UNKNOWN_ROLE -> logger.warn(
+                        "Attempted to revoke a temporary {} but the {} role can not be found.",
+                        actionName, actionName);
+                case MISSING_PERMISSIONS -> logger.warn(
+                        "Attempted to revoke a temporary {} but the bot lacks permission.",
+                        actionName);
+                default -> {
+                    return FailureIdentification.UNKNOWN;
+                }
```
Zabuzard, line 47: Sonar correctly complains about the unecessary `{...}` here.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
Zabuzard: noice

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/RevocableRoleBasedAction.java
MaiTheLord: on a `return`? I believe a `return` requires the block, as it won't compile without it.

Zabuzard: return is implicit for one-liners. just `default -> FailureIdentification.UNKNOWN;`

Zabuzard: then dont directly return out of it but assign some variable so that both flow into the same return statement.
```java
FailureIdentification identification = FailureIdentification.KNOWN;
switch (...) {
  ...
  default -> identification = FailureIdentification.UNKNOWN;
}

return identification;
```

Dunno, try around, maybe you find something nicer. But I want to avoid merging something that introduces Sonar flags.

Tais993: Sonar is wrong, like usually.

Zabuzard: In this case, it is. We can click on "ignore", thats not an issue. But I still think that we could improve here, as reported above.

Tais993: Finally I can say it! fuck you Sonar \:)

And I'm fine with changing it so 👍

MaiTheLord: This is my other idea, but that will cause another sonar violation.
```java
FailureIdentification handlingResult = FailureIdentification.UNKNOWN;

if (failure instanceof ErrorResponseException errorResponseException) {
    handlingResult = FailureIdentification.KNOWN;

    switch (errorResponseException.getErrorResponse()) {
        case UNKNOWN_USER -> logger.debug(
                "Attempted to revoke a temporary {} but user '{}' does not exist anymore.",
                actionName, targetId);
        case UNKNOWN_MEMBER -> logger.debug(
                "Attempted to revoke a temporary {} but user '{}' is not a member of the guild anymore.",
                actionName, targetId);
        case UNKNOWN_ROLE -> logger.warn(
                "Attempted to revoke a temporary {} but the {} role can not be found.",
                actionName, actionName);
        case MISSING_PERMISSIONS -> logger.warn(
                "Attempted to revoke a temporary {} but the bot lacks permission.",
                actionName);
        default -> handlingResult = FailureIdentification.UNKNOWN;
    }
}

return handlingResult;
```
```
Assignments should not be made from within sub-expressions

Assignments within sub-expressions are hard to spot and therefore make the code less readable.
Ideally, sub-expressions should not have side-effects.
```

I think the best solution is to `ignore` this one.

Zabuzard: Can you double check whether `default -> return foo;` actually works without curlies? just to make sure.

Zabuzard: That said, I marked it as false-positive:

sonar

Zabuzard (approved):

Tais993 (approved):

PR #419, opened 2022-03-16
Auto-free inactive help channels

### LINKED ISSUE #392: Auto-free inactive help channels
### Overview

A help channel that has no activity for 2+ hours should automatically be marked **free** (as if using `/free`) by the system.

A small explanation message should also be posted, i.e. as if using `/free`. But maybe with some minor adjusted text like _"was automatically marked free because ..."_ to explain the situation.

### Alternative

Wait until we change our help system completely (thread/form based). But that might still take a couple of months and it would be nice to benefit from this low hanging fruit today already.

MaiTheLord:
## Overview

Implements and closes #392.

Automatically frees all help channels that are **inactive for more than 2 hours**:

auto-free

## Details

Solved by introducing a `AutoFreeRoutine` which executes each 5 minutes, processing the channels.

Functionality that has to be used by both, `FreeCommand` and the `AutoFreeRoutine`, has been moved over to `FreeChannelMonitor`, which both now take as constructor parameter.

## Config changes

The config slightly changed, introducing two new args to the `FreeCommandConfig`, looks like this:
```json
    "freeCommand": [
        {
            "inactiveChannelDuration": "PT2H",
            "messageRetrieveLimit": 10,
            "statusChannel": <put_a_channel_id_here>,
            "monitoredChannels": [
                    <put_a_channel_id_here>
                  ]
        }
   ],
```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
@@ -6,6 +6,8 @@ import org.togetherjava.tjbot.commands.basic.PingCommand;
 import org.togetherjava.tjbot.commands.basic.RoleSelectCommand;
 import org.togetherjava.tjbot.commands.basic.SuggestionsUpDownVoter;
 import org.togetherjava.tjbot.commands.basic.VcActivityCommand;
+import org.togetherjava.tjbot.commands.free.AutoFreeRoutine;
+import org.togetherjava.tjbot.commands.free.FreeChannelMonitor;
 import org.togetherjava.tjbot.commands.free.FreeCommand;
 import org.togetherjava.tjbot.commands.mathcommands.TeXCommand;
 import org.togetherjava.tjbot.commands.moderation.*;
@@ -59,6 +61,7 @@ public enum Features {
         ModerationActionsStore actionsStore = new ModerationActionsStore(database);
         ModAuditLogWriter modAuditLogWriter = new ModAuditLogWriter(config);
         ScamHistoryStore scamHistoryStore = new ScamHistoryStore(database);
+        FreeChannelMonitor freeChannelMonitor = new FreeChannelMonitor(config);
 
         // NOTE The system can add special system relevant commands also by itself,
         // hence this list may not necessarily represent the full list of all commands actually
@@ -71,6 +74,7 @@ public enum Features {
         features.add(new TopHelpersPurgeMessagesRoutine(database));
         features.add(new RemindRoutine(database));
         features.add(new ScamHistoryPurgeRoutine(scamHistoryStore));
+        features.add(new AutoFreeRoutine(freeChannelMonitor));
 
         // Message receivers
         features.add(new TopHelpersMessageListener(database, config));
@@ -102,7 +106,7 @@ public enum Features {
         features.add(new UnquarantineCommand(actionsStore, config));
 
         // Mixtures
-        features.add(new FreeCommand(config));
+        features.add(new FreeCommand(config, freeChannelMonitor));
 
         return features;
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/free/AutoFreeRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/free/AutoFreeRoutine.java
@@ -1,0 +1,55 @@
+package org.togetherjava.tjbot.commands.free;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.TextChannel;
+import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.commands.Routine;
+
+import java.util.Collection;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Routine that automatically marks busy help channels free after a certain time without any
+ * activity.
+ */
+public final class AutoFreeRoutine implements Routine {
+    private final FreeChannelMonitor channelMonitor;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param channelMonitor used to monitor and control the free-status of channels
+     */
+    public AutoFreeRoutine(@NotNull FreeChannelMonitor channelMonitor) {
+        this.channelMonitor = channelMonitor;
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        channelMonitor.guildIds()
+            .map(jda::getGuildById)
+            .filter(Objects::nonNull)
+            .forEach(this::processGuild);
+    }
+
+    private void processGuild(@NotNull Guild guild) {
+        // Mark inactive channels free
+        Collection<TextChannel> inactiveChannels = channelMonitor.freeInactiveChannels(guild);
+
+        // Then update the status
+        channelMonitor.displayStatus(guild);
+
+        // Finally, send the messages (the order is important to ensure sane behavior in case of
+        // crashes)
+        inactiveChannels.forEach(inactiveChannel -> inactiveChannel
+            .sendMessage(UserStrings.AUTO_MARK_AS_FREE.message())
+            .queue());
+    }
+
+    @Override
+    public @NotNull Schedule createSchedule() {
+        return new Schedule(ScheduleMode.FIXED_RATE, 1, 5, TimeUnit.MINUTES);
+    }
+}
--- a/application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
@@ -1,38 +1,56 @@
 package org.togetherjava.tjbot.commands.free;
 
-import net.dv8tion.jda.api.entities.Category;
-import net.dv8tion.jda.api.entities.Guild;
-import net.dv8tion.jda.api.entities.GuildChannel;
-import net.dv8tion.jda.api.entities.TextChannel;
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.entities.*;
+import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.api.utils.TimeUtil;
 import org.jetbrains.annotations.NotNull;
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.config.FreeCommandConfig;
 
+import java.awt.Color;
+import java.time.Instant;
 import java.util.*;
+import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
 
 /**
  * A class responsible for monitoring the status of channels and reporting on their busy/free status
  * for use by {@link FreeCommand}.
- *
+ * <p>
  * Channels for monitoring are added via {@link #addChannelToMonitor(long)} however the monitoring
  * will not be accessible/visible until a channel in the same {@link Guild} is registered for the
  * output via {@link #addChannelForStatus(TextChannel)}. This will all happen automatically for any
  * channels listed in {@link org.togetherjava.tjbot.config.FreeCommandConfig}.
- *
+ * <p>
  * When a status channel is added for a guild, all monitored channels for that guild are tested and
  * an {@link IllegalStateException} is thrown if any of them are not {@link TextChannel}s.
- *
+ * <p>
  * After successful configuration, any changes in busy/free status will automatically be displayed
  * in the configured {@code Status Channel} for that guild.
  */
-final class ChannelMonitor {
+public final class FreeChannelMonitor {
     // Map to store channel ID's, use Guild.getChannels() to guarantee order for display
     private final Map<Long, ChannelStatus> channelsToMonitorById;
     private final Map<Long, Long> guildIdToStatusChannel;
+    private final Map<Long, Long> channelIdToMessageIdForStatus;
 
-    ChannelMonitor() {
+    private static final String STATUS_TITLE = "**__CHANNEL STATUS__**\n\n";
+    private static final Color MESSAGE_HIGHLIGHT_COLOR = Color.decode("#CCCC00");
+
+    private final Config config;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param config the config to use
+     */
+    public FreeChannelMonitor(@NotNull Config config) {
         guildIdToStatusChannel = new HashMap<>(); // JDA required to populate map
         channelsToMonitorById = new HashMap<>();
+        channelIdToMessageIdForStatus = new HashMap<>();
+        this.config = config;
     }
 
     /**
@@ -48,7 +66,7 @@ final class ChannelMonitor {
      * Method for adding the channel that the status will be printed in. Even though the method only
      * stores the long id it requires, the method requires the actual {@link TextChannel} to be
      * passed because it needs to verify it as well as store the guild id.
-     *
+     * <p>
      * This method also calls a method which updates the status of the channels in the
      * {@link Guild}. So always add the status channel <strong>after</strong> you have added all
      * monitored channels for the guild, see {@link #addChannelToMonitor(long)}.
@@ -57,14 +75,14 @@ final class ChannelMonitor {
      */
     public void addChannelForStatus(@NotNull final TextChannel channel) {
         guildIdToStatusChannel.put(channel.getGuild().getIdLong(), channel.getIdLong());
-        updateStatusFor(channel.getGuild());
+        freeInactiveChannels(channel.getGuild());
     }
 
     /**
      * This method tests whether a guild id is configured for monitoring in the free command system.
      * To add a guild for monitoring see {@link org.togetherjava.tjbot.config.FreeCommandConfig} or
      * {@link #addChannelForStatus(TextChannel)}.
-     * 
+     *
      * @param guildId the id of the guild to test.
      * @return whether the guild is configured in the free command system or not.
      */
@@ -77,7 +95,7 @@ final class ChannelMonitor {
      * system. To add a channel for monitoring see
      * {@link org.togetherjava.tjbot.config.FreeCommandConfig} or
      * {@link #addChannelToMonitor(long)}.
-     * 
+     *
      * @param channelId the id of the channel to test.
      * @return {@code true} if the channel is configured in the system, {@code false} otherwise.
      */
@@ -107,36 +125,38 @@ final class ChannelMonitor {
 
     /**
      * This method tests if a channel is currently active by fetching the latest message and testing
-     * if it was posted more recently than the configured time limit, see
-     * {@link FreeUtil#inactiveTimeLimit()} and
-     * {@link org.togetherjava.tjbot.config.FreeCommandConfig#INACTIVE_DURATION},
-     * {@link org.togetherjava.tjbot.config.FreeCommandConfig#INACTIVE_UNIT}.
+     * if it was posted more recently than the configured time limit.
      *
      * @param channel the channel to test.
+     * @param when the reference moment, usually "now"
      * @return {@code true} if the channel is inactive, false if it has received messages more
      *         recently than the configured duration.
      * @throws IllegalArgumentException if the channel passed is not monitored. See
      *         {@link #addChannelToMonitor(long)}
      */
-    public boolean isChannelInactive(@NotNull final TextChannel channel) {
+    public boolean isChannelInactive(@NotNull final TextChannel channel, @NotNull Instant when) {
         requiresIsMonitored(channel.getIdLong());
 
-        // TODO change the entire inactive test to work via rest-actions
-        return FreeUtil.getLastMessageId(channel)
-            // black magic to convert OptionalLong into Optional<Long> because OptionalLong does not
-            // have .map
+        OptionalLong maybeLastMessageId = FreeUtil.getLastMessageId(channel);
+        if (maybeLastMessageId.isEmpty()) {
+            return true;
+        }
+
+        FreeCommandConfig configForChannel = config.getFreeCommandConfig()
             .stream()
-            .boxed()
-            .findFirst()
-            .map(FreeUtil::timeFromId)
-            .map(createdTime -> createdTime.isBefore(FreeUtil.inactiveTimeLimit()))
-            .orElse(true); // if no channel history could be fetched assume channel is free
+            .filter(freeConfig -> freeConfig.getMonitoredChannels().contains(channel.getIdLong()))
+            .findAny()
+            .orElseThrow();
+
+        return TimeUtil.getTimeCreated(maybeLastMessageId.orElseThrow())
+            .toInstant()
+            .isBefore(when.minus(configForChannel.getInactiveChannelDuration()));
     }
 
     /**
      * This method sets the channel's status to 'busy' see {@link ChannelStatus#setBusy(long)} for
      * details.
-     * 
+     *
      * @param channelId the id for the channel status to modify.
      * @param userId the id of the user changing the status to busy.
      * @throws IllegalArgumentException if the channel passed is not monitored. See
@@ -149,7 +169,7 @@ final class ChannelMonitor {
     /**
      * This method sets the channel's status to 'free', see {@link ChannelStatus#setFree()} for
      * details.
-     * 
+     *
      * @param channelId the id for the channel status to modify.
      * @throws IllegalArgumentException if the channel passed is not monitored. See
      *         {@link #addChannelToMonitor(long)}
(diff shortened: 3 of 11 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
```diff
@@ -281,6 +293,97 @@ public void updateStatusFor(@NotNull Guild guild) {
...
+        Consumer<Message> editStatus = message -> deleteIfNotLatest(message)
+            .ifPresentOrElse(ignored -> message.editMessageEmbeds(embed).queue(), sendStatus);
+
+        if (!channelIdToMessageIdForStatus.containsKey(channel.getIdLong())) {
+            findExistingStatusMessage(channel).ifPresentOrElse(editStatus, sendStatus);
+        } else {
+            Optional.ofNullable(channelIdToMessageIdForStatus.get(channel.getIdLong()))
+                .map(channel::retrieveMessageById)
+                .ifPresentOrElse(restAction -> restAction.queue(editStatus), sendStatus);
+        }
```
Tais993, line 333: This is really bad read, there are too many "branches"
Also besides of the lambda variables it;s getting even harder, take a look into merging the optional's, or something else to improve readability

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getTextChannels()
+            .stream()
+            .filter(channel -> channelMonitor.isMonitoringChannel(channel.getIdLong())
+                    && channelMonitor.isChannelBusy(channel.getIdLong()))
+            .forEach(channel -> channel.retrieveMessageById(channel.getLatestMessageId())
+                .queue(latestMessage -> {
```
Tais993, line 36: and I don't know, but if possible look for a way to refactor this?
This is also very hard to read

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
MaiTheLord: Actually, I did it like this to make sure there's no duplicate code, and also, I got rid of the `complete` that was in the original code, so now this is not blocking at all. I'll try to make it more readable though.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: can you be a bit more specific? do you suggest to get the channels first, put them in a some kind of list, then iterate over them, or something very different?

Tais993: 2 points you could take a look at

- getLatestMessageId return's the last message **send** their ID, so if a message gets removed, it's still that ID, so this will error
- Take a look at TimeUtil#getTimeCreated, an ID contains the creation date already

application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
MaiTheLord: 6e27d4f

Tais993: You can undo that commit and keep it like it used to be for now, maybe someone else has suggestions

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: 24331a6c1f9f6796f890c40ba5e820cacb2c3b44

application/src/main/java/org/togetherjava/tjbot/commands/Features.java
```diff
@@ -52,6 +57,7 @@
         TagSystem tagSystem = new TagSystem(database);
         ModerationActionsStore actionsStore = new ModerationActionsStore(database);
         ModAuditLogWriter modAuditLogWriter = new ModAuditLogWriter(config);
+        ChannelMonitor channelMonitor = new ChannelMonitor();
```
Zabuzard, line 60: if this class escapes its package now and goes into such a prominent place, it has to be renamed so that its context is clear. `FreeChannelMonitor` for example.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,56 @@
...
+ * Routine that automatically marks inactive help channels as free.
+ * <p>
+ * The routine runs every 10 minutes, and checks for help channels that have been inactive for more
+ * than 2 hours.
+ * </p>
+ */
+public class FreeHelpChannelsRoutine implements Routine {
+    private final ChannelMonitor channelMonitor;
+
+
```
Zabuzard, line 21: two newlines

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getTextChannels()
+            .stream()
+            .filter(channel -> channelMonitor.isMonitoringChannel(channel.getIdLong()))
+            .filter(channel -> channelMonitor.isChannelBusy(channel.getIdLong()))
+            .forEach(channel -> {
+                if (TimeUtil.getTimeCreated(channel.getLatestMessageIdLong())
+                    .isBefore(OffsetDateTime.now().minusHours(2))) {
```
Zabuzard, line 39: move the 2 into a constant. its too easy to overlook otherwise

application/src/main/java/org/togetherjava/tjbot/commands/free/ChannelMonitor.java
MaiTheLord: you say its worse now? I think it makes the branching way more understandable.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,56 @@
...
+import org.togetherjava.tjbot.commands.Routine;
+
+import java.time.OffsetDateTime;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Routine that automatically marks inactive help channels as free.
+ * <p>
+ * The routine runs every 10 minutes, and checks for help channels that have been inactive for more
+ * than 2 hours.
```
Zabuzard, line 15: To prevent the doc from going out-of-date with the actual numbers, you can move both numbers into constants and then reference them here using `for more than {@value Foo} hours`. It will susbtitute the vlaue in the text then.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
Tais993: Mhm, where did you learn that from? :p

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getTextChannels()
+            .stream()
+            .filter(channel -> freeChannelMonitor.isMonitoringChannel(channel.getIdLong()))
+            .filter(channel -> freeChannelMonitor.isChannelBusy(channel.getIdLong()))
+            .forEach(channel -> {
+                if (TimeUtil.getTimeCreated(channel.getLatestMessageIdLong())
+                    .isBefore(OffsetDateTime.now().minusHours(2))) {
```
Tais993: use filter instead

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,56 @@
...
+    public FreeHelpChannelsRoutine(@NotNull FreeChannelMonitor freeChannelMonitor) {
+        this.freeChannelMonitor = freeChannelMonitor;
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getTextChannels()
+            .stream()
+            .filter(channel -> freeChannelMonitor.isMonitoringChannel(channel.getIdLong()))
+            .filter(channel -> freeChannelMonitor.isChannelBusy(channel.getIdLong()))
```
Tais993, line 38: I don't know what your opinion is on adding overloads that take a TextChannel?
Would make the filters a lot more readable, and wouldn't add a lot more code.

Could also create a private method

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: I am not sure what do you mean by adding overloads.

Tais993: You do not know what an overload is?

Imagine you having a print method:

```java
void print(String)
void print(Integer)
```

These are overloads, overloads are multiple methods with the same name.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: done.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: I know what's method overloading. I didn't understand what methods did you suggest to overload :)

Tais993: Should have made that clear the next time "what do you mean by adding overloads" sounds like you don't know what an overload is.
And see the message above, I'm explaining this case there

MaiTheLord: Yea, I saw that.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getTextChannels()
+            .stream()
+            .filter(freeChannelMonitor::isMonitoringChannel)
+            .filter(freeChannelMonitor::isChannelBusy)
+            .filter(channel -> TimeUtil.getTimeCreated(channel.getLatestMessageIdLong())
+                .isBefore(OffsetDateTime.now().minusMinutes(INACTIVE_TIME_MINUTES)))
```
Tais993, line 40: Create private utility method for this?

Tais993 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
```diff
@@ -105,6 +130,18 @@ public boolean isChannelBusy(final long channelId) {
         return requiresIsMonitored(channelId).isBusy();
     }
 
+    /**
+     * This method tests if channel status to busy, see {@link ChannelStatus#isBusy()} for details.
```
Zabuzard, line 134: "tests if channel status **to** busy" is incorrect english. probably meant "is busy".

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
```diff
@@ -25,14 +28,23 @@
  * After successful configuration, any changes in busy/free status will automatically be displayed
  * in the configured {@code Status Channel} for that guild.
  */
-final class ChannelMonitor {
+public final class FreeChannelMonitor {
```
Zabuzard, line 31: The monitor class is totally overloaded now. It was meant to only monitor channels (see its current Javadoc which u also forgot to adjust accordingly). But now its essentially the core business logic. It can not only monitor channels but also change status, build messages, send messages, find history, ...

I would advise to undo all changes on this class and introduce a `FreeSystem` or `FreeChannelSystem` class instead (similar to how `TagCommand` and similar use a `TagSystem` class), where you have all those utility methods. And then you would pass around the FreeSystem, not the monitor, to whoever is interested.

Also reads better in `Features` when you pass around a FooSystem rather than a monitor.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Routine that automatically marks inactive help channels as free.
+ * <p>
+ * The routine runs every {@value CHECK_INTERVAL_MINUTES} minutes, and checks for help channels that
+ * have been inactive for more than {@value INACTIVE_TIME_MINUTES} minutes.
+ * </p>
+ */
+public class FreeHelpChannelsRoutine implements Routine {
```
Zabuzard, line 19: Bad class name. It should be called according to its purpose. So sth like `FreeInactiveHelpChannelsRoutine`. I.e. the emphasis is on the fact that it operates on inactive channels.

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final int INACTIVE_TIME_MINUTES = 120;
```
Zabuzard, line 20: i would rename this to `INACTIVE_AFTER_MINUTES`. Or even better, dont use `int` but `Duration`: `Duration INACTIVE_AFTER = Duration.ofHours(2)`

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final int CHECK_INTERVAL_MINUTES = 10;
```
Zabuzard, line 21: its not really obvious based on the name that this controls the routine schedule. maybe rename to SCHEDULE_INTERVAL_MINUTES. or RUN_EVERY_MINUTES.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeHelpChannelsRoutine.java
MaiTheLord: how do I show it in the javadoc then?

Tais993: Not, `{@value}` only works with primitives and String's.
Can replace it with a `{@link}`?

MaiTheLord: probably, but that would make it less readable.

Tais993: And the advantage is superior maintainability, instead of having to check a lot of documentation for hardcoded values it just works.

The disadvantage of readability isn't even that big, it's very small

borgrel:
u do realise that all of the logic for this has already been written?? (its currently set to an hr but thats easy to change)

all u need to do is create a routine, give it a reference to to free channel and call `checkBusyStatusAllChannels(jda)`

so its litterally like 5 lines of code

new Routine().setAction(10 minutes, freecommand -> checkBusyStatusAllChannels)

thats all thats needed

Zabuzard:
Cool. Nice to see how well you prepared for this feature already. Unfortunate, that we overlooked it 😆

MaiTheLord:
how do you suggest we send a message then? or maybe we don't?

borgrel:
> how do you suggest we send a message then? or maybe we don't?

@MaiTheLord the purpose of programming is NOT to do each and every tiny thing that needs doing to perform a task as part of that task each and every time u want the task done .... this leads to inconsistency, bugs, unmaintainable code, no chain or responsibility and just about every bad practice that exists for programming!

the purpose is to group all actions required for a task in a single place _so that u never have to do each tiny thing manually_

since the channel should NEVER be free'd without users getting a message telling them that it is free .... u dont have to do it urself because its already written as part of the code to free a channel

borgrel:
it calls other methods and when it does (eventually) change the status of a channel, the method that is called to do so ALSO sends a channel message to notify users

thats the point ... the checkStatus method IS NOT RESPONSIBLE for changing the status of a channel, so it SHOULD NOT be responsible to sending a message that states the channel is free

tasks get grouped and called in a logical structured manner so that things like sending a message when the status changes does not get forgotten

by expecting checkStatus to send messages is the same as saying "my apartment building is so cool its got a doorman to open the doors for visitors. If u want to visit just call 5 mins before so i go down and unlock the door so he can let u in" ..... whats the _point_ of the doorman if ur still doing all of the work urself

borgrel:
checking .......

```
    @Override
    public void onSlashCommand(@NotNull final SlashCommandEvent event) {
        ......
        channelMonitor.setChannelFree(id);
        displayStatus(channelMonitor.getStatusChannelFor(requiresGuild(event)));
        event.reply(UserStrings.MARK_AS_FREE.message()).queue();
    }
```

oh bleh :(
its because a slash command is an interaction so the reply message is confined to replying to an interaction, which is a fundamental (bad) restriction of the api.

So yeah, more than just 5 lines of code, but not a lot more.

That still doesn't change the point that the routine _should not_ be sending messages to tell that the the channel is free. That is outside the scope of its reponsibility (which is to keep track of time and get the free system to update itself at a certain interval)

Rewriting all of the code that is basically already found in free command for the routine was not ideal.

.
.
TL:DR
i would suggest seeing if the bot can trigger slash commands itself (like a regular user) and change the updateStatusFor to trigger new `/free` SlashCommandEvents for the channels that need updating (i dont know if its possible)

if not then u just move the action outside of the slashcommand

```
    public void freeChannel(TextChannel channel, @Nullable Interaction event) {
        channelMonitor.setChannelFree(channel.getIdLong());
        final String message = UserStrings.MARK_AS_FREE.message();

        if (event != null) {
            displayStatus(channelMonitor.getStatusChannelFor(requiresGuild(event)));
            event.reply(message).queue();
        } else {
            channel.sendMessage(message).queue();
        }
   }
```
The point is keep the message sending all in one place instead of having some in free command, some in routine, and some who knows where else.

You will never debug it if the same thing happens in more than one place, only one of them is making a problem and you dont know which one.

borgrel:
dont use peek, its **_strictly !!!_** for debugging

it can be short circuited and all sorts of other nasty side effects.

u have identified the common method though, well done
change the setChannelFree methods' signature to accept an interaction
if the interaction is null send a message
if the interaction exists, reply with the same message

marko-radosavljevic:
@MaiTheLord Hey, are you still working on this PR? ^^

MaiTheLord:
I was busy lately (and still am), so not really.

marko-radosavljevic:
I see, thanks for the update. :heart: 

You wouldn't mind if someone takes over?

MaiTheLord:
I wouldn't mind.
Thanks for asking though.

Zabuzard (dismissed):

Zabuzard:
Taken over, rewritten based on the previous general approach by @MaiTheLord .

MaiTheLord:
Looks good. can't approve as I originally opened the PR

application/src/main/java/org/togetherjava/tjbot/commands/free/AutoFreeRoutine.java
```diff
@@ -0,0 +1,72 @@
...
+        inactiveChannels.forEach(inactiveChannel -> inactiveChannel
+            .sendMessage(UserStrings.AUTO_MARK_AS_FREE.message())
+            .queue());
+    }
+
+    private static boolean isInactive(@NotNull MessageChannel channel, @NotNull Instant when) {
+        return TimeUtil.getTimeCreated(channel.getLatestMessageIdLong())
+            .toInstant()
+            .isBefore(when.minus(INACTIVE_CHANNEL_DURATION));
+    }
```
Tais993, line 66: Why not just FreeChannelMonitor#updateStatusFor?

Zabuzard: Not sure what you mean. I wanted to process the channels grouped by guild to prevent issues. For example, it is more important that the channels are marked free and that the status message is updated than to send the actual "channel is free" message.

Because of that, I split those steps. So if the last step (sending messages) fails for some reason, the system is at least in a sane state.

Zabuzard: Also, I did not want other guilds to interfer with the process. If a guild fails, it shouldnt influence another guild.

Zabuzard: Dont forget that `channelMonitor.displayStatus(guild);` is an operation that should happen once per guild, not per channel. The rest of the operations are indeed per-channel only, but this one is per-guild.

Tais993: Once I get home I'll check.

But isn't this code just a very long & selfmade version of updateStatusFor + forEach of all guilds?

Pretty sure updateStatusFor checks each channel, once inactive sets it to free/available?

Zabuzard: oh... for fucks sake. All the logic was already implemented as dead-code... I just saw this lmao... 😆 FFS

Tais993: If you read this issue you saw that borgrel made a comment about it lol

Zabuzard: yeah i know, i didnt read it 😆

Zabuzard: done. please check again.

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
```diff
@@ -232,26 +264,30 @@ public String statusMessage(@NotNull final Guild guild) {
...
+     * and determines if the last time it was updated is more recent than the configured time. If so
+     * it changes the channel's status to free, see
+     * {@link FreeChannelMonitor#isChannelInactive(TextChannel, Instant)}.
      * <p>
-     * This method is run automatically during startup and should be run on a set schedule, as
-     * defined in {@link org.togetherjava.tjbot.config.FreeCommandConfig}. The scheduled execution
-     * is not currently implemented
-     * 
+     * This method is run automatically during startup and on a set schedule, as defined in
+     * {@link org.togetherjava.tjbot.config.FreeCommandConfig}.
```
Tais993, line 272:
```suggestion
     * {@link org.togetherjava.tjbot.config.FreeCommandConfig FreeCommandConfig}.
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/free/FreeChannelMonitor.java
Zabuzard: done, thanks.

Zabuzard (approved):

PR #422, opened 2022-03-29
Made remindme command's implementation global to MessageChannel

### LINKED ISSUE #405: /remind doesnt work in threads (NPE)
## Overview

The `/remind` command can not send a reminder that was set off from within a Discord-Thread. Worse, it crashes with a NPE:

log

## Proposed action

First of all, it is important to get rid of the error.

Either, we find a way to make it work inside threads (maybe there are some sort of thread ids?), or we just actively try to detect whether it is inside a thread (e.g. the thing is null, as the error says) and respond with something like "sorry, can not remind you from within a thread".

Tais993:
In the current implementation of the remindme command the code is meant so it only works in TextChannel's.
The type has been adjusted to MessageChannel to support newschannels, threads, and technically DM's aswell.

Fixes #405


Zabuzard (approved):
does it work now? if so, approved

Tais993:
I'm 100% sure that it works, but I haven't had the chance to test it yet.

Zabuzard:
@Tais993 does it work? shall we push?

Zabuzard:
Tested, works:

thread test

PR #296, opened 2021-12-03
Logging the /tag-manage command

MaiTheLord:
## What's the change?
`/tag-manage` actions are now logged on `#mod_audit_log`.
## Why is it necessary?
It's enhancing the command `/tag-manage`. solving issue #281 
## What exactly did I add?
When executing the command successfully, the bot sends an embed on `#mod_audit_log` that's showing info about the executed command and the user who used it.
## This is how it looks like:
image

PR #431, opened 2022-04-06
Added missing Javadoc for public return value

### LINKED ISSUE #430: Sonar - Missing Javadoc for public return value
## Overview

Our Sonar is currently failing due to a missing Javadoc for a return value of a public method:

report

Direct link: https://sonarcloud.io/project/issues?issues=AX-Zl-AX7wbRsH2qYsgw&open=AX-Zl-AX7wbRsH2qYsgw&id=Together-Java_TJ-Bot

This was originally merged by #296. But nobody noticed back then, since our Sonar config was incorrect until recently.

Zabuzard:
## Overview

Closes #430.

Our Sonar is currently failing due to a missing Javadoc for a return value of a public method:

report

This was originally merged by #296. But nobody noticed back then, since our Sonar config was incorrect until recently.

This fixes the issue by adding the Javadoc, easy money.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/moderation/ModAuditLogWriter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/moderation/ModAuditLogWriter.java
@@ -83,6 +83,7 @@ public final class ModAuditLogWriter {
      * this method will return an empty optional, and a warning message will be written.
      *
      * @param guild the guild to look for the channel in
+     * @return the channel used for moderation audit logs, if present
      */
     public Optional<TextChannel> getAndHandleModAuditLogChannel(@NotNull Guild guild) {
         Optional<TextChannel> auditLogChannel = guild.getTextChannelCache()


Tais993:
When has our sonarcloud config been changed then?

Zabuzard:
> When has our sonarcloud config been changed then?

I changed it maybe a 2-4 weeks ago. It used `master` as **main branch** instead of `develop` and the quality gate was missing a rule to fail for "critical issues" (such as the public javadoc rule). It only had security issues as reason to fail the gate.

marko-radosavljevic (approved):

application/src/main/java/org/togetherjava/tjbot/moderation/ModAuditLogWriter.java
```diff
@@ -83,6 +83,7 @@ public void write(@NotNull String title, @NotNull String description, @NotNull U
      * this method will return an empty optional, and a warning message will be written.
      *
      * @param guild the guild to look for the channel in
+     * @return the channel used for moderation audit logs, if present
```
Tais993, line 86: > the channel used for moderation audit logs, if present

Sounds like

> the channel, or null when not present.

But actually it returns an Optional of the channel, maybe it has to do with the way I read

Zabuzard: Well, I guess the return **type** is obvious from the signature. I usually write it the same way as if it would be null, since the Javadoc operates a level higher than Javas typesystem. Like, it doesnt care how its represented 🤷

Tais993: Disagree with that, don't like it.
I'd rather read 

```java
@return An {@link Optional} of {@link TextChannel} which is the channel used for moderation audit logs.
```

Tais993: You are documenting the method, so it should reflect the method?

I don't understand this and sadly never noticed it, but if no-one else cares we can keep it like this ig

PR #446, opened 2022-05-27
Top helpers based on message length

Zabuzard:
## Overview

Since determining top helpers on message count is indeed quite unfair to helpers who prefer to type out their messages instead of posting multiple mini messages, this changes the metric to measure and determine them based on the **length of the messages**.

Not perfect, but gets the job done and is fairer.

example

Existing data stays untouched. We could revert back at any time. Also, for previous months, the new code can still output the same results as before (by defaulting to message length 1 for migrated content).

## When to release

Note that this should ideally be released near the end or start of a month, to not mess up the metric for that month. It should definitely not be released in the middle of a month. Alternatively, it can of course still be manually retrieved from the database with the old query.


Zabuzard:
Merging this early since its low risk and I want to have it deployed at 1st of June (see PR description).

PR #459, opened 2022-06-21
Added /change-help-title command

Zabuzard:
## Overview

This adds the `/change-help-title` command, upon request. Pretty straightforward:

example

While at it, the code for dealing with categories and titles has been simplified a bit by introducing a helper record.

Also, changed the cooldown of other commands to 30mins as well.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
@@ -1,0 +1,105 @@
+package org.togetherjava.tjbot.commands.help;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.ThreadChannel;
+import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
+import net.dv8tion.jda.api.interactions.InteractionHook;
+import net.dv8tion.jda.api.interactions.commands.OptionType;
+import net.dv8tion.jda.api.requests.RestAction;
+import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
```diff
@@ -0,0 +1,105 @@
...
+                .reply("Please wait a bit, this command can only be used once per %d %s.".formatted(
+                        COOLDOWN_DURATION_VALUE,
+                        COOLDOWN_DURATION_UNIT.toString().toLowerCase(Locale.US)))
+                .setEphemeral(true)
+                .queue();
+            return;
+        }
+        helpThreadIdToLastTitleChange.put(helpThread.getIdLong(), Instant.now());
+
+        event.deferReply().queue();
```
Tais993, line 82: Why first deferring? This shouldn't take a few minutes, this is more of a 50ms task

Zabuzard: Maybe a leftover from copypasta of the change-category command, where we piped more commands. Ill take a look.

Zabuzard:
> What's the reasoning for this change? It can be abused and trigger rate limits?

I reduced it from 60mins to 30mins. Cause people, like Ala, complained. And Tijs said that the rate limits should be able to handle it still.

But yes, the reason for having the cooldown in the first place are rate limits. The UX is terrible if you hit the rate limit and we have no way of knowing when we hit it. JDA just sends out ur request but Discord does not process it yet. Its not like ur method crashes with an exception or anything.

So I am trying to prevent them from happening in the first place, for regular usage.

Tais993:
> Its not like ur method crashes with an exception or anything.

https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/requests/RestAction.html#timeout(long,java.util.concurrent.TimeUnit)

It'll fail with a [TimeoutException](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeoutException.html), not sure or this goes in the failure consumer or just throws.

If we timeout to like, 20s we know for certain whenever we got ratelimited (unless Discord servers crashed)

PR #464, opened 2022-07-07
Bugfix with temp actions being revoked too early

Zabuzard:
## Overview

Fixes a bug with temporary moderation actions being revoked too early in certain edge cases.

Basically, the issue occurs if a user already had a temporary action in the past, which was also revoked already. The next temporary action will then be revoked too early, since the code picks up the old expired action and figures _"oh, lets revoke it"_.

This introduces a small fix where the code will always look at the last apply-action first and only proceeds revoking if that one has already expired.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
@@ -81,13 +81,14 @@ public final class TemporaryModerationRoutine implements Routine {
 
     private void processGroupedActions(@NotNull RevocationGroupIdentifier groupIdentifier) {
         // Do not revoke an action which was overwritten by a permanent action that was issued
-        // afterwards
+        // afterwards, or if the last action has not even expired yet
         // For example if a user was perm-banned after being temp-banned
         ActionRecord lastApplyAction = actionsStore
             .findLastActionAgainstTargetByType(groupIdentifier.guildId, groupIdentifier.targetId,
                     groupIdentifier.type)
             .orElseThrow();
-        if (lastApplyAction.actionExpiresAt() == null) {
+        if (lastApplyAction.actionExpiresAt() == null
+                || lastApplyAction.actionExpiresAt().isAfter(Instant.now())) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


Tais993 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
```diff
@@ -81,13 +81,14 @@ private void checkExpiredActions() {
 
     private void processGroupedActions(@NotNull RevocationGroupIdentifier groupIdentifier) {
         // Do not revoke an action which was overwritten by a permanent action that was issued
-        // afterwards
+        // afterwards, or if the last action has not even expired yet
         // For example if a user was perm-banned after being temp-banned
         ActionRecord lastApplyAction = actionsStore
             .findLastActionAgainstTargetByType(groupIdentifier.guildId, groupIdentifier.targetId,
                     groupIdentifier.type)
             .orElseThrow();
```
java-coding-prodigy, line 89:
```suggestion
        // For example if a user was perm-banned after being temp-banned
        Instant actionExpiresAt = actionsStore
            .findLastActionAgainstTargetByType(groupIdentifier.guildId, groupIdentifier.targetId,
                    groupIdentifier.type)
            .orElseThrow()
            .actionExpiresAt();
            if (actionExpiresAt == null
                 || actionExpiresAt.isAfter(Instant.now())) {
```

java-coding-prodigy (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
Zabuzard: You are overlooking that its also used down below:

usage

Zabuzard:
Tested, works:

logs

application/src/main/java/org/togetherjava/tjbot/commands/moderation/temp/TemporaryModerationRoutine.java
java-coding-prodigy: Oh, I missed that.
But maybe we can have a method in `ActionRecord` which checks whether it has expired, and then call that method itself. Since we need to check that more than once.

Zabuzard: Thats fair, I guess. Ill add such a method.

Zabuzard: done. please check again, thanks.

java-coding-prodigy (approved):
LGTM

PR #465, opened 2022-07-09
Bugfix when creating help threads, no bot-messages

Zabuzard:
## Overview

This is an attempt to fix the current, frequent, bug where help threads are missing the bot explanation messages.

## Explanation

They are caused by an exception:

exception

The source of this is most likely that `overviewChannel.createThreadChannel(...)` is rate limitted and hence executed a few seconds later. At which point the subsequent `event.refer(...)` crashes with **Unknown interaction** cause it is past the 3 seconds threshold. The extra messages are then missing, cause they are `flatMap`'ed right after the failing `notifyUser(...)`:

code

## Solution

The solution introduced in this PR is to simply make the whole interaction **deferred**. That way, the code has more time to respond, enough to fit in a delayed thread creation - hopefully.

There are also other alternatives available, such as changing the order of the statements or simply not letting the chain collapse on a single failure. But I would like to try out this fix first.


illuminator3 (approved):
LGTM

Zabuzard:
Week old, one approval, merging.

PR #466, opened 2022-07-09
Bugfix with /change_help_title allowing too long titles

Zabuzard:
## Overview

This addresses two bugs with long titles:

1. the `/change_help_title` had no verification on title length and hence had a really bad UX for too long titles with exceptions and no user response
2. the max title length in our verification was `80`, which is too long and exceeds the true limit of `100` when long categories are added, such as `[Together Java Bot]`

Both issues have been fixed with this PR.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
@@ -4,6 +4,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import net.dv8tion.jda.api.entities.ThreadChannel;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
+import net.dv8tion.jda.api.interactions.callbacks.IReplyCallback;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


marko-radosavljevic (approved):

application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
```diff
@@ -57,6 +61,9 @@ public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
         if (!helper.handleIsHelpThread(event)) {
             return;
         }
+        if (!handleIsValidTitle(title, event)) {
+            return;
+        }
```
Tanish-Azad, line 66: Both if statements do the same thing should reduce it to
```java
if (!helper.handleIsHelpThread(event) || !handleIsValidTitle(title, event)) return;
```

Tanish-Azad (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
Zabuzard: I personally like it separated to make it clear that its actually different topics. But your suggestion is fair, will do.

Zabuzard: Done, thanks.

Zabuzard:
Week old, one approval, merging

PR #469, opened 2022-07-19
Auto create threads for suggestions

### LINKED ISSUE #468: Auto creation of threads in suggestion channel
Right now the bot adds up and down reaction on every message also on messages where someone created a thread. Ex:-
image

Let the bot create a thread for each suggestion and rest discussion will go inside it

**Describe alternatives you've considered**
Analyse the message if it is an thread created message then don't add reactions to it

Tanish-Azad:
Auto create thread for suggestions in suggestions channel.

Closes #468 .

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -40,16 +40,18 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
 
     @Override
     public void onMessageReceived(@NotNull MessageReceivedEvent event) {
-        if (event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild()) {
+        if (event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild()
+                || event.getChannel().getType().isThread()) {
             return;
         }
 
         Guild guild = event.getGuild();
         Message message = event.getMessage();
-        
-        if (!message.getChannel().getType().isThread()) {
-            message.createThreadChannel("Discussion for " + message.getAuthor().getName() + "'s suggestion").queue();
-        }
+
+        message
+            .createThreadChannel(
+                    "Discussion for " + message.getAuthor().getName() + "'s suggestion")
+            .queue();
 
         reactWith(config.getUpVoteEmoteName(), FALLBACK_UP_VOTE, guild, message);
         reactWith(config.getDownVoteEmoteName(), FALLBACK_DOWN_VOTE, guild, message);
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -46,6 +46,10 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
 
         Guild guild = event.getGuild();
         Message message = event.getMessage();
+        
+        if (!message.getChannel().getType().isThread()) {
+            message.createThreadChannel("Discussion for " + message.getAuthor().getName() + "'s suggestion").queue();
+        }
 
         reactWith(config.getUpVoteEmoteName(), FALLBACK_UP_VOTE, guild, message);
         reactWith(config.getDownVoteEmoteName(), FALLBACK_DOWN_VOTE, guild, message);
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -40,8 +40,7 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
 
     @Override
     public void onMessageReceived(@NotNull MessageReceivedEvent event) {
-        if (event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild()
-                || event.getChannel().getType().isThread()) {
+        if (event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild()) {
             return;
         }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -53,7 +53,7 @@ public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
         reactWith(config.getDownVoteEmoteName(), FALLBACK_DOWN_VOTE, guild, message);
     }
 
-    private static void createThread(Message message) {
+    private static void createThread(@NotNull Message message) {
         message
             .createThreadChannel(
                     "Discussion for " + message.getAuthor().getName() + "'s suggestion")
--- a/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
@@ -22,6 +22,7 @@ import java.util.regex.Pattern;
  */
 public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
     private static final Logger logger = LoggerFactory.getLogger(SuggestionsUpDownVoter.class);
+    private static final int TITLE_MAX_LENGTH = 60;
     private static final String FALLBACK_UP_VOTE = "👍";
     private static final String FALLBACK_DOWN_VOTE = "👎";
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -46,6 +46,10 @@ public void onMessageReceived(@NotNull MessageReceivedEvent event) {
 
         Guild guild = event.getGuild();
         Message message = event.getMessage();
+        
+        if (!message.getChannel().getType().isThread()) {
+            message.createThreadChannel("Discussion for " + message.getAuthor().getName() + "'s suggestion").queue();
+        }
```
Zabuzard, line 52: Have you tested this locally in your test server already? This appears to be wrong. If I read it right, it says "if someone created a thread, lets create another thread".

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Zabuzard: Nvm. I read it wrong. I would still be interested in how it looks in practice. Can you post a screenshot?

Tanish-Azad: when the bot creates a thread it does not sends a thread creation message
https://drive.google.com/file/d/1-JGBDCSbo2IYxgn-PKY8szrBlFTbMrcZ/view?usp=sharing
the testing bot have admin perms

Zabuzard: Thats good to know, but could you please send a screenshot of this piece of code, your current implementation, running live? It looks like you just created another mini project to showcase it, instead of actually trying out the real code with our real project.

Zabuzard: Also, the code is failing the spotless check right now.

Tanish-Azad: image
This is from config file:
```json
"suggestions": {
    "channelPattern": "suggestions",
    "upVoteEmoteName": "thumbsup",
    "downVoteEmoteName": "thumbsdown"
  }
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -46,6 +46,10 @@ public void onMessageReceived(@NotNull MessageReceivedEvent event) {
 
         Guild guild = event.getGuild();
         Message message = event.getMessage();
+        
+        if (!message.getChannel().getType().isThread()) {
```
Zabuzard, line 50: This `if` should not only affect thread creation but also the emoji-reactions. All in all, it should basically affect the whole method. Please move it up into the general checks at the top, i.e. the initial `if` that `return`s.

Tanish-Azad:
what to do about the spotless check?

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
```diff
@@ -46,6 +46,10 @@ public void onMessageReceived(@NotNull MessageReceivedEvent event) {
...
+            message.createThreadChannel("Discussion for " + message.getAuthor().getName() + "'s suggestion").queue();
```
Zabuzard, line 51: Please put it into a method, such as `createDiscussionThread(...);`

Ideally the code would then look like:
```java
if (...) {
  return;
}

...

createDiscussionThread(...);
reactWith(...);
reactWith(...);
```

Zabuzard (changes requested):
Also, the code is not passing spotless.

Zabuzard:
> what to do about the spotless check?

If you have set up the code locally and use the gradle commands, it would run spotless already for you. See our wiki for details on that.

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Tanish-Azad: ok 👍🏻

Tanish-Azad:
```java
        if ((event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild()) && !event.getChannel().getType().isThread()) {
            return;
        }
```
like this?

Tanish-Azad: got it
```java
        if (event.getAuthor().isBot() || event.isWebhookMessage() || !event.isFromGuild() || event.getChannel().getType().isThread()) {
            return;
        }
```

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Tanish-Azad: done

Tanish-Azad:
what the thread title should be "Discussion on User's suggestion" or "Discussion for User's suggestion" or something else

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Zabuzard: Better, but still not exactly what I was looking for. You are still checking whether the message is a thread-message. But instead, you should just look for "not ordinary message". I.e. the code should only react on ordinary text messages, not on any other fancy type. Right now you only exclude thread-types but what about all other unwanted types.

Zabuzard:
> what the thread title should be "Discussion on User's suggestion" or "Discussion for User's suggestion" or something else

I dont care. Imo it is fine as u did it :)

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Tanish-Azad: i don't see any other type of message is suggestions channel there is only the actual suggestions and thread creation messages, we already handled the thread creation message. What is left now?

Zabuzard: See, your logic is just the reverse. You do not actually want to check "is not a thread channel" but instead "is a message/guild/text channel" or whatever that channel is. I.e. something that exclusively identifies the channel you are interested in.

https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/ChannelType.html

Tanish-Azad: > is a message/guild/text channel

it is a message because MessageReceivedEvent only happens when it is a message
it is a guild channel because we already have the condition to check that inside the if statement
it is a text channel because messages can only be sent in a text channel + we know that suggestions channel is going to be a text channel

so now all we need to check is that it is not a thread inside the suggestions channel

hope that makes sense

Zabuzard: Sure that this can even happen? Due to the name-pattern it only listens to messages in this particular channel matching that name anyways.

Zabuzard: I just tried it out. It seems that the condition is obsolete. Thread messages do not enter the `onMessageReceived` due to the channel name pattern.

Zabuzard (changes requested):
This is a bit more tricky, but currently the UX suffers from the generic thread names:

issue

I think we can not release it like that and need some way to distinguish them.

I would suggest we determine the title based on the first x characters from the message. It is not perfect, but enough to make it clear what the thread is about. So in this case `Discussion for 'I have a great idea'` or just `I have a great idea` without the fuzz around it.

Code-wise, it could look something like `message.getContent()` and then `substring(0, Math.min(MAX_LENGTH, content.length())`. Where the max length could be something around 60 maybe.

If you want, you can also make the substring smarter by cutting off at the last "word" to not have something like `I have a gre` where its cut in the middle of the word. The code for something like already exists in `ImplicitAskListener#createTitle` where you can get some inspiration, if you want to.

application/src/main/java/org/togetherjava/tjbot/commands/basic/SuggestionsUpDownVoter.java
Tanish-Azad: hmm yes that can be the case i didn't though of that, so should i remove the if condition and test it again?

Zabuzard: yeah, thanks

Zabuzard:
> image is this better?

Yes, but you have to ensure that its within the limit of thread names. So limit it to 60 characters for example. And if you do not want to cut it off in between a word, use a logic as shown earlier.

Tanish-Azad:
yes, i did
```java
    private static void createThread(@NotNull Message message) {
        String messageContent = message.getContentRaw();

        String title = messageContent;
        if (messageContent.length() >= TITLE_MAX_LENGTH) {
            int lastWordEnd = messageContent.lastIndexOf(' ', TITLE_MAX_LENGTH);

            if (lastWordEnd == -1) {
                lastWordEnd = TITLE_MAX_LENGTH;
            }

            title = messageContent.substring(0, lastWordEnd);
        }

        message
            .createThreadChannel(title)
            .queue();
    }
```

Tanish-Azad:
image

Tanish-Azad:
Any thread created will automatically invite the author and the member who created the thread

Zabuzard (approved):
Seems to work well

tests

Zabuzard:
Waiting for a second approval. Otherwise merging in latest 7 days.

Heatmanofurioso (approved):

PR #472, opened 2022-07-27
Audit rework

### LINKED ISSUE #335: /audit crashes if more than 10 cases
### Overview

When using `/audit` on a user who has more than 10 reported cases, the command crashes with an exception:

> Cannot have more than 10 embeds in a message!

```java
8:02:04.098 [pool-2-thread-4] ERROR net.dv8tion.jda.api.requests.RestAction - RestAction queue returned failure
java.lang.IllegalArgumentException: Cannot have more than 10 embeds in a message!
	at net.dv8tion.jda.internal.utils.Checks.check(Checks.java:48) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.WebhookMessageActionImpl.addEmbeds(WebhookMessageActionImpl.java:138) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.entities.AbstractWebhookClient.sendMessageEmbeds(AbstractWebhookClient.java:67) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.entities.AbstractWebhookClient.sendMessageEmbeds(AbstractWebhookClient.java:40) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at org.togetherjava.tjbot.commands.moderation.AuditCommand.lambda$auditUser$5(AuditCommand.java:159) ~[main/:?]
	at net.dv8tion.jda.internal.requests.restaction.operator.FlatMapRestAction.supply(FlatMapRestAction.java:44) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.FlatMapRestAction.lambda$queue$0(FlatMapRestAction.java:54) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.RestActionOperator.doSuccess(RestActionOperator.java:44) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.lambda$queue$0(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.RestActionOperator.doSuccess(RestActionOperator.java:44) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.lambda$queue$0(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.RestActionOperator.doSuccess(RestActionOperator.java:44) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.CombineRestAction.lambda$queue$6(CombineRestAction.java:136) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.api.utils.MiscUtil.locked(MiscUtil.java:146) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.CombineRestAction.lambda$queue$7(CombineRestAction.java:129) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.RestActionOperator.doSuccess(RestActionOperator.java:44) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.lambda$queue$0(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.DeferredRestAction.queue(DeferredRestAction.java:138) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.queue(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.CombineRestAction.queue(CombineRestAction.java:129) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.queue(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.MapRestAction.queue(MapRestAction.java:41) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.FlatMapRestAction.queue(FlatMapRestAction.java:51) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.api.requests.RestAction.queue(RestAction.java:573) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.api.requests.RestAction.queue(RestAction.java:539) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at org.togetherjava.tjbot.commands.moderation.AuditCommand.auditUser(AuditCommand.java:160) ~[main/:?]
	at org.togetherjava.tjbot.commands.moderation.AuditCommand.onSlashCommand(AuditCommand.java:126) ~[main/:?]
	at org.togetherjava.tjbot.commands.system.CommandSystem.lambda$onSlashCommand$7(CommandSystem.java:121) ~[main/:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
Caused by: net.dv8tion.jda.api.exceptions.ContextException
	at net.dv8tion.jda.api.exceptions.ContextException.here(ContextException.java:54) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.RestActionOperator.contextWrap(RestActionOperator.java:105) ~[JDA-4.4.0_351.jar:4.4.0_351]
	at net.dv8tion.jda.internal.requests.restaction.operator.FlatMapRestAction.queue(FlatMapRestAction.java:50) ~[JDA-4.4.0_351.jar:4.4.0_351]
	... 8 more
```

### Possible solutions

The cause is quite clear, it attempts to display all cases and then fails cause there are too many. Obviously, this case is not supposed to be triggered really often and its not thaaat big of a deal but we should find at least _a solution_.

Dyno has **pagination support** to solve this, so there is an **optional** option on the command to show page 2, page 3, ... Would probably be a clean solution to do it that way ourselves as well.

Taz03:
- New UI for Audit command
- Unlimited audit length (almost)

Closes #335 

[Screencast from 28-07-22 11:45:23 PM IST.webm](https://user-images.githubusercontent.com/73871477/182092384-07fccf33-cbee-4e07-96cc-cbb3ecb93487.webm)

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -40,8 +40,8 @@ public final class AuditCommand extends SlashCommandAdapter {
     private static final String COMMAND_NAME = "audit";
     private static final String ACTION_VERB = "audit";
     private static final int MAX_PAGE_LENGTH = 25;
-    private static final String PREVIOUS_BUTTON = "⬅";
-    private static final String NEXT_BUTTON = "➡";
+    private static final String PREVIOUS_BUTTON_LABEL = "⬅";
+    private static final String NEXT_BUTTON_LABEL = "➡";
     private final Predicate<String> hasRequiredRole;
     private final ModerationActionsStore actionsStore;
 
@@ -196,14 +196,14 @@ public final class AuditCommand extends SlashCommandAdapter {
     private @NotNull ActionRow makeActionRow(long guildId, long targetId, long callerId,
             int pageNumber, int totalPages) {
         int previousButtonTurnPageBy = -1;
-        Button previousButton = createPageTurnButton(PREVIOUS_BUTTON, guildId, targetId, callerId,
+        Button previousButton = createPageTurnButton(PREVIOUS_BUTTON_LABEL, guildId, targetId, callerId,
                 pageNumber, previousButtonTurnPageBy);
         if (pageNumber == 1) {
             previousButton = previousButton.asDisabled();
         }
 
-        int nextButtonTurnPageBy = +1;
-        Button nextButton = createPageTurnButton(NEXT_BUTTON, guildId, targetId, callerId,
+        int nextButtonTurnPageBy = 1;
+        Button nextButton = createPageTurnButton(NEXT_BUTTON_LABEL, guildId, targetId, callerId,
                 pageNumber, nextButtonTurnPageBy);
         if (pageNumber == totalPages) {
             nextButton = nextButton.asDisabled();
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -57,7 +57,7 @@ public final class AuditCommand extends SlashCommandAdapter {
     }
 
     private static @NotNull EmbedBuilder createSummaryEmbed(@NotNull User user,
-                                                            @NotNull Collection<ActionRecord> actions) {
+            @NotNull Collection<ActionRecord> actions) {
         return new EmbedBuilder().setTitle("Audit log of **%s**".formatted(user.getAsTag()))
             .setAuthor(user.getName(), null, user.getAvatarUrl())
             .setDescription(createSummaryMessageDescription(actions))
@@ -68,7 +68,8 @@ public final class AuditCommand extends SlashCommandAdapter {
             @NotNull Collection<ActionRecord> actions) {
         int actionAmount = actions.size();
 
-        String shortSummary = "There are **%s actions** against the user.".formatted(actionAmount == 0 ? "no" : actionAmount);
+        String shortSummary = "There are **%s actions** against the user."
+            .formatted(actionAmount == 0 ? "no" : actionAmount);
 
         // Summary of all actions with their count, like "- Warn: 5", descending
         Map<ModerationAction, Long> actionTypeToCount = actions.stream()
@@ -86,18 +87,22 @@ public final class AuditCommand extends SlashCommandAdapter {
     }
 
     private static @NotNull MessageEmbed.Field actionToField(@NotNull ActionRecord action,
-                                                               @NotNull JDA jda) {
-        Function<Instant, String> formatTime = instant -> TimeUtil.getDateTimeString(instant.atOffset(ZoneOffset.UTC));
+            @NotNull JDA jda) {
+        Function<Instant, String> formatTime = instant -> {
+            if (instant == null)
+                return "";
+            return TimeUtil.getDateTimeString(instant.atOffset(ZoneOffset.UTC));
+        };
 
-        String expiresAt = action.actionExpiresAt() == null ? "" : "\nTemporary action, Expires at: " + formatTime.apply(action.actionExpiresAt());
-
-        User targetUser = jda.getUserById(action.authorId());
+        User author = jda.getUserById(action.authorId());
 
         return new MessageEmbed.Field(
-                action.actionType().name() + " by " + (targetUser == null ? "(unknown user)" : targetUser.getAsTag()),
-                action.reason() + "\nIssued at: " + formatTime.apply(action.issuedAt()) + expiresAt,
-                false
-        );
+                action.actionType().name() + " by "
+                        + (author == null ? "(unknown user)" : author.getAsTag()),
+                action.reason() + "\nIssued at: " + formatTime.apply(action.issuedAt())
+                        + "\nTemporary action, Expires at: "
+                        + formatTime.apply(action.actionExpiresAt()),
+                false);
     }
 
     @Override
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -75,6 +75,10 @@ public final class AuditCommand extends SlashCommandAdapter {
         String shortSummary = "There are **%s actions** against the user."
             .formatted(actionAmount == 0 ? "no" : actionAmount);
 
+        if (actionAmount == 0) {
+            return shortSummary;
+        }
+
         // Summary of all actions with their count, like "- Warn: 5", descending
         Map<ModerationAction, Long> actionTypeToCount = actions.stream()
             .collect(Collectors.groupingBy(ActionRecord::actionType, Collectors.counting()));
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -159,7 +159,8 @@ public final class AuditCommand extends SlashCommandAdapter {
         return groupedActions;
     }
 
-    private @NotNull Message auditUser(long guildId, long targetId, int pageNumber, @NotNull JDA jda) {
+    private @NotNull Message auditUser(long guildId, long targetId, int pageNumber,
+            @NotNull JDA jda) {
         List<ActionRecord> actions = actionsStore.getActionsByTargetAscending(guildId, targetId);
         List<List<ActionRecord>> groupedActions = groupActions(actions);
         int totalPages = groupedActions.size();
@@ -178,17 +179,20 @@ public final class AuditCommand extends SlashCommandAdapter {
             .build();
     }
 
-    private @NotNull ActionRow makeActionRow(long guildId, long targetId, int pageNumber, int totalPages) {
+    private @NotNull ActionRow makeActionRow(long guildId, long targetId, int pageNumber,
+            int totalPages) {
         String stringGuildId = String.valueOf(guildId);
         String stringTargetId = String.valueOf(targetId);
         String stringPageNumber = String.valueOf(pageNumber);
 
-        Button previousButton = Button.primary(generateComponentId(stringGuildId, stringTargetId, stringPageNumber, "-1"), "⬅");
+        Button previousButton = Button.primary(
+                generateComponentId(stringGuildId, stringTargetId, stringPageNumber, "-1"), "⬅");
         if (pageNumber == 1) {
             previousButton = previousButton.asDisabled();
         }
 
-        Button nextButton = Button.primary(generateComponentId(stringGuildId, stringTargetId, stringPageNumber, "1"), "➡");
+        Button nextButton = Button.primary(
+                generateComponentId(stringGuildId, stringTargetId, stringPageNumber, "1"), "➡");
         if (pageNumber == totalPages) {
             nextButton = nextButton.asDisabled();
         }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -101,12 +101,15 @@ public final class AuditCommand extends SlashCommandAdapter {
 
         User author = jda.getUserById(action.authorId());
 
+        Instant expiresAt = action.actionExpiresAt();
+        String expiresAtFormatted = expiresAt == null ? ""
+                : "\nTemporary action, expires at: " + formatTime.apply(expiresAt);
+
         return new MessageEmbed.Field(
                 action.actionType().name() + " by "
                         + (author == null ? "(unknown user)" : author.getAsTag()),
                 action.reason() + "\nIssued at: " + formatTime.apply(action.issuedAt())
-                        + "\nTemporary action, Expires at: "
-                        + formatTime.apply(action.actionExpiresAt()),
+                        + expiresAtFormatted,
                 false);
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -127,7 +127,7 @@ public final class AuditCommand extends SlashCommandAdapter {
             return;
         }
 
-        event.reply(auditMessage(guild.getIdLong(), target.getIdLong(), 1, event.getJDA())).queue();
+        event.reply(auditUser(guild.getIdLong(), target.getIdLong(), 1, event.getJDA())).queue();
     }
 
     @SuppressWarnings("BooleanMethodNameMustStartWithQuestion")
@@ -141,7 +141,7 @@ public final class AuditCommand extends SlashCommandAdapter {
         return ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event);
     }
 
-    private Message auditMessage(long guildId, long targetId, int pageNo, @NotNull JDA jda) {
+    private Message auditUser(long guildId, long targetId, int pageNo, @NotNull JDA jda) {
         List<ActionRecord> actions = actionsStore.getActionsByTargetAscending(guildId, targetId);
 
         List<List<ActionRecord>> groupedActions = new ArrayList<>();
@@ -195,7 +195,7 @@ public final class AuditCommand extends SlashCommandAdapter {
     @Override
     public void onButtonClick(@NotNull ButtonInteractionEvent event, @NotNull List<String> args) {
         event
-            .editMessage(auditMessage(Long.parseLong(args.get(0)), Long.parseLong(args.get(1)),
+            .editMessage(auditUser(Long.parseLong(args.get(0)), Long.parseLong(args.get(1)),
                     Integer.parseInt(args.get(2)) + Integer.parseInt(args.get(3)), event.getJDA()))
             .queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -39,6 +39,7 @@ public final class AuditCommand extends SlashCommandAdapter {
     private static final String TARGET_OPTION = "user";
     private static final String COMMAND_NAME = "audit";
     private static final String ACTION_VERB = "audit";
+    private static final int MAX_PAGE_LENGTH = 25;
     private final Predicate<String> hasRequiredRole;
     private final ModerationActionsStore actionsStore;
 
@@ -145,19 +146,23 @@ public final class AuditCommand extends SlashCommandAdapter {
         return ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event);
     }
 
-    private Message auditUser(long guildId, long targetId, int pageNo, @NotNull JDA jda) {
-        List<ActionRecord> actions = actionsStore.getActionsByTargetAscending(guildId, targetId);
-
+    private @NotNull List<List<ActionRecord>> groupActions(@NotNull List<ActionRecord> actions) {
         List<List<ActionRecord>> groupedActions = new ArrayList<>();
         for (int i = 0; i < actions.size(); i++) {
-            if (i % 25 == 0) {
-                groupedActions.add(new ArrayList<>(25));
+            if (i % AuditCommand.MAX_PAGE_LENGTH == 0) {
+                groupedActions.add(new ArrayList<>(AuditCommand.MAX_PAGE_LENGTH));
             }
 
             groupedActions.get(groupedActions.size() - 1).add(actions.get(i));
         }
 
-        int totalPage = groupedActions.size();
+        return groupedActions;
+    }
+
+    private @NotNull Message auditUser(long guildId, long targetId, int pageNumber, @NotNull JDA jda) {
+        List<ActionRecord> actions = actionsStore.getActionsByTargetAscending(guildId, targetId);
+        List<List<ActionRecord>> groupedActions = groupActions(actions);
+        int totalPages = groupedActions.size();
 
         EmbedBuilder audit = createSummaryEmbed(jda.retrieveUserById(targetId).complete(), actions);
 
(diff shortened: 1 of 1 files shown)


Zabuzard:
Are you "fixing" the issue by getting rid of the embeds or whats the approach here?

Zabuzard:
Some other user attempted to fix this bug before already. The approach there was to add "Next and Previous" buttons at the bottom which would let u go to the next page of issues.

Tanish-Azad:
> Are you "fixing" the issue by getting rid of the embeds or whats the approach here?

Kind of, max fields limit is 25 too so it is still limited

Tanish-Azad:
> Some other user attempted to fix this bug before already. The approach there was to add "Next and Previous" buttons at the bottom which would let u go to the next page of issues.

That is a better solution

Zabuzard:
It is making good progress. Nice. Please tell me once it is ready for a review :)

Tanish-Azad:
it is, i want someone to review the code and test the command

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -53,27 +60,25 @@ public AuditCommand(@NotNull ModerationActionsStore actionsStore, @NotNull Confi
...
     private static @NotNull String createSummaryMessageDescription(
             @NotNull Collection<ActionRecord> actions) {
         int actionAmount = actions.size();
-        if (actionAmount == 0) {
-            return "There are **no actions** against the user.";
-        }
 
-        String shortSummary = "There are **%d actions** against the user.".formatted(actionAmount);
+        String shortSummary = "There are **%s actions** against the user."
+            .formatted(actionAmount == 0 ? "no" : actionAmount);
```
Zabuzard, line 76: Why does this not return anymore? Now, it has a trailing newline symbol `shortSummary + "\n" + typeCountSummary` if there are no actions, no?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -85,30 +90,24 @@ public AuditCommand(@NotNull ModerationActionsStore actionsStore, @NotNull Confi
...
+        };
+
+        User author = jda.getUserById(action.authorId());
+
+        return new MessageEmbed.Field(
+                action.actionType().name() + " by "
+                        + (author == null ? "(unknown user)" : author.getAsTag()),
+                action.reason() + "\nIssued at: " + formatTime.apply(action.issuedAt())
+                        + "\nTemporary action, Expires at: "
+                        + formatTime.apply(action.actionExpiresAt()),
```
Zabuzard, line 109: What if it is not temporary? This string should be missing entirely then. It looks to me as if it is still there now, no? As in `Temporary action, Expires at: "` and then no text anymore.

Btw. it should be `expires` not `Expires`.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
         return ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event);
     }
 
-    private void auditUser(@NotNull User user, @NotNull ISnowflake guild,
-            @NotNull IReplyCallback event) {
-        List<ActionRecord> actions =
-                actionsStore.getActionsByTargetAscending(guild.getIdLong(), user.getIdLong());
+    private Message auditMessage(long guildId, long targetId, int pageNo, @NotNull JDA jda) {
```
Zabuzard, line 141: the method name reads odd. you cant "audit a message". you can "audit an user" though. I would keep this as `auditUser` or rename it to `createAuditMessageForUser` or similar.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+        List<ActionRecord> actions = actionsStore.getActionsByTargetAscending(guildId, targetId);
 
-        MessageEmbed summary = createSummaryMessage(user, actions);
-        if (actions.isEmpty()) {
-            event.replyEmbeds(summary).queue();
-            return;
+        List<List<ActionRecord>> groupedActions = new ArrayList<>();
+        for (int i = 0; i < actions.size(); i++) {
+            if (i % 25 == 0) {
+                groupedActions.add(new ArrayList<>(25));
```
Zabuzard, line 147: 25 is a magic number and should be put into a constant.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+            }
+
+            groupedActions.get(groupedActions.size() - 1).add(actions.get(i));
+        }
```
Zabuzard, line 151: I would prefer this pagination thing to go into a separate utility method for readability.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard, line 141: Please dont abbreviate. `pageNumber`, not `pageNo`. Is it 0 or 1 based? What is the behavior if the number is invalid (negative, too high, ...)? I think this deserves a quick javadoc:
```java
/**
  * @param pageNumber blablabla
  */
```

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+        if (pageNo == 1) {
+            previousButton = previousButton.asDisabled();
+        }
+
+        return previousButton;
+    }
+
+    private Button nextButton(long guildId, long targetId, int pageNo, int totalPage) {
+        Button nextButton = Button.primary(generateComponentId(String.valueOf(guildId),
+                String.valueOf(targetId), String.valueOf(pageNo), "1"), "➡");
```
Zabuzard, line 183: Prefer putting the button creation itself into a helper method instead of duplicating it:
```java
private Button createButton(String label, ...) {
  return Button.primary(...);
}
```

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+
+        return new MessageBuilder(audit.build())
+            .setActionRows(ActionRow.of(previousButton(guildId, targetId, pageNo),
+                    nextButton(guildId, targetId, pageNo, totalPage)))
+            .build();
+    }
+
+    private Button previousButton(long guildId, long targetId, int pageNo) {
+        Button previousButton = Button.primary(generateComponentId(String.valueOf(guildId),
+                String.valueOf(targetId), String.valueOf(pageNo), "-1"), "⬅");
```
Zabuzard, line 172: What is the -1 and the 1 in nextButton for? If this is to remember which button is which, use an enum instead.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +138,58 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
 
-        RestAction<List<MessageEmbed>> messagesTask = RestAction
-            .allOf(actions.stream().map(action -> actionToMessage(action, jda)).toList());
-        messagesTask.map(messages -> prependElement(summary, messages))
-            .flatMap(messages -> event.getHook().sendMessageEmbeds(messages))
+    @Override
+    public void onButtonClick(@NotNull ButtonInteractionEvent event, @NotNull List<String> args) {
+        event
+            .editMessage(auditMessage(Long.parseLong(args.get(0)), Long.parseLong(args.get(1)),
+                    Integer.parseInt(args.get(2)) + Integer.parseInt(args.get(3)), event.getJDA()))
```
Zabuzard, line 196: Looking at the `onButtonClick` method, it is not understandable what these args are for. Please put them into well named variables beforehand instead:
```java
Foo foo = args.get(0);
long bar = Long.parseLong(args.get(1));
int baz = Integer.parseInt(args.get(2));
int buz = Integer.parseInt(args.get(3));

event.editMessage(auditMessage(foo, bar, baz, buz, ...);
```
(but with proper names)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: Mh. Okay. Fair enough, I guess.

Personally, I would still add a early-return. I think it makes the code easier to think about if special cases exit the method as soon as possible.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: this for how much this button increments the page number by

Zabuzard: Ah. Thats cool. Consider putting it into a well named variable instead.
```java
int turnPageBy = -1;
```
or something like that.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: fixed

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: fixed

Tanish-Azad: > What is the behavior if the number is invalid (negative, too high, ...)

i don't this that can ever be a case

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: i made a new method for this

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: Not in your specific case. This is about making the code easy to understand for readers though. And to forge robust software, robust with respect to future modifications.

Someone might for example misinterpret the code and use `1` for the first page, where maybe the first page is `0`. Or might think that setting the page to 1000 would make it just use the highest page instead.

Please just make the behavior and intention clear, cause your code is not self-explanatory.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: this problem is still there though. To a reader of the code, it is absolutely unclear what 1 and -1 means. You first need to scroll around to the onButtonClick to understand what it does.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: You solved this only partially. It is still unclear, looking at just the code, what the role of args 3 and args 4 is.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +149,78 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+            int pageNumber = Integer.parseInt(args.get(3)) + Integer.parseInt(args.get(4));
+
+            event
+                .editMessage(auditUser(guildId, targetId, interactorId, pageNumber, event.getJDA()))
+                .queue();
+        } else {
+            event.reply("Only the user who triggered the command can use these buttons.")
+                .setEphemeral(true)
+                .queue();
+        }
```
Zabuzard, line 224: Prefer using early-return to reduce the nesting:

```java
if (callerId != interactorId) {
  event.reply("Only the user ...")...
  return;
}

long guildId = ...
...
```

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +149,78 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
         return ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event);
     }
 
-    private void auditUser(@NotNull User user, @NotNull ISnowflake guild,
-            @NotNull IReplyCallback event) {
-        List<ActionRecord> actions =
-                actionsStore.getActionsByTargetAscending(guild.getIdLong(), user.getIdLong());
+    private @NotNull List<List<ActionRecord>> groupActions(@NotNull List<ActionRecord> actions) {
```
Zabuzard, line 152: I feel like this method name misses an important opportunity to clarify the actual intention. As in, what are you grouping? Why are you grouping? How are you grouping?

My proposal: `groupActionsByPages` or `paginateActions`.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +149,78 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+            .setActionRows(makeActionRow(guildId, targetId, callerId, pageNumber, totalPages))
+            .build();
+    }
+
+    private @NotNull ActionRow makeActionRow(long guildId, long targetId, long callerId,
+            int pageNumber, int totalPages) {
+        String stringGuildId = String.valueOf(guildId);
+        String stringTargetId = String.valueOf(targetId);
+        String stringCallerId = String.valueOf(callerId);
+        String stringPageNumber = String.valueOf(pageNumber);
```
Zabuzard, line 190: These extra variables feel obsolete and overly verbose. Due to having a method, you already gave each of the vars a good name. So there is no need to do it again. Inline those please.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +149,78 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+                stringCallerId, stringPageNumber, "-1"), "⬅");
+        if (pageNumber == 1) {
+            previousButton = previousButton.asDisabled();
+        }
+
+        Button nextButton = Button.primary(generateComponentId(stringGuildId, stringTargetId,
+                stringCallerId, stringPageNumber, "1"), "➡");
+        if (pageNumber == totalPages) {
+            nextButton = nextButton.asDisabled();
+        }
```
Zabuzard, line 202: I do not like how this code section is still almost 100% identical. This is code duplication. Introduce a helper method that creates a button with a label and a turnPagesBy argument and then use that:

```java
private static Button createPageTurnButton(String label, int turnPagesBy, ...) {
  return Button.primary(...)
}
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad:
```java
        String previousButtonTurnPageBy = "-1";
        Button previousButton = Button.primary(generateComponentId(stringGuildId, stringTargetId,
                stringCallerId, stringPageNumber, previousButtonTurnPageBy), "⬅");

        String nextButtonTurnPageBy = "1";
        Button nextButton = Button.primary(generateComponentId(stringGuildId, stringTargetId,
                stringCallerId, stringPageNumber, nextButtonTurnPageBy), "➡");
```

is this better?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad:
```java
        // Handles the case of too low page number and too high page number
        pageNumber = Math.max(1, pageNumber);
        pageNumber = Math.min(totalPages, pageNumber);
```
so i should add this?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: adding comment for this will do the thing?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: yes

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Zabuzard: Or you just put it into variables. Code should be self-explanatory.
```java
int currentPage = ...
int turnPageBy = ...

int pageToDisplay = currentPage + turnPageBy;
```

Dont be scared of writing long code. If it actually helps readability, it is always worth it.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: i did that because i had to use them twice, for both buttons

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: > I think this deserves a quick javadoc:

```java
    /**
     * @param pageNumber page number to display when actions are divided into pages and each page can contain {@link AuditCommand#MAX_PAGE_LENGTH} actions
     */
```

this better?

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: it'd look the same, even if i introduced that new method

Zabuzard: But you get rid of the duplication. Lets say you want to make both buttons red, currently you need to adjust 2 locations and worst case you forget one and they are out of sync...

By introducing a helper, you merge both into a single code location and reduce the duplication. That will also make Sonar happy, which hates code duplication.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -215,22 +219,20 @@ public void onButtonClick(@NotNull ButtonInteractionEvent event, @NotNull List<S
...
-            event
-                .editMessage(
-                        auditUser(guildId, targetId, interactorId, pageToDisplay, event.getJDA()))
-                .queue();
-        } else {
+        if (callerId != interactorId) {
             event.reply("Only the user who triggered the command can use these buttons.")
                 .setEphemeral(true)
                 .queue();
         }
```
Zabuzard, line 226: This has to go all the way to the top of the method and `return;` then.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Tanish-Azad: ohh yea i forgot sry

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -32,12 +39,15 @@ public final class AuditCommand extends SlashCommandAdapter {
     private static final String TARGET_OPTION = "user";
     private static final String COMMAND_NAME = "audit";
     private static final String ACTION_VERB = "audit";
+    private static final int MAX_PAGE_LENGTH = 25;
+    private static final String PREVIOUS_BUTTON = "⬅";
+    private static final String NEXT_BUTTON = "➡";
```
Zabuzard, line 44: The names can be confusing. Those are not the buttons but their labels. Add a `_LABEL` suffix to the names please.

Its especially confusing in ur lines `Button nextButton = createButton(NEXT_BUTTON, ...)`, as in having two variables with essentially the same name.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -139,26 +151,95 @@ private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
...
+    private @NotNull ActionRow makeActionRow(long guildId, long targetId, long callerId,
+            int pageNumber, int totalPages) {
+        int previousButtonTurnPageBy = -1;
+        Button previousButton = createPageTurnButton(PREVIOUS_BUTTON, guildId, targetId, callerId,
+                pageNumber, previousButtonTurnPageBy);
+        if (pageNumber == 1) {
+            previousButton = previousButton.asDisabled();
+        }
+
+        int nextButtonTurnPageBy = +1;
```
Zabuzard, line 205: Please just write `1` and not `+1` here, thanks.

Zabuzard (changes requested):

Zabuzard (approved):

Heatmanofurioso (approved):

PR #491, opened 2022-08-12
Gist auto filesharing in help threads 

### LINKED ISSUE #489: auto upload files like .txt/.java/.gradle to a paste website like paste.gg
This feature automatically **uploads** files from **help threads** like crash.txt, MyClass.java, build.gradle etc to a 
**paste website**, so the helper don't need to download the file if it reaches a specific length.

I think this is a nice idea cause discord cuts files which are too big, so the helper needs to download it. Also paste websites are useable on any device. 

Example from MMD Discord Server:
image
if files are too big:
image

I would try to implement this feature, just want to know your opinion and get some feedback about this.
Also want to know which paste website I should use? It should have a Java Wrapper/API like pastebin!

SquidXTV:
Closes #489. 
# About
Automatically uploads files in help threads to gist and reply with the link to that gist.
It's easier to read files like that and longer files don't need to be downloaded.

image

## ToDo:
* [x] proper API_KEY
* [x] multiple uploads to one gist

## Config changes:
Added gistApiKey to config for file-sharing uploads to gist using GitHub's API.

### How to create this key:
GitHub -> Settings -> Developer Settings -> Personal Access Token -> Create new token

you can choose note and expiration date yourself, and in scopes select gist:
image

after that copy your key and add it to your config.json
see also: [Create GitHub personal access token](https://docs.github.com/en/enterprise-server@3.4/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)

## Release

Config changes, added the following:
```json
"gistApiKey": "<your_gist_personal_access_token>"
```


Zabuzard:
(could you also share some screenshots?)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+        attachment.retrieveInputStream()
+            .thenApply(this::readAttachment)
+            .thenApply(content -> uploadToSharingService(event, attachment, content))
+            .thenAccept(response -> sendMessage(event, attachment, response));
+    }
+
+    private String readAttachment(@NotNull InputStream stream) {
+        try {
+            byte[] bytes = stream.readAllBytes();
+            return new String(bytes, StandardCharsets.UTF_8);
```
Zabuzard, line 83: NIT: If you want, you can put this into a single line `return new String(stream.readAllbytes(), ...)`. Your choice though. That said, the name `bytes` is bad, `rawData` or `rawContent` would be better. Or even `rawAttachmentContent`.

Variable names should reflect what they logically represent in the context of the code and what they are used for. And not their type. `byte[]` already tells me that its `bytes`. Use the variable name to explain the context instead.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private HttpResponse<String> uploadToSharingService(MessageReceivedEvent event,
```
Zabuzard, line 89: From a design POV, imo the website should return the final URL string, i.e. the result. And not the API response. Its part of the methods responsibility to understand how to work with the gist API, not the repsonsibility of the method user.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+            Message.Attachment attachment, String content) {
+        GistFile file = new GistFile();
+        file.setContent(content);
```
Zabuzard, line 92: You could use a constructor for this and remove the setter (since u only use jackson to serialize and not deserialize this, jackson only needs the getter)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+        String fileName = attachment.getFileName();
+        String fileExtension = attachment.getFileExtension();
+
+        if (fileExtension == null || fileExtension.equals("txt")) {
+            fileExtension = "java";
+        } else if (fileExtension.equals("fxml")) {
+            fileExtension = "xml";
+        }
+
+        fileName = fileName.substring(0, fileName.lastIndexOf(".")) + "." + fileExtension;
```
Zabuzard, line 110: what if the filename has no extension, then `lastIndexOf` gives -1 and ur code doesnt gracefully handle this.

For testing, try to upload a file to discord without extension and see what discord does, i suppose it ends up without extension and then u have a problem here.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+
+        String body = """
+                {
+                  "description": "%s",
+                  "public": false,
+                  "files": {
+                      "%s": %s
+                  }
+                }""".formatted(event.getAuthor(), fileName, parsedContent);
```
Zabuzard, line 119: This part needs a comment why you did not went for jackson fully (i.e. the fact that the file names are dynamic and hence parts of the request are manually created)

Also, you miss the url encoding on the `fileName` and `author` name. So if any of those contains bad characters, your request will mess up again.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+
+
```
Zabuzard, line 121: Two newlines, one is enough.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            return CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void sendMessage(MessageReceivedEvent event, Message.Attachment attachment,
```
Zabuzard, line 138: maybe `sendResponse`, so that its clear what this message is about.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+            HttpResponse<String> response) {
+        if (response.statusCode() < 200 || response.statusCode() > 299) {
```
Zabuzard, line 140: there are constants for these magic numbers. sth like `< HttpStatus.OK`, (see WolframAlphaCommand)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+            logger.warn("Gist API unexpected response: " + response.body());
```
Zabuzard, line 141: use the loggers formatter syntax instead: `warn("Gist API unexpectes response: {}", response.body())`. that way its computed lazy.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+            throw new IllegalStateException("Gist API unexpected response: " + response.body());
+        }
+
+        Message message = event.getMessage();
+        String fileName = attachment.getFileName();
+
+        GistResponse gistResponse = null;
```
Zabuzard, line 148: the `= null;` is obsolete and confuses. remove it.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            gistResponse = JSON.readValue(response.body(), GistResponse.class);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Couldn't parse response...", e);
+        }
+
+        String url = gistResponse.getHtmlUrl();
+
+        message.reply(
+                "I uploaded your file as gist, it is much easier to read for everyone like that especially for mobile users!")
```
Zabuzard, line 158: * missing comma `that, especially for`
* please remove the `!` that sounds unfriendly. maybe even add a thumbsup smiley (unicode)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * This package offers all the functionality for automatically uploading files to gist The core
```
Zabuzard, line 2: dont specify the provider here, that might drift apart from the actual implementation. `for automatically uploading files to sharing services`. also, missing `.` (fullstop) at the end of the sentence.

Zabuzard (changes requested):

Zabuzard:
Good job and thanks for the contribution, its an awesome feature 👍

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/ShareLongAttachmentsMessageListener.java
SquidXTV: do I need to make another jackson like class for fileName and author to parse the variables? also what should I name it?

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistFile.java
```diff
@@ -0,0 +1,20 @@
+package org.togetherjava.tjbot.commands.filesharing;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * @see <a href="https://docs.github.com/en/rest/gists/gists#create-a-gist">Create a Gist via
+ *      API</a>
+ */
+final class GistFile {
```
Taz03, line 9: make this class a record

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistFiles.java
```diff
@@ -0,0 +1,26 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Map;
+
+/**
+ * @see <a href="https://docs.github.com/en/rest/gists/gists#create-a-gist">Create a Gist via
+ *      API</a>
+ */
+final class GistFiles {
```
Taz03, line 13: make this class a record

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistRequest.java
```diff
@@ -0,0 +1,31 @@
+package org.togetherjava.tjbot.commands.filesharing;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jetbrains.annotations.NotNull;
+
+final class GistRequest {
```
Taz03, line 6: make this class a record

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistRequest.java
SquidXTV: what about GistResponse?

Taz03: fields of a record r final, and `GistResponse` class have a set method so u won't be able to change the value after creating an instance if using record

SquidXTV: yea true

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistFiles.java
```diff
@@ -12,14 +12,3 @@
  */
 record GistFiles(@NotNull @JsonIgnore @JsonAnyGetter Map<String, GistFile> nameToContent) {
```
Zabuzard, line 13: Is this one really working? Super weird haha

SquidXTV: yea I tested it lmao

SquidXTV: let me test

SquidXTV: it worked

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        return gistResponse.getHtmlUrl();
+    }
+
+    private void sendResponse(@NotNull MessageReceivedEvent event, @NotNull String url) {
+        Message message = event.getMessage();
+        String replyContent =
+                "I uploaded your file(s) as gist, it is much easier to read for everyone like that, especially for mobile users";
+
+        message.reply(replyContent).setActionRow(Button.of(ButtonStyle.LINK, url, "gist")).queue();
```
Taz03, line 165: Suggestion:
https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/interactions/components/buttons/Button.html#link(java.lang.String,java.lang.String)

`Button` class have a specific method to create a link style button

Taz03: actually just do it

it makes the code look better

```java
.setActionRow(Button.link(url, "Gist"))
```

SquidXTV: oh didn't know that this exists, gonna change that

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        User author = event.getAuthor();
+        if (author.isBot() || event.isWebhookMessage()) {
+            return;
+        }
+
+        if (!isHelpThread(event)) {
+            return;
+        }
```
Taz03, line 70: u don't need this check, see #512

SquidXTV: I think I need it, cause it should only upload it if the file is in a help thread

with this check:
image
image

without:
image
image

Taz03: ohh yea, sry i misunderstood it

Taz03:
whats up with the delete button?

SquidXTV:
> whats up with the delete button?
look here:
https://discord.com/channels/272761734820003841/895717328359153664/1008150388534411344

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistResponse.java
```diff
@@ -0,0 +1,23 @@
...
+    @JsonProperty("html_url")
+    private String htmlUrl;
+
+    public @NotNull String getHtmlUrl() {
+        return this.htmlUrl;
+    }
+
+    public void setHtmlUrl(@NotNull String htmlUrl) {
+        this.htmlUrl = htmlUrl;
+    }
```
Taz03, line 22: this method is never used in your code, so u can remove it

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/GistResponse.java
SquidXTV: I thought jackson uses it for creating the json response or smth

Taz03: hmm yea makes sense, when the field is not public

Taz03 (dismissed):

Taz03:
what about adding a slash command `/remove-gist messageId` for now?

u need `messageId` to delete the message
also u need `gistId` to delete the gist

the `messageId` is provided in the command
after getting the messageId u can get the message by using [MessageChannel.retrieveMessageById(messageId)](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/MessageChannel.html#retrieveMessageById(java.lang.String)), after getting the Message u can use [Message.getButtons()](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/Message.html#getButtons()) to get all the buttons on that message and from there u can get the Button(linking to the gist) use the [Button.getUrl()](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/interactions/components/buttons/Button.html#getUrl()) method to get its url and u can get the id from that, then delete the gist

delete the message after that too if u want to or u can just disable the button, i prefer deleting the message(to keep the chat clutter free)

SquidXTV:
> what about adding a slash command `/remove-gist messageId` for now?
> 
> u need `messageId` to delete the message also u need `gistId` to delete the gist
> 
> the `messageId` is provided in the command after getting the messageId u can get the message by using [MessageChannel.retrieveMessageById(messageId)](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/MessageChannel.html#retrieveMessageById(java.lang.String)), after getting the Message u can use [Message.getButtons()](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/Message.html#getButtons()) to get all the buttons on that message and from there u can get the Button(linking to the gist) use the [Button.getUrl()](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/interactions/components/buttons/Button.html#getUrl()) method to get its url and u can get the id from that, then delete the gist
> 
> delete the message after that too if u want to or u can just disable the button, i prefer deleting the message(to keep the chat clutter free)

only sender of file + mods should have rights to use that command?

Taz03:
> what about adding a slash command `/remove-gist messageId` for now?

also u can take message link and get the messageId from it, the reason for using message link over message id can be that for copying message id of a message u need to turn on developer mode on in discord settings, and some new users might not have that turned on so it'll save some time solving the "no copy id button" problem

Taz03:
> only sender of file + mods should have rights to use that command?

yes

Zabuzard:
soft mod is fine for that

Zabuzard:
Anyways, I would like to wrap this up as it is for now, get it shipped and then work on deletion in a separate PR, that okay for you @SquidXTV ?

Long-living and big branches can be problematic to mantain, better ship smaller units. And mods can always manually delete gists by logging in to the bot account anyways.

application/config.json.template
```diff
@@ -1,5 +1,6 @@
 {
     "token": "<put_your_token_here>",
+    "gistApiKey": "<your_gist_personal_access_token>",
```
Zabuzard, line 3: Since you are changing the config, please add to your PR description that a config change is needed (and what to change).

That way it is easier for someone doing the releases (like me) to know what to do. And to not forget.

application/src/main/java/org/togetherjava/tjbot/config/Config.java
```diff
@@ -100,6 +103,15 @@ public String getToken() {
         return token;
     }
 
+    /**
+     * Gets the API Key of GitHub to upload pastes via the API.
```
Zabuzard, line 107: A link to the URL or a github tutorial of how to create this token could be helpful.

Think of a new user who wants to use our code and needs to understand how to obtain such a token. Ideally, the Javadoc hints them in the right direction here.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Listener that receives all sent help messages and uploads them to a share service if the message
+ * contains a file with the given extension in the
+ * {@link FileSharingMessageListener#extensionFilter}
```
Zabuzard, line 34: missing dot at the end of the sentence.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        List<Message.Attachment> attachments =
+                event.getMessage().getAttachments().stream().filter(attachment -> {
+                    String extension = attachment.getFileExtension();
+                    if (extension == null) {
+                        return false;
+                    }
+                    return extensionFilter.contains(extension);
+                }).toList();
```
Zabuzard, line 79: Put the filtering into a helper method to simplify reading the `onMessageReceived` method. It is already doing quite a lot of stuff anyways.

So maybe add a `.filter(isAttachmentRelevant)` or similar.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+            return;
+        }
+
+        List<Message.Attachment> attachments =
+                event.getMessage().getAttachments().stream().filter(attachment -> {
+                    String extension = attachment.getFileExtension();
+                    if (extension == null) {
+                        return false;
+                    }
+                    return extensionFilter.contains(extension);
```
Zabuzard, line 78: Might be wrong, but I think `Set#contains` works well if you give it `null`. I.e. `extensionFilter.contains(null)` might work even, making the previous check obsolete.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+    private void processAttachments(@NotNull MessageReceivedEvent event,
+            @NotNull List<Message.Attachment> attachments) {
+        Map<String, GistFile> filesAsJson = new HashMap<>();
+
+        for (Message.Attachment attachment : attachments) {
+            attachment.retrieveInputStream()
+                .thenApply(this::readAttachment)
+                .thenAccept(content -> filesAsJson.put(createFileName(attachment),
+                        new GistFile(content)))
+                .join();
```
Zabuzard, line 93: Dont join on this directly. That way you compute it sequentially and not in-parallel. Huge bummers.
Also, this actually has a bug. You are modifying the map (which is not thread-safe) in a threaded environment.

Instead, stop at `thenAccept(...)`. The method gives you back a future, which you can then collect in a list of futures. Then, after launching all tasks, you can start waiting on them. So:

```java
Map<String, GistFile> nameToFile = Collections.synchronizedMap(new HashMap<>());

List<CompletableFuture<Void>> tasks = new ArrayList<>();
  CompletableFuture<Void> task = attachment.retrieveInputStream()
    .thenApply(this::readAttachment)
    .thenAccept(content -> filesAsJson.put(createFileName(attachment), new GistFile(content)));

  tasks.add(task);
}

tasks.forEach(CompletableFuture::join);
```

Zabuzard: That said, now you are actually running something blocking on a JDA thread (the thread of `onMessageReceived`), that is not acceptable and can block JDA itself if its very busy.

You have to make sure that those `join`s are actually executed on some other thread. The easiest way to achieve this is by posting the whole `proccessAttachments` method on a service:
```java
CompletableFuture.runAsync(() -> processAttachments(event, attachments));
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private @NotNull String readAttachment(@NotNull InputStream stream) {
+        try {
+            return new String(stream.readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private @NotNull String createFileName(@NotNull Message.Attachment attachment) {
```
Zabuzard, line 110: This method name explains what it does, but doesnt really do a good job at giving specific context.

I vote for `getNameOf`.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        String fileName = attachment.getFileName();
+        String fileExtension = attachment.getFileExtension();
+
+        if (fileExtension == null || fileExtension.equals("txt")) {
+            fileExtension = "java";
+        } else if (fileExtension.equals("fxml")) {
+            fileExtension = "xml";
+        }
+
+        fileName = fileName.substring(0, fileName.lastIndexOf(".")) + "." + fileExtension;
```
Zabuzard, line 120: this still has the problem of crashing if `fileName` has no `.` in it (`lastIndexOf` retuns `-1` then, which `substring(...)` doesnt understand).

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        return fileName;
+    }
+
+    private @NotNull String uploadToGist(@NotNull GistRequest jsonRequest) {
+        String body;
+        try {
+            body = JSON.writeValueAsString(jsonRequest);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Couldn't parse json request!", e);
```
Zabuzard, line 129: Try to add more context, so that a reader of that exception (looking at it in console) directly understands where this is coming from:
```java
"Attempting to upload a file to gist, but unable to create the JSON request."
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        HttpRequest request = HttpRequest.newBuilder()
+            .uri(URI.create(SHARE_API))
+            .header("Accept", "application/json")
+            .header("Authorization", "token " + gistApiKey)
```
Zabuzard, line 135: Might be wrong, but is there maybe an overload that accepts multiple headers directly? Such as
```java
.header(
  key1, value1,
  key2, value2
)
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+            .POST(HttpRequest.BodyPublishers.ofString(body))
+            .build();
+
+        HttpResponse<String> apiResponse;
+        try {
+            apiResponse = CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
```
Zabuzard, line 141: the blocking send now also happens on the JDA thread. so u definitely have to ensure that the whole flow (`processAttachments`) now happens on another thread (see previous comment)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IllegalStateException("HttpResponse interrupted!", e);
```
Zabuzard, line 146: Give more context:
```java
"Attempting to upload a file to gist, but the request got interrupted."
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        if (statusCode < HttpURLConnection.HTTP_OK
+                || statusCode >= HttpURLConnection.HTTP_MULT_CHOICE) {
+            throw new IllegalStateException("Gist API unexpected response: " + apiResponse.body());
+        }
+
+        GistResponse gistResponse;
+        try {
+            gistResponse = JSON.readValue(apiResponse.body(), GistResponse.class);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Couldn't parse response!", e);
```
Zabuzard, line 160: Context:
```java
"Attempting to upload file to gist, but unable to parse its JSON response."
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        return gistResponse.getHtmlUrl();
+    }
+
+    private void sendResponse(@NotNull MessageReceivedEvent event, @NotNull String url) {
+        Message message = event.getMessage();
+        String replyContent =
+                "I uploaded your file(s) as gist, it is much easier to read for everyone like that, especially for mobile users";
```
Zabuzard, line 168: Missing dot at the end.

Also, I do not like that `file(s)` thing. Maybe go for another wording, such as

```java
"I uploaded your attachments as **gist**. That way, they are easier to read for everyone, especially mobile users 👍"
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+            gistResponse = JSON.readValue(apiResponse.body(), GistResponse.class);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Couldn't parse response!", e);
+        }
+        return gistResponse.getHtmlUrl();
+    }
+
+    private void sendResponse(@NotNull MessageReceivedEvent event, @NotNull String url) {
+        Message message = event.getMessage();
+        String replyContent =
```
Zabuzard, line 167: i dont like the var name. maybe go for `messageContent`?

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,183 @@
...
+    private boolean isHelpThread(@NotNull MessageReceivedEvent event) {
+        if (event.getChannelType() != ChannelType.GUILD_PUBLIC_THREAD) {
+            return false;
+        }
+
+        ThreadChannel thread = event.getThreadChannel();
+        String rootChannelName = thread.getParentChannel().getName();
+        return isStagingChannelName.test(rootChannelName)
+                || isOverviewChannelName.test(rootChannelName);
+    }
```
Zabuzard, line 182: Unsure, but isnt such a method already available in `HelpSystemHelper` (or whatever its called)? If so, please use the helper instead.

Zabuzard (changes requested):

SquidXTV:
> Anyways, I would like to wrap this up as it is for now, get it shipped and then work on deletion in a separate PR, that okay for you @SquidXTV ?
> 
> Long-living and big branches can be problematic to mantain, better ship smaller units. And mods can always manually delete gists by logging in to the bot account anyways.

ok no problem for me, gonna work on your pr comments and then I think it's done for this pr.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: yea true, do I even need the isAttachmentRelevant method then if I only have return extensionFilter.contains(...)?

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: but fileName always got a . + extension at the end cause of my extensionFilter at the beginning which filters these

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: there is actually

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: couldn't find it, but I stole it somewhere but I think it was a non-static method thats why I simply copied it

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: maybe try to give me an example file name + extension which you think could break

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: this null check is needed:

image

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Taz03: > Unsure, but isnt such a method already available in HelpSystemHelper (or whatever its called)? If so, please use the helper instead.

its in a diff package and the method is not public, also the use case r different the method in HelpSystemHelper sends a message if its not a help thread + it is getting removed in my pr so its gud if they make their own method

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Taz03: > but fileName always got a .

no

SquidXTV: > > but fileName always got a .
> 
> no

give me an example pls

Taz03: > > > but fileName always got a .
> > 
> > 
> > no
> 
> give me an example pls

Dockerfile

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Zabuzard: fair

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.regex.Pattern;
+
+public final class DeleteGistCommand extends SlashCommandAdapter {
+
+    private static final Logger logger = LoggerFactory.getLogger(DeleteGistCommand.class);
+
+    private static final String MESSAGE_ID_OPTION = "message-id";
+
+    public DeleteGistCommand(@NotNull Config config) {
+        super("remove-gist", "Deletes gist's from auto file-sharing feature!",
```
Zabuzard: the description sounds more like a description for a programmer and less for an actual user. `"Delete a previously uploaded gist"`

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+                SlashCommandVisibility.GUILD);
```
Zabuzard: u forgot actually adding the message-id option

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
+        String messageId = event.getOption(MESSAGE_ID_OPTION).getAsString();
+
+        Message gistMessage = event.getMessageChannel().retrieveMessageById(messageId).complete();
```
Zabuzard: what if used in a different channel? if this is a limitation you cant get around, then it has to be described to the user in the command description or they will be super confused

Zabuzard: you stepped into the next multithreading pitfall. `complete()` blocks the current (JDA) thread. an absolute no-go.

instead, use the returned RestAction and its chaining operations `map`. RestAction is JDAs CompletableFuture.

From there on, all your next actions will be in this chain instead.

Also, you need to react on the failure of the message retrieval to provide the user with good UX in case they either gave the wrong ID or used the command in the wrong channel. They need a proper error message that explains them what they did wrong.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+
+        String content =
+                "I uploaded your file(s) as gist, it is much easier to read for everyone like that, especially for mobile users";
+
+        if (!gistMessage.getAuthor().isBot() || !gistMessage.getContentRaw().equals(content)) {
+            return;
```
Zabuzard: dont just stop without giving the user any response to their command.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        User fileSender = gistMessage.getReferencedMessage().getAuthor();
```
Zabuzard: what if the given message was not a reply, this code can fail and u didnt write it robust.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        User user = event.getUser();
+
+        if (!user.equals(fileSender)) {
+            return;
```
Zabuzard: forgot any error messages

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        String url = gistMessage.getButtons()
+            .stream()
+            .filter(button -> button.getLabel().equals("gist"))
+            .toList()
+            .get(0)
```
Zabuzard: bad, use the streams instead `.findAny()`.

Also, you have to correctly handle the case that the stream is empty. (no buttons on the message)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+            .getUrl();
+        url = url.replace("https://api.github.com/gists/", "");
```
Zabuzard: what is that? are u trying to extract the gist ID? use a capturing group regex instead. and move this into a helper method `getGistId(gistUrl)`

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
```diff
@@ -0,0 +1,57 @@
...
+
+        // send api stuff to
```
Zabuzard: use `// TODO` or `// FIXME` to not forget and prevent this PR from merging without this

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -69,30 +67,42 @@ public void onMessageReceived(@NotNull MessageReceivedEvent event) {
...
+
+        CompletableFuture.runAsync(() -> processAttachments(event, attachments));
+    }
+
+    private boolean isAttachmentRelevant(@NotNull Message.Attachment attachment) {
+        String extension = attachment.getFileExtension();
+        if (extension == null) {
+            return false;
+        }
+        return extensionFilter.contains(extension);
```
Zabuzard, line 85: u forgot the improvement with `contains(null)`

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -126,13 +136,14 @@ private void processAttachments(@NotNull MessageReceivedEvent event,
...
+            throw new IllegalStateException(
+                    "Attempting to upload a file to gist, but unable to create the JSON request.",
+                    e);
         }
 
         HttpRequest request = HttpRequest.newBuilder()
             .uri(URI.create(SHARE_API))
-            .header("Accept", "application/json")
-            .header("Authorization", "token " + gistApiKey)
+            .headers("Accept", "application/json", "Authorization", "token " + gistApiKey)
```
Zabuzard, line 146: is that auto alignment from spotless? in that case the previous version was better, id revert back.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Zabuzard: nvm, u wrote that its needed.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/DeleteGistCommand.java
Taz03: isn't this file is deleted now?

SquidXTV: yea should be
it was WIP

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: yea I think spotless aligned it differently this time

SquidXTV: how can I get the old alignment?

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Listener that receives all sent help messages and uploads them to a share service if the message
+ * contains a file with the given extension in the
+ * {@link FileSharingMessageListener#extensionFilter}.
+ */
+public class FileSharingMessageListener extends MessageReceiverAdapter {
+
+    private static final ObjectMapper JSON = new ObjectMapper();
+
+    private final String gistApiKey;
```
Tais993, line 38: Odd order, put this variable below the static final one's

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,197 @@
...
+            return false;
+        }
+        return extensionFilter.contains(extension);
+    }
+
+
+    private void processAttachments(@NotNull MessageReceivedEvent event,
+            @NotNull List<Message.Attachment> attachments) {
+
+        Map<String, GistFile> filesAsJson = Collections.synchronizedMap(new HashMap<>());
```
Tais993, line 92: Why are you not using a ConcurrentHashMap?

SquidXTV: ask Zabu tbh, he told me to update it like this

SquidXTV: also does it make a difference?

Taz03: ConcurrentHashMap is to make the app thread safe. So u should use it if doing multi-threading

Tais993: Reason why I ask is because ConcurrentHashMap is faster, there's some more differences (like that Collections#syncronizedMap allows null keys, etc), but unless you need one of those I don't see why you should use it.

SquidXTV: but Collections.synchronizedMap returns a thread safe map says the docs

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+
+        List<Message.Attachment> attachments = event.getMessage()
+            .getAttachments()
+            .stream()
+            .filter(this::isAttachmentRelevant)
+            .toList();
+
+        CompletableFuture.runAsync(() -> processAttachments(event, attachments));
```
Zabuzard, line 78: This suppresses failures, unfortunately.

Gonna add a:
```java
runAsync(() -> {
  try {
    processAttachments(event, attachmenets);
  } catch (Exception e) {
    logger.error("Unknown error while processing attachments", e);
  }
}
```

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -0,0 +1,198 @@
...
+            return false;
+        }
+        return extensionFilter.contains(extension);
+    }
+
+
+    private void processAttachments(@NotNull MessageReceivedEvent event,
+            @NotNull List<Message.Attachment> attachments) {
+
+        Map<String, GistFile> filesAsJson = new ConcurrentHashMap<>();
```
Zabuzard, line 93: I dont like the name of this map. its not a map of json after all, but regular java objects.

Maybe go for `nameToFile`

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Zabuzard: ur constantly resolving this, but the problem is still not fixed:
> this still has the problem of crashing if `fileName` has no `.` in it (`lastIndexOf` retuns `-1` then, which `substring(...)` doesnt understand).

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

PR #492, opened 2022-08-13
Added help thread activity detection

### LINKED ISSUE #478: Help Thread activity overview
Help threads should somehow show an activity indicator so that it is easier for helpers to know which help thread needs their attention and which not.

This could for example be realized using channel renaming, prefixing an emoji (note rate limitations).

The actual metrics are up for decision, but could be something along the lines of:

example

Decided based on how many messages and how many different authors (other than OP) posted already.

Zabuzard:
## Overview

Closes #478 . This adds a activity indicator to each help thread:

* 🔻: needs help
* 🔸: likely needs help
* 🔹: seems good

They indicate towards helpers which channels are in most need of help and which likely already received attention by helpers.

The code updates these indicators every 30 minutes (to not hit renaming rate limits). New threads start with 🔻.

discord overview

our overview

## Details

The code determines activity based on a simple heuristic (bot messages are filtered):

* 🔻: default
* 🔸: `<= 10` messages, at least 2 different people and one of them wrote `>= 2` messages
* 🔹: `> 10` messages

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
@@ -75,6 +75,7 @@ public enum Features {
         features.add(new RemindRoutine(database));
         features.add(new ScamHistoryPurgeRoutine(scamHistoryStore));
         features.add(new BotMessageCleanup(config));
+        features.add(new HelpThreadActivityUpdater(helpSystemHelper));
 
         // Message receivers
         features.add(new TopHelpersMessageListener(database, config));
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -97,7 +97,10 @@ public final class AskCommand extends SlashCommandAdapter {
         Guild guild = event.getGuild();
         event.deferReply(true).queue();
 
-        overviewChannel.createThreadChannel("[%s] %s".formatted(category, title))
+        HelpSystemHelper.HelpThreadName name = new HelpSystemHelper.HelpThreadName(
+                HelpSystemHelper.ThreadActivity.NEEDS_HELP, category, title);
+
+        overviewChannel.createThreadChannel(name.toChannelName())
             .flatMap(threadChannel -> handleEvent(eventHook, threadChannel, author, title, category,
                     guild))
             .queue(any -> {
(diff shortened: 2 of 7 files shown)


marko-radosavljevic:
Since we don't use emojis anywhere (in channel names, etc), this would add a lot of visual clutter and random colours and shapes everywhere. They will even update periodically, and flicker. I feel like lots of people will dislike that, I know I would.

So maybe we can first go with something unoffensive like utf-8 characters.

Maybe:
×  - needs help
–  - likely needs help
✓  - seems good

Or something along those lines. We have a ton of options with utf chars.

Also, red-yellow-blue is unorthodox and possibly confusing, especially with the different shapes. Would rather have red-yellow-green with shapes that maybe represent their intended purpose better. I understand that we have somewhat limited options here. Not really a huge deal tho, since we would have a legend explaining the system, and regular helpers that use the inbox system will get accustomed relatively fast.

Zabuzard:
> 

yes. i understand and i would be okay with ur suggestions. maybe we make a voting somewhere, to get a few more opinions.

Zabuzard:
ur symbols look like this @marko-radosavljevic :

changed

I feel like color is needed for it to really be of help. Just my opinion though.

marko-radosavljevic:
Yeah, I think it works. It's definitely unoffensive, and I think I would get used to it pretty quickly. Maybe it needs some detail to make it stand out better, dash or one more space, or circled characters.. Because it blends a bit with the title and the tag, it's supposed to be 'left' of it, as a separate element

Will see if I can think of anything else, and we can have a quick vote or something. Thanks :relaxed:

Taz03:
i prefer red green and yellow with uniform shapes

Taz03 (approved):

Zabuzard:
the "problem" with green and ✅ or ✔️ is that it feels like its for sure OK, handled and doesnt need any help anymore. But we are still talking about a heuristic only and just because there is a lot of traffic, doesnt mean the question has been answered. Hence I tried to go for something 🟦 that looks like "okay" but not "handled for sure".

mikerasch:
I feel like there's no way to deliver an emoji-based system without it being a little confusing. Is it possible we could incorporate custom emojis? I feel like the current default emotes, it's lacking the goal of a non-confusing system.

Zabuzard:
custom emojis are technically possibly, but harder to implement (u have to retrieve them via rest actions etc and have a fallback in case they are not available). and its even harder for the parsing part when u have to allow multiple symbols.

i would really like to avoid going down that rabbit hole.

Zabuzard:
We will go for this design for today and change it based on user feedback in the early days after release, i guess

PR #494, opened 2022-08-13
Adding uncaught exception handling

### LINKED ISSUE #490: Add unhandled exception listener
Right now, when exceptions are thrown outside of the main thread and outside of JDAs own threads (for example via our routines), exceptions are not logged at all and go silently unoticed.

We have to add a global exception listener for all unhandled exceptions and ensure they are correctly logged with our existing logging utility. Otherwise we have no visibility on many issues.

Zabuzard:
## Overview

Closes #490. Some exceptions slipped through the logging system. Most notable, anything happening inside `Routine`s. Even worse, the routine actually stopped forever (until the next bot restart).

This fixes the issue by:
* adding a global default uncaught handler
* additionally logging for scheduled executor services, which otherwise supress and wrap the exceptions into the future

An example of a quick test exception from a routine, which is now correctly logged:

example exception


marko-radosavljevic (dismissed):
:heart:

Taz03 (approved):

Zabuzard:
quick rebase, no code changes

Zabuzard:
(merging with 2 prior approvals)

PR #495, opened 2022-08-13
Auto-prune of full helper roles

### LINKED ISSUE #483: Auto-prune helpers
## Problem

This is somewhat urgent. Due to a technical limitation, our helper-role system breaks once a helper role has more than 100 users (usually the JAVA helper role).

It fails silently, probably an issue on Discord-end, not on ours. Anyways, we need to mitigate this problem until a proper solution has been found.

## Solution

A way to do this is to automatically prune users from the role who have not posted messages for some time.

For a good UX, we should also attempt to DM the user in order to inform them what has happened. Such as

> You seem to have been inactive for X, hence we removed you from the Y helper role. If that was a mistake, just head back to #active_questions and select the role again. Sorry for any inconvenience caused by this.

Also, this should only activate for roles that are actually approaching the 100-limit (so maybe as soon as they go beyond 90). So maybe create a `Routine` that runs every hour, checking the member count of all helper roles (get via `Config`), and if beyond 90, start pruning it.

Zabuzard:
## Overview

Closes #483. This is a somewhat urgent mitigation due to a technical limitation. Our helper-role system breaks once a helper role has more than 100 users (usually the JAVA helper role).

This mitigates the problem by automatically removing some users from helper roles.

## Details

Roles are pruned once they go beyond `95` users (the hard limit is `100`). It will then attempt to prune `10` **random** users who fulfill:
* not recently joined (**7 days**)
* no recent message in a help thread (**90 days** - this is also the lookback limit of our DB as of today)

Users are send a DM to inform them about what happened:

DM

The routine also logs on `INFO` level who got pruned, so we can always double check in the logs. It will also tell us on `WARN` level if a role is, despite auto-pruning, approaching or even going beyond the limit. So that we can start to prune manually.

Latter warnings are also forwarded to the `#mod_audit_log` channel, so that we really do not miss it:

audit log

## Other stuff

This PR also addresses two unrelated things:

* Added hints on how to change the logging level (also for bot only) in `log4j2.xml`
* Fixed a bug with routines starting too early (they must only be started once the system is ready, not directly during startup)
* Fixed a bug where `BotCore#onReady` is not always triggered because the core is added too late to JDA

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
@@ -145,8 +145,6 @@ public final class AutoPruneHelperRoutine implements Routine {
 
     private void pruneMemberFromRole(@NotNull Member member, @NotNull Role role,
             @NotNull TextChannel overviewChannel) {
-        if (Math.random() <= 0.999)
-            return;
         Guild guild = member.getGuild();
 
         String dmMessage =
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
@@ -11,6 +11,7 @@ import org.slf4j.LoggerFactory;
 import org.togetherjava.tjbot.commands.Routine;
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.moderation.ModAuditLogWriter;
 
 import java.time.Duration;
 import java.time.Instant;
@@ -37,6 +38,7 @@ public final class AutoPruneHelperRoutine implements Routine {
     private static final int RECENTLY_JOINED_DAYS = 7;
 
     private final HelpSystemHelper helper;
+    private final ModAuditLogWriter modAuditLogWriter;
     private final Database database;
     private final List<String> allCategories;
 
@@ -45,12 +47,14 @@ public final class AutoPruneHelperRoutine implements Routine {
      *
      * @param config to determine all helper categories
      * @param helper the helper to use
+     * @param modAuditLogWriter to inform mods when manual pruning becomes necessary
      * @param database to determine whether an user is inactive
      */
     public AutoPruneHelperRoutine(@NotNull Config config, @NotNull HelpSystemHelper helper,
-            @NotNull Database database) {
+            @NotNull ModAuditLogWriter modAuditLogWriter, @NotNull Database database) {
         allCategories = config.getHelpSystem().getCategories();
         this.helper = helper;
+        this.modAuditLogWriter = modAuditLogWriter;
         this.database = database;
     }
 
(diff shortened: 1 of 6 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
```diff
@@ -0,0 +1,168 @@
...
+        jda.getGuildCache().forEach(this::pruneForGuild);
+    }
+
+    private void pruneForGuild(@NotNull Guild guild) {
+        TextChannel overviewChannel = guild.getTextChannels()
+            .stream()
+            .filter(channel -> helper.isOverviewChannelName(channel.getName()))
+            .findAny()
+            .orElseThrow();
+        Instant when = Instant.now();
```
marko-radosavljevic, line 77:
```suggestion

        Instant now = Instant.now();
```
I think `now` reads better in this context, with usages:

```java

if (Duration.between(memberJoined, when).toDays() <= RECENTLY_JOINED_DAYS)
```
and

```java
Instant latestActiveMoment = when.minus(INACTIVE_AFTER);
```
Your choice, I'm just nitpicking.

marko-radosavljevic (dismissed):
Clean. :relaxed:

application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
Zabuzard: agreed

application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
```diff
@@ -0,0 +1,168 @@
...
+            @NotNull Instant when) {
+        role.getGuild().findMembersWithRoles(role).onSuccess(members -> {
+            if (isRoleFull(members)) {
+                logger.debug("Helper role {} is full, starting to prune.", role.getName());
+                pruneRole(role, members, overviewChannel, when);
+            }
+        });
+    }
+
+    private boolean isRoleFull(@NotNull Collection<Member> members) {
```
Tais993, line 96: Why not `? extends`

application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
```diff
@@ -0,0 +1,168 @@
...
+        return members.size() >= ROLE_FULL_THRESHOLD;
+    }
+
+    private void pruneRole(@NotNull Role role, @NotNull List<? extends Member> members,
+            @NotNull TextChannel overviewChannel, @NotNull Instant when) {
+        List<Member> membersShuffled = new ArrayList<>(members);
+        Collections.shuffle(membersShuffled);
```
Tais993, line 103: Why the shuffle?

Zabuzard: Because we decide to kick out, lets say 10 users out of 30 users eligible for the prune. I dont want this to happen alphabetically or anything, to not put anyone at a disadvantage.

So i am picking the people to prune at random

application/src/main/java/org/togetherjava/tjbot/commands/help/AutoPruneHelperRoutine.java
Zabuzard: good catch

PR #496, opened 2022-08-13
Increase dependabot support

Zabuzard:
We initially started with a dependabot config that posted every change, daily. That was too spammy for us and we tried out to only have major versions, monthly.

The bot is now running fine since many months and dependabot went almost silent since then.

We should bump it a bit, so that it starts giving us more updates again - we do not want to miss on any cool stuff or security updates.

This time, I am going for:
* daily
* major, minor (but no patch)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,8 @@ updates:
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:
-      interval: "monthly"
+      interval: "daily"
     ignore:
       - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


marko-radosavljevic (approved):
Oh man, I was enjoing the silence.. And I'm still recovering from vaadin flashbacks :cries:

image


But if you want to try it out, sure,  I guess ¯\\_(ツ)_/¯

~~at least we got rid of vaadin :mhh:~~

Taz03 (approved):

PR #497, opened 2022-08-14
Auto-close help thread if OP leaves server

### LINKED ISSUE #480: Auto-close help thread if OP leaves server
Sometimes question askers leave the server while their question is still active. This leads to helpers wasting time, not knowing that OP left already.

Hence, the bot should react to the event of people leaving the server and check if there are any question threads created by them. If so, close them (see how `/close` works).

Implementation-wise, challenges here are that we do not associate help-threads to authors yet. The person implementing this has to explore:
* can we easily automatically derive the thread author? (probably not?)
* otherwise, we should maybe setup a small database that associates thread channel IDs with author IDs (auto-evicted every week or so)

For latter, there are already similar setups for the top-helper system. Can copy pasta some stuff.

mikerasch:
## Overview
Closes #480 
When a user leaves the server, their help thread stays open. This leads to not only the waste of moderators' time when, but also leads to helpers contributing to nobody.

## Details
- Store users in a database upon creation. Storing user id, thread id, and date it was created.
- When the user leaves the guild, inform the channel that it will be closed.
- Archive all thread channels made by the user.
- Every 30 days, purge the database of all thread creation entities.


marko-radosavljevic:
Thank you for your contribution, and welcome to the project! :relaxed: :heart:

Zabuzard:
(you can use interactive git rebase to change the author btw, not a big deal though)

Taz03:
I think, u misread the problem. We want to close the help thread if the user leaves not delete it

mikerasch:
> I think, u misread the problem. We want to close the help thread if the user leaves not delete it

Fixed 👍 For some reason, I didn't think about archiving the thread and my mind just went to delete. Thanks for letting me know :)

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void onEvent(@NotNull GenericEvent event) {
+        if (event instanceof GuildMemberRemoveEvent leaveEvent) {
+            onGuildMemberRemoved(leaveEvent);
+        }
+    }
+
+    private void onGuildMemberRemoved(GuildMemberRemoveEvent leaveEvent) {
+        ThreadChannel threadChannel;
+        Set<Long> channelId = getThreadsAssociatedToLeaver(leaveEvent);
```
Taz03, line 38: make the name plural coz its a collection
```suggestion
        Set<Long> channelIds = getThreadsAssociatedToLeaver(leaveEvent);
```

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+        if (channelId.isEmpty()) {
+            logger.warn("Failed to get channelID user was associated with");
+        } else {
+            for (Long channel : channelId) {
```
Taz03, line 42: since this is a channel id the var name should say that
```suggestion
            for (Long channelId : channelIds) {
```

Taz03 (changes requested):
also y clear db after every 7 days y not remove each thread when it gets closed

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
mikerasch: > also y clear db after every 7 days y not remove each thread when it gets closed

I feel like adding a listener for every closed app would complicate the process. For example, if a user closes a thread and re-opens the thread via typing in it, I don't see an easy way in storing the data again.

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class LeaversPurgeDatabaseRoutine implements Routine {
+    private final Database database;
+    private static final Logger logger = LoggerFactory.getLogger(LeaversPurgeDatabaseRoutine.class);
+    private static final Period DELETE_MESSAGE_RECORDS_AFTER = Period.ofDays(7);
+
+    public LeaversPurgeDatabaseRoutine(@NotNull Database database) {
+        this.database = database;
+    }
+
```
Zabuzard, line 26: two newlines is one too much

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+        if (event instanceof GuildMemberRemoveEvent leaveEvent) {
+            onGuildMemberRemoved(leaveEvent);
+        }
+    }
+
+    private void onGuildMemberRemoved(GuildMemberRemoveEvent leaveEvent) {
+        ThreadChannel threadChannel;
+        Set<Long> channelId = getThreadsAssociatedToLeaver(leaveEvent);
+        if (channelId.isEmpty()) {
+            logger.warn("Failed to get channelID user was associated with");
```
Zabuzard, line 40: what if the user simply had no channels? logging a warning in such a case would be incorrect.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
Zabuzard: should be `long` not `Long`, otherwise it sounds as if u want to support it being null

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+        } else {
+            for (Long channel : channelId) {
+                threadChannel = leaveEvent.getGuild().getThreadChannelById(channel);
+                if (threadChannel == null) {
+                    logger.warn("Thread channel ID: '{}' is already deleted.", channel);
```
Zabuzard: this is missing context and hence would be weird to read in a big log file. like, it has to say what it is and what it attempted.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+                } else {
```
Zabuzard, line 46: ur nesting too much, that hurts readability. get used to early-returns/guard-statements. As in:
```java
if (...) {
  ...
  return;
}
... // rest of the code without nesting
```

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+                    MessageEmbed embed = new EmbedBuilder().setTitle("Original Thread Creator Left")
```
Zabuzard, line 47: U can just write `OP left`, thats more idiomatic

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+                        .setDescription("Closing ticket...")
```
Zabuzard, line 48: its not a "ticket", call it channel.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+                        .setColor(HelpSystemHelper.AMBIENT_COLOR)
+                        .build();
+                    ThreadChannel finalThreadChannel = threadChannel;
```
Zabuzard, line 51: whats this variable for? i dont quite see its purpose.

application/src/main/resources/db/V11__Add_Help_Channel_Plus_Creator.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE add_help_channel
```
Zabuzard, line 1: i dont like the concept of this table. i would find it better if it would be a general purpose help channel list.

hence a better naming would be `help_channels` or `help_threads`

application/src/main/resources/db/V11__Add_Help_Channel_Plus_Creator.sql
```diff
@@ -0,0 +1,6 @@
...
+(
+    channel_id BIGINT    NOT NULL PRIMARY KEY,
+    user_id    BIGINT    NOT NULL,
```
Zabuzard, line 4: user is a bad name, it would be better to give the context of who the user is. i.e. `author`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
mikerasch: Yeah, makes a ton more sense. I knew it was wrong and should have fixed it.

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public @NotNull Schedule createSchedule() {
+        return new Schedule(ScheduleMode.FIXED_RATE, 0, 4, TimeUnit.HOURS);
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        int recordsDeleted = database
+            .writeAndProvide(content -> content.deleteFrom(AddHelpChannel.ADD_HELP_CHANNEL))
```
Tais993, line 36: Maybe it's me, but I really don't understand when it gets deleted from the database.

Would appreciate some comments on when it gets deleted.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final Logger logger =
+            LoggerFactory.getLogger(OnGuildLeaveCloseThreadListener.class);
+    private final Database database;
+
+    public OnGuildLeaveCloseThreadListener(@NotNull Database database) {
+        this.database = database;
+    }
+
+    @Override
+    public void onEvent(@NotNull GenericEvent event) {
```
Tais993, line 30: extend ListenerAdapter and use onGuildMemberRemoveEvent(GuildMemberRemoveEvent) instead

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
mikerasch: Intellj yells at me saying: Variable used in lambda expression should be final or effectively final. So I decided to just make a new variable effectively final.

application/src/main/resources/db/V11__Add_Help_Channel_Plus_Creator.sql
Tais993: I agree, struggled reading some things because of the constant "ADD"

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
mikerasch: Well if the database is going to be used for all help_threads related, it would probably make sense to not purge all the help channels and maybe only the ones which are tied to the user leaving.

Zabuzard: (its okay if we purge our metadata at regular intervals - maybe after a month). we dont want our database to pile up (disk space is limited) and we also dont want to hold too much info, privacy-wise

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
Zabuzard: im not sure if this is possible with the current bot design. i vaguely remember having had problems with that

mikerasch: Originally had the problem also so I stuck to this way. I can try to come up with a solution though. I have to look into what caused the original issue though. If I remember, it was not being able to add this to the features. Going to sleep now, but will update you tomorrow.

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
mikerasch: Well realistically, what would make the most sense interval-wise? 7 days might be too short but I also think a month might be too long.

Zabuzard: its not a big deal if the table lists archived threads though. with the renaming, it will function as general purpose metadata table for help threads - so other functionalities/commands will consume it as well, not just ur auto-close thing.

Tais993: I'm not saying we shouldn't purge, I'm just saying I do not understand this code, might be because I'm reading from GitHub.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
Tais993: > im not sure if this is possible with the current bot design. i vaguely remember having had problems with that

I don't see how it's an issue, the class doesn't extend anything else, it only implements the interface.

And ListenerAdapter implements EventListener, so don't think there will be an issue.


I only remember issues with SlashCommmands using the adapter, and you not being able to combine that win the ListenerAdapter

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
Taz03: i was thinking about this too

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
mikerasch: Fixed it, thanks for the input. Much better way of getting the event

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
mikerasch: > I'm not saying we shouldn't purge, I'm just saying I do not understand this code, might be because I'm reading from GitHub.

- Get access to read and write to the database
- delete all from help_threads where the help time it was created at is <= the difference between now and the time the thread was created.

mikerasch:
> also y clear db after every 7 days y not remove each thread when it gets closed

Eventually, the database will require a complete purge regardless. I feel like it would help ease the transactions done in the DB if it were to only purge every 30 days. Also, note the database will also be used for other purposes so I think keeping it saved for 30 days would have no harm.

Also, we need to take into account that any user is able to re-open the thread after it was closed which would complicate the process. I see no easy way to store the re-opened thread.

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -133,11 +137,28 @@ private boolean handleIsValidTitle(@NotNull CharSequence title, @NotNull IReplyC
     private @NotNull RestAction<Message> handleEvent(@NotNull InteractionHook eventHook,
             @NotNull ThreadChannel threadChannel, @NotNull Member author, @NotNull String title,
             @NotNull String category, @NotNull Guild guild) {
+        mapUserToDatabase(author, threadChannel);
         return sendInitialMessage(guild, threadChannel, author, title, category)
             .flatMap(any -> notifyUser(eventHook, threadChannel))
             .flatMap(any -> helper.sendExplanationMessage(threadChannel));
     }
 
+    private void mapUserToDatabase(Member author, ThreadChannel threadChannel) {
```
Zabuzard, line 146: the naming on this method is confusing. it is totally missing the context. instead, id go for `writeHelpThreadToDatabase` or similar

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -55,10 +59,10 @@ public final class AskCommand extends SlashCommandAdapter {
      * @param config the config to use
      * @param helper the helper to use
      */
-    public AskCommand(@NotNull Config config, @NotNull HelpSystemHelper helper) {
+    public AskCommand(@NotNull Config config, @NotNull HelpSystemHelper helper, Database database) {
```
Zabuzard, line 62: You forgot to also adjust `ImplicitAskListener`, which also creates help threads.

At that point, instead of duplicating the logic twice, it should be moved into `HelpSystemHelper`.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -133,11 +137,28 @@ private boolean handleIsValidTitle(@NotNull CharSequence title, @NotNull IReplyC
...
+                    .setAuthorId(author.getIdLong())
+                    .setChannelId(threadChannel.getIdLong())
+                    .setCreatedAt(threadChannel.getTimeCreated().toInstant());
+                if (helpThreadsRecord.update() == 0) {
+                    helpThreadsRecord.insert();
+                }
+            });
+        } catch (DatabaseException e) {
+            logger.error("Attempted to add user and thread channel to database failed", e);
+        }
```
Zabuzard, line 159: You dont have to catch this, issues are auto-logged already anyways.

application/src/main/java/org/togetherjava/tjbot/commands/help/LeaversPurgeDatabaseRoutine.java
```diff
@@ -0,0 +1,43 @@
...
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.HelpThreads;
+import java.time.Instant;
+import java.time.Period;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Purge Routine to get rid of old thread creations in the database.
+ */
+public class LeaversPurgeDatabaseRoutine implements Routine {
```
Zabuzard, line 17: This is still called `Leavers...` but now the database table is for help threads in general. Id rename it to `HelpThreadMetadataPurger` or similar

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,69 @@
...
+import org.togetherjava.tjbot.commands.EventReceiver;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.HelpThreads;
+
+import javax.annotation.Nonnull;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Remove all thread channels associated to a user when they leave the guild
```
Zabuzard, line 20: missing dot at the end of the sentence.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,69 @@
...
+    public OnGuildLeaveCloseThreadListener(@NotNull Database database) {
+        this.database = database;
+    }
+
+    @Override
+    public void onGuildMemberRemove(@Nonnull GuildMemberRemoveEvent leaveEvent) {
+        Set<Long> channelIds = getThreadsAssociatedToLeaver(leaveEvent);
+        if (channelIds.isEmpty()) {
+            return;
+        }
```
Zabuzard, line 36: obsolete. looping over an empty collection is effectively doing nothing already anyways

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,69 @@
...
+        for (long channelId : channelIds) {
+            closeThread(channelId, leaveEvent);
+        }
+    }
+
+    public Set<Long> getThreadsAssociatedToLeaver(GuildMemberRemoveEvent leaveEvent) {
```
Zabuzard, line 42: id rename it to `getThreadsCreatedByLeaver`, gives more context to what the association actually is.

Zabuzard, line 42: the method shouldnt take the whole event in, it only needs the user id. it should take the user id instead

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,69 @@
...
+        return database
+            .readTransaction(context -> context.select(HelpThreads.HELP_THREADS.CHANNEL_ID))
+            .from(HelpThreads.HELP_THREADS)
+            .where(HelpThreads.HELP_THREADS.AUTHOR_ID.eq(leaveEvent.getUser().getIdLong()))
+            .fetch()
+            .stream()
+            .map(databaseId -> databaseId.getValue(HelpThreads.HELP_THREADS.CHANNEL_ID))
```
Zabuzard, line 49: unsure, but i feel like jooq has a better way to write this pipeline in a way that u dont have to use another extra mapping that seems to be obsolete.

might want to check other usages of the database in the code base to find something cooler

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
```diff
@@ -0,0 +1,69 @@
...
+            .collect(Collectors.toSet());
+
+    }
+
+    public void closeThread(long channelId, GuildMemberRemoveEvent leaveEvent) {
+        ThreadChannel threadChannel = leaveEvent.getGuild().getThreadChannelById(channelId);
+        if (threadChannel == null) {
+            logger.warn("Attempted to archive thread id: '{}' but could not find thread in guild.",
```
Zabuzard, line 57: which guild? has to be part of the log message (the bot can run in multiple guilds)

application/src/main/resources/db/V11__Add_Help_Channel_Plus_Creator.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE help_threads
```
Zabuzard, line 1: i dont like the name of this migration file. its too specific and easily outdated. rename to `...Add_Help_Thread_Metadata...` maybe

Zabuzard, line 1: You gotta adjust the PP.md file (privacy policy) and add to it what we are logging extra now (associating user id with thread id and implicitly when and how many questions they ask) - for 30 days.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildLeaveCloseThreadListener.java
mikerasch: Ah true, nothing to loop over so nothing happens lol

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

PR #516, opened 2022-08-14
fixed package naming

### LINKED ISSUE #418: `commands` package should be named `features`

Taz03:
closes #418


Zabuzard (changes requested):
Thanks for this. But it is a bad timing for that. Please hold it back until we have less activity going on.

The problem is that this causes heavy merge conflicts for everyone currently working on anything. And in particular for beginners, resolving those conflicts will be quite difficult.

Taz03:
> Thanks for this. But it is a bad timing for that. Please hold it back until we have less activity going on.
> 
> The problem is that this causes heavy merge conflicts for everyone currently working on anything. And in particular for beginners, resolving those conflicts will be quite difficult.

understandable 👍🏻

PR #517, opened 2022-08-15
remove unnecessary checks

### LINKED ISSUE #512: Disable the commands in unvalid channels
Not really a problem, just an enhancement. I saw many commands have usage in specific channel and it managed in the code that in which channel it is visible. Which makes the code longer and hard to understand and maintain.

Use discord Integration settings to manage where the command can be used. This way we can make the code concise and won't loose any functionality.
This also comes with some benefits like right now the bot will send a message saying something along the lines of the commands can't be used here, but if a command is blocked from discord integration settings it won't be even visible in unnecessary channels

**Additional context**
https://github.com/Together-Java/TJ-Bot/blob/e7e98f068cf95aebd3135ac4445a48481e2a05c6/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpCategoryCommand.java#L74-L76

Current Behaviour:-
image

After disabling the command from integration settings:-
image

Taz03:
closes #512

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -24,6 +24,7 @@
    },
    "wolframAlphaAppId": "79J52T-6239TVXHR7",
    "helpSystem": {
+       "stagingChannelPattern": "ask_here",
        "overviewChannelPattern": "active_questions",
        "categories": [
            "Java",
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -24,7 +24,6 @@
    },
    "wolframAlphaAppId": "79J52T-6239TVXHR7",
    "helpSystem": {
-       "stagingChannelPattern": "ask_here",
        "overviewChannelPattern": "active_questions",
        "categories": [
            "Java",
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpCategoryCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpCategoryCommand.java
@@ -71,10 +71,6 @@ public final class ChangeHelpCategoryCommand extends SlashCommandAdapter {
     public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
         String category = event.getOption(CATEGORY_OPTION).getAsString();
 
-        if (!helper.handleIsHelpThread(event)) {
-            return;
-        }
-
         ThreadChannel helpThread = event.getThreadChannel();
         if (helpThread.isArchived()) {
             event.reply("This thread is already closed.").setEphemeral(true).queue();
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
@@ -58,7 +58,7 @@ public final class ChangeHelpTitleCommand extends SlashCommandAdapter {
     public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
         String title = event.getOption(TITLE_OPTION).getAsString();
 
-        if (!helper.handleIsHelpThread(event) || !handleIsValidTitle(title, event)) {
+        if (!handleIsValidTitle(title, event)) {
             return;
         }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/CloseCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/CloseCommand.java
@@ -26,31 +26,22 @@ public final class CloseCommand extends SlashCommandAdapter {
     private static final int COOLDOWN_DURATION_VALUE = 30;
     private static final ChronoUnit COOLDOWN_DURATION_UNIT = ChronoUnit.MINUTES;
 
-    private final HelpSystemHelper helper;
     private final Cache<Long, Instant> helpThreadIdToLastClose;
 
     /**
      * Creates a new instance.
-     *
-     * @param helper the helper to use
      */
-    public CloseCommand(@NotNull HelpSystemHelper helper) {
+    public CloseCommand() {
         super("close", "Close this question thread", SlashCommandVisibility.GUILD);
 
         helpThreadIdToLastClose = Caffeine.newBuilder()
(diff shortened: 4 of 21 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpCategoryCommand.java
```diff
@@ -71,10 +71,6 @@ public ChangeHelpCategoryCommand(@NotNull Config config, @NotNull HelpSystemHelp
     public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
         String category = event.getOption(CATEGORY_OPTION).getAsString();
 
-        if (!helper.handleIsHelpThread(event)) {
-            return;
-        }
-
```
Zabuzard, line 77: Now it could also be used in the overview channel itself, not an option. I.e. we have to at least check that its a thread and not a regular channel.

application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpTitleCommand.java
```diff
@@ -58,7 +58,7 @@ public ChangeHelpTitleCommand(@NotNull HelpSystemHelper helper) {
     public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
         String title = event.getOption(TITLE_OPTION).getAsString();
 
-        if (!helper.handleIsHelpThread(event) || !handleIsValidTitle(title, event)) {
```
Zabuzard, line 61: Now it could also be used in the overview channel itself, not an option. I.e. we have to at least check that its a thread and not a regular channel.

application/src/main/java/org/togetherjava/tjbot/commands/help/CloseCommand.java
```diff
@@ -26,31 +26,22 @@ public final class CloseCommand extends SlashCommandAdapter {
...
-
-        this.helper = helper;
     }
 
     @Override
     public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
-        if (!helper.handleIsHelpThread(event)) {
-            return;
-        }
-
```
Zabuzard, line 53: Now it could also be used in the overview channel itself, not an option. I.e. we have to at least check that its a thread and not a regular channel.

Zabuzard (changes requested):

Zabuzard:
While at it, might also want to remove the role checks on most of the moderation commands. Would be interesting how much we code we can remove there :)

application/src/main/java/org/togetherjava/tjbot/commands/help/ChangeHelpCategoryCommand.java
Taz03: normal members can't send messages in overview channel

DevSerendipity:
How did you figure out what what was an unnecessary check?

Taz03:
> How did you figure out what what was an unnecessary check?

one which can and already is replaced by discord integration settings

application/config.json.template
```diff
@@ -24,7 +24,6 @@
    },
    "wolframAlphaAppId": "79J52T-6239TVXHR7",
    "helpSystem": {
-       "stagingChannelPattern": "ask_here",
```
Zabuzard, line 27: I would like to keep this setting in the config, even if its maybe not used. Just because its part of the help system.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
```diff
@@ -144,11 +137,10 @@ public void onSlashCommand(@NotNull SlashCommandInteractionEvent event) {
     private boolean handleChecks(@NotNull Member bot, @NotNull Member author,
             @Nullable Member target, @NotNull IReplyCallback event) {
         // Member doesn't exist if attempting to audit a user who is not part of the guild.
-        if (target != null && !ModerationUtils.handleCanInteractWithTarget(ACTION_VERB, bot, author,
-                target, event)) {
+        if (target == null) {
             return false;
         }
-        return ModerationUtils.handleHasAuthorRole(ACTION_VERB, hasRequiredRole, author, event);
+        return ModerationUtils.handleCanInteractWithTarget(ACTION_VERB, bot, author, target, event);
```
Zabuzard, line 143: You changed the logic here, this is incorrect. Right now you removed the ability to audit users who left the guild, but the command is supposed to support that.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Taz03:
```java
return !(target != null && !ModerationUtils.handleCanInteractWithTarget(ACTION_VERB, bot, author, target, event))
```
will this work?

Zabuzard: Whatever that is, it is super unreadable. Instead I would do:
```java
// Auditing users outside of the guild always works
if (target == null) {
  return true;
}

return ModerationUtils.handleCanInteractWithTarget(...);
```

application/config.json.template
Zabuzard: (by the way, when you remove something from the config, you would also need to update the actual code `Config.java`)

application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
Taz03: > Whatever that is, it is super unreadable.

agree

application/config.json.template
Taz03: > (by the way, when you remove something from the config, you would also need to update the actual code `Config.java`)

it was in `HelpSystemHelper` no? i did removed it from there

Zabuzard: Right. overlooked that. anyways, thanks for restoring it.

Zabuzard (approved):

marko-radosavljevic (approved):

PR #523, opened 2022-08-22
fixed unknown author bug

### LINKED ISSUE #522: Audit can not retrieve users anymore
The recent merge #472 messed up the ability for `/audit` to retrieve the actual users who used the actions against the target.

See the following example:

example

## Reason

The bug comes from exchanging the previously correct
```java
jda.retrieveUserById(action.authorId())
```
by the wrong
```java
jda.getUserById(action.authorId());
```

Latter only checks the JDA cache instead of asking Discord. The cache is rarely reliable (only for the list of guilds).

Changing this back is fairly simple, but requires to make the code properly async using `RestAction`.

Taz03:
closes #522

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -106,7 +106,7 @@ public final class AuditCommand extends SlashCommandAdapter {
             return TimeUtil.getDateTimeString(instant.atOffset(ZoneOffset.UTC));
         };
 
-        User author = jda.getUserById(action.authorId());
+        User author = jda.retrieveUserById(action.authorId()).complete();
 
         Instant expiresAt = action.actionExpiresAt();
         String expiresAtFormatted = expiresAt == null ? ""


Zabuzard (changes requested):
now you are blocking the thread. thats wrong. you have to do a proper RestAction chain like the code did in the past.

```java
jda.retrieve...(...)
  .map(...)
  .map(...)
  .flatMap(...)
  .queue()
```
u gotta learn how to do proper multithreading when working with such stuff :)

Taz03:
i've used the same method at one more place in the code [here](https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java#L182) should that be changed too?

Zabuzard:
> i've used the same method at one more place in the code [here](https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java#L182) should that be changed too?

oof, yes. you are blocking the JDA thread there.

Taz03:
> now you are blocking the thread. thats wrong. you have to do a proper RestAction chain like the code did in the past.
> 
> ```java
> jda.retrieve...(...)
>   .map(...)
>   .map(...)
>   .flatMap(...)
>   .queue()
> ```
> 
> u gotta learn how to do proper multithreading when working with such stuff :)

as of now i dont know how to do that so maybe someone else should do this

PR #528, opened 2022-08-24
Adding org website

Zabuzard:
## Motivation

We want to have a simple website to represent our community. Mostly for SEO purposes, it should simply link to our Discord and thats it.

## Design

Decided to go with a simple Spring Boot Web application that serves static content, nothing fancy - as simple as it could be.

The logviewer project showed how to create UI with pure Java, but its mostly way too complex for most people here to maintain.  It is more reasonable to go with a simple pure html/css/js site and maybe sprinkle a bit of spring-magic over it, if really needed.

## Website

The site itself is, as of now, just a simple Hello World example. The goal for now is to setup the CI/CD and the VPS correctly, so that the website is actually up and running on the server.

website

Once that is done, we can make the website itself pretty. After that we can also get hands on a top level domain.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -6,3 +6,5 @@ include 'formatter'
 // NOTE The logviewer does not properly work as of today.
 //  But it is causing major build slowdowns, so we exclude it for the time being.
 // include 'logviewer'
+include 'website'
+
--- a/website/build.gradle
+++ b/website/build.gradle
@@ -1,0 +1,27 @@
+plugins {
+    id 'org.springframework.boot' version '2.7.3'
+    id 'io.spring.dependency-management' version '1.0.13.RELEASE'
+    id "com.google.cloud.tools.jib" version "3.1.4"
+    id 'java'
+}
+
+dependencies {
+    implementation 'org.springframework.boot:spring-boot-starter-web'
+}
+
+var outputImage = 'togetherjava.duckdns.org:5001/togetherjava/website:' + System.getenv('BRANCH_NAME') ?: 'latest'
+
+jib {
+    from.image = 'eclipse-temurin:18'
+    to {
+        image = outputImage
+        auth {
+            username = System.getenv('REGISTRY_USER') ?: ''
+            password = System.getenv('REGISTRY_PASSWORD') ?: ''
+        }
+    }
+    container {
+        setPorts(["5051"].asList())
+        setCreationTime(Instant.now().toString())
+    }
+}
--- a/website/src/main/java/org/togetherjava/tjbot/website/Application.java
+++ b/website/src/main/java/org/togetherjava/tjbot/website/Application.java
@@ -1,0 +1,19 @@
+package org.togetherjava.tjbot.website;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+/**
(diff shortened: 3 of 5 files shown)


marko-radosavljevic:
You think it's a good decision to couple our website with the bot? They are clearly two distinct entities, and have nothing in common.

Because when I see the big boy Spring, my immediate association is more serious website than a basic static webpage that would redirect to our discord. :smile: 

If we can foresee a more serious website down the line, it's probably wise to start the project in a new repo, with its own build file and docker container for deployment. That way, we can deploy quick sponsorship and event information, without doing any bot deployments. Also, we can deploy to a separate VPS, if the need arises in the future. Build times would be faster on both projects. We would have separate wiki documentation for our spring website, how to build, how to test, technologies, backend/frontend..

If you think it will just be a minimal static page, it's perfectly fine to be in the tjbot repo. If we need something more out of it, it should be trivial to extract it to a new project. In that case, we wouldn't need spring, just some html + css goodness?

marko-radosavljevic:
Also, if all we need is static website, and we might want to have write-ups about events, sponsorships, java things, or any content whatsoever. And we find html+css cumbersome for that use-case, we can use static website generators.

We can use [jekyll](https://jekyllrb.com/) or [hugo](https://gohugo.io/) that will generate static website from markdown, lots of templates, themes, link checking features, etc. Automates the whole thing for us, and looks pretty.

Zabuzard:
@marko-radosavljevic Good points. Here is my thinking:

1. Separate project

Absolutely agree. But I felt like setting up the whole infrastructure is not trivial and I wanted a fast solution. Like, you need the whole watchtower setup, jib, secrets, ... create proper docker stuff on the VPS and whatever. I was lazy and just wanted to reuse the existing infrastructure 🤷 So maybe we go with this solution for today and then make it proper at a later point?

2. Website complexity

I dont think we should have an actual real website presence. The websites sole purpose is to get indexed high on search engines to lurk people into the Discord server. Similar to what [javadiscord.net](https://javadiscord.net/) did, they are the first entry on google for "java discord" just because of their website that has proper SEO.

So a minimal website that looks proper and has a join Discord button is all we need (for today, and likely also for the future).

3. Technology

I wanted something that is trivial to maintain and most people are familiar with vanilla HTML/CSS/JS. So I thought we just write a simplistic HTML5 page with some modern CSS theme and call it a day.

I consider Vaadin/Logviewer a failure in terms of its original goals - it is not really maintainable by us, even though it is pure Java. It is complex and nobody here (except Krankkk) knows the framework.

Generating something outside of markdown also sounds good, but I really dont want to make this setup complex. So I thought Ill just add the bare minimum to serve static a static HTML/CSS/JS - which Spring gives us trivially in 5 lines of code.

The advantage of Spring would be that it gives us (Java) flexibility in case we actually do want to do more in the future - but I doubt we want.

marko-radosavljevic:
Alright, that seems fair to me. Since gradle project is pretty modular, extracting to a new repo if the need arises shouldn't be too difficult.

Agree on the tech stack and the complexity of the project. A nice static page that redirects to discord is all we need for now. I can see an information board, with current events and sponsors, and maybe a contact form down the line (or just mailto link), and that's about it. Should be pretty simple, static, and easy to maintain and update.

Glad we are on the same page about everything. :relaxed:

marko-radosavljevic (approved):

PR #530, opened 2022-08-24
Fixed attachment error + close input stream correctly

### LINKED ISSUE #525: Gist feature sends wrong requests
## Issue

Our recent gist feature (#491) seems to send incorrect requests to the API, which then fail:

error

This happens many times per day, likely on (almost?) each help message.

## Action

Unfortunately, the log is not too helpful. Proposal:

* try to reproduce it locally, it happens super often per day after all, so maybe it is easy to reproduce by trying out a bit
* if not, we can extend the logging (add the request, add details to find the message that triggered it) - deploy the improved logging and then check again

we should do latter anyways to improve handling future bugs with the feature.

SquidXTV:
# About

Fixes #525 and closes input stream correctly now, cause it produced this before:

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
@@ -185,8 +185,8 @@ public class FileSharingMessageListener extends MessageReceiverAdapter {
 
         if (statusCode < HttpURLConnection.HTTP_OK
                 || statusCode >= HttpURLConnection.HTTP_MULT_CHOICE) {
-            LOGGER.warn("Request JSON: {}", body);
-            throw new IllegalStateException("Gist API unexpected response: " + apiResponse.body());
+            throw new IllegalStateException("Gist API unexpected response: " + apiResponse.body()
+                    + ". Request JSON: " + body);
         }
 
         GistResponse gistResponse;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -179,6 +185,7 @@ private void processAttachments(@NotNull MessageReceivedEvent event,
 
         if (statusCode < HttpURLConnection.HTTP_OK
                 || statusCode >= HttpURLConnection.HTTP_MULT_CHOICE) {
+            LOGGER.warn("Request JSON: {}", body);
             throw new IllegalStateException("Gist API unexpected response: " + apiResponse.body());
```
Zabuzard, line 189: Please put the body into the exception instead. Remember that the bot is multithreaded, so the warning message can drift away from the exceptions log. then you have something like
```java
WARN Request JSON: ...
INFO dasdasdasdas
WARN dasd
INFO asd
INFO asd
ERROR asd
INFO as
WARN das
INFO das
INFO das
ERROR Gist API unexpected response: ...
```
and nobody can connect the dots anymore. Better to put everything into one message instead of creating multiple messages.

Zabuzard (changes requested):

Zabuzard (dismissed):

Zabuzard (approved):

marko-radosavljevic (approved):

PR #540, opened 2022-08-27
Adding help thread auto archiver

Zabuzard:
## Motivation

Discord did some changes to the auto-archive system. They now only _hide_ a channel after the set period and archive it only later (a week). That is a problem for us, since we relied on automatic archival after 24h for our help threads. Without it, they pile up and UX goes down.

## Solution

Since Discord does not automatically close our channels properly anymore, we do it ourselves.

This PR adds a simple routine that periodically checks all help threads and archives them if there was no message sent since 24h.

example

## Other

While at it, added a utility for the `handleRequireOverviewChannel` functionality that has been duplicated around 5 times.

Same with `getActiveThreadsIn` which also has been duplicated a couple of times.

## Tests

The solution is well tested. Here are some edge cases (the archive period has been set to 2 minutes for testing):

* Keep it alive by sending messages
activity

* Reopen after close
closes again


Tais993 (approved):

Zabuzard:

PR #541, opened 2022-08-27
Introduce @ParametersAreNonnullByDefault

### LINKED ISSUE #539: Introduce @ParametersAreNonnullByDefault﻿
After a full year of writing `@NonNull` everywhere, we should move over to `@ParametersAreNonnullByDefault﻿` and only annotate `@Nullable`, for parameters that is.

The annotation goes into all `package-info.java` files on each package. Here is a SO explaining how to do it: https://stackoverflow.com/a/7658232/2411243

It is okay if we integrate it step by step. For example first add it to all existing `package-info.java`, then removing all `@NotNull` on params in these packages and later converting packages that dont have a `package-info.java` yet (because someone was too lazy writing a proper package doc? 😄)

The advantage of this is clearly: Less boilerplate > less to read > easier to understand

Zabuzard:
## Overview

Closes #539.

This cleans up the usage of `@NotNull/@Nonnull/@Nullable` and improves the situation by adding `@ParametersAreNonnullByDefault` to all packages.

The changes are:
* adding `@ParametersAreNonnullByDefault` to all existing `package-info.java` (in application and database)
* removing all `@NotNull` from all method parameters (its the default now)
* replacing all `@NotNull` on return types and similar by `@Nonnull` to consistently use `javax.annotations`
* replacing all `@Nullable` from Jetbrains by `@Nullable` from `javax.annotations`
* adding some missing `@Nonnull/@Nullable` on some return types
* removed some suppressions that do not seem to be important for Sonar and only cause confusion for people who do not have the inspections enabled

Fun fact, with these changes, Sonar even found a missing `@Nullable` in `TopHelperCommand`.

## Review

This is a huge PR, but with very minor and simple changes. It will also confuse a lot of merge conflicts (but simple conflicts). Ideally we merge this PR as soon as possible, as it will get conflicts with each merge while it is open.

For reviewing this, it is only really important to make sure that I neither removed a `@Nullable`, nor replaced one with `@Nonnull` or similar.

It is not important to actually really read the code. Skimming over the changes should be sufficient.

It is possible that some formatting went weird due to the heavy annotation removal + spotless. If you found any odd looking spot, tell in review and I will correct it.

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -39,6 +39,7 @@ shadowJar {
 }
 
 dependencies {
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
     implementation 'org.jetbrains:annotations:23.0.0'
 
     implementation project(':database')
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -2,20 +2,17 @@ package org.togetherjava.tjbot;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.JDABuilder;
-import net.dv8tion.jda.api.requests.GatewayIntent;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
-import org.jetbrains.annotations.NotNull;
+import net.dv8tion.jda.api.requests.GatewayIntent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.togetherjava.tjbot.commands.Features;
+import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.commands.system.BotCore;
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.db.Database;
-import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 
 import javax.security.auth.login.LoginException;
-
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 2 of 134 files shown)


Tais993:
Add a way so sonarcloud flags notnull/nullable from Jetbrains?

Zabuzard:
> Add a way so sonarcloud flags notnull/nullable from Jetbrains?

Good idea. I just checked but didnt find any rule. Unfortunately, we can not add custom rules with Sonar**cloud** (need -qube for that).

So we would need a self-written test in the code using reflection. Not a big deal, but welp...

Mom0aut:
Does the @ParametersAreNonnullByDefault work with public and private methods?

Zabuzard:
> 

It works with all methods in the whole package with the annotation. The way this PR sets it up, all code in the applications and database gradle modules has not-null as default for all method parameters now.

Mom0aut:
> > 
> 
> It works with all methods in the whole package with the annotation. The way this PR sets it up, all code in the applications and database gradle modules has not-null as default for all method parameters now.

Rly nice change, just as curiosity is the possible to add for example @Nullable to private methods or it is now impossible to have null parameters in private methods?

Zabuzard:
> 

Yes. It is just about the implicit default, which is now `@Nonnull`. If you want a nullable parameter, just annotate with `@Nullable`.

Note that its still necessary to always annotate return types - no default for that.

As a side note, these annotations actually dont add checks in the code. You can still pass `null` into something annotated with `@NotNull`. However, clever IDEs like Intellij and tools like Sonar will just heavily shout around - so usually its hard to mess up like that.

Zabuzard:

Mom0aut:
I tried to use the `@ParametersAreNonnullByDefault` in my other project and if i wanna call a method with a null parameter the ide warns me about this null paramater. But in the end the method could be executed with null, is this the expected behavior or should the annotions create a null check and forbid the method call with null?

Zabuzard:
> 

Its expected. The annotations only add IDE warnings and nothing else. They do not actually manipulate the code at all, nor add real security in the code.

An alternative for that would be Lombok.

PR #518, opened 2022-08-19
Only allow media posts in media channels

Mom0aut:
closes #485 

## Config

The config was slightly changed, the following was added:

```json
"mediaOnlyChannelPattern": "memes"
```

PR #544, opened 2022-08-28
Allow linked media posts

### LINKED ISSUE #543: Messages with linked media are blocked
The recent PR #518 added functionality to disallow posts in `#memes` without media attached.

It works well, but there is an edge case it blocks, which it shouldnt. Namely if the message contains an URL (which Discord will then, later, transform into an attachment):

message

The bot shouldnt block these. Its enough to check the message content for `http` and then allow it.

For reference, the affected classes are:
* `MediaOnlyChannelListener`
* `MediaOnlyChannelListenerTest`

Zabuzard:
Closes #543.

## Problem

The recent PR #518 added functionality to disallow posts in `#memes` without media attached.

It works well, but there is an edge case it blocks, which it shouldnt. Namely if the message contains an URL (which Discord will then, later, transform into an attachment):

message

## Fix

This PR adds a simple check whether the message contains `http`, then its not blocked. Easy and effective, all we need. Also added an unit test.

example


Zabuzard:

PR #549, opened 2022-08-29
Add @MethodsReturnNonnullByDefault

### LINKED ISSUE #548: Add @MethodsReturnNonnullByDefault
Similar to #541 , we should take the next step and make our methods also return `@Nonnull` by default. Fortunately, this is fairly simple with a custom annotation:

```java
@Nonnull
@TypeQualifierDefault({ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
public @interface MethodsReturnNonnullByDefault {
}
```
(credits go to Minecraft where I snacked this annotation from - kudos to Quinteger#4169 who raised awareness)

The annotation then goes into all `package-info.java` just like seen in #541 . In the same wave, we should go through the code base and eliminate all explicit `@Nonnull`, hurray 🙌

---

If there is an actual dependency introducing this type already, we should use it instead of _"reinventing the wheel"_.

Zabuzard:
Closes #548. Similar to #541, this adds an annotation `@MethodsReturnNonnullByDefault` to all packages.

Effectively making `@Nonnull` completely obsolete in our code base - hence also removing it everywhere.

The annotation is self-written (idea snacked from Minecraft who has this annotation as well - credits go to Quinteger for raising awareness).

The PR will of course generate a lot of (simple) merge conflicts - again. But I think people will like it 👍

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/BootstrapLauncher.java
+++ b/application/src/main/java/org/togetherjava/tjbot/BootstrapLauncher.java
@@ -20,7 +20,6 @@ public class BootstrapLauncher {
         Application.main(args);
     }
 
-
     /**
      * Sets any system-properties before anything else is touched.
      */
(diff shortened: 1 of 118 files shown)


Zabuzard:
auto merging this on the weekend if no approves come in sooner 👍

Issue #481, opened 2022-08-09
Auto-post advice on empty help threads


Zabuzard:
Sometimes new users create a help thread and then dont post any details in it.

In such a situation, the bot should (after some waiting time, maybe 5 minutes) post a message where it pings the author and asks them to post some details, shortly explaining how the system works.

This could be setup as scheduled task, scheduled by `ImplictAskListener.java` and `AskCommand.java` (via a separate helper class). Where we could schedule this "explain-job" in 5 minutes after creating the threads. The job would then check the thread history to see if there are any messages already (other than from the bot) - and if not, it would post this message.

Also, we would have to setup a small database to memorize help thread channel ID <-> author ID in order to be able to know who to ping.

mikerasch:
We could potentially use the current database implemented where it removes the thread if a member leaves the guild.

Zabuzard:
Yes, thats the plan. Hence the:

> Also, we would have to setup a small database to memorize help thread channel ID <-> author ID in order to be able to know who to ping.

🧠 👍

java-coding-prodigy:
I think I will do this

java-coding-prodigy:
I will make this functionality

marko-radosavljevic:
Any progress on this feature @java-coding-prodigy? ^^

java-coding-prodigy:
Oops I totally forgot about this!
I had worked  on this previously but I can conplete it now.
I was unsure whether I would have to use a ScheduledFuture or if the RestAction API provides the required functionality, could someone please clarify that bit?

Tais993:
Just use a ScheduledExecutor, where you run it after 5 minutes, and in there you do the check whenever you've to post advice.

Zabuzard:
we already have a similar feature that u can copy pasta. we auto-post advice on selecting a category for questions in `ImplicitAskListener`. it runs as a scheduled task, 5 minutes after creation.

PR #554, opened 2022-09-01
Auto posting advice on uncategorized help threads

### LINKED ISSUE #487: Auto-post category advice on uncategorized help threads
Sometimes new users create a help thread and then dont select any category at all.

In such a situation, the bot should (after some waiting time, maybe 5 minutes) post a message where it pings the author and asks them to select a category, shortly explaining how the system works.

This could be setup as scheduled task, scheduled by `ImplictAskListener.java` and `AskCommand.java` (via a separate helper class). Where we could schedule this "explain-job" in 5 minutes after creating the threads.

Also, we would have to setup a small database to memorize help thread channel ID <-> author ID in order to be able to know who to ping.

(This is similar to #481 )

Zabuzard:
## Overview

Closes #487 .

Reminds the user of selecting a category for their help threads if they still dont have one after **5 minutes**:

example

Technically, its solved with a simple scheduled task - scheduled after creation of threads without categories (`ImplicitAskListener`).

## Notes

I put the code for that into `HelpSystemHelper` since the executor service (with 3 threads) should be reused for #481 and similar issues.

The ping is outside of the embed to ensure it actually pings the author, instead of just a soft-ping. But I still wanted to keep using the embed instead of a raw message for the better UI.

Before someone asks during CR, there is no overload for `retrieveThreadChannelById` or similar. JDAs cache seems to be a reliable source for channels, similar to guilds (i.e. JDA seems to receive all updates to them and keeps the cache up2date - unlike for users).

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -1,6 +1,8 @@
 package org.togetherjava.tjbot.commands.help;
 
 import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.MessageBuilder;
 import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.requests.restaction.MessageAction;
@@ -19,6 +21,9 @@ import java.io.InputStream;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -239,6 +250,50 @@ List<ThreadChannel> getActiveThreadsIn(TextChannel channel) {
...
+                        threadChannelId, authorId);
+                return new CompletedRestAction<>(jda, null);
+            }
+
+            // Still no category, send advice
+            MessageEmbed embed = HelpSystemHelper.embedWith(
+                    """
+                            Hey there 👋 You have to select a category for your help thread, otherwise nobody can see your question.
+                            Please use the `/change-help-category` slash-command and pick what fits best, thanks 🙂
+                            """);
```
java-coding-prodigy, line 290: Can't we make this a field? At the very least make the string a field.

Zabuzard: Why? What would be the advantage? Its already very isolated with it being a single argument to the factory.

Zabuzard:
one week, no progress. merging

PR #556, opened 2022-09-02
fixes thread creation warns in media only channels

### LINKED ISSUE #546: Creating a thread in an educational only channel will make the bot falsely send a dm warning to not send normal messages in the channel
**Context**
Educational channels are channel where only images, videos, etc are supposed to be posted, and to discuss them, a thread is supposed to be created to not flood the channel.
Since recently, in order to help mods, the bot will delete and warn the user in a message without an image, link, etc.
The warning is a dm containing an explanation and the message that was deleted.

**Describe the bug**
If an user create a thread in an educational channel, the bot will fasely claim that the "message" shouldn't be here and so will send a warning in a dm to the user, containing the title of the thread.
But, while the bot normaly delete the message, this time, it won't delete the thread (I guess because it can't since it's not a message ?), it's important to note, the bot will only send a warning, it won't delete the thread.
Also note that here, I am specifically explaining things if the user create a thread on an existing message, but if the user create a thread without an existing message, thread creation message will be deleted but the thread won't, so the thread will be unreachable from the channel... so in this case, the thread should be deleted.

**Expected behavior**
An user should not get a warning for creating a thread on an existing message.

**To Reproduce**
- Mouseover a message in an educational channel
- Click on "create a thread" on this message
- Give a title and a first message for the thread
- You should get a warning in dm

And for the alternative, which will delete the thread creation message but not the attached thread which will make it unreachable :
- Click on the "+" symbol at the left of the text area
- Click on "create a thread"
- Give a name and a first message
- The text creation message will be deleted, but not the thread, making it unreachable basically and a warning will be sent in a dm

**Screenshots**
image
image

Taz03:
fixes #546 
more info on the issue

[Screencast from 02-09-22 10:53:12 PM IST.webm](https://user-images.githubusercontent.com/73871477/188206210-b345e5d1-4ef6-419e-b81d-97f96bbc2dd8.webm)

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
@@ -39,7 +39,7 @@ public final class MediaOnlyChannelListener extends MessageReceiverAdapter {
 
         Message message = event.getMessage();
         // Checks if it's a thread creation message. No need to do anything in that case then
-        if (message.getType().equals(MessageType.THREAD_CREATED)) {
+        if (message.getType() == MessageType.THREAD_CREATED) {
             return;
         }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
@@ -4,6 +4,7 @@ import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.MessageBuilder;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.entities.MessageType;
 import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import net.dv8tion.jda.api.requests.RestAction;
 import org.togetherjava.tjbot.commands.MessageReceiverAdapter;
@@ -38,7 +39,7 @@ public final class MediaOnlyChannelListener extends MessageReceiverAdapter {
 
         Message message = event.getMessage();
         // Checks if it's a thread creation message. No need to do anything in that case then
-        if (message.toString().contains("THREAD_CREATED")) {
+        if (message.getType().equals(MessageType.THREAD_CREATED)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -37,39 +35,29 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
-                    .setEmbeds(originalMessageEmbed)
-                    .build();
+        if (message.toString().contains("THREAD_CREATED")) {
+            return;
+        }
 
-        return originalMessage.getAuthor()
-            .openPrivateChannel()
-            .flatMap(channel -> channel.sendMessage(dmMessage));
+        if (message.getAttachments().isEmpty() && message.getEmbeds().isEmpty()
```
Mom0aut, line 43: rly dislike the method deletion and the whole logic into one big clunky section, i prefer that you extract methods

Mom0aut (approved):

application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -37,27 +36,24 @@ public void onMessageReceived(MessageReceivedEvent event) {
             return;
         }
 
-        if (messageHasNoMediaAttached(event)) {
-            deleteMessage(event).flatMap(any -> dmUser(event)).queue();
+        Message message = event.getMessage();
+        // Checks if it's a thread creation message. No need to do anything in that case then
+        if (message.toString().contains("THREAD_CREATED")) {
```
Zabuzard, line 41: Thats not the proper JDA way to test this. Also, in general, never do any business logic based on `toString()`, this method may change its output at any time.

Instead, check the type of the message `message.getType() == ...`. As an example, see the class that cleans up bot messages in the help channel, it can filter thread creation messages already.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -38,7 +39,7 @@ public void onMessageReceived(MessageReceivedEvent event) {
 
         Message message = event.getMessage();
         // Checks if it's a thread creation message. No need to do anything in that case then
```
Zabuzard, line 41: This comment feels obsolete to me. The code is kinda self-explanatory.

Zabuzard (changes requested):

Zabuzard (approved):

Mom0aut (approved):

PR #557, opened 2022-09-02
fixed package names

### LINKED ISSUE #418: `commands` package should be named `features`

Taz03:
fixes #418 

made the package names more reasonable


Zabuzard:
Thanks. But there are too many PRs open right now. In particular, this would get massively in the way of Tais current work. Lets wait for a better moment.

PR #565, opened 2022-09-08
Use slash-command mentions (#559)

### LINKED ISSUE #559: Use slash-command mentions
## Mentions

Discord recently released a new feature called [slash command mentions](https://netcord.site/mention-slash-commands-in-chat/).

mention

The syntax for this is simple, its pure text in a specific format like `</ask:987249686383591447>` where the number is the ID of the command. The ID can be get in developer mode by right clicking the command when about to use it:

copy_id

## Proposal

The bot should make use of this when writing messages that contain such mentions, for example when explaining the help system:

example1

example2

and possibly more. Its likely easy to identify with a text search of `/` in the code base.

### Utility

When implementing, we should add utility to not have to write this syntax all the time. Similar to how we can do `author.getAsMention()`, we should have some sort of `DiscordUtils.mentionSlashCommand("ask", 987249686383591447)`.

We already have such an utility class somewhere, which can be reused for this.

### IDs

A bit annoying, but we can not just hardcode those IDs. They likely have to be part of the `Config.java`/`config.json` file. They might be different for each command/guild/...

However, we should have a quick check whether we can retrieve the ID of a command via JDA. Then we could possibly make this dynamic and add it to our slash commands in-code.

Gyoo:
# Changes in this PR 

- Added method MessageUtils.mentionSlashCommand to convert a command name to a mentioned command using the Guild's commands list

I have found 5 text slash commands in strings, I replaced them all with mentioned commands

Closes #559 , check it for details

Result : 
Discord message with mentioned command

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -314,7 +314,7 @@ public final class HelpSystemHelper {
 
             // Still no category, send advice
             return MessageUtils.mentionSlashCommand(threadChannel.getGuild(),
-                    String.join(HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
+                    String.join(" ", HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
                             HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName()))
                 .flatMap(command -> {
                     MessageEmbed embed = HelpSystemHelper.embedWith(
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -7,6 +7,7 @@ import net.dv8tion.jda.api.entities.channel.Channel;
 import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.entities.channel.middleman.GuildChannel;
+import net.dv8tion.jda.api.entities.channel.middleman.GuildMessageChannel;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
@@ -93,12 +94,15 @@ public final class HelpSystemHelper {
     }
 
     RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
-        return MessageUtils.mentionSlashCommand(threadChannel.getGuild(), CloseCommand.COMMAND_NAME)
-            .flatMap(command -> sendExplanationMessage(threadChannel, command));
+        return MessageUtils
+            .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                    HelpThreadCommand.Subcommand.CLOSE.getCommandName())
+            .flatMap(closeCommandMention -> sendExplanationMessage(threadChannel,
+                    closeCommandMention));
     }
 
     private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel,
-            String command) {
+            String closeCommandMention) {
         boolean useCodeSyntaxExampleImage = true;
         InputStream codeSyntaxExampleData =
                 AskCommand.class.getResourceAsStream("/" + CODE_SYNTAX_EXAMPLE_PATH);
@@ -124,7 +128,7 @@ public final class HelpSystemHelper {
                                     With enough info, someone knows the answer for sure."""),
                 HelpSystemHelper.embedWith(
                         "Don't forget to close your thread using the command %s when your question has been answered, thanks."
-                            .formatted(command)));
+                            .formatted(closeCommandMention)));
 
         MessageCreateAction action = threadChannel.sendMessage(message);
         if (useCodeSyntaxExampleImage) {
@@ -307,16 +311,24 @@ public final class HelpSystemHelper {
             }
 
             // Still no category, send advice
-            MessageEmbed embed = HelpSystemHelper.embedWith(
-                    """
-                            Hey there 👋 You have to select a category for your help thread, otherwise nobody can see your question.
-                            Please use the `/help-thread change category` slash-command and pick what fits best, thanks 🙂
-                            """);
-            MessageCreateData message = new MessageCreateBuilder().setContent(author.getAsMention())
-                .setEmbeds(embed)
-                .build();
+            return MessageUtils
+                .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                        HelpThreadCommand.CHANGE_SUBCOMMAND,
+                        HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
+                .flatMap(command -> {
+                    MessageEmbed embed = HelpSystemHelper.embedWith(
+                            """
+                                    Hey there 👋 You have to select a category for your help thread, otherwise nobody can see your question.
+                                    Please use the %s slash-command and pick what fits best, thanks 🙂
+                                    """
+                                .formatted(command));
+                    MessageCreateData message =
+                            new MessageCreateBuilder().setContent(author.getAsMention())
+                                .setEmbeds(embed)
+                                .build();
 
-            return threadChannel.sendMessage(message);
+                    return threadChannel.sendMessage(message);
+                });
         }).queue();
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -92,7 +92,7 @@ public final class HelpSystemHelper {
         categoryRoleSuffix = helpConfig.getCategoryRoleSuffix();
     }
 
-    RestAction<Message> sendExplanationMessage(Guild guild, MessageChannel threadChannel) {
+    RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
         boolean useCodeSyntaxExampleImage = true;
         InputStream codeSyntaxExampleData =
                 AskCommand.class.getResourceAsStream("/" + CODE_SYNTAX_EXAMPLE_PATH);
@@ -117,8 +117,11 @@ public final class HelpSystemHelper {
                                     **provide details**, context, more code, examples and maybe some screenshots. \
                                     With enough info, someone knows the answer for sure."""),
                 HelpSystemHelper.embedWith(
-                        "Don't forget to close your thread using the command **%s close** when your question has been answered, thanks."
-                            .formatted(MessageUtils.mentionSlashCommand(guild, "help-thread"))));
+                        "Don't forget to close your thread using the command %s when your question has been answered, thanks."
+                            .formatted(MessageUtils
+                                .mentionSlashCommand(threadChannel.getGuild(),
+                                        CloseCommand.COMMAND_NAME)
+                                .complete())));
 
         MessageCreateAction action = threadChannel.sendMessage(message);
         if (useCodeSyntaxExampleImage) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -17,6 +17,7 @@ import net.dv8tion.jda.api.utils.messages.MessageCreateData;
 import net.dv8tion.jda.internal.requests.CompletedRestAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import org.togetherjava.tjbot.commands.utils.MessageUtils;
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.config.HelpSystemConfig;
@@ -95,8 +96,9 @@ public final class HelpSystemHelper {
 
     RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
         return MessageUtils
-            .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
-                    HelpThreadCommand.Subcommand.CLOSE.getCommandName())
+            .mentionSlashCommand(threadChannel.getGuild(),
+                    String.join(" ", HelpThreadCommand.COMMAND_NAME,
+                            HelpThreadCommand.Subcommand.CLOSE.getCommandName()))
             .flatMap(closeCommandMention -> sendExplanationMessage(threadChannel,
                     closeCommandMention));
     }
@@ -311,10 +313,9 @@ public final class HelpSystemHelper {
             }
 
             // Still no category, send advice
-            return MessageUtils
-                .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
-                        HelpThreadCommand.CHANGE_SUBCOMMAND,
-                        HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
+            return MessageUtils.mentionSlashCommand(threadChannel.getGuild(),
+                    String.join(HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
+                            HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName()))
                 .flatMap(command -> {
                     MessageEmbed embed = HelpSystemHelper.embedWith(
                             """
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -93,6 +93,12 @@ public final class HelpSystemHelper {
     }
 
     RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
+        return MessageUtils.mentionSlashCommand(threadChannel.getGuild(), CloseCommand.COMMAND_NAME)
+            .flatMap(command -> sendExplanationMessage(threadChannel, command));
+    }
+
+    private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel,
+            String command) {
         boolean useCodeSyntaxExampleImage = true;
         InputStream codeSyntaxExampleData =
                 AskCommand.class.getResourceAsStream("/" + CODE_SYNTAX_EXAMPLE_PATH);
@@ -118,10 +124,7 @@ public final class HelpSystemHelper {
                                     With enough info, someone knows the answer for sure."""),
                 HelpSystemHelper.embedWith(
                         "Don't forget to close your thread using the command %s when your question has been answered, thanks."
-                            .formatted(MessageUtils
-                                .mentionSlashCommand(threadChannel.getGuild(),
-                                        CloseCommand.COMMAND_NAME)
-                                .complete())));
+                            .formatted(command)));
 
         MessageCreateAction action = threadChannel.sendMessage(message);
         if (useCodeSyntaxExampleImage) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -97,8 +97,8 @@ public final class HelpSystemHelper {
     RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
         return MessageUtils
             .mentionSlashCommand(threadChannel.getGuild(),
-                    String.join(" ", HelpThreadCommand.COMMAND_NAME,
-                            HelpThreadCommand.Subcommand.CLOSE.getCommandName()))
+                    HelpThreadCommand.COMMAND_NAME,
+                            HelpThreadCommand.Subcommand.CLOSE.getCommandName())
             .flatMap(closeCommandMention -> sendExplanationMessage(threadChannel,
                     closeCommandMention));
     }
@@ -314,8 +314,8 @@ public final class HelpSystemHelper {
 
             // Still no category, send advice
             return MessageUtils.mentionSlashCommand(threadChannel.getGuild(),
-                    String.join(" ", HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
-                            HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName()))
+                    HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                            HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
                 .flatMap(command -> {
                     MessageEmbed embed = HelpSystemHelper.embedWith(
                             """
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -96,9 +96,8 @@ public final class HelpSystemHelper {
 
     RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
         return MessageUtils
-            .mentionSlashCommand(threadChannel.getGuild(),
-                    HelpThreadCommand.COMMAND_NAME,
-                            HelpThreadCommand.Subcommand.CLOSE.getCommandName())
+            .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                    HelpThreadCommand.Subcommand.CLOSE.getCommandName())
             .flatMap(closeCommandMention -> sendExplanationMessage(threadChannel,
                     closeCommandMention));
     }
@@ -313,9 +312,10 @@ public final class HelpSystemHelper {
             }
 
             // Still no category, send advice
-            return MessageUtils.mentionSlashCommand(threadChannel.getGuild(),
-                    HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
-                            HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
+            return MessageUtils
+                .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                        HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                        HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
                 .flatMap(command -> {
                     MessageEmbed embed = HelpSystemHelper.embedWith(
                             """
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -16,7 +16,7 @@ import net.dv8tion.jda.api.utils.messages.MessageCreateData;
 import net.dv8tion.jda.internal.requests.CompletedRestAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.togetherjava.tjbot.commands.utils.MessageUtils;
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.config.HelpSystemConfig;
 import org.togetherjava.tjbot.db.Database;
@@ -92,7 +92,7 @@ public final class HelpSystemHelper {
         categoryRoleSuffix = helpConfig.getCategoryRoleSuffix();
     }
 
-    RestAction<Message> sendExplanationMessage(MessageChannel threadChannel) {
+    RestAction<Message> sendExplanationMessage(Guild guild, MessageChannel threadChannel) {
         boolean useCodeSyntaxExampleImage = true;
         InputStream codeSyntaxExampleData =
                 AskCommand.class.getResourceAsStream("/" + CODE_SYNTAX_EXAMPLE_PATH);
@@ -117,7 +117,8 @@ public final class HelpSystemHelper {
                                     **provide details**, context, more code, examples and maybe some screenshots. \
                                     With enough info, someone knows the answer for sure."""),
                 HelpSystemHelper.embedWith(
-                        "Don't forget to close your thread using the command **/help-thread close** when your question has been answered, thanks."));
+                        "Don't forget to close your thread using the command **%s close** when your question has been answered, thanks."
+                            .formatted(MessageUtils.mentionSlashCommand(guild, "help-thread"))));
 
         MessageCreateAction action = threadChannel.sendMessage(message);
         if (useCodeSyntaxExampleImage) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
@@ -119,15 +119,15 @@ public final class ImplicitAskListener extends MessageReceiverAdapter {
         String threadDescription = lastHelpThread == null ? "your previously created help thread"
                 : lastHelpThread.getAsMention();
 
-        message.getChannel()
-            .sendMessage("""
-                    %s Please use %s to follow up on your question, \
-                    or use %s to ask a new questions, thanks.""".formatted(author.getAsMention(),
-                    threadDescription,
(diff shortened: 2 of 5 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
```diff
@@ -181,24 +182,23 @@ private static MessageAction sendInitialMessage(ThreadChannel threadChannel,
...
-            .sendMessage(
-                    """
-                            %s Please use `/ask` to ask questions. Don't worry though, I created %s for you. \
-                            Please continue there, thanks."""
-                        .formatted(message.getAuthor().getAsMention(),
-                                threadChannel.getAsMention()));
+            .sendMessage("""
+                    %s Please use %s to ask questions. Don't worry though, I created %s for you. \
+                    Please continue there, thanks.""".formatted(message.getAuthor().getAsMention(),
+                    MessageUtils.mentionSlashCommand(message.getGuild(), "ask"),
```
Zabuzard, line 200: Id like to be able to introduce a compile-time dependency between this and the actual command. So that we dont have the raw string `"ask"` floating around here.

I.e. ideally there should be some form of linking this with the actual `AskCommand` which knows its name and has a `getName()` method.

To make this nice, we probably need a non-trivial rework and Im not sure if u can follow my thoughts, but:
* some sort of class that knows all commands and their names `CommandMentionService`
* all commands have to be registered there (for example by `BotCore`)
* the service is then created in `Features.java` like the other helpers
* people who need it, can then request it in their constructor params
* and then simply do `commandMentionService.mention(AskCommand.class)`

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
         return afterEscape.replace("\\```", "\\`\\`\\`");
     }
 
+    /**
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord
```
Zabuzard, line 64: missing dot at the end of the sentence

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
...
+     *
+     * @param guild guild that owns the command
+     * @param command command name
+     * @return Formatted string for mentioned slash command
```
Zabuzard, line 68: missing "for [the] mentioned"

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
...
+        AtomicLong commandId = new AtomicLong();
+        guild.retrieveCommands()
+            .queue(commands -> commands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(command))
+                .findFirst()
+                .ifPresentOrElse(c -> commandId.set(c.getIdLong()), () -> {
+                    throw new IllegalArgumentException("Command %s does not exist in guild %s"
+                        .formatted(command, guild.getId()));
+                }));
+        return String.format("</%s:%d>", command, commandId.get());
```
Zabuzard, line 82: I do see one problem with this design, which we should address. The method makes an actual request to JDA/Discord, but blocks on it.

We should avoid blocking and instead return a `RestAction<String>` out of this. I know that this will uglify the call-sites and make it more complex, but we dont have much choice here.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
...
+    public static String mentionSlashCommand(Guild guild, String command) {
+        AtomicLong commandId = new AtomicLong();
+        guild.retrieveCommands()
+            .queue(commands -> commands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(command))
+                .findFirst()
+                .ifPresentOrElse(c -> commandId.set(c.getIdLong()), () -> {
+                    throw new IllegalArgumentException("Command %s does not exist in guild %s"
+                        .formatted(command, guild.getId()));
+                }));
```
Zabuzard, line 81: This would be more readable and idiomatic if split into:
```java
.map(c -> commandId.set(...))
.orElseThrow(() -> new IllegalArgumentException(...))
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
Zabuzard: Btw, sorry if this is too much. I definitely did not anticipate this when tagging the issue with "good first issue". So its totally okay if you want someone else to take over from here.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -83,7 +84,7 @@ public HelpSystemHelper(JDA jda, Config config, Database database) {
         categoryRoleSuffix = helpConfig.getCategoryRoleSuffix();
     }
 
-    RestAction<Message> sendExplanationMessage(MessageChannel threadChannel) {
+    RestAction<Message> sendExplanationMessage(Guild guild, MessageChannel threadChannel) {
```
Tais993, line 87: I feel like using a GuildMessageChannel, and removing Guild would be nicer.

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
Tais993: Or we add a constant with the name to AskCommand? By far simplest, and I personally find adding a CommandMentionService sound overkill

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
         return afterEscape.replace("\\```", "\\`\\`\\`");
     }
 
+    /**
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord
+     *
+     * @param guild guild that owns the command
```
Tais993, line 66:
```suggestion
     * @param guild the {@link Guild} that contains the command
```

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,27 @@ public static String escapeMarkdown(String text) {
...
+     * @param command command name
```
Tais993, line 67:
```suggestion
     * @param command the command's name
```

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Tais993: This doesn't block, this code actually simply doesn't work.
Queue runs after the `return`  statement

But yes, rewriting it so it returns a `RestAction<String>`  would be better

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
Zabuzard: Yes. We discussed this in the server and Gyoo came up with the same idea 👍

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Gyoo: Made a change to return a `RestAction<String>` as suggested, but then how should I handle it on the caller side ? Right now I've used `MessageUtils.mentionSlashCommand(...).complete()` but it makes it blocking on the caller's side instead of inside the method, so I'm quite unsure this is what you expect.

Zabuzard: ur supposed to have a single chain. all actions will be mapped on it. at the end of everything, there will be a `queue()` at some point, which sends out the whole chain of actions.

Zabuzard:
is it done again? or should i wait with a full CR?

Gyoo:
I'd like a review of my work on my latest commit as I'm not entirely sure I've done things the right way. If this is good for you we can carry on. I'll have to fix a few sonar issues though.

Zabuzard:
im not really sure which aspect of it ur unsure about. it looks fine to me. all u did was making it `RestAction` and then using `RestAction` accordingly on all call-sites (if they havent used them yet). which is what i expected for this.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -83,7 +84,13 @@ public HelpSystemHelper(JDA jda, Config config, Database database) {
...
     }
 
-    RestAction<Message> sendExplanationMessage(MessageChannel threadChannel) {
+    RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel) {
+        return MessageUtils.mentionSlashCommand(threadChannel.getGuild(), CloseCommand.COMMAND_NAME)
+            .flatMap(command -> sendExplanationMessage(threadChannel, command));
+    }
+
+    private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadChannel,
+            String command) {
```
Zabuzard, line 93: the name `command` is confusing. what command? just call it `closeCommandMention`. then its clear what it is. and also what its purpose is.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,25 @@ public static String escapeMarkdown(String text) {
...
+     * @param command the command's name
+     * @return Formatted string for the mentioned slash command
+     * @throws IllegalArgumentException when the command name doesn't match with the guild's
+     *         commands
+     */
+    public static RestAction<String> mentionSlashCommand(Guild guild, String command) {
+        return guild.retrieveCommands().map(commands -> {
+            Command foundCommand = commands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(command))
+                .findFirst()
```
Zabuzard, line 76: use `findAny()`. its potentially faster. and u dont care which it picks, since u dont know the order either.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,25 @@ public static String escapeMarkdown(String text) {
...
+                .orElseThrow(() -> new IllegalArgumentException(
+                        "Command %s does not exist in guild %s".formatted(command, guild.getId())));
+            return String.format("</%s:%d>", command, foundCommand.getIdLong());
```
Zabuzard, line 79: never rely on the toString representation of objects. JDA might change in the next update what `command.toString()` gives and then this will break.

Instead, build the string you expect urself. I suspect u intended to do `command.getName()` here or similar.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
```diff
@@ -114,13 +114,15 @@ private boolean handleIsNotOnCooldown(Message message) {
...
-                .formatted(author.getAsMention(), threadDescription))
-            .flatMap(any -> message.delete())
+        MessageUtils.mentionSlashCommand(message.getGuild(), AskCommand.COMMAND_NAME)
+            .map(command -> message.getChannel()
+                .sendMessage("""
+                        %s Please use %s to follow up on your question, \
+                        or use %s to ask a new questions, thanks."""
+                    .formatted(author.getAsMention(), threadDescription, command))
+                .flatMap(any -> message.delete()))
             .queue();
```
Tais993, line 124: Use flatmap instead of a map, with a flatmap inside the map
```suggestion
            .flatMap(command -> message.getChannel()
                .sendMessage("""
                        %s Please use %s to follow up on your question, \
                        or use %s to ask a new questions, thanks."""
                    .formatted(author.getAsMention(), threadDescription, command)))
             .flatMap(any -> message.delete()))
            .queue();
```

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Tais993: I'd personally also use toString here, at least, what is the purpose for toString if it's not logging?

Zabuzard: im not talking about the log/error message. but the `String.format(..., command, ...)`, which is application logic.

Tais993: OWH nvm yes

the to string method returns the type & name & id, so does that even work?

Gyoo: The `command` parameter here comes from the static field `COMMAND_NAME` in the caller's classes, so it's not related to JDA. I think the relevant thing to do would be the change the field's name to something more precise

Zabuzard: ah `command` vs `foundCommand`... yeah... prime example of bad variable naming haha. u got us both 😝

Gyoo: Naming fields is always difficult 😅
I'll try to find something more relevant when I have time to work on that again

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -64,19 +64,21 @@ public static String escapeMarkdown(String text) {
      * in Discord.
      *
      * @param guild the {@link Guild} that contains the command
-     * @param command the command's name
+     * @param commandInput the command's name
```
Zabuzard, line 67: u forgot to adjust the description

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -64,19 +64,21 @@ public static String escapeMarkdown(String text) {
...
-                .filter(c -> c.getName().equalsIgnoreCase(command))
-                .findFirst()
-                .orElseThrow(() -> new IllegalArgumentException(
-                        "Command %s does not exist in guild %s".formatted(command, guild.getId())));
-            return String.format("</%s:%d>", command, foundCommand.getIdLong());
+    public static RestAction<String> mentionSlashCommand(Guild guild, String... commandInput) {
+        String commandWithOptions = String.join(" ", commandInput);
+        return guild.retrieveCommands().map(guildCommands -> {
+            Command guildCommand = guildCommands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(commandInput[0]))
```
Zabuzard, line 76: This method is easy to get wrong. Someone could think its correct to use it as `mentionSlashCommand(guild, "foo bar", "baz")` and then it would fail since it assumes the first argument to be the base name.

Also, it actually has a bug. It accesses `[0]` without length checking, nor did it require a length of > 1. I would change it to:
```java
mentionSlashCommand(Guild guild, String baseCommandName, String... subcommands)
```
or keep it a single string `String fullCommand` and then split on space for extraction urself.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -64,19 +64,21 @@ public static String escapeMarkdown(String text) {
...
+                .findAny()
+                .orElseThrow(
+                        () -> new IllegalArgumentException("Command %s does not exist in guild %s"
```
Zabuzard, line 79: the error message can be hard to read with, considering u have spaces in the command now. add a single quotes around it, such as `'%s'`. then its easier to read.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Gyoo: Agreed, this signature makes it easier to understand

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +60,31 @@ public static String escapeMarkdown(String text) {
...
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param baseCommandName the command's name
+     * @param subcommands optional subcommands, depending on the base command used
+     * @return Formatted string for the mentioned slash command
+     * @throws IllegalArgumentException when the command name doesn't match with the guild's
+     *         commands
+     */
+    public static RestAction<String> mentionSlashCommand(Guild guild, String baseCommandName,
+            String... subcommands) {
```
Taz03, line 75: y u taking multiple subcommands?

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: a command can only have one subcommand no? i mean a mentioned command can only have one subcommand, then y multiple???

Gyoo: A Subcommand group can have Subcommands itself. For the `help-thread` command, there is a `change` group, that has `category` and `title` bound to it. However, it also has a `close` subcommand, so here's a list of possible outcomes :
`/help-thread change category`
`/help-thread change title`
`/help-thread close`

That's why you can have a list of "subcommands"

Taz03: so `change` is actually a subcommandgroup not subcommand, what u should do is take only one input as `commandPath` which is the full path, this is also the name jda uses so it'll be a better option then u'll get `help-thread change category` as one input then u can split it and do ur thing.

Taz03: looking at ur implementation it should be an easy change, other things looks good to me

Zabuzard: kinda agree. its a bit hard to give this a nice and non-confusing signature without going the `fullCommand` approach. Maybe its indeed easier to just go that route instead.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +60,31 @@ public static String escapeMarkdown(String text) {
...
+        List<String> fullCommand = new ArrayList<>(List.of(baseCommandName));
+        fullCommand.addAll(List.of(subcommands));
+        return guild.retrieveCommands().map(guildCommands -> {
+            Command guildCommand = guildCommands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(baseCommandName))
+                .findAny()
+                .orElseThrow(
+                        () -> new IllegalArgumentException("Command '%s' does not exist in guild %s"
+                            .formatted(String.join(" ", fullCommand), guild.getId())));
```
Tais993, line 84: It only checks for the base command

So this should be

```suggestion
                        () -> new IllegalArgumentException("Command '%s' does not exist in guild %s"
                            .formatted(baseCommandName, guild.getId())));
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +60,27 @@ public static String escapeMarkdown(String text) {
...
+    /**
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord.
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param fullCommand the command's name with its optional subcommand
+     * @return Formatted string for the mentioned slash command
+     * @throws IllegalArgumentException when the command isn't found in the guild
+     */
+    public static RestAction<String> mentionSlashCommand(Guild guild, @NotNull String fullCommand) {
```
Taz03, line 72: remove notnull annotation and instead of `fullCommand` name it `commandPath`

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +60,27 @@ public static String escapeMarkdown(String text) {
...
+        List<String> fullCommandSplit = List.of(fullCommand.split(" "));
+        return guild.retrieveCommands().map(guildCommands -> {
+            Command guildCommand = guildCommands.stream()
+                .filter(c -> c.getName().equalsIgnoreCase(fullCommandSplit.get(0)))
+                .findAny()
+                .orElseThrow(
+                        () -> new IllegalArgumentException("Command '%s' does not exist in guild %s"
+                            .formatted(fullCommandSplit.get(0), guild.getId())));
+            return String.format("</%s:%d>", String.join(" ", fullCommandSplit),
+                    guildCommand.getIdLong());
```
Taz03, line 82:
```suggestion
            return String.format("</%s:%d>", commandPath, guildCommand.getId());
```

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +60,27 @@ public static String escapeMarkdown(String text) {
...
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord.
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param fullCommand the command's name with its optional subcommand
+     * @return Formatted string for the mentioned slash command
+     * @throws IllegalArgumentException when the command isn't found in the guild
+     */
+    public static RestAction<String> mentionSlashCommand(Guild guild, @NotNull String fullCommand) {
+        List<String> fullCommandSplit = List.of(fullCommand.split(" "));
```
Taz03, line 73: y r u making it a list?

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: y even have all these just do
```java
String commandName = commandPath.split(" ")[0];
```
u never need the other 2 parts (if any)

Zabuzard: ... in which case u should limit the `split`, i.e. `split(" ", 1)` 😛 @Taz03

Taz03: yea this too

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -65,21 +64,20 @@ public static String escapeMarkdown(String text) {
...
      * @param guild the {@link Guild} that contains the command
-     * @param fullCommand the command's name with its optional subcommand
+     * @param commandPath the command's name with its optional subcommand
      * @return Formatted string for the mentioned slash command
      * @throws IllegalArgumentException when the command isn't found in the guild
      */
-    public static RestAction<String> mentionSlashCommand(Guild guild, @NotNull String fullCommand) {
-        List<String> fullCommandSplit = List.of(fullCommand.split(" "));
+    public static RestAction<String> mentionSlashCommand(Guild guild, String commandPath) {
+        String commandName = commandPath.split(" ", 1)[0];
```
Zabuzard, line 72: a bit fragile. what if there is no space? like when u do `mentionSlashCommand(guild, "ban")`. does it still work then or crash? it should be safe.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Gyoo: Good thing you double checked on this. I did a bunch of tests in JShell and it turns out `split(" ", 1)` doesn't split anything. However, `split(" ", 2)[0]` will extract the first word of the command path properly. If the command is only one word, it works fine too. Considering this, what would you suggest to make it safer ? I can only think about a null check.

Zabuzard: null check is obsolete, since your method param is implicitly annotated with `@Nonnull`.

If that `[0]` is always safe, im fine with it

Taz03: > what if there is no space?

yes it'll work in that case

Taz03: it'll just return the word

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,26 @@ public static String escapeMarkdown(String text) {
...
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord.
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param commandPath the command's name with its optional subcommand
+     * @return Formatted string for the mentioned slash command
+     * @throws IllegalArgumentException when the command isn't found in the guild
+     */
+    public static RestAction<String> mentionSlashCommand(Guild guild, String commandPath) {
+        String commandName = commandPath.split(" ", 2)[0];
```
Taz03, line 72: y 2?

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,26 @@ public static String escapeMarkdown(String text) {
         return afterEscape.replace("\\```", "\\`\\`\\`");
     }
 
+    /**
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord.
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param commandPath the command's name with its optional subcommand
```
Taz03, line 67:
```suggestion
     * @param commandPath full command path with subcommand groups and subcommands
```

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: > so it basically does nothing

what u want to split in a one word string anyways

Taz03: if theres nothing to split it'll return an array with the string as first element and thats what we are expecting here so change it back to 1
```suggestion
        String commandName = commandPath.split(" ", 1)[0];
```

Gyoo: But then `c.getName().equalsIgnoreCase(commandName)` doesn't work anymore (line 75) if the command path has more than one word. That's why we need to extract the first word in all cases.

Taz03: ohh yea, i'll just just remove the limit at this point

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -311,10 +313,9 @@ private void executeUncategorizedAdviceCheck(long threadChannelId, long authorId
             }
 
             // Still no category, send advice
-            return MessageUtils
-                .mentionSlashCommand(threadChannel.getGuild(), HelpThreadCommand.COMMAND_NAME,
-                        HelpThreadCommand.CHANGE_SUBCOMMAND,
-                        HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
+            return MessageUtils.mentionSlashCommand(threadChannel.getGuild(),
+                    String.join(HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
```
Zabuzard: this is missing the `" "`?

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
```diff
@@ -185,9 +185,9 @@ private static RestAction<Message> sendInitialMessage(ThreadChannel threadChanne
             .build();
 
         return MessageUtils
-            .mentionSlashCommand(originalMessage.getGuild(), HelpThreadCommand.COMMAND_NAME,
-                    HelpThreadCommand.CHANGE_SUBCOMMAND,
-                    HelpThreadCommand.Subcommand.CHANGE_CATEGORY.getCommandName())
+            .mentionSlashCommand(originalMessage.getGuild(),
+                    String.join(HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND,
```
Zabuzard: same

Zabuzard: i dont like that everyone has to do `String.join(" ", ...)` now. and the fact that it already lead u to two bugs by forgetting `" "`.

We should probably go back and add a vararg or List overload (sorry for the back and forth) - what do u think?

Zabuzard (changes requested):
can u have a quick double-check whether we added new string-mentions by now? (have a quick look over the recently merged PRs maybe)

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/ImplicitAskListener.java
Gyoo: Well I just did a full search on `String.join` and I still managed to miss some, so other people will probably miss that too when adding new command mentions in the future. I'm fine going back to varargs.

Gyoo:
> can u have a quick double-check whether we added new string-mentions by now? (have a quick look over the recently merged PRs maybe)

Went through all commits since early September, couldn't find anything new.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -68,16 +68,16 @@ public static String escapeMarkdown(String text) {
      * @return Formatted string for the mentioned slash command
      * @throws IllegalArgumentException when the command isn't found in the guild
      */
-    public static RestAction<String> mentionSlashCommand(Guild guild, String commandPath) {
-        String commandName = commandPath.split(" ", 2)[0];
+    public static RestAction<String> mentionSlashCommand(Guild guild, String... commandPath) {
+        String commandName = commandPath[0];
```
Zabuzard, line 72: now it can crash if not given any value though. u gotta make the first one mandatory by `String commandName, String... subCommands` or similar

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Gyoo: Right, sorry I forgot about that

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -64,19 +64,24 @@ public static String escapeMarkdown(String text) {
...
         return guild.retrieveCommands().map(guildCommands -> {
             Command guildCommand = guildCommands.stream()
                 .filter(c -> c.getName().equalsIgnoreCase(commandName))
                 .findAny()
                 .orElseThrow(
                         () -> new IllegalArgumentException("Command '%s' does not exist in guild %s"
                             .formatted(commandName, guild.getId())));
+            String commandPath = commandName;
+            if (subCommands.length > 0) {
+                commandPath = commandPath + " " + String.join(" ", subCommands);
```
Zabuzard, line 83: use `+=`

Zabuzard (changes requested):

Zabuzard (dismissed):

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -56,4 +59,31 @@ public static String escapeMarkdown(String text) {
         return afterEscape.replace("\\```", "\\`\\`\\`");
     }
 
+    /**
+     * Converts a slash command text to a mentioned slash command, which you can directly click on
+     * in Discord.
+     *
+     * @param guild the {@link Guild} that contains the command
+     * @param commandName the command's name
+     * @param subCommands optional subcommands, depending on the base command used
```
Tais993, line 68:
```suggestion
     * @param subCommands optional subcommand group & subcommand, depending on the base command used
```

Zabuzard (approved):

Taz03 (approved):

Tais993 (approved):

PR #574, opened 2022-09-18
Grouped help thread commands

### LINKED ISSUE #532: Group the help thread commands
# Group the help thread commands

by using sub slash commands and slash command groups

```
/help-thread change title
/help-thread change category
/help-thread close
/help-thread pin-message
...
```
Advantages:-
1. It is clear that these commands are help-thread commands and will only be used there
2. Easy to maintain (e.g. only one command to set to help threads in discord integration settings)

Taz03:
closes #532
more info on the issue


Zabuzard (changes requested):
They shouldnt be merged into a single class. Prefer doing it like `TagManageCommand` which has them still all separated

Taz03:
> They shouldnt be merged into a single class. Prefer doing it like `TagManageCommand` which has them still all separated

isn't that is in a single class too?

Taz03:
> What is this `BOOKMARK` stuff? Please do not add stuff from a different PR/issue to this

ok 😟

Zabuzard:
> > What is this `BOOKMARK` stuff? Please do not add stuff from a different PR/issue to this
> 
> ok 😟

Sorry, but its better to have things separated. U can branch off this branch for the other PR, if u want to not have to wait.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+ * Implements the {@code /help-thread} command, which are contains special command for help threads
+ * only
+ */
+public final class HelpThreadCommand extends SlashCommandAdapter {
+    private static final int COOLDOWN_DURATION_VALUE = 30;
+    private static final ChronoUnit COOLDOWN_DURATION_UNIT = ChronoUnit.MINUTES;
+    private static final String CATEGORY = "category";
+    private static final String TITLE = "title";
+    private static final String CLOSE = "close";
+    private static final String BOOKMARK = "bookmark";
```
Zabuzard, line 39: These constants miss **what** they are and hence are confusing. For example `CATEGORY_SUBCOMMAND`, `TITLE_SUBCOMMAND`, `CLOSE_SUBCOMMAND`.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private static final Supplier<Cache<Long, Instant>> newCaffeine = () -> Caffeine.newBuilder()
+        .maximumSize(1_000)
+        .expireAfterAccess(COOLDOWN_DURATION_VALUE, TimeUnit.of(COOLDOWN_DURATION_UNIT))
+        .build();
+
+    private final HelpSystemHelper helper;
+    private final Cache<Long, Instant> helpThreadIdToLastCategoryChange = newCaffeine.get();
+    private final Cache<Long, Instant> helpThreadIdToLastTitleChange = newCaffeine.get();
+    private final Cache<Long, Instant> helpThreadIdToLastClose = newCaffeine.get();
```
Zabuzard, line 49: This doesnt scale well and is kinda ugly. Id introduce an enum for ur subcommands (3 as of now) and then create a `Map<Subcommand, Cache<...>>` or similar.

The idea is to reduce the locations one has to adjust when introducing a new subcommand

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Creates a new instance.
+     *
+     * @param config the config to use
+     * @param helper the helper to use
+     */
+    public HelpThreadCommand(Config config, HelpSystemHelper helper) {
+        super("help-thread", "Help thread specific commands", SlashCommandVisibility.GUILD);
+
```
Zabuzard, line 59: "Help thread specific commands" reads a bit odd. Maybe `"Commands for help threads, such as changing the category"`

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+                new SubcommandData(BOOKMARK, "Sends a dm linking thins help thread"));
+
+        this.helper = helper;
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        ThreadChannel helpThread = event.getThreadChannel();
+        if (helpThread.isArchived()) {
+            event.reply("This thread is already closed.").setEphemeral(true).queue();
```
Zabuzard, line 83: This might reopen the channel. Double-check that ur not doing something not intended here

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+                }
+                close(event, helpThread);
+            }
+            case BOOKMARK -> {
+                bookmark(event, helpThread);
+            }
+            default -> {
+                // This can never be the case
+            }
+        }
```
Zabuzard, line 127: This switch is way too complex. It should simply forward to the corresponding method and thats it.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        ThreadChannel helpThread = event.getThreadChannel();
+        if (helpThread.isArchived()) {
+            event.reply("This thread is already closed.").setEphemeral(true).queue();
+            return;
+        }
+
+        boolean isOnCooldown = false;
```
Zabuzard, line 87: Move the whole cooldown logic up here, or into an extra method. It shouldnt be mixed with the stuff below.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+            }
+            case BOOKMARK -> {
+                bookmark(event, helpThread);
+            }
+            default -> {
+                // This can never be the case
+            }
+        }
+
+        if (isOnCooldown) {
```
Zabuzard, line 129: Right now ur code does the change even if its on cooldown. But sends this message extra on top, after doing the changes...
Instead, u have to check this before and end the method before u do any changes.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,212 @@
...
+        return Optional.ofNullable(helpThreadIdToLastAction.getIfPresent(helpThread.getIdLong()))
+            .map(lastAction -> lastAction.plus(COOLDOWN_DURATION_VALUE, COOLDOWN_DURATION_UNIT))
+            .filter(Instant.now()::isBefore)
+            .isPresent();
+    }
+
+    private void changeCategory(SlashCommandInteractionEvent event, ThreadChannel helpThread) {
+        String category = event.getOption(CATEGORY).getAsString();
+
+        helpThreadIdToLastCategoryChange.put(helpThread.getIdLong(), Instant.now());
```
Zabuzard, line 150: All methods have this in common, the duplication should be removed by moving this stuff a level up.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: ig it doesn't really matter as the end user is not gonna see this anyways when using the commands

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: yes it does, what to do

Taz03: yes it does, actually it never really gets triggered, coz when someone uses a command that makes a archived thread unarchived. So ig we should remove this

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: should i make a new interface and all the subcommand will extend that interface??

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: or u reclose it. otherwise people might accidentally open threads

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: no. just move it up a level in the call hierarchy

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -158,7 +124,21 @@ private void changeCategory(SlashCommandInteractionEvent event, ThreadChannel he
     }
 
     private void changeTitle(SlashCommandInteractionEvent event, ThreadChannel helpThread) {
-        String title = event.getOption(TITLE).getAsString();
+        if (isHelpThreadOnCooldown(helpThreadIdToLastTitleChange, helpThread)) {
+            sendCooldownMessage(event);
+            return;
+        }
```
Zabuzard, line 130: all commands have this duplicated. move it one up in the hierarchy

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: > or u reclose it. otherwise people might accidentally open threads

how am i gonna know if it was closed before using the command??

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: > no. just move it up a level in the call hierarchy

i don't understand can u give some outline

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: good point. u probably cant. alright, then we dont deal with it extra.

anyways, i would still keep this check, in case discord changes the ux to not allow slash commands when its closed already. as it was in the past.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: u have method A() and method A() calls method B(), C() and D(). and B(), C() and D() do the same thing X. so move X from B, C, D one level up into A(), and the duplication is gone.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: i don't think thats how we should do things. If discord does that then we'll update our code accordingly

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: so move everything in the switch statements??

Taz03: by everything i mean the logic

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: a new subcommand might won't have a cooldown and won't need cache

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: it was like that just a few weeks ago though. chances are that its maybe just a bug and they recover the old behavior

Taz03: > it was like that just a few weeks ago

was it?? from what i remember discord always unarchives an archived threads if someone sends something in it

Zabuzard: im 100% that it was like that. otherwise i wouldnt have coded this in the first place. the code, including this feature, was tested.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: the duplicate part has to go one up. im not telling u how to write it exactly, thats ur job. im telling u to get rid of code duplication. or should i take over?

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: i dont see a problem with that. just dont include such a future command in the map then.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: > or should i take over?

yes, please

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: fixed

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: fixed

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: fixed

Zabuzard (dismissed):

Taz03:
wow u changed a lot, looking nice

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        ThreadChannel helpThread = event.getThreadChannel();
+
+        Subcommand invokedSubcommand = Arrays.stream(Subcommand.values())
+            .filter(subcommand -> subcommand.getCommandName().equals(event.getSubcommandName()))
+            .findAny()
+            .orElseThrow();
```
Tais993, line 96: This is pretty bad if we're talking about performance, I also find this to be overcomplicated.

I much rather see an EnumMap with a subcommand name -> Subcommand instance relation.

It'd make this a simple

```suggestion
Subcommand invokedSubcommand  = subcommands.get(event.getSubcommandName());
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -0,0 +1,185 @@
...
+
+        getData().addSubcommands(new SubcommandData(CLOSE_SUBCOMMAND, "Close this help thread"));
+
+        this.helper = helper;
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        ThreadChannel helpThread = event.getThreadChannel();
+
```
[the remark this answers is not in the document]
Taz03: what if the user used change category command, this will still check the cooldown of close command

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Taz03: what is `subcommands` here?

Tais993: An enum map, where the key is the name and the value is the instancee

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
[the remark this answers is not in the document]
Tais993: So you'd want to make a method that returns the Map based on the subcommands name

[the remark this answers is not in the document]
Zabuzard: lolwhat. seems i forgot to push a change. oopsie

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: thats fair. i can do that

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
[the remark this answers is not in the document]
Zabuzard: wait. is this a github anomaly? i dont see any of the duplication u mentioned anymore @Tais993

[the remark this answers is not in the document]
Tais993: There's no duplication anymore, you fixed it, this was just a quick comment with the intention to explain what the duplication was exactly

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: done, please check

Zabuzard (dismissed):

Zabuzard (approved):

Tais993 (approved):

PR #579, opened 2022-09-21
Forward logs to Discord

Zabuzard:
## Overview

This is a convenience replacement for the (broken/deactivated) Logviewer project.

It forwards all logs to a Discord channel, via webhooks.

example

## Info/Error channel

The day to day routine will be that we want to be notified when an error/warning happened, but not for info logs. Therefore, we set up 2 channels with separate webhooks. One for errors, one for infos:

channel split

Now, users can mute one channel and keep notifications on for the other.

## Security

Note that logs can contain sensitive information (for example when banning an user). Therefore, a new log marker `LogMarkers.SENSITIVE` has been added. Logs with that marker are not forwarded to Discord. All existing logs have been checked and adjusted. Looks something like this:

```java
logger.warn(LogMarkers.SENSITIVE, "Banned user {} for reason: {}", user, reason);
```

### Queue

The implementation is relatively straightforward, but has some non trivial complexity. In a nutshell, it buffers all incoming logs in a `PriorityQueue` (ordered on the timestamp to ensure correct log order). The queue has a max size of `10_000` logs and ideally is never hit in practice (this is to protect RAM, in case).

The queue is then processed each 5 seconds with a `ScheduledExecutorService`. The service then polls and sends a maximum of 10 logs in a batch request to Discord.

### Failures

If the request fails, the logs are put back into the queue. If a batch fails 3 times, it is discarded to not entirely block the whole mechanism forever.

### Rate limitting

When we send too much, Discord hits us with rate limits (`429` http status). When this happens, we extract the end of the limitation from the `retry-after` header and skip the `ScheduledExecutorService` calls until `Instant.now()` is after that time.

### Logging the logger

The whole system also has logging/error reporting itself. Its logs all have a `LogMarkers.NO_DISCORD` though, to prevent that it forwards its own logs - which could lead to a potential infinite loop of spawning logs in an erroneous scenario.

## Config

The config has been changed, add:

```json
"logInfoChannelWebhook": "<put_your_webhook_here>",
"logErrorChannelWebhook": "<put_your_webhook_here>"
```

## Checklist

* [x] Appender
* [x] Add to config
* [x] Configurable webhook
* [x] Proper embeds
* [x] Error handling
* [x] Rate limitting / queue / buffer
* [x] Split info/error channel
* [x] Javadoc


application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,228 @@
...
+        LogMessage log = LogMessage.ofEvent(event);
+
+        synchronized (pendingLogsLock) {
+            pendingLogs.add(log);
+        }
+    }
+
+    private void processPendingLogs() {
+        try {
+            if (rateLimitExpiresAt != null) {
```
Mom0aut, line 111: could u just return if `rateLimitExpiresAt == null` ?

Taz03: thats a different thing

Zabuzard: The method has to do extra logic in all cases:
* if its not null, there is rate limitting and we either
  * have to return if its still active
  * reset it if its expired already
* otherwise we execute the actual forwarding

Zabuzard: i agree though, the nested if can be hard to comprehend at first sight. ill create a helper method

Zabuzard: pushed, have a check

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -108,14 +108,9 @@ void forwardLogEvent(LogEvent event) {
...
         try {
-            if (rateLimitExpiresAt != null) {
-                if (Instant.now().isBefore(rateLimitExpiresAt)) {
-                    // Still rate limited, try again later
-                    return;
-                }
-
-                // Rate limit has expired, try again
-                rateLimitExpiresAt = null;
+            if (handleIsRateLimitActive()) {
```
Mom0aut, line 111: like the method extraction but maybe the name `isRateLimitActive` is better suited 😄

marko-radosavljevic: we use this `handle` prefix a lot in the codebase, the intention is to make it clear that the method does more than a simple `is/has` check.

in this case, there is a side effect with `rateLimitExpiresAt` which you wouldn't expect from a method that _checks if the rate limit is active_ (and nothing else).

marko-radosavljevic: I agree that it reads weird, especially if you are not used to it, but it's the convention we went with. And it's better to be a bit weird, than to be a liar, I guess ^^

Zabuzard: yeah. i couldnt come up with a better name, but `isRateLimitActive()` would be misleading and thats dangerous. but having the method vs not having it at all is definitely better, it did improve the code readability.

maybe someone else has a better name idea. it does:
* `isRateLimitActive`
* but additionally maintains the internal state for this.
point is, it must be called every now and then. if people just check the field themselves instead of using the method, they will see outdated information.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,236 @@
...
+ * Logs are forwarded in correct order, based on their timestamp. They are not forwarded
+ * immediately, but at a fixed schedule in batches of {@value MAX_BATCH_SIZE} logs.
+ * <p>
+ * Failed logs are repeated {@value MAX_RETRIES_UNTIL_DISCARD} times until eventually discarded.
+ * <p>
+ * Although unlikely to hit, the class maximally buffers {@value MAX_PENDING_LOGS} logs until
+ * discarding further logs. Under normal circumstances, the class can easily handle high loads of
+ * logs.
+ */
+final class DiscordLogForwarder {
```
Tais993, line 40: This feels like a matter of reinventing the wheel, you're just making a smaller Discord webhooks library, maintaining this can only be done by you, or other expert programmers that want to spend their time into figuring out what you exactrly did.

Why did you not just use an existing library, that does the rate limiting and such for you?

Zabuzard: > Why did you not just use an existing library, that does the rate limiting and such for you?

Unless you find the wheel I am reinventing here, I am not reinventing anything. I couldnt find anything like that.

But maybe I just looked for the wrong stuff, I was specifically looking for SLF4J2 Log Appenders for Discord. And I only found 2 very unsatisfying implementations (which also did not address this aspect).

If you can link me anything that can reduce the complexity in this code, I would be very glad to replace it.

Zabuzard: That said, I did my best to make this class maintainable by others. I dont think its _that bad_. It has a few flaws, yes, but it should be doable.

Tais993: It's not that bad, but I still believe that using a library will help.

I'm not talking about a Logback <-> Discord library, buy just a Discord webhook library (https://github.com/MinnDevelopment/discord-webhooks)

This will remove the need of rate limiting, and all the Http stuff.

Zabuzard: i wasnt aware of such libs. thanks for letting me know. this is a good idea and ill check it out

Zabuzard: done. recheck please

Zabuzard:
a week old, a semi approval. merging this

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,154 @@
...
+     * This method is thread-safe.
+     *
+     * @param event the log to forward
+     */
+    void forwardLogEvent(LogEvent event) {
+        if (pendingLogs.size() >= MAX_PENDING_LOGS) {
+            logger.warn(LogMarkers.NO_DISCORD,
+                    """
+                            Exceeded the max amount of logs that can be buffered. \
+                            Logs are forwarded to Discord slower than they pile up. Discarding the latest log...""");
```
Tais993, line 80: I feel like it should be logged if the logs are starting to pile up already, before logs start to get discarded

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,154 @@
...
+            logger.warn(LogMarkers.NO_DISCORD,
+                    "Unknown error when forwarding pending logs to Discord.", e);
+        }
+    }
+
+    private List<LogMessage> pollLogsToProcessBatch() {
+        int batchSize = Math.min(pendingLogs.size(), MAX_BATCH_SIZE);
+        synchronized (pendingLogsLock) {
+            return Stream.generate(pendingLogs::remove).limit(batchSize).toList();
+        }
```
Tais993, line 112: See comment above, add such handling here for that?

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,154 @@
...
+        private static String abbreviate(String text, int maxLength) {
+            if (text.length() < maxLength) {
+                return text;
+            }
+
+            if (maxLength < 3) {
+                return text.substring(0, maxLength);
+            }
+
+            return text.substring(0, maxLength - 3) + "...";
```
Tais993, line 146: I don't understand why the behaviour changes if the maxLength is above 3

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static final int MAX_PENDING_LOGS = 10_000;
+    private static final int MAX_BATCH_SIZE = WebhookMessage.MAX_EMBEDS;
+    private static final ScheduledExecutorService SERVICE =
+            Executors.newSingleThreadScheduledExecutor();
+    /**
+     * Has to be small enough for fitting all {@value MAX_BATCH_SIZE} embeds contained in a batch
+     * into the total character length of ~6000.
+     */
+    private static final int MAX_EMBED_DESCRIPTION = 1_000;
```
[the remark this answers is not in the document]
Zabuzard: We are sending multiple embeds in a single message (10). And Discord seems to apply a limit to the total length of all embed contents summed up.

Without that `...` support that I added, you get error messages by the API saying that the total length is 6000.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Zabuzard: Thats fair and makes sense. I just consider all of this a really low priority. The logs in Discord arent supposed to be a point of truth.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Zabuzard: It had to do with the expected UX for that `...` stuff on low limits. Lets say u want to abbreviate `hello` to `3`. Then you dont want it to become just `...` but `hel`. For `4` u want `h...`.

That way it ensures that the message is `maxLength` long. 🤷

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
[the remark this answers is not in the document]
Tais993: Your math doesn't really work-out here

10 embeds, 1k max per embed

This means 10k max

but you get an error stating that 6k is the max.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Tais993: While I understand that the logs aren't supposed to be a point of truth.
I still don't want to just "discard logs randomly"

Even better would be, to log *on Discord* stating that logs are getting piled up. 

This way it actually gets noticed.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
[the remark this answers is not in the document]
Zabuzard: yes. thats intended. logs rarely go above the limit, just sometimes.

its okay if its not accepted by discord and errors out. happens very rarely.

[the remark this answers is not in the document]
Tais993: Yea I'm uncertain on this one.

If I were in your shoes, I'd imagine you'd tell me exactly what I just told you.

I don't like the "it'll be fine, doesn't matter", but I do understand where it's coming from.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Tais993: 👍

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Zabuzard: It gets noticed by mods regardless, since we have warning-alerting on the VPS itself. But yes, this is a good suggestion.

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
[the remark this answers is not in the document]
Zabuzard: yeah. i can see myself saying exactly this to myself as well 😆

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
Zabuzard: fixed in #607 by sending a warning log message when reaching a threshold of `8_000` (hard limit is `10_000`)

application/src/main/java/org/togetherjava/tjbot/logging/discord/DiscordLogForwarder.java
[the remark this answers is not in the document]
Zabuzard: fixed in #607 by shortening the embeds hard enough to fit into the limit - if the limit is reached

PR #580, opened 2022-09-22
fixed bot positing advice on closed uncategorized help threads

### LINKED ISSUE #571: Bot sends advice on closed uncategorized help threads
Bot sends advice on uncategorized help threads, even after the thread gets closed. This makes it open again

image

Taz03:
fixes #571 

Currently the bot sends a advice to select a category on help threads with no category.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -275,7 +275,7 @@ public final class HelpSystemHelper {
             }
 
             if (threadChannel.isArchived()) {
-                return null;
+                return new CompletedRestAction<>(jda, null);
             }
 
             Optional<String> category = getCategoryOfChannel(threadChannel);
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -274,6 +274,10 @@ public final class HelpSystemHelper {
                 return new CompletedRestAction<>(jda, null);
             }
 
+            if (threadChannel.isArchived()) {
+                return null;
+            }
+
             Optional<String> category = getCategoryOfChannel(threadChannel);
             if (category.isPresent()) {
                 logger.debug(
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -274,6 +274,10 @@ private void executeUncategorizedAdviceCheck(long threadChannelId, long authorId
                 return new CompletedRestAction<>(jda, null);
             }
 
+            if (threadChannel.isArchived()) {
+                return null;
```
Zabuzard, line 278: `return null` is the wrong action. this leads to the action crashing with an exception in JDA. instead, do it exactly like the `if` right before this one, with an already-completed action.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
Taz03: yes, i wanted to ask this. Why i need to return something at all?? when the method returns void shouldn't it be just `return;`??

Zabuzard: ur not returning in the method. ur returning in the lambda thats used for the `flatMap` thats expecting a `RestAction<Message>`.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
```diff
@@ -274,6 +274,10 @@ private void executeUncategorizedAdviceCheck(long threadChannelId, long authorId
                 return new CompletedRestAction<>(jda, null);
             }
 
+            if (threadChannel.isArchived()) {
+                return new CompletedRestAction<>(jda, null);
```
Zabuzard, line 278: please add a debug-level log, similar to the one above. otherwise its hard for us to debug.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
Taz03: ohh yea i see, thanks

Zabuzard (approved):

Mom0aut (approved):

PR #583, opened 2022-09-23
Changing auto archiving from 24 to 12h

### LINKED ISSUE #582: Reduce auto archive time from 24h to 12h
As proposed in suggestions, we should reduce the auto archive time from 24h of inactivity to 12h.

We expect that this will reduce the amount of actually inactive threads, while hopefully not closing too much of still valid threads. Ideally, this is a better sweetspot than 24h.

Zabuzard:
Closes #582 .

As proposed in suggestions, we should reduce the auto archive time from 24h of inactivity to 12h.

We expect that this will reduce the amount of actually inactive threads, while hopefully not closing too much of still valid threads. Ideally, this is a better sweetspot than 24h.


Zabuzard:
not much to talk about this, code-wise. and the decision has been discussed and approved in discord already. so ill just merge it

PR #587, opened 2022-09-27
Migrate to JDA alpha 20

### LINKED ISSUE #586: Update JDA to v5.0.0-alpha.20
In particular for the forum migration, we want to update to the latest JDA:

https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-alpha.20

There are a few internal reworks which could cause some minor migration work here and there.

Also see #536 which hints changes in the
* Emote system
* Message system
* Command visibility system

Zabuzard:
## Overview

Closes #586 .

This migrates our JDA dependency from alpha 9 to the latest **alpha 20**. JDA had several breaking changes that required work on our side. The changes are done in multiple steps (see commits):

### 9 -> 10

They introduced a new overload for `event.getOption(...)` that accepts a fallback value directly. So instead of checking for `null`, we can provide the fallback directly. Adjusted in `RoleSelectCommand`.

The max amount of embeds (10) is now a constant, using that instead of hardcoding in `WolframAlphaHandler`.

Breaking changes on moderation methods, such as `guild.ban(user)`. Those methods now dont accept a pure ID anymore, but require a proper user object. Fortunately, this was trivial to adjust since we always had those at hand already:
* `TemporaryMuteAction`
* `TemporaryQuarantineAction`

### 10 -> 13

Breaking changes for emojis. They are now not called `Emote` anymore, but `Emoji`. And a few minor changes on the API. Biggest changes in `SuggestionsUpDownVoter`, mostly just changing `Emote` to `RichCustomEmoji` and similar. Just some simple renamings.

Slash command visibility has now to be set on the command itself, using `setGuildOnly`. Done so directly in `BotCommandAdapter`.

### 13 -> 18

Massive breaking changes from 17 to 18. They split the hierarchy of `Message`, `MessageBuilder` and everything that belongs to that. Essentially, the class was logically split in 3:
* messages from discord to client, `Message`
* messages from bot to discord, `MessageCreateAction`
* existing messages from bot to discord, `MessageEditAction`

This change bleeds heavily into a lot of classes. For example, all builders had to be replaced by either `MessageCreateBuilder` or `MessageEditBuilder`. But not only that, some classes did both (edit and create) and now had to be properly split (in some cases, generics worked quite fine here).

The biggest hit however is the whole `JdaTester` mocking setup for our unit tests. A lot of stuff had to be adjusted there.

Also, `ActionRow` was basically inlined to `List<Button>` and file uploads like `message.addFile(data, name)` where replaced with a wrapper `message.addFiles(FileUpload.fromData(data, name))` and a few more subtle, but breaking changes like these here and there.

### 18 -> 19

`LoginException` was replaced by `InvalidTokenException`.

The `guild.ban(...)` and `kick(...)` methods have been deprecated to slightly change their usage. `ban` now takes a `TimeUnit` instead of implicitly always being `TimeUnit.DAYS` and `reason` has been moved out to `.ban(...).reason(reason)` instead of being directly in the `ban/kick` call.

Apart from that, this change introduces massive noise on the imports. They essentially moved around all `Channel` classes.

### 19 -> 20

No breaking changes for us. This adds `RestAction#onSuccess`, which replaces one ugly `.<Void>map(... -> { ... return null;})` occurence 👍

## Checklist

Details at: https://github.com/DV8FromTheWorld/JDA/releases

* [x] 5 -> 6
* [x] 6 -> 10
* [x] 10 -> 13
* [x] 13 -> 18
* [x] Fix tests with 18
* [x] 18 -> 19
* [x] 19 -> 20


application/src/main/java/org/togetherjava/tjbot/commands/BotCommandAdapter.java
```diff
@@ -51,7 +51,7 @@ public abstract class BotCommandAdapter implements BotCommand {
      * @param visibility the visibility of the command
      */
     protected BotCommandAdapter(CommandData data, CommandVisibility visibility) {
-        this.data = Objects.requireNonNull(data, "The data shouldn't be null");
+        this.data = data.setGuildOnly(visibility == CommandVisibility.GUILD);
```
Taz03, line 54: y u removed the non null check?

Taz03 (changes requested):

Tais993 (approved):
Me when no issues

(although gotta fix the removed requireNonNull)

image

Zabuzard:
one approve, one week old, merging

PR #558, opened 2022-09-03
Command rework, attempt number 2

Tais993:
# Changes in this PR

### Added interfaces
BotCommand (contains logic for every Discord command)
MessageContextCommand
UserContextCommand

### Added adapters
BotCommandAdapter

### Added general classes
CommandReloadUtil (reload command has been deleted, commands will be updated automatically every restart)

### Renamed
SlashCommandVisibility to CommandVisibility
SlashCommandProvider to CommandProvider

For more info, see #368

PR #588, opened 2022-09-29
Added ComponentIdInteractor delegate

Zabuzard:
## Overview

This adds a new class called `ComponentIdInteractor`, which effectively is a delegate-pattern solution to provide component-ID generation based out of a class that can be used as delegate for composition, instead of relying on the existing `BotCommandAdapter` inheritance.

Essentially, the existing code to generate component IDs has been moved out of `BotCommandAdapter` into this class instead. `BotCommandAdapter` now **delegates** to `ComponentIdInteractor` for this functionality.

## Advantage

The benefit is that non-`BotCommand` classes, in particular `Routine`s can now easily access this functionality as well. For example, a routine might now easily create buttons like this:

```java
public class FooRoutine implements Routine, UserInteractor {
  // Create the delegate
  private final ComponentIdInteractor componentIdInteractor = new ComponentIdInteractor(getInteractionType(), getName());

  // Stuff from Routine
  @Override
  public Schedule createSchedule() { ... }

  @Override
  public void runRoutine(JDA jda) {
    ...
    // Create some button using the delegate, and later attach it to some message
    Button button = Button.success(
      componentIdInteractor.generateComponentId("John"),
      "click me");
    ...
  }

  // Stuff from UserInteractor
  @Override
  public String getName() {
    return "foo";
  }

  @Override
  public UserInteractionType getInteractionType() {
    return UserInteractionType.OTHER;
  }

  @Override
  public void acceptComponentIdGenerator(ComponentIdGenerator generator) {
    componentIdInteractor.acceptComponentIdGenerator(generator); // delegate
  }

  @Override
    System.out.println("Hello " + args.get(0)); // Hello John
  }

  @Override
  public void onSelectionMenu(SelectMenuInteractionEvent event, List<String> args) {
    // Not interested in
  }
}
```

## Redesign prefix

While at it, the whole topic with `UserInteractorPrefix` was redesigned. It is now essentially renamed to `UserInteractionType` and will not only be used for prefixes, but essentailly just represents the type of interaction.

This is then used by `BotCore` for prefix-based-forwarding. Implementations of `UserInteractor` now have to explicitly tell their type instead of the system being based on the classes and reflection.

This gives maximum flexibility. At the same time, it is still convenient though, since everything is hidden behind the adapters for 99% of the use cases. `BotCommandAdapter` now auto-infers the type based on the JDA command types, for the commands its responsible for (slash, user context, message context).

## Checklist

* [x] Implement
* [x] Let `ScamBlocker` use the delegate
* [x] Wait for #558 
* [x] redesign prefix stuff


Zabuzard:
(FYI @SquidXTV )

application/src/main/java/org/togetherjava/tjbot/commands/componentids/ComponentIdInteractor.java
```diff
@@ -0,0 +1,92 @@
...
+        return name;
+    }
+
+    /**
+     * Must be used before generating component IDs with {@link #generateComponentId(String...)} and
+     * similar.
+     * <p>
+     * It will provide the interactor a component ID generator through this method, which can be
+     * used to generate component IDs, as used for button or selection menus. See
+     * {@link SlashCommand#onSlashCommand(SlashCommandInteractionEvent)} for details on how to use
```
Tais993, line 48: In the command rework the explanation gets moved to the generateComponentId method.

So either add a TODO comment / wait for the command rework to update.

Zabuzard: ill wait, yeah.

Zabuzard: done, recheck

application/src/main/java/org/togetherjava/tjbot/commands/componentids/ComponentIdInteractor.java
```diff
@@ -0,0 +1,92 @@
...
+     * this.
+     *
+     * @param generator the provided component ID generator
+     */
+    public void acceptComponentIdGenerator(ComponentIdGenerator generator) {
+        this.generator = generator;
```
Tais993, line 50: null check

Tais993 (commented):
The issue with the prefix enum before was that the component ID generation needed the prefixed name.

But that still is the case, so what have you actually changed?

application/src/main/java/org/togetherjava/tjbot/commands/BotCommandAdapter.java
```diff
@@ -54,17 +52,17 @@ protected BotCommandAdapter(CommandData data, CommandVisibility visibility) {
...
-        this.name = data.getName();
-        this.type = data.getType();
+        type = data.getType();
+        componentIdInteractor = new ComponentIdInteractor(data.getName());
     }
 
     @Override
     public final String getName() {
-        return name;
+        return componentIdInteractor.getName();
```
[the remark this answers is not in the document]
Zabuzard: same. i already redesigned. its not part of the component-id-interactors job to provide a name. it needs a name, but shouldnt provide it.

the componentIdInteractor does add prefixes, thats its job. therefore, the botcommandadapter doesnt have to think about this anymore since it just delegates.

application/src/main/java/org/togetherjava/tjbot/commands/componentids/ComponentIdInteractor.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/BotCommandAdapter.java
```diff
@@ -54,17 +49,27 @@ protected BotCommandAdapter(CommandData data, CommandVisibility visibility) {
...
+    }
+
+    private static UserInteractionType commandTypeToInteractionType(Command.Type type) {
+        return switch (type) {
+            case SLASH -> UserInteractionType.SLASH_COMMAND;
+            case USER -> UserInteractionType.USER_CONTEXT_COMMAND;
+            case MESSAGE -> UserInteractionType.MESSAGE_CONTEXT_COMMAND;
+            case UNKNOWN -> throw new IllegalArgumentException(
+                    "Can not infer the type of user interaction for this command, it is unknown.");
+        };
```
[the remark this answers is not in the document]
Zabuzard: @Tais993 i intentionally did it like that, since its not meant to cover all `UserInteractionType`s, only the ones that `BotCommandAdapter` is interested in.

Like, `ComponentIdInteractor` wants a `UserInteractionType` and `BotCommandAdapter` has to provide one. In this particular case, the adapter decided to extract that out of the `CommandData`. So its its choice and job to do that.

On the other side, `UserInteractionType` can contain more enum entries that have nothing to do with this and are not compatible with that. Its not a 1:1 translation. In fact, it already contains `Other` which cant be mapped by this.

application/src/main/java/org/togetherjava/tjbot/commands/componentids/ComponentIdInteractor.java
Zabuzard: yes. done.

PR #596, opened 2022-09-30
fixed context commands not working

Tais993:
I might forgot to test it properly after copying the initial PR, slash commands worked but forgot to check context commands (stupid, bad example I agree)

So therefor a fix: this also contains minor improvements of BotCore and UserInteractorPrefix

In BotCore:
Removed requireSlashCommand, and added requireUserInteractor. This works based on generics, very useful.

Also added UserInteractorPrefix#getPrefixedName(String)


Zabuzard (approved):
Tested, can confirm it works.

PR #597, opened 2022-09-30
Bugfix command completion not working

Tais993:
In BotCore:
Added onCommandAutoCompleteInteraction listener, this way autocompletion events will actually get forwarded.

Funny, didnt think of this during the previous PR, was probably too hasty.

(maybe this was all planned for free commits \s

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
@@ -2,6 +2,7 @@ package org.togetherjava.tjbot.commands.system;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Channel;
+import net.dv8tion.jda.api.events.interaction.command.CommandAutoCompleteInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.UserContextInteractionEvent;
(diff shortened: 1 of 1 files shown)


Taz03 (approved):

Taz03:
works, tested it :+1:

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -241,6 +242,18 @@ public void onSlashCommandInteraction(SlashCommandInteractionEvent event) {
                     .onSlashCommand(event));
     }
 
+    @Override
+    public void onCommandAutoCompleteInteraction(
+            @NotNull final CommandAutoCompleteInteractionEvent event) {
```
Taz03, line 247: remove not null
```suggestion
    @Override
    public void onCommandAutoCompleteInteraction(
            final CommandAutoCompleteInteractionEvent event) {
```

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
```diff
@@ -241,6 +242,18 @@ public void onSlashCommandInteraction(SlashCommandInteractionEvent event) {
...
+        String name = event.getName();
+
+        logger.debug("Received auto completion from command-subcommand '{}-{}' (#{}) on guild '{}'",
+                name, event.getSubcommandName(), event.getId(), event.getGuild());
```
Taz03, line 251: what will happen if the command doesn't have a subcommand??

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/system/BotCore.java
Tais993: Good point, I should replace this with getPath or whatever it's called (shows command, groups and subcommand)

If anyone can do this it'd be highly appreciated, not home till tomorrow evening.

Taz03 (approved):

illuminator3 (approved):
LGTM

PR #598, opened 2022-10-01
added auto-complete for tag command

### LINKED ISSUE #594: Add auto-completion to /tag
Slashcommands now have autocompletion support.

`/tag` is the perfect candidate for this and should support it.

Therefore, simply implement `onAutoComplete` which is already avaialble to `SlashCommandAdapter` and incorperate existing tags into it.

Taz03:
Added auto complete for tag command

image

closes #594

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -76,12 +76,11 @@ public final class TagCommand extends SlashCommandAdapter {
             return;
         }
 
-        Collection<Command.Choice> choices = StringDistances
-            .autocompleteSuggestions(focusedOption.getValue(), tagSystem.getAllIds(), 0.5)
-            .stream()
-            .map(id -> new Command.Choice(id, id))
-            .limit(OptionData.MAX_CHOICES)
-            .toList();
+        Collection<Command.Choice> choices =
+                StringDistances.closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), OptionData.MAX_CHOICES)
+                    .stream()
+                    .map(id -> new Command.Choice(id, id))
+                    .toList();
 
         event.replyChoices(choices).queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -26,8 +26,8 @@ public final class TagCommand extends SlashCommandAdapter {
     private final TagSystem tagSystem;
 
     private static final String COMMAND_NAME = "tag";
-    private static final String ID_OPTION = "id";
-    private static final String REPLY_TO_USER_OPTION = "reply-to";
+    static final String ID_OPTION = "id";
+    static final String REPLY_TO_USER_OPTION = "reply-to";
     private static final int MAX_OPTIONS = 25;
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -75,14 +75,16 @@ public final class TagCommand extends SlashCommandAdapter {
         AutoCompleteQuery focusedOption = event.getFocusedOption();
 
         if (focusedOption.getName().equals(ID_OPTION)) {
-            Collection<Command.Choice> choices = StringDistances
-                .autocompleteSuggestions(focusedOption.getValue(), tagSystem.getAllIds(), 0.5)
-                .stream()
-                .map(id -> new Command.Choice(id, id))
-                .limit(OptionData.MAX_CHOICES)
-                .toList();
-
-            event.replyChoices(choices).queue();
+            return;
         }
+
+        Collection<Command.Choice> choices = StringDistances
+            .autocompleteSuggestions(focusedOption.getValue(), tagSystem.getAllIds(), 0.5)
+            .stream()
+            .map(id -> new Command.Choice(id, id))
+            .limit(OptionData.MAX_CHOICES)
+            .toList();
+
+        event.replyChoices(choices).queue();
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -25,7 +25,7 @@ import java.util.Collection;
  */
 public final class TagCommand extends SlashCommandAdapter {
     private final TagSystem tagSystem;
-    private static final int MAX_OPTIONS = 5;
+    private static final int MAX_SUGGESTIONS = 5;
     static final String ID_OPTION = "id";
     static final String REPLY_TO_USER_OPTION = "reply-to";
 
@@ -78,7 +78,7 @@ public final class TagCommand extends SlashCommandAdapter {
         }
 
         Collection<Command.Choice> choices = StringDistances
-            .closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), MAX_OPTIONS)
+            .closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), MAX_SUGGESTIONS)
             .stream()
             .map(id -> new Command.Choice(id, id))
             .toList();
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -72,15 +72,15 @@ public final class TagCommand extends SlashCommandAdapter {
     public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
         AutoCompleteQuery focusedOption = event.getFocusedOption();
 
-        if (focusedOption.getName().equals(ID_OPTION)) {
-            return;
+        if (!focusedOption.getName().equals(ID_OPTION)) {
+            throw new IllegalArgumentException("Unexpected option");
         }
 
-        Collection<Command.Choice> choices = StringDistances
-            .closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), OptionData.MAX_CHOICES)
-            .stream()
-            .map(id -> new Command.Choice(id, id))
-            .toList();
+        Collection<Command.Choice> choices =
+                StringDistances.closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), 5)
+                    .stream()
+                    .map(id -> new Command.Choice(id, id))
+                    .toList();
 
         event.replyChoices(choices).queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -11,13 +11,10 @@ import net.dv8tion.jda.api.interactions.commands.build.OptionData;
 import net.dv8tion.jda.api.requests.restaction.interactions.ReplyCallbackAction;
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.utils.StringDistances;
 
 import java.time.Instant;
 import java.util.Collection;
-import java.util.Comparator;
-import java.util.List;
-
-import static org.togetherjava.tjbot.commands.utils.StringDistances.prefixEditDistance;
 
 /**
  * Implements the {@code /tag} command which lets the bot respond content of a tag that has been
@@ -78,19 +75,14 @@ public final class TagCommand extends SlashCommandAdapter {
         AutoCompleteQuery focusedOption = event.getFocusedOption();
 
         if (focusedOption.getName().equals(ID_OPTION)) {
-            List<Command.Choice> choices =
-                    matches(focusedOption.getValue(), tagSystem.getAllIds()).stream()
-                        .map(id -> new Command.Choice(id, id))
-                        .limit(OptionData.MAX_CHOICES)
-                        .toList();
+            Collection<Command.Choice> choices = StringDistances
+                .autocompleteSuggestions(focusedOption.getValue(), tagSystem.getAllIds(), 0.5)
+                .stream()
+                .map(id -> new Command.Choice(id, id))
+                .limit(OptionData.MAX_CHOICES)
+                .toList();
 
             event.replyChoices(choices).queue();
         }
-    }
-
-    private static List<String> matches(String prefix, Collection<String> candidates) {
-        return candidates.stream()
-            .sorted(Comparator.comparingInt(candidate -> prefixEditDistance(prefix, candidate)))
-            .toList();
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -12,6 +12,7 @@ import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -42,9 +43,10 @@ public final class TagCommand extends SlashCommandAdapter {
         // TODO Think about adding an ephemeral selection menu with pagination support
         // if the user calls this without id or similar
         getData().addOptions(
-                new OptionData(OptionType.STRING, ID_OPTION, "The id of the tag to display", true, true),
-                new OptionData(OptionType.USER, REPLY_TO_USER_OPTION, "Optionally, the user who you want to reply to", false)
-        );
+                new OptionData(OptionType.STRING, ID_OPTION, "The id of the tag to display", true,
+                        true),
+                new OptionData(OptionType.USER, REPLY_TO_USER_OPTION,
+                        "Optionally, the user who you want to reply to", false));
     }
 
     @Override
@@ -71,14 +73,16 @@ public final class TagCommand extends SlashCommandAdapter {
 
     @Override
     public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
-        if (event.getName().equals(COMMAND_NAME) && event.getFocusedOption().getName().equals(ID_OPTION)) {
-            event.replyChoices(
-                    tagSystem.getAllIds().stream()
-                            .filter(id -> id.startsWith(event.getFocusedOption().getValue()))
-                            .map(id -> new Command.Choice(id, id))
-                            .limit(MAX_OPTIONS)
-                            .toList()
-            ).queue();
+        if (event.getName().equals(COMMAND_NAME)
+                && event.getFocusedOption().getName().equals(ID_OPTION)) {
+            List<Command.Choice> choices = tagSystem.getAllIds()
+                .stream()
+                .filter(id -> id.startsWith(event.getFocusedOption().getValue()))
+                .map(id -> new Command.Choice(id, id))
+                .limit(MAX_OPTIONS)
+                .toList();
+
+            event.replyChoices(choices).queue();
         }
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -76,11 +76,11 @@ public final class TagCommand extends SlashCommandAdapter {
             return;
         }
 
-        Collection<Command.Choice> choices =
-                StringDistances.closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), OptionData.MAX_CHOICES)
-                    .stream()
-                    .map(id -> new Command.Choice(id, id))
-                    .toList();
+        Collection<Command.Choice> choices = StringDistances
+            .closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), OptionData.MAX_CHOICES)
+            .stream()
+            .map(id -> new Command.Choice(id, id))
+            .toList();
 
         event.replyChoices(choices).queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -39,8 +39,6 @@ public final class TagCommand extends SlashCommandAdapter {
 
         this.tagSystem = tagSystem;
 
-        // TODO Think about adding an ephemeral selection menu with pagination support
-        // if the user calls this without id or similar
         getData().addOptions(
                 new OptionData(OptionType.STRING, ID_OPTION, "The id of the tag to display", true,
                         true),
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -28,7 +28,6 @@ public final class TagCommand extends SlashCommandAdapter {
     private static final String COMMAND_NAME = "tag";
     static final String ID_OPTION = "id";
     static final String REPLY_TO_USER_OPTION = "reply-to";
-    private static final int MAX_OPTIONS = 25;
 
     /**
      * Creates a new instance, using the given tag system as base.
@@ -73,13 +72,12 @@ public final class TagCommand extends SlashCommandAdapter {
 
     @Override
     public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
-        if (event.getName().equals(COMMAND_NAME)
-                && event.getFocusedOption().getName().equals(ID_OPTION)) {
+        if (event.getFocusedOption().getName().equals(ID_OPTION)) {
             List<Command.Choice> choices = tagSystem.getAllIds()
                 .stream()
                 .filter(id -> id.startsWith(event.getFocusedOption().getValue()))
                 .map(id -> new Command.Choice(id, id))
-                .limit(MAX_OPTIONS)
+                .limit(OptionData.MAX_CHOICES)
                 .toList();
 
             event.replyChoices(choices).queue();
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -3,6 +3,7 @@ package org.togetherjava.tjbot.commands.tags;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.events.interaction.command.CommandAutoCompleteInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
+import net.dv8tion.jda.api.interactions.AutoCompleteQuery;
 import net.dv8tion.jda.api.interactions.commands.Command;
 import net.dv8tion.jda.api.interactions.commands.OptionMapping;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
@@ -12,8 +13,11 @@ import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 
 import java.time.Instant;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
-import java.util.Objects;
+
+import static org.togetherjava.tjbot.commands.utils.StringDistances.prefixEditDistance;
 
 /**
  * Implements the {@code /tag} command which lets the bot respond content of a tag that has been
@@ -49,7 +53,7 @@ public final class TagCommand extends SlashCommandAdapter {
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
-        String id = Objects.requireNonNull(event.getOption(ID_OPTION)).getAsString();
+        String id = event.getOption(ID_OPTION).getAsString();
         OptionMapping replyToUserOption = event.getOption(REPLY_TO_USER_OPTION);
 
         if (tagSystem.handleIsUnknownTag(id, event)) {
@@ -71,15 +75,22 @@ public final class TagCommand extends SlashCommandAdapter {
 
     @Override
     public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
-        if (event.getFocusedOption().getName().equals(ID_OPTION)) {
-            List<Command.Choice> choices = tagSystem.getAllIds()
-                .stream()
-                .filter(id -> id.startsWith(event.getFocusedOption().getValue()))
-                .map(id -> new Command.Choice(id, id))
-                .limit(OptionData.MAX_CHOICES)
-                .toList();
+        AutoCompleteQuery focusedOption = event.getFocusedOption();
+
+        if (focusedOption.getName().equals(ID_OPTION)) {
+            List<Command.Choice> choices =
+                    matches(focusedOption.getValue(), tagSystem.getAllIds()).stream()
+                        .map(id -> new Command.Choice(id, id))
+                        .limit(OptionData.MAX_CHOICES)
+                        .toList();
 
             event.replyChoices(choices).queue();
         }
+    }
+
+    private static List<String> matches(String prefix, Collection<String> candidates) {
+        return candidates.stream()
+            .sorted(Comparator.comparingInt(candidate -> prefixEditDistance(prefix, candidate)))
+            .toList();
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
@@ -25,7 +25,6 @@ import java.util.Objects;
 public final class TagCommand extends SlashCommandAdapter {
     private final TagSystem tagSystem;
 
-    private static final String COMMAND_NAME = "tag";
     static final String ID_OPTION = "id";
     static final String REPLY_TO_USER_OPTION = "reply-to";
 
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -62,4 +70,19 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
         }
         message.queue();
     }
+
+    @Override
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        if (event.getName().equals(COMMAND_NAME)
```
Tais993, line 76: This is useless

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -62,4 +70,19 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+                && event.getFocusedOption().getName().equals(ID_OPTION)) {
+            List<Command.Choice> choices = tagSystem.getAllIds()
+                .stream()
+                .filter(id -> id.startsWith(event.getFocusedOption().getValue()))
```
Tais993, line 80: I'd instead refactor the StringDistance method to get *good* autocompletion instead of a simple startsWith

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -21,24 +25,28 @@
 public final class TagCommand extends SlashCommandAdapter {
     private final TagSystem tagSystem;
 
+    private static final String COMMAND_NAME = "tag";
     static final String ID_OPTION = "id";
     static final String REPLY_TO_USER_OPTION = "reply-to";
+    private static final int MAX_OPTIONS = 25;
```
Tais993, line 31: OptionData#MAX_OPTIONS or OptionData#MAX_CHOICES

Not sure what it's called

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Taz03: yes, i figured that out already, will remove it

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Taz03: MAX_CHOICES it is

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Taz03: i don't get what u mean, can u elaborate

Tais993: There's a StringDistance class or something, take a look at that class.

Taz03: can't find any

Tais993: Look again, StringDistances

IntelliJ has a feature which allows you to look for classes, you cannot miss it when using that

Taz03: 2022-10-01_115204-751x578_scrot

Tais993: In IntelliJ I said, maybe it's a class in our project (yes it is :p

Taz03: ok found it

Taz03: it won't work i believe it only returns a single item not a collection

Taz03: startsWith should be just fine

Tais993: Pretty sure I tagged you in the Discord about this.

Add an autocompletion method which does the same as autocomplete, but then it returns a collection

Tais993: It is "fine" but I much rather have better
See my messages in the Discord

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -36,14 +44,16 @@ public TagCommand(TagSystem tagSystem) {
...
+                new OptionData(OptionType.STRING, ID_OPTION, "The id of the tag to display", true,
+                        true),
+                new OptionData(OptionType.USER, REPLY_TO_USER_OPTION,
+                        "Optionally, the user who you want to reply to", false));
     }
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
-        String id = Objects.requireNonNull(event.getOption(ID_OPTION)).getAsString();
+        String id = event.getOption(ID_OPTION).getAsString();
```
Tais993, line 56: Removed require non null?

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Taz03: yes, coz it can't be null

Tais993: Thumbsup

Zabuzard:
Does this conflict with #602 ? Im confused.

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -62,4 +69,20 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
         }
         message.queue();
     }
+
+    @Override
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        AutoCompleteQuery focusedOption = event.getFocusedOption();
+
+        if (focusedOption.getName().equals(ID_OPTION)) {
```
Zabuzard, line 77: 1. use early-return to reduce the nesting
2. use fail-fast and not defensive programming here. u dont expect other options to end up here, so if it does, it should throw

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -51,6 +51,29 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
...
+            .min()
+            .orElse(0);
+        return candidates.stream()
+            .map(candidate -> new CandidateScore(candidate, prefixEditDistance(prefix, candidate)))
+            .sorted()
+            .takeWhile(candidateScore -> candidateScore.score() <= firstCandidateScore
+                    + firstCandidateScore * errorMargin)
+            .map(CandidateScore::candidate)
+            .toList();
+    }
```
Zabuzard, line 75: the whole design for this method is not really "idiomatic". and there are multiple issues with it. lets refactor it a bit:
```java
public static Collection<String> closestMatches(CharSequence source, Collection<String> destinations, int limit) {
        // TODO some argument validation ...
        
        Queue<String> matches = new PriorityQueue<>(Comparator.comparingInt(x -> prefixEditDistance(source, x)));
        matches.putAll(destinations);
        
        return Stream.generate(matches::poll)
          .limit(limit)
          .toList();
}
```

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -51,6 +51,29 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
...
 
+    /**
+     * Gives sorted suggestion to autocomplete a prefix string from the given options
+     *
+     * @param prefix the prefix to give suggestion for
+     * @param candidates all the possible suggestions
+     * @param errorMargin error margin of suggestions, error percentage / 100
+     * @return collection of autocomplete suggestions
+     */
+    public static Collection<String> autocompleteSuggestions(CharSequence prefix,
```
Zabuzard, line 62: Expand the existing unit tests for this. please dont leave it untested.

application/src/main/java/org/togetherjava/tjbot/commands/utils/CandidateScore.java
```diff
@@ -0,0 +1,8 @@
+package org.togetherjava.tjbot.commands.utils;
+
+record CandidateScore(String candidate, int score) implements Comparable<CandidateScore> {
```
Zabuzard, line 3: This class is obsolete when u use the comparator-api (as shown in my rework example)

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -62,4 +69,20 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+            Collection<Command.Choice> choices = StringDistances
+                .autocompleteSuggestions(focusedOption.getValue(), tagSystem.getAllIds(), 0.5)
+                .stream()
+                .map(id -> new Command.Choice(id, id))
+                .limit(OptionData.MAX_CHOICES)
```
Zabuzard, line 82: this limit should be given to the `autocompleteSuggestions` method instead, it shouldnt sort ALL tags (which could theoretically be thousands). (see my suggested rework)

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -62,4 +69,20 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+                .toList();
+
+            event.replyChoices(choices).queue();
```
Zabuzard, line 85: `tagSystem.getAllIds()` can take a bit since it has to check the database. not sure how autocomplete works with long methods, but in regular commands u have to use `deferReply()` and then respond via the interactionhook. maybe u have to do sth like that here as well?

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Tais993: Can you elaborate on the issues, as it seems youre removing part of the functionality.

There's a reason it gets mapped to a CandidateScore, take a look at the takeWhile method.

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Tais993: Auto completion should be fast, I don't think there's deferReply to begin with, but even if there was, having to wait 3s for auto completion is way too long if you ask me.

Taz03: `CommandAutoCompleteInteractionEvent` have no method called `deferReply()` also i don't think `replyChoices()` timesout

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: The way this math attempts to provide a good UX isnt really working out imo. Its completely missing the string length for example.

I am okay with adding this feature again. But then incorperate it into the proposed design with the PQ instead. The `CandidateScore` class should be nested, nobody else needs it. And should be renamed to `MatchScore` or `ScoredMatch`, depending on what it contains.

Tais993: I certainly believe this is far from the best solution.
but I do believe it's better than your suggestion of deleting it completely.

Since I personally lack experience with algorithms, you're on your own here.

Zabuzard: To begin with, i dont even understand how this math is supposed work at all. Its definitely in need for its own little method/comments/javadoc. And I would appreciate it if someone can give me a quick example calculation, so that i understand it.

Zabuzard: probably bc u made it public? hard to tell for me without seeing what u saw.

Taz03: should i put the record in the method?

Taz03:
> Sth I just saw: This PR should remove the TODO at `TagCommand` line `42` that essentially asks about adding exactly this.

it says `if the user calls this without id or similar` and this pr never deals with that ig #602 does that

Zabuzard:
@Taz03 that TODO was added bc we never knew that autocomplete will exist. the purpose of the TODO was to essentially provide autocomplete.

i know that bc i wrote it, lol 😆 its okay, please remove it, thanks

Taz03:
ok

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -177,4 +164,17 @@ private static int[][] computeLevenshteinDistanceTable(CharSequence source,
 
         return table;
     }
+
+    private record MatchScore(String candidate, int score) implements Comparable<MatchScore> {
+        @Override
+        public int compareTo(MatchScore otherMatchScore) {
+            int compare = Integer.compare(this.score, otherMatchScore.score);
```
Zabuzard, line 171: `compare` is a bad name. if at all, then `compareResult`. that said, u can call it `byScore` or `byScoreResult` maybe

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: what about scoreDiff?

Zabuzard: `diff` suggests that it would be the difference, like `foo - bar`. But its not.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: not resolved. still missing the priority queue approach with k-sort. and all the explanation/rework regarding that odd math.

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Zabuzard: not resolved. u didnt do this at all

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: not resolved. u didnt do this at all

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Zabuzard: not resolved. u didnt do this at all

Zabuzard (changes requested):
please resolve my existing issues, thanks

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: I thought u and tais agreed on keeping the current implementation, sry english is not my first language

Zabuzard: no worries. we agreed on keeping a filtering to not display non-sense suggestions. but:
* this can still be added to the PQ k-sort approach perfectly fine
* and im not okay with this odd way of doing it and would like to get a small explanation of how it works to understand it and then suggest something better

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: i did added some tests

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Tais993: The idea of this was made up by me, I don't know or there's any official algorithm like this.

The most confusing method is 
```java
            .takeWhile(candidateScore -> candidateScore.score() <= firstCandidateScore
                    + firstCandidateScore * errorMargin)
```

The idea of this method is pretty simple.

If the most accurate completion is 2 characters away, the most inaccurate completion can only be a max of 6 characters away. A slight improvement would do this.

```java
            .takeWhile(candidateScore -> candidateScore.score() <= firstCandidateScore * 2 * errorMargin)
```

Maybe a helper method would help even more?

```java
isCandidateInErrorRange(candidateScore , firstCandidateScore, errorMargin)
```

Do you understand it now?

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: im sorry, i didnt see them.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: sorry, no. can u give me a concrete example with some toy data?

Tais993: We have input `autocomplet`

the most accurate completion is `autocomplete` which is a total of 1 character away

1 * 2 * 0.5 =  3
The farthest distance a completion can be away is 3, so any completion that has a stringdistance of `1/2/3` will be accepted.



Another completion, `autocom` is total of 4 character, this means this won't get accepted.
But `autocomp` will get accepted, since it's only 3 characters away.


This way, we won't just have 25 random completion.
Makes more sense?

Zabuzard: yes. i understand the math now. it feels reversed to what i would have done:
```java
top10Candidates
  .filter(candidateWithPed -> candidateWithPed.ped / input.length()  < 0.5)
```
example, user inputs `"aaa"` and we have candidates `"aab"` and `"foo"`. latter is totally off, the PEDs are:
* `"aab" -> 1
* `"foo"` -> 3

the filter divides the PED with the original length `"aaa".length() -> 3` to compute how far away the candidate is:
* `"aab" -> `1 / 3` -> `0.333` is `< 0.5`, accepted
* `"foo"` -> `3 / 3` -> `1.0` is not `< 0.5`, discarded

Tais993: I'm noticing some knowledge about algorithms, sounds better than my suggestion. \:p

Zabuzard: this can ofc be optimized to already execute on the original stream with a `takeWhile`, like u suggested.

Taz03:
```java
.takeWhile(candidateScore -> candidateScore.score() <= firstCandidateScore * 2 * errorMargin)
```
@Tais993 this is not the same thing

Tais993: what's the difference between `firstCandidateScore  * 2` and `firstCandidateScore + firstCandidateScore `

Zabuzard: Actually, the way it was written (urs, but also mine) both have a performance flaw. We have to compute the PED for all entries once and then remember that instead. Using the PED in the Comparator will result in it being recomputed many times (more for sorted() than for a PQ, but still way too often). So we should do it like this:
```java
List<MatchScore> scoredMatches = getAllTags().stream()
  .map(match -> computePed(input, match))
  .toList();
```
and now we can sort based on the already-computed score:
```java
Queue<MatchScore> bestMatches = new PriorityQueue<>();
bestMatches.addAll(scoredMatches);

List<String> results = Stream.generate(bestMatches::poll)
  .takeWhile(match -> match.score / input.length() <= 0.5)
  .map(MatchScore::getMatch)
  .limit(10)
  .toList();
```
or similar (writing this from memory)

Taz03: > what's the difference between `firstCandidateScore * 2` and `firstCandidateScore + firstCandidateScore `

thats not what it is, i first multiply errorMargin in firstCandidateScore and then add it to firstCandidateScore

Tais993: Ah I completely forgot about mathematical rules LMFAO

Tais993: Why not just do `firstCandidateScore * (errorMargin + 1)`, since this is identical

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Taz03: which error to throw?

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -1,15 +1,19 @@
...
 
 /**
  * Utility class for computing string distances, for example the edit distance between two words.
  */
 public class StringDistances {
+    /**
+     * Matches that are further off than this are not considered as match anymore. The value is
+     * between 0.0 (full match) and 1.0 (completely different).
+     */
+    private static final double offByPercentageThreshold = 0.5;
```
Zabuzard, line 15: constant naming convention `OFF_BY_PERCENTAGE_THRESHOLD`

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -56,24 +60,28 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
...
      * @return collection of autocomplete suggestions
      */
-    public static Collection<String> autocompleteSuggestions(CharSequence prefix,
-            Collection<String> candidates, double errorMargin) {
-        int firstCandidateScore = candidates.stream()
-            .mapToInt(candidate -> prefixEditDistance(prefix, candidate))
-            .min()
-            .orElse(0);
-        return candidates.stream()
+    public static Collection<String> closeMatches(CharSequence prefix,
```
Zabuzard, line 65: `limit` is missing javadoc

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -56,24 +60,28 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
      *
      * @param prefix the prefix to give suggestion for
      * @param candidates all the possible suggestions
-     * @param errorMargin error margin of suggestions, error percentage / 100
      * @return collection of autocomplete suggestions
```
Zabuzard, line 63: `* @return collection of autocomplete suggestions` this text assumes context that is incorrect from the POV of `StringDistances`. rephrase to `the matches closest to the given prefix, limited to the given limit`

application/src/test/java/org/togetherjava/tjbot/commands/utils/StringDistancesTest.java
```diff
@@ -10,25 +10,17 @@
...
-        List<TestCase> tests =
-                List.of(new TestCase("empty_candidates", List.of(), "prefix", List.of(), 0),
-                        new TestCase("empty_prefix", List.of("one", "three", "two"), "",
-                                List.of("one", "two", "three"), 0),
-                        new TestCase("all_empty", List.of(), "", List.of(), 1),
-                        new TestCase("max_error", List.of("aa"), "a",
-                                List.of("json", "one", "aa", "two", "++"), 1),
-                        new TestCase("real_test", List.of("j", "java", "js", "one"), "jo",
-                                List.of("java", "xj", "bs", "one", "yes", "js", "a", "j"), 0.5));
+        List<TestCase> tests = List.of();
```
Zabuzard, line 18: TODO tests

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
Zabuzard: `IllegalArgumentException` since u got an option that u didnt expect.

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -72,15 +72,15 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
     public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
         AutoCompleteQuery focusedOption = event.getFocusedOption();
 
-        if (focusedOption.getName().equals(ID_OPTION)) {
-            return;
+        if (!focusedOption.getName().equals(ID_OPTION)) {
+            throw new IllegalArgumentException("Unexpected option");
```
Zabuzard, line 76: the exception would be a lot more helpful if u actually add what it was to the message. `"Unexpected option, was: " + ...`

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -72,15 +72,15 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
         }
 
-        Collection<Command.Choice> choices = StringDistances
-            .closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), OptionData.MAX_CHOICES)
-            .stream()
-            .map(id -> new Command.Choice(id, id))
-            .toList();
+        Collection<Command.Choice> choices =
+                StringDistances.closeMatches(focusedOption.getValue(), tagSystem.getAllIds(), 5)
```
Zabuzard, line 80: magic number `5`

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -65,6 +65,14 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
      */
     public static Collection<String> closeMatches(CharSequence prefix,
             Collection<String> candidates, int limit) {
+        if (prefix.isEmpty()) {
```
Zabuzard, line 68: if the prefix is empty, the returned list should be empty, since any suggestion can only be a bad suggestion.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -65,6 +65,14 @@ public static <S extends CharSequence> Optional<S> autocomplete(CharSequence pre
...
+            return candidates.stream().sorted().limit(limit).toList();
+        }
+
+        if (candidates.isEmpty()) {
+            return candidates;
```
Zabuzard, line 73: personally, i prefer `return List.of();` in such a case. makes it more explicit and clear that the returned list is empty.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -73,14 +81,14 @@ public static Collection<String> closeMatches(CharSequence prefix,
...
+            .limit(limit)
             .takeWhile(matchScore -> isCloseEnough(matchScore, prefix))
             .map(MatchScore::candidate)
-            .limit(limit)
             .toList();
     }
 
     private static boolean isCloseEnough(MatchScore matchScore, CharSequence prefix) {
-        return matchScore.score / prefix.length() <= OFF_BY_PERCENTAGE_THRESHOLD;
+        return (double) matchScore.score / prefix.length() <= OFF_BY_PERCENTAGE_THRESHOLD;
```
Zabuzard, line 91: voting for `score` itself to be `double` aready. based on the name, one would except that at least.

application/src/test/java/org/togetherjava/tjbot/commands/utils/StringDistancesTest.java
```diff
@@ -15,7 +15,16 @@ record TestCase(String name, Collection<String> expectedSuggestions, String pref
                 Collection<String> candidates, int limit) {
         }
 
-        List<TestCase> tests = List.of();
+        List<String> exampleTags = List.of("c", "c#", "c++", "emacs", "foo", "hello", "java", "js",
```
Zabuzard, line 18: call it `candidates` maybe. or `existingTags`

application/src/test/java/org/togetherjava/tjbot/commands/utils/StringDistancesTest.java
```diff
@@ -15,7 +15,16 @@ record TestCase(String name, Collection<String> expectedSuggestions, String pref
...
+                "key", "nvim", "py", "tag", "taz", "vi", "vim");
+
+        List<TestCase> tests = List.of(new TestCase("no_tags", List.of(), "foo", List.of(), 5),
+                new TestCase("no_prefix", List.of("c", "c#", "c++", "emacs", "foo"), "",
+                        exampleTags, 5),
+                new TestCase("both_empty", List.of(), "", List.of(), 5),
+                new TestCase("test0", List.of("vi", "vim"), "v", exampleTags, 5),
+                new TestCase("test1", List.of("java", "js"), "j", exampleTags, 5),
+                new TestCase("test2", List.of("c", "c#", "c++"), "c", exampleTags, 5));
```
Zabuzard, line 27: lots of hardcoded `5` here. add that as a variable beforehand

Zabuzard, line 27: `test0-2` arent really helpful names. maybe u can be a bit more precise as to what it tests. maybe `withPrefix0-2`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: yes, but i dont like to see the no auto-complete suggestions even before i did anything
image

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: when PED method returns `int` it should be `int` ig but idk, i'll make it a `double` if u want

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: lol whats that. cant it just tell discord that it has nothing and discord doesnt show anything extra then? like without autocomplete? @Tais993

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Zabuzard: the PED should stay int but the score itself should be double. in the future, it might also include relevance or other stuff, BM-25 matching etc blabla. it doesnt necessarily have to be a PED only. otherwise, the name should be changed from score to distance imo - then it could stay an int 🤷

application/src/main/java/org/togetherjava/tjbot/commands/tags/TagCommand.java
```diff
@@ -25,7 +25,7 @@
  */
 public final class TagCommand extends SlashCommandAdapter {
     private final TagSystem tagSystem;
-
+    private static final int MAX_OPTIONS = 5;
```
Zabuzard, line 28: `MAX_OPTIONS` feels a bit confusing, naming-wise. i think it would be clearer if its `MAX_SUGGESTIONS` or `MAX_MATCHES`.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
Taz03: no

Taz03: that would be nice tho

Zabuzard: anyways, ur doing sorted() again instead of k-sort. thats the same mistake than before, a performance-problem. use a PQ.

application/src/main/java/org/togetherjava/tjbot/commands/utils/StringDistances.java
```diff
@@ -80,6 +76,13 @@ public static Collection<String> closeMatches(CharSequence prefix,
         Queue<MatchScore> bestMatches = new PriorityQueue<>();
         bestMatches.addAll(scoredMatches);
 
+        if (prefix.isEmpty()) {
+            return Stream.generate(bestMatches::poll)
+                .limit(limit)
+                .map(MatchScore::candidate)
+                .toList();
+        }
+
```
Zabuzard, line 85: I feel like this feature is better put into `isCloseEnough` to always say `0.0` when the `prefix.length()` is `0`.

Right now, this feels like an useless duplication

Zabuzard (changes requested):

Zabuzard (approved):
before u merge, please have a final test over all edge cases. its easy to break stuff through longer rounds of CR if it doesnt have 100% unit test coverage 👍

Taz03:
@Zabuzard one approve and been a whole week so merge now?

Tais993:
Okay so I learned something new.
If GitHub says "change requested" it's useless and means nothing, as my review has been requested already.
I'll review RN

Tais993 (approved):

PR #599, opened 2022-10-02
GitHub Referencing & GitHub Command

illuminator3:
unknown
image
image

## changes `gistApiKey` to `githubApiKey` in the config


Tais993:
I LOVE YOU ILLU

Taz03:
what about not having the ping when the bot replies back?

illuminator3:
> what about not having the ping when the bot replies back?

why?

Taz03:
> why?

coz its a bot reply

illuminator3:
> > why?
> 
> coz its a bot reply

@Tais993 what do you think?

Tais993:
Don't have an opinion, either is fine to me

illuminator3:
I'll remove it since the command usage also has no ping I guess

application/config.json.template
```diff
@@ -1,6 +1,6 @@
 {
     "token": "<put_your_token_here>",
-    "gistApiKey": "<your_gist_personal_access_token>",
+    "githubApiKey": "<your_github_personal_access_token>",
```
Tais993, line 3: Add disclaimer in the PR description about this so it doesnt get missed

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+ * embed that contains info on the issue/pr
+ */
+@ParametersAreNonnullByDefault
+@MethodsReturnNonnullByDefault
+public class GitHubReference extends MessageReceiverAdapter {
+    /**
+     * The pattern used to determine whether a message is referencing an issue
+     */
+    protected static final Pattern ISSUE_REFERENCE_PATTERN =
+            Pattern.compile("(?:\\s+|[()\\[\\]<>]|^)#(\\d+)(?:\\s+|[()\\[\\]<>]|$)");
```
Tais993, line 37: is there anyway this can become readable, like, groups with names or such?

illuminator3: there's only one _real_ group. ?: indicates a noncapturing group

Zabuzard:
> changes gistApiKey to githubApiKey in the config

note that this requires that people update their api key settings. gonna mention that when merging this.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Tais993: Is there any way you could document this,. since I really don't understand what everything is exactly for.

I'd imagine all you need is a check which goes for: anycharacter#anynumbers

While here, it's a lot, don't know what it all does?

application/config.json.template
Zabuzard, line 3: which permission checkboxes have to be clicked on the token now? before, it was just `gist`. and now?

application/src/main/java/org/togetherjava/tjbot/config/Config.java
```diff
@@ -134,8 +140,8 @@ public String getToken() {
      *      "https://docs.github.com/en/enterprise-server@3.4/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token">Create
      *      a GitHub key</a>
      */
-    public String getGistApiKey() {
-        return gistApiKey;
+    public String getGitHubApiKey() {
```
Zabuzard, line 145: adjust javadoc, its incorrect now

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+     * The cache updates every minute (see the ONE_MINUTE_IN_MILLIS constant)
+     */
+    private long lastCache;
+    /**
+     * The currently cached auto completion
+     */
+    private List<String> autocompleteCache;
+
+    public GitHubCommand(GitHubReference reference) {
+        super("github", "Search GitHub for an issue", CommandVisibility.GUILD);
```
Zabuzard, line 50: the name is a bit too unspecified, could clash with other future commands.
maybe go with `github-issue`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final String COMMAND_OPTION_NAME = "title";
+
+    private final GitHubReference reference;
+
+    /**
+     * The last time the cache was updated (millis after Jan 1st 1970)
+     * <p>
+     * The cache updates every minute (see the ONE_MINUTE_IN_MILLIS constant)
+     */
+    private long lastCache;
```
Zabuzard, line 43: Time units have to be part of the variable name to avoid confusion `lastCacheMillis`. Also, u wouldnt need javadoc here if u would make the var name better, add the "update" part to it: `lastCacheUpdateMillis`. That said, please go with something high level instead, namely `Instant lastCacheUpdate;`.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+     * lol)
+     */
+    private static final String COMMAND_OPTION_NAME = "title";
+
+    private final GitHubReference reference;
+
+    /**
+     * The last time the cache was updated (millis after Jan 1st 1970)
+     * <p>
+     * The cache updates every minute (see the ONE_MINUTE_IN_MILLIS constant)
```
Zabuzard, line 41: You can use `{@value ONE_MINUTE_IN_MILLIS}` for such stuff instead. that will substitute the value into it. that way it cant rot

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard, line 50: the description could be confusing, it sounds as if it searches whole github. maybe make it clear that its for bot development only.
`Search GitHub repositories of this bot for an issue`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+        updateCache();
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        String title = event.getOption(COMMAND_OPTION_NAME).getAsString();
+        Matcher matcher = GitHubReference.ISSUE_REFERENCE_PATTERN.matcher(title);
+
+        if (!matcher.find()) {
+            event.reply("Could not parse your query").setEphemeral(true).queue();
```
Zabuzard, line 66: why? what should the user do? what was wrong? where can they find info about how to do it right?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+
+            return;
+        }
+
+        int id = Integer.parseInt(matcher.group(1));
```
Zabuzard, line 71: u can use named-groups for better readability.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+        Optional<GHIssue> issue = reference.findIssue(id);
+
+        if (issue.isPresent()) {
+            try {
+                event.replyEmbeds(reference.generateReply(issue.get())).queue();
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        } else {
+            event.reply("Could not find the issue you are looking for").setEphemeral(true).queue();
```
Zabuzard, line 81: missing dot or smiley at the end of the sentence

Zabuzard, line 81: the api lookup could take too long for `event.reply...`. u gotta use `event.deferReply()` here.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        String title = event.getOption(COMMAND_OPTION_NAME).getAsString();
+
+        if (title.isEmpty()) {
+            event.replyChoiceStrings(autocompleteCache.stream().limit(25).toList()).queue();
+        } else {
+            event
+                .replyChoiceStrings(autocompleteCache.stream()
+                    .sorted(Comparator.comparingInt(s -> StringDistances.editDistance(title,
```
Zabuzard, line 94: sure u want to sort all of them? think about using a `PriorityQueue` with that as `Comparator` and then calling `poll()` `25-times`:
```java
Queue<Foo> foos = new PriorityQueue<>(Comparator...);

List<Foo> foosBla = Stream.generate(foos::poll)
        .limit(25)
        .toList();
```

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+                            s.replaceFirst("\\[\\d+] ", ""))))
+                    .limit(25)
+                    .toList())
+                .queue();
+        }
+
+        if (lastCache <= System.currentTimeMillis() - ONE_MINUTE_IN_MILLIS) {
+            updateCache();
+        }
```
Zabuzard, line 103: u only update after an usage. that means that the usage itself is, in practice, always outdated.

instead, run this on a schedule in the background. use `implements Routine` for this.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: I simplified the pattern

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+            event
+                .replyChoiceStrings(autocompleteCache.stream()
+                    .sorted(Comparator.comparingInt(s -> StringDistances.editDistance(title,
+                            s.replaceFirst("\\[\\d+] ", ""))))
+                    .limit(25)
+                    .toList())
+                .queue();
+        }
+
+        if (lastCache <= System.currentTimeMillis() - ONE_MINUTE_IN_MILLIS) {
```
Zabuzard, line 101: this check is too low level. prefer using `Instant.now().minus(...).isAfter(...)` and similar

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+            updateCache();
+        }
+    }
+
+    /**
+     * Updates the cache (has no time check, i.e. it does not check whether ONE_MINUTE_IN_MILLIS has
+     * passed since the last cache)
+     */
```
Zabuzard, line 109: The `(...)` part is a bit confusing to read. Actually, I would remove it entirely.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+                return repo.getIssues(GHIssueState.ALL);
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        }).flatMap(List::stream).sorted((i1, i2) -> {
+            try {
+                return i2.getUpdatedAt().compareTo(i1.getUpdatedAt());
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
```
Zabuzard, line 122: create a small helper method for this, then the comparator reads nicely.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+                        .map(this::getUserNameOrFailAtRuntime)
+                        .collect(Collectors.joining(", "))))
+            .build();
+    }
+
+    // this is utterly stupid
+    /**
+     * Either properly gathers the name of a user or throws a UncheckedIOException
+     */
+    private String getUserNameOrFailAtRuntime(GHUser user) throws UncheckedIOException {
```
Zabuzard, line 138: that `AtRuntime` is obsolete. rename to `getUserNameOrThrow`, that would be idiomatic

Tais993 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+            Pattern.compile("#(\\d+)");
+
+    private final Config config;
+
+    /**
+     * The repositories that are searched when looking for an issue
+     */
+    private List<GHRepository> repositories;
+
+    public GitHubReference(Config config) {
```
Zabuzard, line 45: I dont like the name of this class. Maybe rename to `GitHubReferenceIssueListener`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+        super(Pattern.compile(config.getGitHubReferenceChannelPattern()));
+
+        this.config = config;
+
+        try {
+            acquireRepositories();
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
```
Zabuzard, line 55: just move the try-catch stuff into the helper method, keeps the constructor clean

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Scans all repositories mentioned in the config and adds their object representations using
+     * kohuske's API to a list
+     */
+    private void acquireRepositories() throws IOException {
+        repositories = new ArrayList<>();
+
+        GitHub githubApi = GitHub.connectUsingOAuth(config.getGitHubApiKey());
+
+        for (long repo : config.getGitHubRepositories()) {
```
Zabuzard, line 66: bad var name. `repoId` would be better

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+
+        Message message = event.getMessage();
+        String content = message.getContentRaw();
+        Matcher matcher = ISSUE_REFERENCE_PATTERN.matcher(content);
+        List<MessageEmbed> embeds = new ArrayList<>();
+
+        while (matcher.find()) {
+            int id = Integer.parseInt(matcher.group(1));
+
+            try {
```
Zabuzard, line 85: u can use named-groups for readability

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+                Optional<GHIssue> issue = findIssue(id);
+
+                if (issue.isPresent()) {
+                    embeds.add(generateReply(issue.get()));
+                }
+            } catch (IOException ex) {
```
Zabuzard, line 91: can use the optional API here. `findIssue(id).ifPresent(...)`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+        for (List<MessageEmbed> messageEmbeds : partition) {
+            if (first) {
+                message.replyEmbeds(messageEmbeds).mentionRepliedUser(false).queue();
+
+                first = false;
+            } else {
+                textChannel.sendMessageEmbeds(messageEmbeds).mentionRepliedUser(false).queue();
+            }
+        }
+    }
```
Zabuzard, line 109: it might make sense to extract this whole blob into some `sendResponse(...)` method or similar.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        }
+
+        List<List<MessageEmbed>> partition = ListUtils.partition(embeds, 10);
+        boolean first = true;
+        TextChannel textChannel = message.getChannel().asTextChannel();
+
+        for (List<MessageEmbed> messageEmbeds : partition) {
```
Zabuzard, line 100: move that var down to where its used

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+                    issue.getAssignees()
+                        .stream()
+                        .map(this::getUserNameOrFailAtRuntime)
+                        .collect(Collectors.joining(", "))))
+            .build();
+    }
+
+    // this is utterly stupid
+    /**
+     * Either properly gathers the name of a user or throws a UncheckedIOException
```
Zabuzard, line 136: what is this?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    /**
+     * Safely (returning an optional instead of throwing an exception) looks through all of the
+     * given repositories for an issue/pr with the given id
+     */
+    protected Optional<GHIssue> findIssue(int id) {
```
Zabuzard, line 150: that safely part (and its expalantion) is obsolete. Its clear to a reader what the method does if it returns `Optional`.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+        return repositories.stream().map(repository -> {
+            try {
+                return Optional.of(repository.getIssue(id));
+            } catch (FileNotFoundException ignored) {
+                return Optional.<GHIssue>empty();
```
Zabuzard, line 155: sure that generic hint is needed? doesnt look like it

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        }).filter(Optional::isPresent).map(Optional::get).findAny();
```
Zabuzard, line 159: dont use `get`, use `orElseThrow` instead. `get` might get deprecated

Zabuzard (changes requested):

application/config.json.template
illuminator3: I selected public repo. not sure if it would even work without it

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: why? to me it feels like that would be like shooting with canons on pigeons. I don't see why we would go all in with instants and stuff like that if a simple currentTimeMillis check works too.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: constant updating is not necessary here. it was never intended for it to update every minute, but instead only once it's really used.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: the priority queue would still sort them internally though no?

application/src/main/java/org/togetherjava/tjbot/config/Config.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: what about `github-search`?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+@ParametersAreNonnullByDefault
+@MethodsReturnNonnullByDefault
+public class GitHubCommand extends SlashCommandAdapter {
+    private static final long ONE_MINUTE_IN_MILLIS = 60_000L;
+
+    /**
+     * The command option name of this slash command (this is only there to make the linter happy
+     * lol)
+     */
+    private static final String COMMAND_OPTION_NAME = "title";
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/config/Config.java
```diff
@@ -258,6 +264,20 @@ public List<String> getBlacklistedFileExtensions() {
...
+     * The REGEX pattern used to identify the channels that support GitHub issue referencing
+     */
+    public String getGitHubReferenceChannelPattern() {
+        return githubReferenceChannelPattern;
+    }
+
+    /**
+     * The list of repositories that are searched when referencing a GitHub issue
+     */
+    public List<Long> getGitHubRepositories() {
```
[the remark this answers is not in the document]
Tais993: While the name is valid, it takes a couple seconds if not longer of thoughts to understand that it's for the PR/issue reference feature

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Tais993: `github-search` or `github search` (as subcommands yes) I'd say make most sense.

Also, it also supports PR's right? If yes, description should be modified?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: but I'm 100% certain that the thing isn't empty no?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: pull requests are technically issues but I'll modify the description

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: it does k-sort, not full-sort. thats MUCH faster. not that its super crucial here, not like we have thousands of issues.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: but that means that the user always sees an outdated result. the chances for the user to have a good experience are thus close to zero.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: its error prone and generally harder to read. but in particular: its error prone. very easy to do mistakes with this raw math.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Tais993: > u only update after an usage. that means that the usage itself is, in practice, always outdated
Fun part is that, in practice, it's actually not always outdated.

Autocompletion starts when you start typing.

So you type `c`, so this command start updating.

While it's updating you finish your sentence `command rework` or `command`.

And by then it's updated.


Alternatively, we just spam the GitHub API while unnecessary (which I myself wasn't very fond of) 
Or we wait with responding, but that's kinda bad UX, since you want to display option ASAP I'd imagine.

Besides that, most of the time it being outdated won't actually matter. For cases where it does matter, once the user notices the issue isn't being autocompleted, they either simply run the command or they retype it / wait a sec.

I believe this is the better solution myself

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * All repositories monitored by this instance
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: it's needed
image

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: the getName method of the api I'm using is declared with throws IOException which would clutter up the stream statement. this is basically just a helper method for a nicer looking stream

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: what is lacking readability here?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    /**
+     * Scans all repositories mentioned in the config and adds their object representations using
+     * kohuske's API to a list
+     */
+    private void acquireRepositories() throws IOException {
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: what about `GitHubReferencingListener`?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done. made it a field

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * The last time the cache was updated (millis after Jan 1st 1970)
+     * <p>
+     * The cache updates every minute (see the ONE_MINUTE_IN_MILLIS constant)
+     */
+    private long lastCache;
+    /**
+     * The currently cached auto completion
+     */
+    private List<String> autocompleteCache;
```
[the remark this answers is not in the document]
illuminator3: it contains the autocompletion results, like the name suggests

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: `Search configured GitHub repositories for an issue/pull request`?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: how would I do that considering whether the message is ephemeral depends on the outcome of the api lookup?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: which var?

application/src/main/java/org/togetherjava/tjbot/config/Config.java
```diff
@@ -258,6 +264,20 @@ public List<String> getBlacklistedFileExtensions() {
         return Collections.unmodifiableList(blacklistedFileExtension);
     }
 
+    /**
+     * The REGEX pattern used to identify the channels that support GitHub issue referencing
+     */
+    public String getGitHubReferenceChannelPattern() {
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/config/Config.java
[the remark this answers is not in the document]
illuminator3: soooooo what name should I use?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: added more info

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: then I'd need another try catch inside the lambda, which would make the code look really really ugly

illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+            try {
+                Optional<GHIssue> issue = findIssue(id);
+
+                if (issue.isPresent()) {
+                    embeds.add(generateReply(issue.get()));
+                }
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        }
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+
+        List<List<MessageEmbed>> partition = ListUtils.partition(embeds, 10);
+        boolean first = true;
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
+package org.togetherjava.tjbot.commands.github;
```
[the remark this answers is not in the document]
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,127 @@
...
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        String title = event.getOption(COMMAND_OPTION_NAME).getAsString();
+
+        if (title.isEmpty()) {
+            event.replyChoiceStrings(autocompleteCache.stream().limit(25).toList()).queue();
+        } else {
+            event
+                .replyChoiceStrings(autocompleteCache.stream()
+                    .sorted(Comparator.comparingInt(s -> StringDistances.editDistance(title,
+                            s.replaceFirst("\\[\\d+] ", ""))))
```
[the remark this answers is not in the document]
illuminator3: it removes the first occurrence of [#somenumber] (I even forgot # in the regex there lol)

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/config/Config.java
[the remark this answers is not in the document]
Zabuzard: I guess the main cause is that `reference` doesnt sound like a feature. as in, the name for this feature itself isnt really unique. maybe use a synonym to "refer". maybe sth along the lines of "link preview" or "issue preview" or "link issue".

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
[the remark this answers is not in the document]
Zabuzard: but what are those autocompletion results? are they issues? names of authors? repositories? thats what im missing.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: * please dont put the name of the bot in it, otherwise the bot couldnt be used by others anymore.
* "configured" gives a bad UX. this command is seen by random users who dont know anything about the bot - they might think they can use this command to search their own github stuff. u gotta tell them somehow that its bot related.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: yeah. in that case u have to decide and drop the ephemeral. no way around that, unfortunately

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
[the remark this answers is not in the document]
Zabuzard: thats fair. just put it into a helper method. maybe with a mini javadoc example like u just gave me.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: we have 3 PRs now that essentially duplicated this pattern of "i have a list and want to get the best x matches". we should probably elevate this into a helper somewhere. but alright, lets do that afterwards, once all PRs are merged.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: i was missing the fact that the update runs in the autocompletion. in that case i am okay with it - but still dont really like it. but ur right, in that case its not a real issue.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: good question. seems the comment is put at the wrong place. ill recheck

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: ah. you could potentially get rid of this issue by putting the contents of the `map(...)` into a helper method.
would likely also simplify readability here.

application/config.json.template
Zabuzard: can u put that info in the PR description please? im going through them when i do releases, to tell people how to migrate

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: consider moving the named group `"id"` into a constant. then people cant do mistakes and we have compile-time-safety.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        String title = event.getOption(TITLE_OPTION).getAsString();
+
+        if (title.isEmpty()) {
+            event.replyChoiceStrings(autocompleteCache.stream().limit(25).toList()).queue();
+        } else {
+            Queue<String> queue = new PriorityQueue<>(Comparator.comparingInt(
```
Zabuzard, line 90: the name isnt really optimal. maybe `closestMatches` or `closestSuggestions`.

it also reads a bit complex since u nested everything. i would at least move out the 25-list into a variable:
```java
Queue<String> closestMatches = ...
List<String> suggestions = ...

event.replyChoiceStrings(suggestions).queue();
```
maybe also the logic u use to compute the score. So that its `Comparator.comparingInt(this::scoreSuggestion)` or similar.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,118 @@
...
+                    s -> StringDistances.editDistance(title, s.replaceFirst("\\[#\\d+] ", ""))));
+
+            queue.addAll(autocompleteCache);
+
+            event.replyChoiceStrings(Stream.generate(queue::poll).limit(25).toList()).queue();
+        }
+
+        if (lastCacheUpdate.isAfter(Instant.now().minus(Duration.ofMinutes(1)))) {
```
Zabuzard, line 98: the `Duration.ofMinutes(1)` could be moved upwards into a constant for readability:
`Duration CACHE_EXPIRES_AFTER = ...`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: better, yes. but i still have problems with the term "reference" for this (see other discussion). maybe we can find a better word that has less confusion potential

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+            return;
+        }
+
+        Message message = event.getMessage();
+        String content = message.getContentRaw();
+        Matcher matcher = ISSUE_REFERENCE_PATTERN.matcher(content);
+        List<MessageEmbed> embeds = new ArrayList<>();
+
+        while (matcher.find()) {
+            findIssue(Integer.parseInt(matcher.group("id")))
```
Zabuzard, line 82: `"id"` should be a constant (also see other comment in the other class). maybe `ID_GROUP = "id"`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+                .ifPresent(issue -> embeds.add(generateReply(issue)));
+        }
+
+        replyBatchEmbeds(embeds, message, false);
+    }
+
+    /**
+     * Replies to the given message with the given embeds in "batches", sending 10 embeds at a time
+     * (the discord limit)
+     */
```
Zabuzard, line 92: that 10 limit is in JDA as constant. somewhere around `Message.MAX_EMBEDS` or similar.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: ah. it was about the `TextChannel`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,169 @@
...
+    private String getUserNameOrThrow(GHUser user) throws UncheckedIOException {
+        try {
+            return user.getName();
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    /**
+     * Looks through all of the given repositories for an issue/pr with the given id
```
Zabuzard, line 149: missing dot

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/config/Config.java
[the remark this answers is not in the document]
Tais993: Something with preview makes sense in my eyes, and would be easier to understand by others.

githubReferencePreviewRepositories?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: the else is called every loop iteration except the first

Zabuzard: ah. fair then. maybe move it right next to the loop then:
```java
Foo foo = ...
for (...) {
  ...
}
```
but at this point its just nitpicking

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
illuminator3: it's already right above the for loop though?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
illuminator3: done

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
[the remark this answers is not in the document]
illuminator3: done

application/config.json.template
```diff
@@ -83,7 +83,11 @@
         "wsc",
         "wsf",
         "wsh"
-  ],
-  "logInfoChannelWebhook": "<put_your_webhook_here>",
-  "logErrorChannelWebhook": "<put_your_webhook_here>"
+    ],
+    "githubReferencingEnabledChannelPattern": "general",
+    "githubRepositories": [
+        <repository id here (e.g. 403389278>
```
Taz03, line 89: missing closing bracket
```suggestion
        <repository id here (e.g. 403389278)>
```

application/config.json.template
Taz03: > not sure if it would even work without it

test it then

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,123 @@
...
+                        .setEphemeral(true)
+                        .queue());
+    }
+
+    @Override
+    public void onAutoComplete(CommandAutoCompleteInteractionEvent event) {
+        String title = event.getOption(TITLE_OPTION).getAsString();
+
+        if (title.isEmpty()) {
+            event.replyChoiceStrings(autocompleteGHIssueCache.stream().limit(25).toList()).queue();
```
Taz03, line 88: don't use the magic number `25` use [OptionData#MAX_CHOICES](https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/interactions/commands/build/OptionData.html#MAX_CHOICES) instead

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,123 @@
...
+        } else {
+            Queue<String> closestSuggestions =
+                    new PriorityQueue<>(Comparator.comparingInt(suggestionScorer(title)));
+
+            closestSuggestions.addAll(autocompleteGHIssueCache);
+
+            event.replyChoiceStrings(Stream.generate(closestSuggestions::poll).limit(25).toList())
```
Taz03, line 95: don't use the magic number `25` use [OptionData#MAX_CHOICES](https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/interactions/commands/build/OptionData.html#MAX_CHOICES) instead

Taz03 (changes requested):

application/config.json.template
```diff
@@ -9,42 +9,42 @@
...
+        "hostWhitelist": ["discord.com", "discord.gg", "discord.media", "discordapp.com", "discordapp.net", "discordstatus.com"],
+        "hostBlacklist": ["bit.ly"],
+        "suspiciousHostKeywords": ["discord", "nitro", "premium"],
+        "isHostSimilarToKeywordDistanceThreshold": 2
+    },
+    "wolframAlphaAppId": "79J52T-6239TVXHR7",
+    "helpSystem": {
+        "stagingChannelPattern": "ask_here",
+        "overviewChannelPattern": "active_questions",
+            "categories": [
```
Zabuzard, line 30: the indent for categories is one too much

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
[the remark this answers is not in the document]
Zabuzard: the method is good, thanks for that. but i would still like to have a small comment on the regex. like `// Remove the ID [#123] and then match`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: ah. fair

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
Zabuzard: could u also move the `Stream.generate(closestSuggestions::poll).limit(25).toList()` into its own variable? Its a bit harder to read when its all packed together with the JDA stuff of message-sending.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * GitHub Referencing feature. If someone sends #id of an issue (e.g. #207) the bot replies with an
+ * embed that contains info on the issue/pr
+ */
+public class GitHubReference extends MessageReceiverAdapter {
+    /**
+     * The pattern used to determine whether a message is referencing an issue
+     */
+    protected static final Pattern ISSUE_REFERENCE_PATTERN = Pattern.compile("#(?<id>\\d+)");
+    protected static final String ID_GROUP = "id";
```
Zabuzard, line 35: gotta use the ID_GROUP in ur pattern as well. `"#(?<%s>\\d+".formatted(ID_GROUP)`

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
Zabuzard: seems u forgot to push this one

java-coding-prodigy:
One year anniversary of this PR 🎉
Also are we going to this or should it be closed?

ankitsmt211:
#981 which continued work of this PR already merged.

PR #601, opened 2022-10-03
Fixed components not working

Tais993:
In BotCommandAdapter
the Interactor's name was used without prefix, but to forward from the button the prefix is needed.

I promise next time I'll create a better PR! :p


illuminator3 (approved):

illuminator3 (approved):

Taz03 (approved):

Zabuzard:
this is a design problem. the prefix was meant to be an internal detail only (inside `BotCore`). but now it leaked into the user API (think about people who go via interface and not by adapter. they are not meant to know the details of their prefixes)

Tais993:
Where does it leak to user API?

Zabuzard:
I fixed what I was annoyed by in #588 👍

PR #604, opened 2022-10-04
added bookmark command

### LINKED ISSUE #570: Bookmark question thread command
## Is your feature request related to a problem? Please describe.

Currently, there are no ways to 'follow' a question thread to track answers to the question or have an automated way of saving a link to the question thread discussion for later review.

## Describe the solution you'd like

A command such as `/bookmark` or `/save` so that the bot can DM the user with the thread link (which can be obtained from the thread ID and can be referred with `<#(the threadID)>`).

For saving some context about each bookmark, implementors may decide to include the first message sent by the thread author after thread creation (assumed as the main question). Issue is that question can be scattered trough 20 messages, with images etc. And since a question can often be poorly titled, having just a title is not enough.

So probably the best bet is to have an optional `tag` or `memo` option for the slash command, so users can organize their bookmark list with a personalized tag. It's searchable, and it can be used as a reminder what the question was about, thus correcting for poorly titled questions.

## Describe alternatives you've considered

Manually do it.

<sub>(Originally suggested by voidcorruptor)</sub>

Taz03:
closes #570 

In DM:-
2022-10-04_201534-529x95_scrot

The command:-
2022-10-04_201618-458x108_scrot
2022-10-04_201702-572x153_scrot

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
@@ -209,12 +209,15 @@ public final class HelpThreadCommand extends SlashCommandAdapter {
         event.getUser()
             .openPrivateChannel()
             .flatMap(channel -> channel.sendMessage(String.format("<#%s>", helpThread.getIdLong())))
-            .queue();
-
-        event.reply(
-                "An attempt has made to send a link of this help thread to your DMs. Check your inbox")
-            .setEphemeral(true)
-            .queue();
+            .queue(onSuccess -> event.reply("A link to this help thread has been sent to your DMs")
+                .setEphemeral(true)
+                .queue(),
+                    onFailure -> event.reply("""
+                            Unable to send a DM. Try doing:-
+                            1. Make sure this bot isn't blocked
+                            2. Make sure you allow DMs from members of this server""")
+                        .setEphemeral(true)
+                        .queue());
     }
 
     private static Stream<Subcommand> streamSubcommands() {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
@@ -208,7 +208,8 @@ public final class HelpThreadCommand extends SlashCommandAdapter {
     private void bookmarkThread(SlashCommandInteractionEvent event, ThreadChannel helpThread) {
         event.getUser()
             .openPrivateChannel()
-            .flatMap(channel -> channel.sendMessage(helpThread.getAsMention()))
+            .flatMap(channel -> channel
+                .sendMessage("You bookmarked the following thread: " + helpThread.getAsMention()))
             .queue(onSuccess -> event.reply("A link to this help thread has been sent to your DMs")
                 .setEphemeral(true)
                 .queue(),
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
@@ -76,7 +76,8 @@ public final class HelpThreadCommand extends SlashCommandAdapter {
                     .addSubcommands(changeCategory, changeTitle);
         getData().addSubcommandGroups(changeCommands);
 
-        getData().addSubcommands(Subcommand.CLOSE.toSubcommandData());
+        getData().addSubcommands(Subcommand.CLOSE.toSubcommandData(),
+                Subcommand.BOOKMARK.toSubcommandData());
 
         this.helper = helper;
 
@@ -89,9 +90,9 @@ public final class HelpThreadCommand extends SlashCommandAdapter {
         subcommandToCooldownCache = new EnumMap<>(streamSubcommands()
             .filter(Subcommand::hasCooldown)
             .collect(Collectors.toMap(Function.identity(), any -> createCooldownCache.get())));
-        subcommandToEventHandler = new EnumMap<>(
-                Map.of(Subcommand.CHANGE_CATEGORY, this::changeCategory, Subcommand.CHANGE_TITLE,
-                        this::changeTitle, Subcommand.CLOSE, this::closeThread));
+        subcommandToEventHandler = new EnumMap<>(Map.of(Subcommand.CHANGE_CATEGORY,
+                this::changeCategory, Subcommand.CHANGE_TITLE, this::changeTitle, Subcommand.CLOSE,
+                this::closeThread, Subcommand.BOOKMARK, this::bookmarkThread));
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -204,6 +205,18 @@ private void closeThread(SlashCommandInteractionEvent event, ThreadChannel helpT
...
+    private void bookmarkThread(SlashCommandInteractionEvent event, ThreadChannel helpThread) {
+        event.getUser()
+            .openPrivateChannel()
+            .flatMap(channel -> channel.sendMessage(String.format("<#%s>", helpThread.getIdLong())))
+            .queue();
+
+        event.reply(
+                "An attempt has made to send a link of this help thread to your DMs. Check your inbox")
+            .setEphemeral(true)
+            .queue();
```
Zabuzard, line 217: Make this a single action queue. with a success and failure action that have a different text.
also, use `deferReply`.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
```diff
@@ -204,6 +205,18 @@ private void closeThread(SlashCommandInteractionEvent event, ThreadChannel helpT
         event.replyEmbeds(embed).flatMap(any -> helpThread.getManager().setArchived(true)).queue();
     }
 
+    private void bookmarkThread(SlashCommandInteractionEvent event, ThreadChannel helpThread) {
+        event.getUser()
+            .openPrivateChannel()
+            .flatMap(channel -> channel.sendMessage(String.format("<#%s>", helpThread.getIdLong())))
```
Zabuzard, line 211: move the part that creates the message into a dedicated help method please.

Zabuzard, line 211: the message is lacking any sort of context. Maybe add some:
```java
You bookmarked the following thread:
* #foobar
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Tais993: why deferReply?

Zabuzard: opening a DM channel and sending a DM there, then waiting for success/failure, could take longer than 3s.

Tais993: It *could* but is very unrealistic, means you're getting heavily rate-limited or our code is really bad, I'm fine with a defer, just preferably don't want every command to defer. As it feels like it's hiding the underlying issue

Zabuzard: i just feel like everything that needs external resources (reading files, reading database, sending http requests, ...) has to `defer`, since it cant respond immediately.

not sure if we did that everywhere though. in particular the "reading database" bit is probably not used "correctly" everywhere

Zabuzard: thats fair. im okay with either for here. its more important that we have a single queue with nice error/success responses. i dont like this "i attempted sth, check ur inbox". as if the bot doesnt know whether it worked out or not - it does know.

Taz03: > Make this a single action queue. with a success and failure action that have a different text

ok but i don't think defer reply makes sense here

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Tais993:
```suggestion
            .flatMap(channel -> channel.sendMessage(helpThread.getAsMention())))
```

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: 🙆‍♂️

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadCommand.java
Zabuzard: the code is easier to read if its separated into groups (methods). and easier to maintain.
the message also might get more complex (see my other suggestion). all in all, its just easier if thinks are separated clean.

Tais993: Even if the message gets a little "bigger"

Separating it into another method only makes sense if it's huge, or contains logic. And 3 lines is really small, nor does it contain any _real logic_.

I mean
"""
your bookmarksssszz! :-
""" + helpChannel.getAsMention();

I personally much rather keep it like this, as separating into methods won't actually help readability, and will only make it harder / more complex when trying to read this command.


I agree with you, but not to this extent, it's I don't see the benefit at this level. 

So I'll leave it up to Taz to decide whenever they want to split it up or not.


Regarding "it's just easier ..."

Well, I very much doubt that, beginners tend to not use many methods at all, the more methods, the more complicated the code is. At least, that's what my experience says with beginners.



All in all, I don't disagree with splitting things up, I only believe that this method is too small to be eligible for that.

Taz03: will that look good after a few bookmarks

marko-radosavljevic: > will that look good after a few bookmarks

Not really, but keep in mind that bot DMs a lot of different things to the user. Having just random thread links in-between other messages doesn't sound like a good UX.

Taz03: > Having just random thread links in-between other messages doesn't sound like a good UX.

yes, i'd like to have something in one single like not multi-line so doing that zab said but in single like

Tais993: That is very much unrelated to what marko said.

Marko is talking about messages in between, not multi-line messages

Tais993:
In all honesty, I don't think is the way it should be implemented.
Now it just sends messages with the thread, and that's all.

Marko was talking about messages being in-between, and that is indeed a concern.
Bookmarks are meant to be permanent, not temporary I'd assume.

So I honestly suggest a rewrite:

`/bookmark save`
`/boomark list`
`/bookmark remove`
and such

The original issue even talks about being able to add `tags`


I think we should go back to discussing what this feature exactly means, and after that implement it.

Taz03:
> In all honesty, I don't think is the way it should be implemented.

i just did what was in the issue, if there is a problem i can close the pr and we can re-discuss things in the issue

Tais993:
> So probably the best bet is to have an optional tag or memo option for the slash command, so users can organize their bookmark list with a personalized tag. It's searchable, and it can be used as a reminder what the question was about, thus correcting for poorly titled questions.

You didn't do anything that part talks about.

The issue itself wasn't the full implementation either, just a vague suggestion.

I'm not telling you to close this, but I'd like to discuss it's implementation, maybe we find we want it differently. And then you can just proceed working from this PR, or worst case restart.

Taz03:
Agree, ig best will be to discuss more on it, and this is a low effort pr so i have no problem closing it

PR #606, opened 2022-10-04
Automatically organize imports with spotless

marko-radosavljevic:
This PR introduces automatic import organizing with spotless, since it's a common pain point in the workflow.
You push → sonarcloud doesn't like the unused import → you have to fix the issue and push again. 

So with this PR, we have automatic fixes for:
- newline missing at the end of files, or too many newlines at the end of files (can be extended to other files if we want, like json config etc)
- unused imports
- import order

Automatically ordering imports is not necessary, but it's a nice thing to have, since everything would be always consistent.
Import are ordered like this: external dependencies, internal org.togetherjava classes, java classes and static imports. Since that's the ordering we have been using so far. There are still small changes in a lot of files, mostly new lines between categories.

Feel free to suggest new ordering, but keep in mind it would introduce heavy merge conflicts since every PR branch has import changes. ^^

* * * 

Diff of the `build.gradle` since there are a lot of changes by spotless:
```diff
diff --git a/build.gradle b/build.gradle
index 9f4a7cb..e04ada7 100644
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,13 @@ subprojects {
         java {
             // Excludes build folder since it contains generated java classes.
             targetExclude("build/**")
+            endWithNewline()
+            removeUnusedImports()
+            // empty string '' covers all imports that aren't explicitly specified,
+            // we use it as catch-all for external dependencies like JDA
+            // '\\#` is prefix for static imports
+            importOrder('','org.togetherjava', 'javax', 'java', '\\#')
+            // TODO Use latest eclipse jdt formatter, issue tracking this: https://github.com/Together-Java/TJ-Bot/issues/605
             eclipse('4.19.0').configFile("${rootProject.rootDir}/meta/formatting/google-style-eclipse.xml")
         }
     }
```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -7,6 +7,7 @@ import net.dv8tion.jda.api.exceptions.InvalidTokenException;
 import net.dv8tion.jda.api.requests.GatewayIntent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import org.togetherjava.tjbot.commands.Features;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.commands.system.BotCore;
--- a/application/src/main/java/org/togetherjava/tjbot/CommandReloading.java
+++ b/application/src/main/java/org/togetherjava/tjbot/CommandReloading.java
@@ -7,6 +7,7 @@ import net.dv8tion.jda.api.requests.restaction.CommandListUpdateAction;
 import org.jetbrains.annotations.Contract;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import org.togetherjava.tjbot.commands.BotCommand;
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.system.CommandProvider;
--- a/application/src/main/java/org/togetherjava/tjbot/commands/BotCommandAdapter.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/BotCommandAdapter.java
@@ -6,6 +6,7 @@ import net.dv8tion.jda.api.events.interaction.component.SelectMenuInteractionEve
 import net.dv8tion.jda.api.interactions.commands.Command;
 import net.dv8tion.jda.api.interactions.commands.build.CommandData;
 import org.jetbrains.annotations.Contract;
+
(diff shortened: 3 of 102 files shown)


Taz03:
> but keep in mind it would introduce heavy merge conflicts since every PR branch has import changes

they are easy to fix so no problem to me

Taz03 (approved):

Zabuzard (changes requested):
Two things to ease migration here:
* can you adjust the `meta/formatting/google-style-eclipse.xml` file?
* can you export an IntelliJ formatter settings file and/or tell us which checkboxes to check so that it conforms?

I want my IDE to conform with the setup as much as possible, to reduce the amount of stuff that Spotless changes.

We also promote such a setup, with a small guide, in our Wiki for beginners.

marko-radosavljevic:
There shouldn't be much or any migration with this. It's not like we are changing code formatting, it's just imports, so workflow shouldn't be affected.

You write code → IDE adds imports for you, you don't even see them because they are collapsed → you commit and push and everything is organized for the repo. So ideally, this should be completely invisible to the user.

The whole idea with our fully automatic workflow was to make it beginner-friendly. So nobody has to worry about formatting and complex configuration importing in all your tooling, only to realize half of the IDEs don't even support it.

Now, I understand wanting to conform to formatting of the code, since you don't want to flip-flop completely different formatting after each commit. But who cares about imports? Since currently it's wild west anyway, everyone does imports however they want or however their tooling does it.

* * *

Now to answer your questions,

> can you adjust the meta/formatting/google-style-eclipse.xml file

No, eclipse has a separate file format for ordering imports.
Spotless can use eclipse exported import order, looks something like this
```
#Organize Import Order
#Wed Oct 05 11:50:58 CEST 2022
0=java
1=javax
2=org
3=com
```

It's a text file, and IntelliJ doesn't understand it, doubt other IDEs either.

> can you export an IntelliJ formatter settings file and/or tell us which checkboxes to check so that it conforms?

Yeah, this should give the exact same result as our spotless import order. Since I copied our current ordering and IntelliJ default as much as possible, the difference is mostly in spacing: 

image

(Unable to resize the window, but the rest is same)

* * *

Ordering is not that important anyway, we can drop that commit as far as I'm concerned. Just some consistency is nice, if it comes at 0 cost.

- newline missing at the end of files, or too many newlines at the end of files (can be extended to other files if we want, like json config etc)
- unused imports

These two are actually important since they waste so much dev hours monthly, and they generate no changes at all.

Zabuzard (approved):
feel free to hit the button @marko-radosavljevic

PR #611, opened 2022-10-06
Replaced overview message to embed

### LINKED ISSUE #533: Overview cant update if too many questions
## Problem

Right now we have so many questions (70) that the message for `#active_questions` is too long for Discord (>2000 characters). Discord prevents the edit and the action fails:

```


    tjbot-develop_1  | 2022-08-25T11:19:46.304526000Z java.lang.IllegalStateException: Cannot build a Message with more than 2000 characters. Please limit your input.
    tjbot-develop_1  | 2022-08-25T11:19:46.304816000Z 	at net.dv8tion.jda.api.MessageBuilder.build(MessageBuilder.java:1022) ~[JDA-5.0.0-alpha.9.jar:5.0.0-alpha.9]
    tjbot-develop_1  | 2022-08-25T11:19:46.305012000Z 	at org.togetherjava.tjbot.commands.help.HelpThreadOverviewUpdater.updateOverview(HelpThreadOverviewUpdater.java:139) ~[classes/:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.305217000Z 	at org.togetherjava.tjbot.commands.help.HelpThreadOverviewUpdater.updateOverviewForGuild(HelpThreadOverviewUpdater.java:105) ~[classes/:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.305422000Z 	at net.dv8tion.jda.internal.utils.cache.AbstractCacheView.forEach(AbstractCacheView.java:101) ~[JDA-5.0.0-alpha.9.jar:5.0.0-alpha.9]
    tjbot-develop_1  | 2022-08-25T11:19:46.305614000Z 	at org.togetherjava.tjbot.commands.help.HelpThreadOverviewUpdater.runRoutine(HelpThreadOverviewUpdater.java:69) ~[classes/:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.306132000Z 	at org.togetherjava.tjbot.commands.system.BotCore.lambda$scheduleRoutines$7(BotCore.java:172) ~[classes/:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.306356000Z 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.306588000Z 	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.306935000Z 	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.307183000Z 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.307430000Z 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:19:46.307664000Z 	at java.lang.Thread.run(Thread.java:833) ~[?:?]
    tjbot-develop_1  | 2022-08-25T11:20:46.302391000Z 11:20:46.294 [pool-3-thread-1] ERROR org.togetherjava.tjbot.commands.system.BotCore - Unknown error in routine HelpThreadOverviewUpdate
```
As a result, our question overview rots and gets no updates anymore.

## Workaround

The only way to work around this is by actively closing threads until its below the limit again.

On the good note, the system heals itself once the message is small enough again.

## Approach

We have to find a solution to this problem. Unfortunately, I can not think of a very simple one.

Possibilities:
* Divide the question into multiple messages, disadvantages: Makes finding and editing the question more cumbersome
* See if we can use embeds to bypass the limit (one embed per category?) - the original reason we did not use embeds is because links in embeds could not be clicked on mobile (Android). The latest update to Android might have fixed that though.

Tais993:
Initially the overview was normal message, but this had the downside of having a limit of 2000 characters.
Embeds have a limit of 4096, which is more than double. But there was a reason for it, embeds didn't have click-through channel mentions on phone.

And that's a major issue, as it'd mean you read the thread channels name, then you find the "Discord" overview, type in the title (assuming you haven't forgotten it), and then you finally reach the channel.

Compared to a simple click, this takes a lot of time.

Since phone now has support for this, there's no reason to not use embeds.
So after this we support more than 2x the amount of questions than before!

Closes #533 

This breaks backwards compatibility, the old message has to get deleted.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadOverviewUpdater.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadOverviewUpdater.java
@@ -1,5 +1,6 @@
 package org.togetherjava.tjbot.commands.help;
 
+import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Message;
@@ -13,6 +14,7 @@ import net.dv8tion.jda.api.utils.messages.MessageCreateData;
 import net.dv8tion.jda.api.utils.messages.MessageEditBuilder;
 import net.dv8tion.jda.api.utils.messages.MessageEditData;
 import net.dv8tion.jda.internal.requests.CompletedRestAction;
+import org.jetbrains.annotations.NonNls;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.togetherjava.tjbot.commands.MessageReceiverAdapter;
@@ -39,7 +41,7 @@ import java.util.stream.Collectors;
 public final class HelpThreadOverviewUpdater extends MessageReceiverAdapter implements Routine {
     private static final Logger logger = LoggerFactory.getLogger(HelpThreadOverviewUpdater.class);
 
-    private static final String STATUS_TITLE = "## __**Active questions**__ ##";
+    private static final @NonNls String STATUS_TITLE = "## Active questions ##";
     private static final int OVERVIEW_QUESTION_LIMIT = 150;
     private static final AtomicInteger FIND_STATUS_MESSAGE_CONSECUTIVE_FAILURES =
             new AtomicInteger(0);
@@ -120,7 +122,9 @@ public final class HelpThreadOverviewUpdater extends MessageReceiverAdapter impl
         logger.debug("Found {} active questions", activeThreads.size());
 
         MessageEditData message = new MessageEditBuilder()
-            .setContent(STATUS_TITLE + "\n\n" + createDescription(activeThreads))
+            .setEmbeds(new EmbedBuilder().setTitle(STATUS_TITLE)
+                .setDescription(createDescription(activeThreads))
+                .build())
             .build();
 
         getStatusMessage(overviewChannel)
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadOverviewUpdater.java
```diff
@@ -39,7 +41,7 @@
 public final class HelpThreadOverviewUpdater extends MessageReceiverAdapter implements Routine {
     private static final Logger logger = LoggerFactory.getLogger(HelpThreadOverviewUpdater.class);
 
-    private static final String STATUS_TITLE = "## __**Active questions**__ ##";
+    private static final @NonNls String STATUS_TITLE = "## Active questions ##";
```
Zabuzard, line 44: The presence of this annotation feels super random and confusing. Mostly because it is the only time its used yet. To prevent people from confusion, Id suggest to remove it (or to add it everywhere it applies instead).

That said, it doesnt even seem correct. This string **is** user facing. It technically would need localization. The annotation says the opposite.

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadOverviewUpdater.java
```diff
@@ -163,12 +167,13 @@ private static RestAction<Optional<Message>> getStatusMessage(MessageChannel cha
     }
 
     private static boolean isStatusMessage(Message message) {
-        if (!message.getAuthor().equals(message.getJDA().getSelfUser())) {
+        if (message.getEmbeds().isEmpty()
```
Zabuzard, line 170: can `getEmbeds()` return `null`? if so, this would be unsafe

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadOverviewUpdater.java
Tais993: I'll remove the annotation.

Initially I was trying or nonce's work, those aren't localised, but it didn't work and forgot to remove afterwards.

Zabuzard (approved):
fix spotless/sonar issues and good 👍

Taz03 (approved):

Zabuzard (approved):

PR #614, opened 2022-10-07
TopHelpersCommand message is too long

### LINKED ISSUE #603: TopHelpersCommand message is too long
## Issue

The `TopHelpersCommand` message is too long and crashes:

crash

Stacktrace:

stacktrace

Gotta make it a bit smaller.

## Fix

We can just reduce the amount of people it lists from `20` to, lets say, `18`:

https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java#L43

To reproduce the issue locally and test whether that did the trick, one would have to fake the result returned by the DB call here:

https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java#L71

Or directly put some fake data into the database in order to hit the limit of 20 people. Also, they should have a message count of around `10000` each. All of this contributes to the size of the message.

Zabuzard:
## Overview

Fixes #603 which describes a bug when using `/top-helpers` resulting in a too long message, not accepted by Discord.

The approach is simple yet effective:
* reduced the rows from 20 to 18
* abbreviate long user names (limit 15)
* move the description out of the column

these factors effectively reduced the message that was too long from 2300 characters down to 1100 characters, which is perfectively within limits.

## Notes

Since this class also needed abbreviation logic, I elevated the `abbreviate` method from within the `DiscordLogForwarder` to `MessageUtils`. And also gave it a unit test.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java
@@ -17,6 +17,7 @@ import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.utils.MessageUtils;
 import org.togetherjava.tjbot.db.Database;
 
 import javax.annotation.Nullable;
@@ -42,7 +43,9 @@ public final class TopHelpersCommand extends SlashCommandAdapter {
     private static final Logger logger = LoggerFactory.getLogger(TopHelpersCommand.class);
     private static final String COMMAND_NAME = "top-helpers";
     private static final String MONTH_OPTION = "at-month";
-    private static final int TOP_HELPER_LIMIT = 20;
+    private static final int TOP_HELPER_LIMIT = 18;
+
+    private static final int MAX_USER_NAME_LIMIT = 15;
 
     private final Database database;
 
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java
```diff
@@ -143,29 +146,30 @@ private static void handleTopHelpers(Collection<TopHelperResult> topHelpers,
                     userIdToMember.get(topHelper.authorId())))
             .toList();
 
-        String message =
-                "```java%n%s%n```".formatted(dataTableToString(topHelpersDataTable, timeRange));
+        String message = """
+                ```java
+                // for %s
```
Tais993, line 151: What does this mean, I need some pictures to understand the context

Tais993 (commented):
Can you add screenshots of how it looks now?

application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersCommand.java
Zabuzard: https://i.imgur.com/bl8MbIP.png

Zabuzard: And some users with very long names, lets say an user named `FooBarBazHelloWorld` would now be abbreviated to `FooBarBazHel...` (we have the ID on the left anyways)

Tais993 (approved):
next time give screenshot of before your changes and after.
Would make it a little easier to understand what actually changed

Zabuzard:
> next time give screenshot of before your changes and after. Would make it a little easier to understand what actually changed

sure. was just a bit difficult, since i dont have 20+ users on my private server to make the problem obvious. and i was too lazy to mock the DB just for screenshots.

and the real tables from PROD contained sensitive info that i also was too lazy to mock lol.

but yeah, agreed.

Taz03 (approved):

Zabuzard:
just rebased, no code changes. already had 2 approvals. merging once pipeline succeeds

PR #616, opened 2022-10-07
Link/Image previews for tags

### LINKED ISSUE #270: Tags dont display image and link previews
## Overview

Right now we use embeds for our tag system, but embeds (by default) dont display previews of image links and links in general.

This results in a major UX drawback, compared to the old Dyno system:

### Install
`/tag install` vs `?tag install`:

tj-bot install
dyno install

### Ask
`/tag ask` vs `?tag ask`

tj-bot ask
dyno ask

## Ideas

At the top of my head, I see two possible fixes:
* the easy fix - we just go back to plain messages instead of embeds
* the advanced fix - we parse the content, identify all links and attach them as proper image/link to the embed, so that a preview is generated
  * this could be just attached to the very end of the message, like it is right now in the dyno system
  * or, but probably more complex, be added exactly at the point of mentioning in the message. but that likely requires splitting the message in multiple.

Zabuzard:
## Overview

Fixes #270 by implementing automatic link and image previews to the `/tag` command.

Image only preview:
example

Link preview:
link

## Utility

This feature is also available as general purpose utility via the `LinkPreviews` class, yielding records `LinkPreview`.

The utility offers two methods:
* `static List<String> extractLinks(String content)`
* `static CompletableFuture<List<LinkPreview>> createLinkPreviews(List<String> links)`

Features like #637 will likely make use of this as well.

## Dependencies

Adds `io.github.url-detector:url-detector:0.1.23` to the deps. URL regex is not something we want to have in our own code base lol.

Also adds **JSoup** for the OpenGraph/Twitter meta-extraction.

## Checklist

* [x] draft
* [x] opengraph extraction
* [x] add warn logs to all failures
* [x] polish
* [x] javadoc
* [x] ~~possibly unit tests~~ (not today)

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -57,6 +57,8 @@ dependencies {
 
     implementation 'io.mikael:urlbuilder:2.0.9'
 
+    implementation 'org.jsoup:jsoup:1.15.3'
+
     implementation 'org.scilab.forge:jlatexmath:1.0.7'
     implementation 'org.scilab.forge:jlatexmath-font-greek:1.0.7'
     implementation 'org.scilab.forge:jlatexmath-font-cyrillic:1.0.7'
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -68,6 +68,8 @@ dependencies {
 
 	implementation 'com.github.freva:ascii-table:1.8.0'
 
+    implementation 'io.github.url-detector:url-detector:0.1.23'
+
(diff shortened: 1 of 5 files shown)


Zabuzard:
> This pull request is stale because it has been open 30 days with no activity. Remove stale label, comment or add the valid label or this will be closed in 5 days.

yikes. ill have a look again 😬

Zabuzard:
8 days since last change, merging

PR #617, opened 2022-10-08
Auto-pin first message in help threads

### LINKED ISSUE #479: Auto-pin first message in help threads
The bot should automatically pin the first message in help threads, this is usually the OPs question, or directly next to it.

Makes it easier for people to navigate in a help thread if they can just jump all the way up.

Implementation-wise, this could be added directly into the flow of `AskCommand.java` and `ImplicitAskListener.java`, where it also creates the threads and posts the first automatic messages.

Taz03:
closes #479 

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -128,6 +128,7 @@ public final class AskCommand extends SlashCommandAdapter {
         helper.writeHelpThreadToDatabase(author, threadChannel);
         return sendInitialMessage(guild, threadChannel, author, title, category)
             .flatMap(any -> notifyUser(eventHook, threadChannel))
+            .flatMap(any -> threadChannel.pinMessageById(threadChannel.getLatestMessageId()))
             .flatMap(any -> helper.sendExplanationMessage(threadChannel));
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -14,7 +14,6 @@ import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.commands.build.OptionData;
 import net.dv8tion.jda.api.requests.ErrorResponse;
 import net.dv8tion.jda.api.requests.RestAction;
-import net.dv8tion.jda.internal.requests.CompletedRestAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -160,18 +160,6 @@ public final class AskCommand extends SlashCommandAdapter {
                 Please ask your question there, thanks.""".formatted(threadChannel.getAsMention()));
     }
 
-    private static RestAction<Void> pinFirstMessage(ThreadChannel threadChannel) {
-        long messageId = threadChannel.getLatestMessageIdLong();
-
-        if (messageId == 0) {
-            logger.warn("Can't find the first message in this help thread (#{})",
-                    threadChannel.getId());
-            return new CompletedRestAction<>(threadChannel.getJDA(), null);
-        }
-
-        return threadChannel.pinMessageById(messageId);
-    }
-
     private static void handleFailure(Throwable exception, InteractionHook eventHook) {
         if (exception instanceof ErrorResponseException responseException) {
             ErrorResponse response = responseException.getErrorResponse();
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -127,8 +127,8 @@ public final class AskCommand extends SlashCommandAdapter {
             Member author, String title, String category, Guild guild) {
         helper.writeHelpThreadToDatabase(author, threadChannel);
         return sendInitialMessage(guild, threadChannel, author, title, category)
-            .flatMap(any -> notifyUser(eventHook, threadChannel))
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -128,6 +128,7 @@ private RestAction<Message> handleEvent(InteractionHook eventHook, ThreadChannel
         helper.writeHelpThreadToDatabase(author, threadChannel);
         return sendInitialMessage(guild, threadChannel, author, title, category)
             .flatMap(any -> notifyUser(eventHook, threadChannel))
+            .flatMap(any -> threadChannel.pinMessageById(threadChannel.getLatestMessageId()))
```
Zabuzard, line 131: the latest message isnt necessarily the first, is it?

does it run against the cache or does it send a blocking request?

i suspect getLatestMessageID can also fail and maybe return null?

Zabuzard (changes requested):

Zabuzard:
could u add some screenshot how it looks like?

Taz03:
> could u add some screenshot how it looks like?

added

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: > the latest message isnt necessarily the first, is it?

i mean, it sends the first message then i pin it so maybe yes it is the first

> does it run against the cache or does it send a blocking request?

yes it uses cache i believe

> i suspect getLatestMessageID can also fail and maybe return null?

it is annotated with nonnull, it returns 0 instead of null

Zabuzard: I see. It should be safe in 99% of cases. Its possible though that it runs into an error message if its `0`. We could:
* add an if-else, log a warning if its 0 and not call `pinMessageById` then

To make that readable, put the stuff into a helper method like `pinFirstMessage`

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -159,6 +161,17 @@ private static RestAction<Message> notifyUser(InteractionHook eventHook,
                 Please ask your question there, thanks.""".formatted(threadChannel.getAsMention()));
     }
 
+    private static RestAction<Void> pinFirstMessage(ThreadChannel threadChannel) {
+        long messageId = threadChannel.getLatestMessageIdLong();
+
+        if (messageId == 0) {
+            logger.debug("Can't find the first message in this help thread (#{})", threadChannel.getId());
```
Zabuzard: should be warn for now, not debug

Zabuzard (changes requested):

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -158,6 +160,18 @@ private static RestAction<Message> notifyUser(InteractionHook eventHook,
...
+    private static RestAction<Void> pinFirstMessage(ThreadChannel threadChannel) {
+        long messageId = threadChannel.getLatestMessageIdLong();
+
+        if (messageId == 0) {
+            logger.warn("Can't find the first message in this help thread (#{})",
+                    threadChannel.getId());
+            return new CompletedRestAction<>(threadChannel.getJDA(), null);
+        }
+
+        return threadChannel.pinMessageById(messageId);
```
Tais993, line 172: This is useless, see above

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -128,6 +129,7 @@ private RestAction<Message> handleEvent(InteractionHook eventHook, ThreadChannel
         helper.writeHelpThreadToDatabase(author, threadChannel);
         return sendInitialMessage(guild, threadChannel, author, title, category)
             .flatMap(any -> notifyUser(eventHook, threadChannel))
+            .flatMap(any -> pinFirstMessage(threadChannel))
```
Tais993, line 132: Undo all changes, and instead use this
```suggestion
        return sendInitialMessage(guild, threadChannel, author, title, category)
            .flatMap(Message::pin)
            .flatMap(any -> notifyUser(eventHook, threadChannel))
```

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: lmao

Taz03: wow ok

Taz03: wait no this didn't worked

Taz03: image

Zabuzard: sounds like ur running it on the wrong message then. review the order of operations. maybe u have to move it one up or down or sth. i.e. right after the message from the thread is send.

Tais993: owh

Tais993: mb
```java
        return sendInitialMessage(guild, threadChannel, author, title, category)
            .flatMap(Message::pin)
            .flatMap(any -> notifyUser(eventHook, threadChannel))
            ```
            should work then

Taz03: image

yes works again

Taz03: yes make a new one @Tais993

Tais993 (approved):

Zabuzard (approved):

PR #628, opened 2022-10-14
Added POC for /google command

### LINKED ISSUE #627: Google/Stackoverflow searching for #active-questions and or independent slash command
There are often a questions that are asked in the #active-questions channel that could we resolved with a quick Google search. An example what I have seen recently was along the lines of "Why do we need a connection pool for JDBC?". The exact question in that exact format, searched on Google gave the questioner the exact answer they were looking for. 

Introducing a feature within the bot that does a quick Google or Stackoverflow search would an excellent feature for such questions. The bot can query the thread title and post back the first x results to the user. In that scenario, if the searches answer the users question, questions can be closed immediatly without the user having to wait for somebody to respond. 

The following API's are available for the implementation:

https://api.stackexchange.com/docs/search
https://developers.google.com/custom-search/v1/overview

Google limits us to 100 searches per day. As discussed with Zabuzard, it should be fine as long as the user inititates the request and not the bot automatically. The end solution may be that the bot sends a message saying "Would you like me to search [title] on Google?". User can react to the message and see a bunch of results. 

With the additional functionality, an introduction to two new slash commands.
/google <search term>
/stackoverflow <search term>

surajkumar:
This pull requests implements 2 new slash commands:

/google <query>
This command searches Google for the input query and provides search results accordingly.

/googleit
This command takes the thread in #active_questions that the command was invoked from and Google's the title of the thread. The command only works in a thread within #active_questions. 

In additon to the /googleit command, a message has been added to the bot that tells users that they can ask the bot to google their question.

Closes #627.

example of /google

example2 the modified bot message in #active_questions

example3 of /googleit


marko-radosavljevic:
Thanks for your contribution, and welcome to the project! :relaxed: :heart:

Can you update your PR description and include some pictures of UI/UX in discord?

surajkumar:
This is for Zabu. I'll add all the relevant information once it's ready. This PR is to not be merged.

marko-radosavljevic:
Alright, will convert it to draft in that case. Thanks ^^

Zabuzard:
id suggest u make the PR a bit prettier. people will be happy to discuss the UX if u add a few pictures. not just me 👍

application/src/main/java/org/togetherjava/tjbot/commands/search/GoogleSearchStrategy.java
```diff
@@ -0,0 +1,28 @@
...
+import io.mikael.urlbuilder.UrlBuilder;
+
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.concurrent.CompletableFuture;
+
+public class GoogleSearchStrategy extends SearchStrategy {
+    private static final String API_KEY =
+            "1d84500b082341aad66de5cf02f7c83d8608411b01b7ffcbb2d40042eea250df";
```
Zabuzard, line 12: that has to move into config. id suggest u invalidate this token. people can now use it to do harm in ur name.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/search/GoogleSearchStrategy.java
surajkumar: That's a test token I generated using dummy credentials. I can change this once you have created that resource file you mentioned earlier.

surajkumar:
Updated the code and PR to be ready for prod. The API key should now be passed as an environment variable called "serpapi-api-key". Once [Zabuzard](https://github.com/Zabuzard) creates the api keys file/store we can migrate to that.

Tais993:
See the "config" class and config.json file

Exists already

marko-radosavljevic:
UI/UX look nice, I would just trim it down considerably. Just the first embed is enough, the other 5 embeds are not really useful, since they are small and don't provide a link.

If beginners knew how to google their answer, they wouldn't come to us for help and guidance, so it's important to not overwhelm them with 20 results. If the answer is not in the first 3 results, it's a badly asked question anyway, and other results won't help either. When was the last time you used 2nd page on google/searx? ^^

surajkumar:
Changed the code to grab API key from the config.json. 


> UI/UX look nice, I would just trim it down considerably. Just the first embed is enough, the other 5 embeds are not really useful, since they are small and don't provide a link.
> 
> If beginners knew how to google their answer, they wouldn't come to us for help and guidance, so it's important to not overwhelm them with 20 results. If the answer is not in the first 3 results, it's a badly asked question anyway, and other results won't help either. When was the last time you used 2nd page on google/searx? ^^

The first embeds are the q&s from "People also ask" on the Google search page. 
[Example](https://i.imgur.com/GwF0Bqd.png)

The 5 embeds after it are clickable. Just click the titles. 
As for your comments regarding the "2nd page". I use it all the time. Though in this case for the bot, it's the first 5 search results on the 1st page.

How would you like it formatted?

surajkumar:
Handing over. I won't have time to continue working on this as I was hoping it would only take a day to complete. All the changes are done. I suppose the only thing left would be the actual visual message. If it's sufficent, then this can be merged otherwise, whoever takes over can make the visual changes. Thanks

marko-radosavljevic:
> The first embeds are the q&s from "People also ask" on the Google search page.

Ah, I see, now it's clear why there is a difference between first and other embeds, thanks.

Does anyone use 'people also ask'? I don't care what other people (also) ask, I care about the thing I asked.

Maybe it's useful for people that ask bad questions, so the engine helps them out with "hey, maybe this is what you are looking for". Or its quick answers to similar questions, that can easily fit in embed? Idea being that people don't have to open links and read long blog posts, answer is right there?

Genuinely curious, since I don't use that thing. But from the examples you posted, it looks useful. Maybe the normal search results would be more useful if they had the same amount of context as the first embed tho.

I would probably go with just 3 results, so it's not too overwhelming, but with enough context. Ideally, with the whole answer in the embed, so the user doesn't have to open links in their browser and read 30-minute guides. Maybe 'people also ask' is not a bad idea here, so 1 of that and 2 regular top results.

surajkumar:
> > The first embeds are the q&s from "People also ask" on the Google search page.
> 
> Ah, I see, now it's clear why there is a difference between first and other embeds, thanks.
> 
> Does anyone use 'people also ask'? I don't care what other people (also) ask, I care about the thing I asked.
> 
> Maybe it's useful for people that ask bad questions, so the engine helps them out with "hey, maybe this is what you are looking for". Or its quick answers to similar questions, that can easily fit in embed? Idea being that people don't have to open links and read long blog posts, answer is right there?
> 
> Genuinely curious, since I don't use that thing. But from the examples you posted, it looks useful. Maybe the normal search results would be more useful if they had the same amount of context as the first embed tho.
> 
> I would probably go with just 3 results, so it's not too overwhelming, but with enough context. Ideally, with the whole answer in the embed, so the user doesn't have to open links in their browser and read 30-minute guides. Maybe 'people also ask' is not a bad idea here, so 1 of that and 2 regular top results.

So the "people also ask" section is usually your same question but rephrased differently (often better). I would actually keep that there, it's very useful and I personally use it alot. Google by default limits this to 4. I reckon we should keep that and limit the organic results to maybe 2. So the bots message isn't too long and overwhelming.

To point out, the "people also ask" is not always visible due to the nature of some question (e.g. if the question is uncommon) in which case, maybe we can keep the top 4 results of the organic results.

So...

4 "people also ask"
2 organic search results

if no "people also ask" then 4 organic search results. 

Opinion?

Zabuzard:
OP wants to transfer this to whoever else wants to work on it. I am going to move it from the fork to a branch.

Zabuzard:
Transferred into #637 Thanks for the efforts @surajkumar 👍

PR #493, opened 2022-08-13
Adding the /modmail command

DevSerendipity:
Commands which redirects command called from Guild or Dm to a dedicated channel in Config#getModAuditLogChannelPattern()

Samples of message being public (can be seen which user called the command)
`/modmail message:test from guild without message private option message-private:true`
image
Sample of message being private (can not be seen which user called the command)
`/modmail message:test from guild with message private option message-private:true`
image

The messages related to the mode Utils

image

PR #630, opened 2022-10-15
Updated dm message for mod actions

### LINKED ISSUE #620: Improve formatting on user facing moderation messages
We have almost no formatting on user facing moderation messages, which hurts readability. I would like 'reason' part to stick out more.

Currently, it looks like this on 1920x1080:

image

And on my phone:

image

Maybe some embed magic, bold on 'reason', or just an empty line to create some space. My eyes have to jump around to find where the important part starts, since reasoning completely blended in with surrounding text.

Taz03:
closes #620 

image

the thumbnail image is server icon

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/BanCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/BanCommand.java
@@ -2,7 +2,6 @@ package org.togetherjava.tjbot.commands.moderation;
 
 import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.entities.*;
-import net.dv8tion.jda.api.events.GenericEvent;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
 import net.dv8tion.jda.api.exceptions.ErrorResponseException;
 import net.dv8tion.jda.api.interactions.InteractionHook;
@@ -83,22 +82,14 @@ public final class BanCommand extends SlashCommandAdapter {
         return event.reply(message).setEphemeral(true);
     }
 
-    private static RestAction<Boolean> sendDm(ISnowflake target,
-            @Nullable ModerationUtils.TemporaryData temporaryData, String reason, Guild guild,
-            GenericEvent event) {
-        String durationMessage =
-                temporaryData == null ? "permanently" : "for " + temporaryData.duration();
-        String dmMessage =
-                """
-                        Hey there, sorry to tell you but unfortunately you have been banned %s from the server %s.
-                        If you think this was a mistake, please contact a moderator or admin of the server.
-                        The reason for the ban is: %s
-                        """
-                    .formatted(durationMessage, guild.getName(), reason);
+    private static RestAction<Boolean> sendDm(User target,
+            @Nullable ModerationUtils.TemporaryData temporaryData, String reason, Guild guild) {
+        String duration = temporaryData == null ? "Permanently" : temporaryData.duration();
 
-        return event.getJDA()
-            .openPrivateChannelById(target.getId())
-            .flatMap(channel -> channel.sendMessage(dmMessage))
+        return target.openPrivateChannel()
+            .flatMap(channel -> channel.sendMessageEmbeds(
+                    ModerationUtils.getModActionEmbed(guild, ACTION_VERB, "", reason, duration)
+                        .build()))
             .mapToResult()
             .map(Result::isSuccess);
     }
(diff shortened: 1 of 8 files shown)


Tais993:
I'm sticking to my opinion that I find the guild title and the thumbnail feel really random / off.
I'd really instead use the author field like I showed in #620.

I'd love some other opinions on this

(also, next time prefix your branch with `feature` or `bug` and such, a.e `feature/moderation-message-improvement`

Taz03:
@Tais993 turns out i can put the second value as `null` and it'll work, doing it then

also zabu said branch name doesn't really matter but yea i can do that it'll just be harder for me to type everytime

marko-radosavljevic:
How would it look like with a slightly bigger reason, for example for:

> A wonderful serenity has taken possession of my entire soul, like these sweet mornings of spring which I enjoy with my whole heart. I am alone, and feel the charm of existence in this spot, which was created for the bliss of souls like mine. I am so happy, my dear friend...

Which would be a medium-sized reasoning, that is very commonly used. Because currently it looks like only few words would fit nicely.

Taz03:
@marko-radosavljevic 
image
ik it is looking trash

what about this one
image

marko-radosavljevic:
Yeah, I think the second one is nice. :relaxed:

Taz03:
> Yeah, I think the second one is nice. relaxed

ok pushing the changes

Zabuzard:
i like the options we have with the formatting. looks much nicer.

however, the actual message is less friendly now. it is very important that those messages are super friendly and with smileys and whatever. the old messages were better.

we want that users who behaved bad will adapt and come back. so we have to be very kind to them when showing them a message that they have been temp banned or warned. they must perceive it as "its all good, dont worry. we want u in our community. just adapt please, thanks"

Zabuzard (changes requested):
ur latest changes miss which server. the bot is designed to be able to run on multiple guilds/servers at the same time

Taz03:
> ur latest changes miss "which server" in the messages. the bot is designed to be able to run on multiple guilds/servers at the same time

actually the server details are already given at the top as author so i thought its ok if just remove it, do u want them back?

Zabuzard:
could u update ur PR description with the up2date pictures? id like to see how everything looks before i review, cause that massively helps the review process.

Zabuzard:
> > could u update ur PR description with the up2date pictures? id like to see how everything looks before i review, cause that massively helps the review process.
> 
> it is already updated

i thought u wanted to make the messages friendlier?

Tais993:
> i thought u wanted to make the messages friendlier?

image

This is the current message, so yes, it's a lot friendlier.

But I don't like that it says "MUTE" in the title, I'd personally prefer something like "You have been muted"

Taz03:
> i thought u wanted to make the messages friendlier?

yea i did that already, u can read the msg in actual code i thought u wanted the updated ui

will update the messages once my frnd gets online and i can mute him again

Taz03:
> But I don't like that it says "MUTE" in the title, I'd personally prefer something like "You have been muted"

i'd like to stay with the current format coz this way the user can know everything at a glance and won't need to read through text

like its a mute and for this duration and for this reason. and now it also explains it in the description

Zabuzard:
Im okay with a short title like `Mute`. But it needs to be user friendly, so `Mute`, not `MUTE`. otherwise u get the feeling that someone is screaming at u.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ModerationUtils.java
```diff
@@ -58,7 +59,7 @@ private ModerationUtils() {
     static EmbedBuilder getModActionEmbed(Guild guild, String action, String description,
             String reason, boolean isPunishAction) {
         return new EmbedBuilder().setAuthor(guild.getName(), null, guild.getIconUrl())
-            .setTitle(action.toUpperCase())
+            .setTitle(action.substring(0, 1).toUpperCase() + action.substring(1))
```
Zabuzard, line 61: The design on this `action` seems incorrect.
1. u called it `action`, but what is it exactly? its supposed to be the **title**. but u give it the `ACTION_VERB`
2. the action verb is sth like "banned", "warned" - a verb. its not a title. You want it to say "Ban" and "Mute", not "Banned".

You should adjust the callsite and definitely improve the naming on your variable here, it should be `actionTitle` and in the javadoc it should give an example, like `"ban"`.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/BanCommand.java
```diff
@@ -90,7 +90,7 @@ private static RestAction<Boolean> sendDm(User target,
 
         return target.openPrivateChannel()
             .flatMap(channel -> channel.sendMessageEmbeds(ModerationUtils
-                .getModActionEmbed(guild, ACTION_VERB, description, reason, duration)
+                .getModActionEmbed(guild, "Ban", description, reason, duration)
```
Zabuzard: better. but please put it into a constant in the class, similar to ACTION_VERB. so maybe ACTION_TITLE. same with the other classes, thanks

Zabuzard (changes requested):

Zabuzard:
just as a headsup, i smell annoying merge conflicts with #493 which also touches those messages

PR #631, opened 2022-10-15
minor fixes in website files

Taz03:
- fixed formatting
- minor code fixing


Tais993:
The formatting changes have to be discussed first. In all honesty, I don't see any reason why we should change the formatting.

Taz03:
good point

PR #632, opened 2022-10-15
website code fixes

Taz03:
- fixed css values
- used proper equals operator


Tais993:
Was this intentionally made a draft?

Zabuzard:
I dont see a point in maintaining code that we dont own. This is all template code. Ideally the website wouldnt even be part of TJ-Bot and then we should somehow clearly clarify what is template and what is our own code.

I actually marked all of the template files as "ignore this file" in Sonar and CodeFactor.

That said, for now we should probably only do UX/content changes on the website, and only fix actual bugs or critical stuff in the code.

PR #649, opened 2022-10-21
Fix Top-Helper message length exploit

### LINKED ISSUE #648: Top-Helper message length exploit
**Describe the bug**
You can easily farm Top-Helper message points by adding thousands of `Zero Width Joiner`s (invisible characters) to your help messages and boost yourself up the Top-Helper list.

**Expected behavior**
Unwanted characters should not increase the message points

**To Reproduce**
1. Go to https://emojipedia.org/zero-width-joiner/ and copy the character
2. Write a help message and add the character until you reach the message limit
3. Take a look at the `help_channel_messages` database table
4.  => The message length is 2000

**Screenshots**
image
after a few messages

## Fix

We can just adjust the value put into the database at `TopHelpersMessageListener`:

code

to not use just `.length()`, but ignore any fancy character. It is probably okay if we just count `a-zA-Z0-9` or similar.

Nxllpointer:
Fixes #648 

One was able to farm Top-Helper message length points by adding invisible characters to help messages

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
@@ -18,7 +18,13 @@ import static org.togetherjava.tjbot.db.generated.tables.HelpChannelMessages.HEL
  * {@link TopHelpersCommand} to pick them up.
  */
 public final class TopHelpersMessageListener extends MessageReceiverAdapter {
-    private static final String UNCOUNTED_CHARS = "[^\\x20-\\x7E]";
+    /**
+     * Matches invisible control characters and unused code points
+     * 
+     * @see <a href="https://www.regular-expressions.info/unicode.html#category">Unicode
+     *      Categories</a>
+     */
+    private static final String UNCOUNTED_CHARS = "\\P{C}";
 
     private final Database database;
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
@@ -18,6 +18,8 @@ import static org.togetherjava.tjbot.db.generated.tables.HelpChannelMessages.HEL
  * {@link TopHelpersCommand} to pick them up.
  */
 public final class TopHelpersMessageListener extends MessageReceiverAdapter {
+    private static final String UNCOUNTED_CHARS = "[^\\x20-\\x7E]";
+
     private final Database database;
 
     private final Predicate<String> isStagingChannelName;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
```diff
@@ -18,6 +18,8 @@
  * {@link TopHelpersCommand} to pick them up.
  */
 public final class TopHelpersMessageListener extends MessageReceiverAdapter {
+    private static final String UNCOUNTED_CHARS = "[^\\x20-\\x7E]";
```
Zabuzard, line 21: that regex is ugly and unmaintainable. there are a few better alternatives. first, a direct translation that is at least readable `[^ -~]` (space to tilde).

but I would actually go for an approach using the named classes, such as `\P{C}`, which are all visible characters.

More here: https://stackoverflow.com/a/66340317/2411243

Also, such unreadable code at least needs a quick comment explaining it (and a link to the SO where u got it from, if any).

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/tophelper/TopHelpersMessageListener.java
Zabuzard: That said, I would also be okay with just `\\W` (non word characters).

java-coding-prodigy: For math and algorithm-based help there would be a significant number of non-word characters.

Nxllpointer: hm i see that is way more readable

Nxllpointer:
> Can you test it quickly? Just to make sure it really works as expected, thanks.

It works with the zero width joiner. Charachters like äöü for example were also excluded but because the discord is english this should not be an issue

Taz03 (approved):

Zabuzard:
@Nxllpointer why are umlaute excluded with `\P{C}`? that category should list all invisible characters. the umlaute are visible. are we really sure we just filtered what we expect and didnt kill everything? please doublecheck that regular characters still count.

java-coding-prodigy:
> @Nxllpointer why are umlaute excluded with `\P{C}`? that category should list all invisible characters. the umlaute are visible. are we really sure we just filtered what we expect and didnt kill everything? please doublecheck that regular characters still count.

You could do `\P{C}|<Whatever the regex for umluate is>` in that case

Zabuzard:
im okay with what it is. im just confused why its not included. which yields me to thinking that there might be a bug in your code. let me check it on regex101.com quickly

Nxllpointer:
Maybe it's an issue with how java handles Unicode categories

Zabuzard:
> Maybe it's an issue with how java handles Unicode categories

Are you sure it messes up umlaute? When I test it locally, it works.

test

Zabuzard:
Btw, you should have used a `Pattern` instead of a raw string for the replacement. Would be faster. But welp, too late.

Nxllpointer:
Hm. I used a string because replaceAll does not accept a pattern. Is there an easy readable solution?

Zabuzard:
> Hm. I used a string because replaceAll does not accept a pattern. Is there an easy readable solution?

replacement via pattern goes through the pattern, not from string. sth like `pattern.replace(...)`.

anyways, lets concentrate on why u think that the code doesnt work for umlaute. please elaborate

Nxllpointer:
Maybe I also can't read a number. If it works for you it should also work for the bot

Nxllpointer:
If you want to I can change it to a pattern later

Zabuzard:
> 

ur not getting me. u claimed that it doesnt work for umlaute. why did u say that? did u try it out and it didnt work? **do we have a bug in the code?**

Nxllpointer:
When I tried it it didn't work. Maybe I just forgot to update the database viewer. You can test if it works.

Nxllpointer:
@Zabuzard oops I just realized that instead of counting the visible chars I counted the invisible ones... I will make a PR with the fix and using Pattern.

Nxllpointer:
PR should fix everything now. Sorry for this stupid mistake.

PR #650, opened 2022-10-22
Added list for reminders

### LINKED ISSUE #641: Add /reminders command
## Motivation

We have the `/remind` command that lets users create reminders.

Its easy to forget about all currently pending reminders though and sometimes an user would like to review them before they are actually triggered.

## Overview

Since we have all pending reminders for all users in the database, we could easily add a `/reminders` command that shows (as **ephemeral** message) all pending reminders of the current user.

## UX

The reminders could be shown as embeds, one by one. And be limited to max ~10, sorted by when they are fired off (soonest first). It should also state the total amount of pending reminders, in case it was limited.

Details can be figured out once the first proof of concept is there.

Taz03:
closes #641 

2022-10-22_204946-530x365_scrot

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
@@ -1,5 +1,6 @@
 package org.togetherjava.tjbot.commands.reminder;
 
+import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.ISnowflake;
 import net.dv8tion.jda.api.entities.User;
@@ -7,19 +8,20 @@ import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEve
 import net.dv8tion.jda.api.interactions.callbacks.IReplyCallback;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.commands.build.OptionData;
-
 import net.dv8tion.jda.api.interactions.commands.build.SubcommandData;
-import org.jooq.Result;
 import org.jooq.SelectWhereStep;
+
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.db.Database;
 import org.togetherjava.tjbot.db.generated.tables.records.PendingRemindersRecord;
 
+import java.awt.*;
 import java.time.*;
 import java.time.temporal.TemporalAmount;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
 
 import static org.togetherjava.tjbot.db.generated.Tables.PENDING_REMINDERS;
@@ -38,7 +40,7 @@ import static org.togetherjava.tjbot.db.generated.Tables.PENDING_REMINDERS;
  * <p>
  * Pending reminders are processed and send by {@link RemindRoutine}.
  */
-public final class RemindCommand extends SlashCommandAdapter {
+public final class ReminderCommand extends SlashCommandAdapter {
     private static final String COMMAND_NAME = "reminder";
     private static final String LIST_SUBCOMMAND = "list";
     static final String ADD_SUBCOMMAND = "add";
@@ -54,14 +56,15 @@ public final class RemindCommand extends SlashCommandAdapter {
     static final int MAX_PENDING_REMINDERS_PER_USER = 100;
 
     private final Database database;
-    private final Map<String, Consumer<SlashCommandInteractionEvent>> subcommandHandler = Map.of(ADD_SUBCOMMAND, this::handleAddCommand, LIST_SUBCOMMAND, this::handleListCommand);
+    private final Map<String, Consumer<SlashCommandInteractionEvent>> subcommandHandler = Map
+        .of(ADD_SUBCOMMAND, this::handleAddCommand, LIST_SUBCOMMAND, this::handleListCommand);
 
     /**
      * Creates an instance of the command.
      *
      * @param database to store and fetch the reminders from
      */
-    public RemindCommand(Database database) {
+    public ReminderCommand(Database database) {
         super(COMMAND_NAME, "Reminds you after a given time period has passed (e.g. in 5 weeks)",
                 CommandVisibility.GUILD);
 
@@ -75,9 +78,11 @@ public final class RemindCommand extends SlashCommandAdapter {
         TIME_UNITS.forEach(unit -> timeUnit.addChoice(unit, unit));
 
         getData().addSubcommands(
-                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount, timeUnit, new OptionData(OptionType.STRING, CONTENT_OPTION, "what to remind you about", true)),
-                new SubcommandData(LIST_SUBCOMMAND, "lists all reminders")
-        );
+                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount,
+                        timeUnit,
+                        new OptionData(OptionType.STRING, CONTENT_OPTION,
+                                "what to remind you about", true)),
+                new SubcommandData(LIST_SUBCOMMAND, "lists all reminders"));
 
         this.database = database;
     }
@@ -118,15 +123,25 @@ public final class RemindCommand extends SlashCommandAdapter {
     }
 
     private void handleListCommand(SlashCommandInteractionEvent event) {
-        Result<PendingRemindersRecord> pendingReminders = database.read(context -> {
-            try (SelectWhereStep<PendingRemindersRecord> select = context.selectFrom(PENDING_REMINDERS)) {
-                return select.where(PENDING_REMINDERS.AUTHOR_ID.eq(event.getUser().getIdLong())).fetch();
+        BiFunction<Long, Instant, String> getDescription = (channelId, remindAt) -> """
+                Channel: <#%s>
+                Remind at: <t:%s>""".formatted(channelId, remindAt.getEpochSecond());
+
+        EmbedBuilder remindersEmbed = new EmbedBuilder().setTitle("Reminders").setColor(RemindRoutine.AMBIENT_COLOR);
+
+        database.read(context -> {
+            try (SelectWhereStep<PendingRemindersRecord> select =
+                    context.selectFrom(PENDING_REMINDERS)) {
+                return select
+                    .where(PENDING_REMINDERS.GUILD_ID.eq(event.getGuild().getIdLong())
+                        .and(PENDING_REMINDERS.AUTHOR_ID.eq(event.getUser().getIdLong())))
+                    .fetch();
             }
+        }).forEach(reminder -> {
+            remindersEmbed.addField(reminder.getContent(), getDescription.apply(reminder.getChannelId(), reminder.getRemindAt()), true);
         });
 
-        for (PendingRemindersRecord pendingReminder : pendingReminders) {
-            
-        }
+        event.replyEmbeds(remindersEmbed.build()).setEphemeral(true).queue();
     }
 
     private static Instant parseWhen(int whenAmount, String whenUnit) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
@@ -17,8 +17,8 @@ import org.togetherjava.tjbot.commands.moderation.scam.ScamBlocker;
 import org.togetherjava.tjbot.commands.moderation.scam.ScamHistoryPurgeRoutine;
 import org.togetherjava.tjbot.commands.moderation.scam.ScamHistoryStore;
 import org.togetherjava.tjbot.commands.moderation.temp.TemporaryModerationRoutine;
-import org.togetherjava.tjbot.commands.reminder.RemindCommand;
 import org.togetherjava.tjbot.commands.reminder.RemindRoutine;
+import org.togetherjava.tjbot.commands.reminder.ReminderCommand;
 import org.togetherjava.tjbot.commands.system.BotCore;
 import org.togetherjava.tjbot.commands.system.LogLevelCommand;
 import org.togetherjava.tjbot.commands.tags.TagCommand;
@@ -120,7 +120,7 @@ public class Features {
         features.add(new TopHelpersCommand(database));
         features.add(new RoleSelectCommand());
         features.add(new NoteCommand(actionsStore));
-        features.add(new RemindCommand(database));
+        features.add(new ReminderCommand(database));
         features.add(new QuarantineCommand(actionsStore, config));
         features.add(new UnquarantineCommand(actionsStore, config));
         features.add(new WhoIsCommand());
--- a/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
@@ -9,9 +9,12 @@ import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.commands.build.OptionData;
 
 import net.dv8tion.jda.api.interactions.commands.build.SubcommandData;
+import org.jooq.Result;
+import org.jooq.SelectWhereStep;
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.records.PendingRemindersRecord;
 
 import java.time.*;
 import java.time.temporal.TemporalAmount;
@@ -72,7 +75,7 @@ public final class RemindCommand extends SlashCommandAdapter {
         TIME_UNITS.forEach(unit -> timeUnit.addChoice(unit, unit));
 
         getData().addSubcommands(
-                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount, timeUnit),
+                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount, timeUnit, new OptionData(OptionType.STRING, CONTENT_OPTION, "what to remind you about", true)),
                 new SubcommandData(LIST_SUBCOMMAND, "lists all reminders")
         );
 
@@ -115,7 +118,15 @@ public final class RemindCommand extends SlashCommandAdapter {
     }
 
     private void handleListCommand(SlashCommandInteractionEvent event) {
+        Result<PendingRemindersRecord> pendingReminders = database.read(context -> {
+            try (SelectWhereStep<PendingRemindersRecord> select = context.selectFrom(PENDING_REMINDERS)) {
+                return select.where(PENDING_REMINDERS.AUTHOR_ID.eq(event.getUser().getIdLong())).fetch();
+            }
+        });
 
+        for (PendingRemindersRecord pendingReminder : pendingReminders) {
+            
+        }
     }
 
     private static Instant parseWhen(int whenAmount, String whenUnit) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindCommand.java
@@ -8,6 +8,7 @@ import net.dv8tion.jda.api.interactions.callbacks.IReplyCallback;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
 import net.dv8tion.jda.api.interactions.commands.build.OptionData;
 
+import net.dv8tion.jda.api.interactions.commands.build.SubcommandData;
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
 import org.togetherjava.tjbot.db.Database;
@@ -15,6 +16,8 @@ import org.togetherjava.tjbot.db.Database;
 import java.time.*;
 import java.time.temporal.TemporalAmount;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
 
 import static org.togetherjava.tjbot.db.generated.Tables.PENDING_REMINDERS;
 
@@ -33,7 +36,9 @@ import static org.togetherjava.tjbot.db.generated.Tables.PENDING_REMINDERS;
  * Pending reminders are processed and send by {@link RemindRoutine}.
  */
 public final class RemindCommand extends SlashCommandAdapter {
-    private static final String COMMAND_NAME = "remind";
+    private static final String COMMAND_NAME = "reminder";
+    private static final String LIST_SUBCOMMAND = "list";
+    static final String ADD_SUBCOMMAND = "add";
     static final String TIME_AMOUNT_OPTION = "time-amount";
     static final String TIME_UNIT_OPTION = "time-unit";
     static final String CONTENT_OPTION = "content";
@@ -46,6 +51,7 @@ public final class RemindCommand extends SlashCommandAdapter {
     static final int MAX_PENDING_REMINDERS_PER_USER = 100;
 
     private final Database database;
+    private final Map<String, Consumer<SlashCommandInteractionEvent>> subcommandHandler = Map.of(ADD_SUBCOMMAND, this::handleAddCommand, LIST_SUBCOMMAND, this::handleListCommand);
 
     /**
      * Creates an instance of the command.
@@ -65,14 +71,20 @@ public final class RemindCommand extends SlashCommandAdapter {
                 "period to remind you in, the unit of time (e.g. 5 [weeks])", true);
         TIME_UNITS.forEach(unit -> timeUnit.addChoice(unit, unit));
 
-        getData().addOptions(timeUnit, timeAmount)
-            .addOption(OptionType.STRING, CONTENT_OPTION, "what to remind you about", true);
+        getData().addSubcommands(
+                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount, timeUnit),
+                new SubcommandData(LIST_SUBCOMMAND, "lists all reminders")
+        );
 
         this.database = database;
     }
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
+        subcommandHandler.get(event.getSubcommandName()).accept(event);
+    }
+
+    private void handleAddCommand(SlashCommandInteractionEvent event) {
         int timeAmount = Math.toIntExact(event.getOption(TIME_AMOUNT_OPTION).getAsLong());
         String timeUnit = event.getOption(TIME_UNIT_OPTION).getAsString();
         String content = event.getOption(CONTENT_OPTION).getAsString();
@@ -100,6 +112,10 @@ public final class RemindCommand extends SlashCommandAdapter {
             .setRemindAt(remindAt)
             .setContent(content)
             .insert());
+    }
+
+    private void handleListCommand(SlashCommandInteractionEvent event) {
+
     }
 
     private static Instant parseWhen(int whenAmount, String whenUnit) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/reminder/RemindRoutine.java
@@ -28,11 +28,11 @@ import static org.togetherjava.tjbot.db.generated.Tables.PENDING_REMINDERS;
 /**
  * Routine that processes and sends pending reminders.
  * <p>
- * Reminders can be set by using {@link RemindCommand}.
+ * Reminders can be set by using {@link ReminderCommand}.
  */
 public final class RemindRoutine implements Routine {
     static final Logger logger = LoggerFactory.getLogger(RemindRoutine.class);
-    private static final Color AMBIENT_COLOR = Color.decode("#F7F492");
+    static final Color AMBIENT_COLOR = Color.decode("#F7F492");
     private static final int SCHEDULE_INTERVAL_SECONDS = 30;
     private final Database database;
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/reminder/package-info.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/reminder/package-info.java
@@ -1,6 +1,6 @@
 /**
  * This packages offers all the functionality for the remind-command. The core class is
- * {@link org.togetherjava.tjbot.commands.reminder.RemindCommand}.
+ * {@link org.togetherjava.tjbot.commands.reminder.ReminderCommand}.
  */
 @MethodsReturnNonnullByDefault
 @ParametersAreNonnullByDefault
--- a/application/src/test/java/org/togetherjava/tjbot/commands/reminder/RemindCommandTest.java
+++ b/application/src/test/java/org/togetherjava/tjbot/commands/reminder/RemindCommandTest.java
@@ -43,6 +43,7 @@ final class RemindCommandTest {
     private SlashCommandInteractionEvent triggerSlashCommand(int timeAmount, String timeUnit,
             String content, Member author) {
         SlashCommandInteractionEvent event = jdaTester.createSlashCommandInteractionEvent(command)
+                .setSubcommand(RemindCommand.ADD_SUBCOMMAND)
             .setOption(RemindCommand.TIME_AMOUNT_OPTION, timeAmount)
             .setOption(RemindCommand.TIME_UNIT_OPTION, timeUnit)
             .setOption(RemindCommand.CONTENT_OPTION, content)
--- a/application/src/test/java/org/togetherjava/tjbot/commands/reminder/ReminderCommandTest.java
+++ b/application/src/test/java/org/togetherjava/tjbot/commands/reminder/ReminderCommandTest.java
@@ -22,7 +22,7 @@ import static org.mockito.ArgumentMatchers.startsWith;
 import static org.mockito.Mockito.verify;
 import static org.togetherjava.tjbot.db.generated.tables.PendingReminders.PENDING_REMINDERS;
 
-final class RemindCommandTest {
+final class ReminderCommandTest {
     private SlashCommand command;
     private JdaTester jdaTester;
     private RawReminderTestHelper rawReminders;
@@ -30,7 +30,7 @@ final class RemindCommandTest {
     @BeforeEach
     void setUp() {
         Database database = Database.createMemoryDatabase(PENDING_REMINDERS);
-        command = new RemindCommand(database);
+        command = new ReminderCommand(database);
         jdaTester = new JdaTester();
         rawReminders = new RawReminderTestHelper(database, jdaTester);
     }
@@ -43,10 +43,10 @@ final class RemindCommandTest {
     private SlashCommandInteractionEvent triggerSlashCommand(int timeAmount, String timeUnit,
             String content, Member author) {
         SlashCommandInteractionEvent event = jdaTester.createSlashCommandInteractionEvent(command)
-                .setSubcommand(RemindCommand.ADD_SUBCOMMAND)
-            .setOption(RemindCommand.TIME_AMOUNT_OPTION, timeAmount)
-            .setOption(RemindCommand.TIME_UNIT_OPTION, timeUnit)
-            .setOption(RemindCommand.CONTENT_OPTION, content)
+            .setSubcommand(ReminderCommand.ADD_SUBCOMMAND)
+            .setOption(ReminderCommand.TIME_AMOUNT_OPTION, timeAmount)
+            .setOption(ReminderCommand.TIME_UNIT_OPTION, timeUnit)
+            .setOption(ReminderCommand.CONTENT_OPTION, content)
             .setUserWhoTriggered(author)
             .build();
 
@@ -83,7 +83,7 @@ final class RemindCommandTest {
     void userIsLimitedIfTooManyPendingReminders() {
         // GIVEN a user with too many reminders still pending
         Instant remindAt = Instant.now().plus(100, ChronoUnit.DAYS);
-        for (int i = 0; i < RemindCommand.MAX_PENDING_REMINDERS_PER_USER; i++) {
+        for (int i = 0; i < ReminderCommand.MAX_PENDING_REMINDERS_PER_USER; i++) {
             rawReminders.insertReminder("foo " + i, remindAt);
         }
 
@@ -93,7 +93,7 @@ final class RemindCommandTest {
         // THEN rejects and responds accordingly, no new reminder was created
         verify(event)
             .reply(startsWith("You have reached the maximum amount of pending reminders per user"));
-        assertEquals(RemindCommand.MAX_PENDING_REMINDERS_PER_USER,
+        assertEquals(ReminderCommand.MAX_PENDING_REMINDERS_PER_USER,
                 rawReminders.readReminders().size());
     }
 
@@ -104,7 +104,7 @@ final class RemindCommandTest {
         // and a second user with no reminders yet
         Member firstUser = jdaTester.createMemberSpy(1);
         Instant remindAt = Instant.now().plus(100, ChronoUnit.DAYS);
-        for (int i = 0; i < RemindCommand.MAX_PENDING_REMINDERS_PER_USER; i++) {
+        for (int i = 0; i < ReminderCommand.MAX_PENDING_REMINDERS_PER_USER; i++) {
             rawReminders.insertReminder("foo " + i, remindAt, firstUser);
         }
 


application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -32,8 +39,10 @@
  * <p>
  * Pending reminders are processed and send by {@link RemindRoutine}.
  */
-public final class RemindCommand extends SlashCommandAdapter {
-    private static final String COMMAND_NAME = "remind";
+public final class ReminderCommand extends SlashCommandAdapter {
+    private static final String COMMAND_NAME = "reminder";
+    private static final String LIST_SUBCOMMAND = "list";
+    static final String ADD_SUBCOMMAND = "add";
```
Zabuzard, line 45: i think `create` would be better

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -65,14 +76,22 @@ public RemindCommand(Database database) {
...
         TIME_UNITS.forEach(unit -> timeUnit.addChoice(unit, unit));
 
-        getData().addOptions(timeUnit, timeAmount)
-            .addOption(OptionType.STRING, CONTENT_OPTION, "what to remind you about", true);
+        getData().addSubcommands(
+                new SubcommandData(ADD_SUBCOMMAND, "adds a reminder").addOptions(timeAmount,
+                        timeUnit,
+                        new OptionData(OptionType.STRING, CONTENT_OPTION,
+                                "what to remind you about", true)),
+                new SubcommandData(LIST_SUBCOMMAND, "lists all reminders"));
```
Zabuzard, line 84: "shows all your currently pending reminders"

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +121,29 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
             .insert());
     }
 
+    private void handleListCommand(SlashCommandInteractionEvent event) {
+        BiFunction<Long, Instant, String> getDescription = (channelId, remindAt) -> """
+                Channel: <#%s>
+                Remind at: <t:%s>""".formatted(channelId, remindAt.getEpochSecond());
```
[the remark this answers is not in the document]
Zabuzard: for the time stuff, JDA already has utility, please reuse:
```java
String text = TimeFormat.DATE_TIME_SHORT.format(Instant.now());
```

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +121,29 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+        EmbedBuilder remindersEmbed =
+                new EmbedBuilder().setTitle("Reminders").setColor(RemindRoutine.AMBIENT_COLOR);
+
+        database.read(context -> {
+            try (SelectWhereStep<PendingRemindersRecord> select =
+                    context.selectFrom(PENDING_REMINDERS)) {
+                return select
+                    .where(PENDING_REMINDERS.GUILD_ID.eq(event.getGuild().getIdLong())
+                        .and(PENDING_REMINDERS.AUTHOR_ID.eq(event.getUser().getIdLong())))
+                    .fetch();
```
Zabuzard, line 138: u might want to sort them ascending by date. so that it displays the soonest reminders first

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03:
```java
.sortAsc(PENDING_REMINDERS.REMIND_AT)
```
like this??

Nxllpointer: I believe it is `.orderBy(PENDING_REMIMDERS.REMIND_AT.asc())`

Nxllpointer: If i got it right sortAsc applys after getting the result and orderBy affects the sql query. If i had to choose i would use orderBy because the database is probably faster at ordering before fetching than you doing it yourself after fetching.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -65,14 +76,22 @@ public RemindCommand(Database database) {
...
         TIME_UNITS.forEach(unit -> timeUnit.addChoice(unit, unit));
 
-        getData().addOptions(timeUnit, timeAmount)
-            .addOption(OptionType.STRING, CONTENT_OPTION, "what to remind you about", true);
+        getData().addSubcommands(
+                new SubcommandData(CREATE_SUBCOMMAND, "creates a reminder").addOptions(timeAmount,
+                        timeUnit,
+                        new OptionData(OptionType.STRING, CONTENT_OPTION,
+                                "what to remind you about", true)),
+                new SubcommandData(LIST_SUBCOMMAND, "shows all your currently pending reminders"));
```
Nxllpointer, line 84: Maybe put the subcommands and options into seperate variables first. This looks really messy.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -65,14 +76,22 @@ public RemindCommand(Database database) {
...
 
         this.database = database;
     }
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
+        subcommandHandler.get(event.getSubcommandName()).accept(event);
+    }
+
+    private void handleAddCommand(SlashCommandInteractionEvent event) {
```
Nxllpointer, line 94: Is having a map here really neccesairy? In my opinion it would be way more readable with a simple switch. You dont even use it anywhere else.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +121,31 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+                Remind at: %s""".formatted(MessageUtils.mentionChannelById(channelId),
+                TimeFormat.DEFAULT.format(remindAt));
+
+        EmbedBuilder remindersEmbed = new EmbedBuilder().setTitle("Pending reminders")
+            .setColor(RemindRoutine.AMBIENT_COLOR);
+
+        database
+            .read(context -> context.selectFrom(PENDING_REMINDERS)
+                .where(PENDING_REMINDERS.GUILD_ID.eq(event.getGuild().getIdLong())
+                    .and(PENDING_REMINDERS.AUTHOR_ID.eq(event.getUser().getIdLong())))
```
Nxllpointer, line 136: Using `.and` makes it unneccesairily unreadable. You can just pass all of the requirements to `.where`
Example `.where(requirement1, requirement2)` :+1:

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +121,31 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+                .orderBy(PENDING_REMINDERS.CREATED_AT.asc())
+                .fetch())
+            .forEach(reminder -> remindersEmbed.addField(reminder.getContent(),
+                    getDescription.apply(reminder.getChannelId(), reminder.getRemindAt()), false));
```
Nxllpointer, line 140: Here again i would be using multiple variables to store the field values. You dont have to put everything in one line. Use some curly brackets and add it in a readable way please.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -109,4 +109,13 @@ public static String abbreviate(String text, int maxLength) {
...
     }
 
+    /**
+     * Mentions a guild channel by its id
+     *
+     * @param channelId the channel id to mention
+     * @return channel mention
+     */
+    public static String mentionChannelById(long channelId) {
+        return "<#%s>".formatted(channelId);
```
Nxllpointer, line 119: %d is used for numbers

Nxllpointer (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: i believe .and makes it more readable and understandable

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: i mean %s can take anything, but yea will do that

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: with a map i can decrease the number of places one will need to change when introducing a new subcommand

Nxllpointer: No. Its just one place anyways

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: But then at least thore the guild and user id in a variable to make the lines shorter

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -32,8 +39,10 @@
  * <p>
  * Pending reminders are processed and send by {@link RemindRoutine}.
  */
-public final class RemindCommand extends SlashCommandAdapter {
-    private static final String COMMAND_NAME = "remind";
+public final class ReminderCommand extends SlashCommandAdapter {
```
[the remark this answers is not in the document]
Zabuzard: @Taz03 can u please also add some lines about the feature you just added?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -123,24 +124,29 @@ private void handleAddCommand(SlashCommandInteractionEvent event) {
...
+
+        long guildId = event.getGuild().getIdLong();
+        long userId = event.getUser().getIdLong();
+
+        database
+            .read(context -> context.selectFrom(PENDING_REMINDERS)
+                .where(PENDING_REMINDERS.GUILD_ID.eq(guildId)
+                    .and(PENDING_REMINDERS.AUTHOR_ID.eq(userId)))
+                .orderBy(PENDING_REMINDERS.CREATED_AT.asc())
+                .fetch())
```
Zabuzard, line 142: id move this into a help method to increase readability of this method and reduce its scope

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Zabuzard: now it sorts ascending by creating date. i meant more like ascending on when its fired. so that u can see them in the order they will be fired. @Taz03

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -109,4 +109,13 @@ public static String abbreviate(String text, int maxLength) {
...
 
+    /**
+     * Mentions a guild channel by its id
+     *
+     * @param channelId the channel id to mention
+     * @return channel mention
+     */
+    public static String mentionChannelById(long channelId) {
+        return "<#%d>".formatted(channelId);
+    }
```
Zabuzard, line 121: same as i said on ur other PR

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +122,34 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+            .forEach(reminder -> remindersEmbed.addField(reminder.getContent(),
+                    getDescription.apply(reminder.getChannelId(), reminder.getRemindAt()), false));
+
+        if (remindersEmbed.getFields().isEmpty()) {
```
Zabuzard, line 146: this condition feels a bit odd. i would instead have expected that you check on the size of the returned DB list. like `if (pendingReminders.isEmpty())`, i think that would make it more readable.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -102,6 +122,34 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+        long userId = event.getUser().getIdLong();
+
+        database
+            .read(context -> context.selectFrom(PENDING_REMINDERS)
+                .where(PENDING_REMINDERS.GUILD_ID.eq(guildId)
+                    .and(PENDING_REMINDERS.AUTHOR_ID.eq(userId)))
+                .orderBy(PENDING_REMINDERS.CREATED_AT.asc())
+                .fetch())
+            .forEach(reminder -> remindersEmbed.addField(reminder.getContent(),
+                    getDescription.apply(reminder.getChannelId(), reminder.getRemindAt()), false));
```
Zabuzard, line 144: what happens if the reminder text is quite long? people can create really long reminders and this field-title for sure has some smaller limit, no? might have to use `MessageUtils.abbreviate`.

and if the limit is too small, using the title might not even be an option anymore. its important that people can read their full reminder here (in most cases).

that said, lets say one creates reminders of the size:
```java
i have to do foo and then bar, please tell me foo and bar. its important that i do foo and bar.
```
how many of those can be created until the embed runs out of its max size and we get exceptions? we might have to add pagination. or at least make sure that the UX is good if this happens, i.e. we have to limit the shown reminders based on total length or similar.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: the field title can hold 256 chars max. so in the worst case a field can have around 300 chars and a embed can have 4k smth chars at max so a embed should be able to hold 13-14 of very big reminders. ig the solution is to setup pagination with 10 reminders on each page

also this is what it'll look like with big reminders each is 250 chars
image
ui is not a problem will need to use the abbreviate method on longer reminders

Nxllpointer: Enjoy adding pagination! Its annoying :D

Nxllpointer: you can take a look at my PR to see how i implemented it

Taz03: > Enjoy adding pagination! Its annoying :D

lol yea, but i've did that before on this repo shouldn't be too hard

Nxllpointer: Hm im interested how you did it

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: image
is this good enough?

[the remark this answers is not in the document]
Zabuzard: yes

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Zabuzard: we can think about adding some helpers for pagination, if u had trouble with it. seems to be sth that we will frequently need.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -110,12 +110,13 @@ public static String abbreviate(String text, int maxLength) {
...
     /**
-     * Mentions a guild channel by its id Clone of JDAs Channel#getAsMention, but unfortunately
-     * channel instances can not be created out of just an ID, unlike User#fromId
+     * Mentions a guild channel by its id.
      *
      * @param channelId the ID of the channel to mention
      * @return the channel as formatted string which Discord interprets as clickable mention
      */
+    // Clone of JDAs Channel#getAsMention, but unfortunately channel instances can not be created
+    // out of just an ID, unlike User#fromId
```
Zabuzard, line 119: goes inside method

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -134,22 +136,31 @@ private void handleListCommand(SlashCommandInteractionEvent event) {
...
+                Instant remindAt = reminder.getRemindAt();
+
+                remindersEmbed.addField(content, getDescription.apply(channelId, remindAt), false);
+            });
         }
 
         event.replyEmbeds(remindersEmbed.build()).setEphemeral(true).queue();
     }
 
+    private Result<PendingRemindersRecord> getReminders(long guildId, long userId) {
```
Zabuzard, line 156: getPendingReminders

also, id change it to accept Guild, User. with two longs, its easier for people to mess up and accidentally give them in the wrong order.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: > we can think about adding some helpers for pagination, if u had trouble with it. seems to be sth that we will frequently need.

Cool idea. Yes it would really make sense and make individual features a lot lighter. I may do that after the bookmarks feature gets merged.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            Result<PendingRemindersRecord> remindersRecords, int pageNumber) {
+        List<PendingRemindersRecord> reminders = new ArrayList<>(MAX_PAGE_LENGTH);
+
+        int start = (pageNumber - 1) * MAX_PAGE_LENGTH;
+        int end = Math.min(start + MAX_PAGE_LENGTH, remindersRecords.size());
+
+        for (int i = start; i < end; i++) {
+            reminders.add(remindersRecords.get(i));
+        }
+
```
Nxllpointer, line 230: You can use `remindersRecords.subList(start, end)`

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+        return reminders;
+    }
+
+    private static void addReminderAsField(PendingRemindersRecord reminder, EmbedBuilder embed) {
+        BiFunction<Long, Instant, String> getDescription = (channelId, remindAt) -> """
+                Channel: %s
+                Remind at: %s""".formatted(MessageUtils.mentionChannelById(channelId),
+                TimeFormat.DEFAULT.format(remindAt));
```
Nxllpointer, line 238: Since this does not access any local variables you can move this to its own method.

Nxllpointer, line 238: Getting back to the topic of timezones, you dont need to format time yourself. Discord added timestamp formatting and the time gets adjusted for the users timezone. How cool! Here is some info on it: https://gist.github.com/LeviSnoot/d9147767abeef2f770e9ddcd91eb85aa

Nxllpointer (changes requested):
Good job so far! Im confident you can make this code really amazing!

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+    static final String TIME_AMOUNT_OPTION = "time-amount";
+    static final String TIME_UNIT_OPTION = "time-unit";
+    static final String CONTENT_OPTION = "content";
+
+    private static final int MIN_TIME_AMOUNT = 1;
+    private static final int MAX_TIME_AMOUNT = 1_000;
+    private static final List<String> TIME_UNITS =
+            List.of("minutes", "hours", "days", "weeks", "months", "years");
+    private static final Period MAX_TIME_PERIOD = Period.ofYears(3);
+    private static final int MAX_PAGE_LENGTH = 10;
```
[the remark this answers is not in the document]
Taz03: yes, the reminders on each page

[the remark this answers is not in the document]
Nxllpointer: Then give it a self explaining name

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+    private static final String LIST_SUBCOMMAND = "list";
+    static final String CREATE_SUBCOMMAND = "create";
+    static final String TIME_AMOUNT_OPTION = "time-amount";
+    static final String TIME_UNIT_OPTION = "time-unit";
+    static final String CONTENT_OPTION = "content";
+
+    private static final int MIN_TIME_AMOUNT = 1;
+    private static final int MAX_TIME_AMOUNT = 1_000;
+    private static final List<String> TIME_UNITS =
+            List.of("minutes", "hours", "days", "weeks", "months", "years");
```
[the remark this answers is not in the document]
Taz03: this is not my code, also this is same as so many other commands, if u want to change make an issue and then change in all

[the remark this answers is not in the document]
Nxllpointer: Here it does make a lot of sense though as you will notice in another conversation.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+        return database.read(context -> context.selectFrom(PENDING_REMINDERS)
+            .where(PENDING_REMINDERS.GUILD_ID.eq(guild.getIdLong())
+                .and(PENDING_REMINDERS.AUTHOR_ID.eq(user.getIdLong())))
+            .orderBy(PENDING_REMINDERS.REMIND_AT.asc())
+            .fetch());
+    }
+
+    private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendingReminders,
+            int pageToShow) {
+        int totalPages = (int) Math.ceil(pendingReminders.size() / (double) MAX_PAGE_LENGTH);
```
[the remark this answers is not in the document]
Taz03: never knew that exists

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+
+        EmbedBuilder remindersEmbed = new EmbedBuilder().setTitle("Pending reminders")
+            .setColor(RemindRoutine.AMBIENT_COLOR);
+        MessageCreateBuilder listMessage = new MessageCreateBuilder();
+
+        if (pendingReminders.isEmpty()) {
+            remindersEmbed.setDescription("No pending reminders");
+        } else {
+            if (totalPages > 1) {
+                remindersEmbed.setFooter("Page: " + pageToShow + "/" + totalPages);
```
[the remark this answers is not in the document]
Taz03: > This looks confusing

i disagree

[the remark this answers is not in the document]
Nxllpointer: But `Page %d/%d` is easier to read

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+
+        return List.of(previousButton, nextButton);
+    }
+
+    private Button createPageTurnButton(String label, long pageNumber, int turnPageBy) {
+        return Button.primary(
+                generateComponentId(String.valueOf(pageNumber), String.valueOf(turnPageBy)), label);
+    }
+
+    private static List<PendingRemindersRecord> getRemindersPage(
```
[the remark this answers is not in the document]
Taz03: yea makes sense

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            case LIST_SUBCOMMAND -> handleListCommand(event);
+            default -> throw new AssertionError(
+                    "Unexpected Subcommand: " + event.getSubcommandName());
+        }
+    }
+
+    @Override
+    public void onButtonClick(ButtonInteractionEvent event, List<String> args) {
+        int pageNumber = Integer.parseInt(args.get(0));
+        int pageTurnBy = Integer.parseInt(args.get(1));
```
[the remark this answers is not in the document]
Taz03: that'll need more processing, rn i just parse it and add it to the pageNumber no need to do extra steps

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+        int pageToShow = pageNumber + pageTurnBy;
+
+        event
+            .editMessage(
+                    MessageEditData.fromCreateData(createListMessage(pendingReminders, pageToShow)))
+            .queue();
+    }
+
+    private void handleCreateCommand(SlashCommandInteractionEvent event) {
+        int timeAmount = Math.toIntExact(event.getOption(TIME_AMOUNT_OPTION).getAsLong());
```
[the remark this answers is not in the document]
Taz03: this method is not mine i just copy pasted it, i don't know what was the thought process when doing this, check git blame of actual code and ask why this was decided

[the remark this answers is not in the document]
Nxllpointer: Would you still change it please?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            case "month", "months" -> Period.ofMonths(whenAmount);
+            case "year", "years" -> Period.ofYears(whenAmount);
+            default -> throw new IllegalArgumentException("Unsupported unit, was: " + whenUnit);
+        };
+
+        return ZonedDateTime.now(ZoneOffset.UTC).plus(period).toInstant();
+    }
+
+    private static boolean handleIsRemindAtWithinLimits(Instant remindAt, IReplyCallback event) {
+        ZonedDateTime maxWhen = ZonedDateTime.now(ZoneOffset.UTC).plus(MAX_TIME_PERIOD);
```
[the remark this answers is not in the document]
Taz03: not my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+
+        if (remindAt.atZone(ZoneOffset.UTC).isBefore(maxWhen)) {
```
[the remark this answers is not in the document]
Taz03: again

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            case "minute", "minutes" -> Duration.ofMinutes(whenAmount);
+            case "hour", "hours" -> Duration.ofHours(whenAmount);
+            case "day", "days" -> Period.ofDays(whenAmount);
+            case "week", "weeks" -> Period.ofWeeks(whenAmount);
+            case "month", "months" -> Period.ofMonths(whenAmount);
+            case "year", "years" -> Period.ofYears(whenAmount);
+            default -> throw new IllegalArgumentException("Unsupported unit, was: " + whenUnit);
+        };
+
+        return ZonedDateTime.now(ZoneOffset.UTC).plus(period).toInstant();
```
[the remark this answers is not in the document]
Taz03: not my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Nxllpointer: Well the new git blame will show you as the developer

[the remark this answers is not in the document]
Nxllpointer: Or do you think changing the old code should be done in a seperate PR?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: i tried looking for a method which does that but i gave up too quickly :joy:

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: that is exactly what this method does, first i did that manually but zabu said theres a method for that in jda

Taz03: i can but i won't, i dont like having a method for every little thing

i can even just make it a normal string and format it in the stream thing

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: i mean this code was approved so y would i change it? if it is approved it has some purpose

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: But what you are doing here is basically creating a method in a method. So why dont you just separate them.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: yes do it in a separate pr after this one

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: Oh ok. I thought TimeFormat was a standard java class

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: what would be proper name according to u?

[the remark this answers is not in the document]
Nxllpointer: `ENTRIES_PER_PAGE` or something like that

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: yea i agree, changing it to a string

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Nxllpointer: Then lets save this for the second cleanup PR

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: coz this is a proper use of functional interface

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: what about `REMINDERS_PER_PAGE`?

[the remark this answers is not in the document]
Nxllpointer: Thats perfect

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            .queue();
+
+        database.write(context -> context.newRecord(PENDING_REMINDERS)
+            .setCreatedAt(Instant.now())
+            .setGuildId(guild.getIdLong())
+            .setChannelId(event.getChannel().getIdLong())
+            .setAuthorId(author.getIdLong())
+            .setRemindAt(remindAt)
+            .setContent(content)
+            .insert());
```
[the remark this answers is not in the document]
Taz03: not my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+            .queue();
+
+        return false;
+    }
+
+    private boolean handleIsUserBelowMaxPendingReminders(ISnowflake author, ISnowflake guild,
+            IReplyCallback event) {
+        int pendingReminders = database.read(context -> context.fetchCount(PENDING_REMINDERS,
+                PENDING_REMINDERS.AUTHOR_ID.equal(author.getIdLong())
+                    .and(PENDING_REMINDERS.GUILD_ID.equal(guild.getIdLong()))));
```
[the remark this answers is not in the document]
Taz03: not my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+        TemporalAmount period = switch (whenUnit) {
+            case "second", "seconds" -> Duration.ofSeconds(whenAmount);
+            case "minute", "minutes" -> Duration.ofMinutes(whenAmount);
+            case "hour", "hours" -> Duration.ofHours(whenAmount);
+            case "day", "days" -> Period.ofDays(whenAmount);
+            case "week", "weeks" -> Period.ofWeeks(whenAmount);
+            case "month", "months" -> Period.ofMonths(whenAmount);
+            case "year", "years" -> Period.ofYears(whenAmount);
+            default -> throw new IllegalArgumentException("Unsupported unit, was: " + whenUnit);
+        };
```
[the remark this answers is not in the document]
Taz03: not my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    private void handleListCommand(SlashCommandInteractionEvent event) {
+        Result<PendingRemindersRecord> pendingReminders =
+                getPendingReminders(event.getGuild(), event.getUser());
+
+        event.reply(createListMessage(pendingReminders, 1)).setEphemeral(true).queue();
+    }
+
+    private Result<PendingRemindersRecord> getPendingReminders(Guild guild, User user) {
```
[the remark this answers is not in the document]
Taz03: others aren't my code

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: no, but will do that anyways

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: I would go with the String approach. It is only used a single time anyways

Nxllpointer:
@Taz03 I will approve if we have both settled on a solution for the last open conversation

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: its not like i just put the channelId and remindAt,, they are not even strings i use some methods against them and then put them in string.

just let it be how it is

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: @Nxllpointer also if i just need to know which button it is i dont need to pass an argument

i can do that like this
```java
switch (event.getButton().getLabel()) {
            case NEXT_BUTTON_LABEL -> pageToShow += 1;
            case PREVIOUS_BUTTON_LABEL -> pageToShow -= 1;
        }
```

but i pass those argument to not do further processing i can just parse it and use it

r u convinced now?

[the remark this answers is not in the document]
Nxllpointer: The label idea is great. How did I not think of that. I get there you just want to add turnBy but as I said above it requires some time to understand. One should be able to understand code instantly without any comments.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: Look here it does exactly the same, except you dont need your BiFunction. If your method used this BiFunction multiple times and used some things that are the same for all descriptions, then it would totally make sense. This does not however.

```java
private static void addReminderAsField(PendingRemindersRecord reminder, EmbedBuilder embed) {
        String content = reminder.getContent();
        long channelId = reminder.getChannelId();
        Instant remindAt = reminder.getRemindAt();

        String description = """
                Channel: %s
                Remind at: %s""".formatted(MessageUtils.mentionChannelById(channelId),
                TimeFormat.DEFAULT.format(remindAt));

        embed.addField(MessageUtils.abbreviate(content, MAX_REMINDER_TITLE_LENGTH), description,
                false);
    }
```

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: @Zabuzard what do u think about this??

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: yes good idea

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -108,12 +108,18 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
-        int pageNumber = Integer.parseInt(args.get(0));
-        int pageTurnBy = Integer.parseInt(args.get(1));
+        int pageToShow = Integer.parseInt(args.get(0));
+
+        switch (event.getButton().getLabel()) {
+            case PREVIOUS_BUTTON_LABEL -> pageToShow--;
+            case NEXT_BUTTON_LABEL -> pageToShow++;
+            default -> {
+                // do nothing
+            }
```
Nxllpointer, line 118: This should never happen so you can just throw an exception here

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -193,28 +199,21 @@ private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendi
...
+        Button previousButton = Button.primary(generateComponentId(String.valueOf(pageNumber)),
+                PREVIOUS_BUTTON_LABEL);
         if (pageNumber <= 1) {
             previousButton = previousButton.asDisabled();
         }
 
-        int nextButtonTurnPageBy = 1;
         Button nextButton =
-                createPageTurnButton(NEXT_BUTTON_LABEL, pageNumber, nextButtonTurnPageBy);
+                Button.primary(generateComponentId(String.valueOf(pageNumber)), NEXT_BUTTON_LABEL);
```
Nxllpointer, line 209: No need to call String.valueOf twice. You should just store the String representation in a variable like `pageNumberString`. This would also make it shorter and your button creation would probably fit in one line.

Nxllpointer (changes requested):
Almost perfect! Also what about our discussion on the button labels?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: y it should never happen?? u can add more buttons in future like jump to last or jump to first

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: last time i did pagination (in audit command) zabu told me to do it this way, so gonna keep it like this

Taz03:
> Also what about our discussion on the button labels?

i mean its a problem with specific person system, so y should i bother about that
anyways did u checked how it looks on mobile?

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: I am pretty sure he did not tell you to duplicate code

Nxllpointer:
I did check it on my android phone. They got converted to the discord-like emojis but worse. We should just use the discord emojis so it looks the same for everyone

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: ohh yea he told me to make a separate method but no method now so ig i can make a variable

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: which exception should i throw tho?

Taz03:
> I did check it on my android phone. They got converted to the discord-like emojis but worse. We should just use the discord emojis so it looks the same for everyone

makes sense

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Nxllpointer: Im not an expert with exception types but i think an IllegalArgumentException will do the job. I think Zabuzard knows his way around exceptions pretty well though.

Taz03: i used if else now so resolving this

Nxllpointer (dismissed):
Everything should be fine now that you changed 👍. Todo: cleanup old code (see resolved conversations)

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+    }
 
-        Result<PendingRemindersRecord> pendingReminders = getReminders(guildId, userId);
+    private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendingReminders,
+            int pageToShow) {
+        int totalPages = Math.ceilDiv(pendingReminders.size(), REMINDERS_PER_PAGE);
+
+        EmbedBuilder remindersEmbed = new EmbedBuilder().setTitle("Pending reminders")
+            .setColor(RemindRoutine.AMBIENT_COLOR);
+        MessageCreateBuilder listMessage = new MessageCreateBuilder();
```
Zabuzard, line 182: (also `pendingRemindersPage` then)

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+                remindersEmbed.setFooter("Page: %d/%d".formatted(pageToShow, totalPages));
 
-                remindersEmbed.addField(content, getDescription.apply(channelId, remindAt), false);
-            });
+                getPageEntries(pendingReminders, pageToShow)
+                    .forEach(reminder -> addReminderAsField(reminder, remindersEmbed));
+
+                listMessage.addActionRow(createPageTurnButtons(pageToShow, totalPages));
+            } else {
+                pendingReminders.forEach(reminder -> addReminderAsField(reminder, remindersEmbed));
```
Zabuzard, line 195: this is duplicated in both branches. try to get rid of that

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+        return listMessage.addEmbeds(remindersEmbed.build()).build();
     }
 
-    private Result<PendingRemindersRecord> getReminders(long guildId, long userId) {
-        return database.read(context -> context.selectFrom(PENDING_REMINDERS)
-            .where(PENDING_REMINDERS.GUILD_ID.eq(guildId)
-                .and(PENDING_REMINDERS.AUTHOR_ID.eq(userId)))
-            .orderBy(PENDING_REMINDERS.REMIND_AT.asc())
-            .fetch());
+    private List<Button> createPageTurnButtons(int pageNumber, int totalPages) {
```
Zabuzard, line 202: who is `pageNumber`? the current page? the page to turn to? ... please clarify

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+                Button.primary(generateComponentId(pageNumberString), NEXT_BUTTON_EMOJI);
+        if (pageNumber >= totalPages) {
+            nextButton = nextButton.asDisabled();
+        }
+
+        return List.of(previousButton, nextButton);
+    }
+
+    private static List<PendingRemindersRecord> getPageEntries(
+            Result<PendingRemindersRecord> remindersRecords, int pageNumber) {
```
Zabuzard, line 221: maybe change to `paginateReminders`. and `pageNumber` should be changed to maybe `pageToDisplay` or `currentPage` or similar

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -52,6 +62,10 @@ public final class ReminderCommand extends SlashCommandAdapter {
     private static final List<String> TIME_UNITS =
             List.of("minutes", "hours", "days", "weeks", "months", "years");
     private static final Period MAX_TIME_PERIOD = Period.ofYears(3);
+    private static final int REMINDERS_PER_PAGE = 10;
+    private static final int MAX_REMINDER_TITLE_LENGTH = 256;
```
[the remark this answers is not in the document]
Taz03: its the limit set by discord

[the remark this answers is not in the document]
Taz03: https://discord.com/developers/docs/resources/channel#embed-object-embed-limits

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: currentPage, yes i also think it is a bit confusing

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: > maybe change to paginateReminders.

it doesn't paginate anything it just gives the reminders which should be shown on the given page

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -52,6 +62,10 @@ public final class ReminderCommand extends SlashCommandAdapter {
...
+    private static final Emoji PREVIOUS_BUTTON_EMOJI = Emoji.fromUnicode("⬅");
+    private static final Emoji NEXT_BUTTON_EMOJI = Emoji.fromUnicode("➡");
```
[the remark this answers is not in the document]
Taz03: but this is emoji y call it label

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+            .where(PENDING_REMINDERS.GUILD_ID.eq(guild.getIdLong())
+                .and(PENDING_REMINDERS.AUTHOR_ID.eq(user.getIdLong())))
+            .orderBy(PENDING_REMINDERS.REMIND_AT.asc())
+            .fetch());
+    }
 
-        Result<PendingRemindersRecord> pendingReminders = getReminders(guildId, userId);
+    private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendingReminders,
+            int pageToShow) {
+        int totalPages = Math.ceilDiv(pendingReminders.size(), REMINDERS_PER_PAGE);
```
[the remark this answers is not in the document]
Taz03: yea so say there are 12 pending reminders and we display 10 reminders per page

so by normal division the answer will be 1 which is wrong we need 2 pages in total

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+    private Result<PendingRemindersRecord> getPendingReminders(Guild guild, User user) {
+        return database.read(context -> context.selectFrom(PENDING_REMINDERS)
+            .where(PENDING_REMINDERS.GUILD_ID.eq(guild.getIdLong())
+                .and(PENDING_REMINDERS.AUTHOR_ID.eq(user.getIdLong())))
+            .orderBy(PENDING_REMINDERS.REMIND_AT.asc())
+            .fetch());
+    }
 
-        Result<PendingRemindersRecord> pendingReminders = getReminders(guildId, userId);
+    private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendingReminders,
```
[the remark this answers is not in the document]
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Zabuzard: yeah and `pageNumber` is ambigious. i didnt understand what it means when i read it. rename it so that its clear. maybe `pageToDisplay` or `pageToShowRemindersFor` or whatever.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Zabuzard: no as in "its not duplicated" or as in "i refuse to do that for no reasons"?

these two lines are identical and thats not good and can easily be get rid of:

duplication

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Zabuzard: so that people understand what its used for. its clear that its an emoji, the type already says that. thats boring. the important bit is "okay, what is this emoji used for?" its used as a label for the button.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Zabuzard: makes sense. add that as comment right above
```java
// 12 reminders, 10 per page, ceilDiv(12, 10) = 2
```
then nobody trips anymore

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Nxllpointer: I agree with Taz's naming. Buttons can have a label **and** an emoji. Calling this label would be wrong

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: ig this comment is too extra

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: > its not duplicated

both are getting called on different things

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Zabuzard: this comment is the prime example of a good comment. it ensures that readers dont trip ur confusing code. it helps readers to understand ur code.

it prevents the next person from reading ur code and asking themselves "why is this ceilDiv?" and then spending 5 minutes to figure it out or asking u.

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Zabuzard: so ur point is that u dont know how to get rid of the duplication? thats fair, let me see myself then

Zabuzard: there u go:
```java
} else {
    List<PendingRemindersRecord> pendingRemindersForPage = pendingReminders.stream().toList();
    if (totalPages > 1) {
        remindersEmbed.setFooter("Page: %d/%d".formatted(pageToShow, totalPages));
        pendingRemindersPage.addActionRow(createPageTurnButtons(pageToShow, totalPages));

        pendingRemindersForPage = getPageEntries(pendingReminders, pageToShow);
    }

    pendingRemindersForPage.forEach(reminder -> addReminderAsField(reminder, remindersEmbed))
}
```

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -63,6 +63,10 @@ public final class ReminderCommand extends SlashCommandAdapter {
             List.of("minutes", "hours", "days", "weeks", "months", "years");
     private static final Period MAX_TIME_PERIOD = Period.ofYears(3);
     private static final int REMINDERS_PER_PAGE = 10;
+    /**
+     * Maximum length of field title set by <a href=
+     * "https://discord.com/developers/docs/resources/channel#embed-object-embed-limits">Discord</a>
+     */
```
Zabuzard, line 69: Get rid of this field and use `MessageEmbed.TITLE_MAX_LENGTH` instead

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: yes i was thinking of something like this, y not we just change the main variable to list??

Zabuzard: also fine by me 👍

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
```diff
@@ -125,40 +159,84 @@ private void handleCreateCommand(SlashCommandInteractionEvent event) {
...
+        return database.read(context -> context.selectFrom(PENDING_REMINDERS)
+            .where(PENDING_REMINDERS.GUILD_ID.eq(guild.getIdLong())
+                .and(PENDING_REMINDERS.AUTHOR_ID.eq(user.getIdLong())))
+            .orderBy(PENDING_REMINDERS.REMIND_AT.asc())
+            .fetch());
+    }
 
-        Result<PendingRemindersRecord> pendingReminders = getReminders(guildId, userId);
+    private MessageCreateData createListMessage(Result<PendingRemindersRecord> pendingReminders,
+            int pageToShow) {
```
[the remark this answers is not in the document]
Taz03: what about instead of an exception we do what we did in audit command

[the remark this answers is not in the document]
Zabuzard: what was that? im for fail-fast. when there is a bug in the code, it has to be detected asap. the best way to do that is with exceptions.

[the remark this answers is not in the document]
Taz03: image
we made it to adjust the page number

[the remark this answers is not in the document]
Taz03: so then it needs to be a util method right? where to put this method then?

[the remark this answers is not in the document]
Zabuzard: is `clamp` right now a local private method in that other class?

i guess `org.togetherjava.tjbot.commands.utils` and then a class called `Pagination`. we can also think about adding more commonly used stuff there later. pagination seems to be frequently used, so i want to add some utility later on.

but for now its fine if its just clamp.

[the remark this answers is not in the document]
Taz03: > is clamp right now a local private method in that other class?

yes

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/reminder/ReminderCommand.java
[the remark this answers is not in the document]
Taz03: done

Zabuzard (approved):

Tais993 (approved):

PR #651, opened 2022-10-23
Added cooldown to ask command

### LINKED ISSUE #555: Add Question cooldown with command-advice
## Problem

People frequently start creating multiple threads for the same question, because they do not understand the system (despite all the existing explanation).

For example, they do a typo in the title or forgot to post any message at all, or want to have a different category and so on. Instead of using the slash commands, they create a new thread.

## Solution

The bot should block such attempts (5 (?) minute cooldown) and additionally provide an explanation message how to use the slash-commands for adjusting the existing message.

A similar feature exists already for `ImplicitAskListener` which blocks attempts for a few seconds and links the user to their existing thread. This existing feature should just be improved, so that `AskCommand` also uses it, it lasts for 5 minutes instead of seconds, and it posts slash-command usage advice as well.

Taz03:
closes #555 

image
no thread was created for the second command

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -144,8 +144,8 @@ public final class AskCommand extends SlashCommandAdapter {
             .read(context -> context.selectFrom(HELP_THREADS)
                 .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
                 .orderBy(HELP_THREADS.CREATED_AT.desc())
-                .fetch())
-            .get(0)
+                .limit(1)
+                .fetchOne())
             .getChannelId();
 
         String message =
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -144,14 +144,9 @@ public final class AskCommand extends SlashCommandAdapter {
             .read(context -> context.selectFrom(HELP_THREADS)
                 .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
                 .orderBy(HELP_THREADS.CREATED_AT.desc())
-                .fetchOne())
+                .fetch())
+            .get(0)
             .getChannelId();
-
-        if (lastThreadByAuthorId == 0) {
-            logger.warn("Can't find the last help thread created by the user with id ({})",
-                    user.getId());
-            return;
-        }
 
         String message =
                 """
@@ -160,7 +155,7 @@ public final class AskCommand extends SlashCommandAdapter {
 
         RestAction.allOf(changeCategory, changeTitle)
             .map(commandMentions -> message.formatted(COOLDOWN_DURATION_VALUE,
-                    COOLDOWN_DURATION_UNIT.name(),
+                    COOLDOWN_DURATION_UNIT.name().toLowerCase(),
                     MessageUtils.mentionChannelById(lastThreadByAuthorId), commandMentions.get(0),
                     commandMentions.get(1)))
             .flatMap(text -> event.reply(text).setEphemeral(true))
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -1,9 +1,8 @@
 package org.togetherjava.tjbot.commands.help;
 
-import net.dv8tion.jda.api.entities.Guild;
-import net.dv8tion.jda.api.entities.IMentionable;
-import net.dv8tion.jda.api.entities.Member;
-import net.dv8tion.jda.api.entities.Message;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
@@ -19,9 +18,13 @@ import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.commands.CommandVisibility;
 import org.togetherjava.tjbot.commands.SlashCommandAdapter;
+import org.togetherjava.tjbot.commands.utils.MessageUtils;
 import org.togetherjava.tjbot.config.Config;
 
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MAX;
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MIN;
@@ -47,10 +50,16 @@ import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPAC
  * </pre>
  */
 public final class AskCommand extends SlashCommandAdapter {
+    private static final int COOLDOWN_DURATION_VALUE = 5;
+    private static final ChronoUnit COOLDOWN_DURATION_UNIT = ChronoUnit.MINUTES;
     private static final Logger logger = LoggerFactory.getLogger(AskCommand.class);
     public static final String COMMAND_NAME = "ask";
     private static final String TITLE_OPTION = "title";
     private static final String CATEGORY_OPTION = "category";
+    private final Cache<Long, Instant> userToLastAsk = Caffeine.newBuilder()
+        .maximumSize(1_000)
+        .expireAfterAccess(COOLDOWN_DURATION_VALUE, TimeUnit.of(COOLDOWN_DURATION_UNIT))
+        .build();
     private final HelpSystemHelper helper;
 
     /**
@@ -77,6 +86,31 @@ public final class AskCommand extends SlashCommandAdapter {
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
+        if (isUserOnCooldown(event.getUser())) {
+            String message =
+                    """
+                            You can only create 1 help thread per 5 minutes. While you are waiting changing the infos of your previous thread may help.
+                            Use these commands in the help thread to change its details:-
+                            %s
+                            %s""";
+
+            RestAction<String> changeTitle = MessageUtils.mentionGuildSlashCommand(event.getGuild(),
+                    HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                    HelpThreadCommand.CHANGE_TITLE_SUBCOMMAND);
+            RestAction<String> changeCategory = MessageUtils.mentionGuildSlashCommand(
+                    event.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                    HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                    HelpThreadCommand.CHANGE_CATEGORY_OPTION);
+
+            RestAction.allOf(changeCategory, changeTitle)
+                .map(commandMentions -> message.formatted(commandMentions.get(0),
+                        commandMentions.get(1)))
+                .flatMap(event::reply)
+                .queue();
+
+            return;
+        }
+
         String title = event.getOption(TITLE_OPTION).getAsString();
         String category = event.getOption(CATEGORY_OPTION).getAsString();
 
@@ -102,8 +136,15 @@ public final class AskCommand extends SlashCommandAdapter {
         overviewChannel.createThreadChannel(name.toChannelName())
             .flatMap(threadChannel -> handleEvent(eventHook, threadChannel, author, title, category,
                     guild))
-            .queue(any -> {
-            }, e -> handleFailure(e, eventHook));
+            .queue(any -> userToLastAsk.put(event.getUser().getIdLong(), Instant.now()),
+                    e -> handleFailure(e, eventHook));
+    }
+
+    private boolean isUserOnCooldown(User user) {
+        return Optional.ofNullable(userToLastAsk.getIfPresent(user.getIdLong()))
+            .map(lastAction -> lastAction.plus(COOLDOWN_DURATION_VALUE, COOLDOWN_DURATION_UNIT))
+            .filter(Instant.now()::isBefore)
+            .isPresent();
     }
 
     private boolean handleIsValidTitle(CharSequence title, IReplyCallback event) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -259,7 +259,7 @@ public final class AskCommand extends SlashCommandAdapter {
     }
 
     private static RestAction<String> mentionHelpChangeCommand(Guild guild, String subcommand) {
-        return MessageUtils.mentionGuildSlashCommand(guild, COMMAND_NAME, CHANGE_SUBCOMMAND_GROUP,
-                subcommand);
+        return MessageUtils.mentionGuildSlashCommand(guild, HelpThreadCommand.COMMAND_NAME,
+                CHANGE_SUBCOMMAND_GROUP, subcommand);
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -103,8 +103,7 @@ public final class AskCommand extends SlashCommandAdapter {
                     HelpThreadCommand.CHANGE_CATEGORY_OPTION);
 
             RestAction.allOf(changeCategory, changeTitle)
-                .map(commandMentions -> message.formatted(commandMentions.get(0),
-                        commandMentions.get(1)))
+                .map(mentions -> message.formatted(mentions.get(0), mentions.get(1)))
                 .flatMap(text -> event.reply(text).setEphemeral(true))
                 .queue();
 
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -23,6 +23,8 @@ import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.db.Database;
 import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
 
+import javax.annotation.Nullable;
+
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Optional;
@@ -138,16 +140,7 @@ public final class AskCommand extends SlashCommandAdapter {
         User user = event.getUser();
         Guild guild = event.getGuild();
 
-        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
-        RestAction<String> changeCategory =
-                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
-
-        HelpThreadsRecord lastThreadByAuthor =
-                database.read(context -> context.selectFrom(HELP_THREADS)
-                    .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
-                    .orderBy(HELP_THREADS.CREATED_AT.desc())
-                    .limit(1)
-                    .fetchOne());
+        HelpThreadsRecord lastThreadByAuthor = getLatestHelpThread(user);
 
         String cooldownDuration =
                 COOLDOWN_DURATION_VALUE + " " + COOLDOWN_DURATION_UNIT.name().toLowerCase();
@@ -156,7 +149,8 @@ public final class AskCommand extends SlashCommandAdapter {
             logger.warn("Can't find the last help thread created by the user with id ({})",
                     user.getId());
             event
-                .reply("Sorry, something went wrong try again after %s".formatted(cooldownDuration))
+                .reply("Sorry, something went wrong. Please try again after %s"
+                    .formatted(cooldownDuration))
                 .queue();
             return;
         }
@@ -166,12 +160,25 @@ public final class AskCommand extends SlashCommandAdapter {
                         Sorry, you can only create a single help thread every %s. Please use your existing thread %s instead.
                         If you made a typo or similar, you can adjust the title using the command %s and the category with %s 👌""";
 
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory =
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
+
         RestAction.allOf(changeCategory, changeTitle)
             .map(commandMentions -> message.formatted(cooldownDuration,
                     MessageUtils.mentionChannelById(lastThreadByAuthor.getChannelId()),
                     commandMentions.get(0), commandMentions.get(1)))
             .flatMap(text -> event.reply(text).setEphemeral(true))
             .queue();
+    }
+
+    @Nullable
+    private HelpThreadsRecord getLatestHelpThread(User user) {
+        return database.read(context -> context.selectFrom(HELP_THREADS)
+            .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
+            .orderBy(HELP_THREADS.CREATED_AT.desc())
+            .limit(1)
+            .fetchOne());
     }
 
     private boolean handleIsValidTitle(CharSequence title, IReplyCallback event) {
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeUnit;
 
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MAX;
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MIN;
+import static org.togetherjava.tjbot.commands.help.HelpThreadCommand.*;
 import static org.togetherjava.tjbot.db.generated.Tables.HELP_THREADS;
 
 /**
@@ -133,18 +134,12 @@ public final class AskCommand extends SlashCommandAdapter {
     }
 
     private void sendCooldownResponse(SlashCommandInteractionEvent event) {
-        Guild guild = event.getGuild();
         User user = event.getUser();
+        Guild guild = event.getGuild();
 
-        String message =
-                """
-                        Sorry, you can only create a single help thread every %s %s. Please use your existing thread %s instead.
-                        If you made a typo or similar, you can adjust the title using the command %s and the category with %s 👌""";
-
-        RestAction<String> changeTitle =
-                mentionHelpChangeCommand(guild, HelpThreadCommand.CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
         RestAction<String> changeCategory =
-                mentionHelpChangeCommand(guild, HelpThreadCommand.CHANGE_CATEGORY_SUBCOMMAND);
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
         long lastThreadByAuthorId = database
             .read(context -> context.selectFrom(HELP_THREADS)
                 .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
@@ -157,6 +152,11 @@ public final class AskCommand extends SlashCommandAdapter {
                     user.getId());
             return;
         }
+
+        String message =
+                """
+                        Sorry, you can only create a single help thread every %s %s. Please use your existing thread %s instead.
+                        If you made a typo or similar, you can adjust the title using the command %s and the category with %s 👌""";
 
         RestAction.allOf(changeCategory, changeTitle)
             .map(commandMentions -> message.formatted(COOLDOWN_DURATION_VALUE,
@@ -236,7 +236,7 @@ public final class AskCommand extends SlashCommandAdapter {
     }
 
     private static RestAction<String> mentionHelpChangeCommand(Guild guild, String subcommand) {
-        return MessageUtils.mentionGuildSlashCommand(guild, HelpThreadCommand.COMMAND_NAME,
-                HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP, subcommand);
+        return MessageUtils.mentionGuildSlashCommand(guild, COMMAND_NAME, CHANGE_SUBCOMMAND_GROUP,
+                subcommand);
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
@@ -27,8 +27,10 @@ import javax.annotation.Nullable;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MAX;
 import static org.togetherjava.tjbot.commands.help.HelpSystemHelper.TITLE_COMPACT_LENGTH_MIN;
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -77,6 +86,30 @@ public AskCommand(Config config, HelpSystemHelper helper) {
...
+                    event.getGuild(), HelpThreadCommand.COMMAND_NAME,
+                    HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                    HelpThreadCommand.CHANGE_CATEGORY_OPTION);
+
+            RestAction.allOf(changeCategory, changeTitle)
+                .map(mentions -> message.formatted(mentions.get(0), mentions.get(1)))
+                .flatMap(text -> event.reply(text).setEphemeral(true))
+                .queue();
+
+            return;
```
Zabuzard, line 110: move into helper method please

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -77,6 +86,30 @@ public AskCommand(Config config, HelpSystemHelper helper) {
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
+        if (isUserOnCooldown(event.getUser())) {
+            String message =
+                    """
+                            You can only create 1 help thread per 5 minutes. While you are waiting changing the infos of your previous thread may help.
+                            Use these commands in the help thread to change its details:-
+                            %s
+                            %s""";
```
Zabuzard, line 95: * It is crucial that we link the existing thread
* Needs some friendlyness and grammar rework.

> Sorry, you can only create a single help thread every 5 minutes. Please use your existing thread #foo instead.
> If you made a typo or similar, you can adjust the title using the command /foo and the category with /foo 👌

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -141,4 +141,13 @@ public static String abbreviate(String text, int maxLength) {
...
     }
 
+    /**
+     * Mentions a guild channel by its id
+     *
+     * @param channelId the channel id to mention
+     * @return channel mention
+     */
+    public static String mentionChannelById(long channelId) {
+        return "<#%s>".formatted(channelId);
```
Zabuzard, line 151: Please add a comment that links to where this is from and why it had to be created. I.e.
```java
// Clone of JDAs Channel#getAsMention, but unfortunately channel instances can not be created out of just an ID, unlike User#fromId
```
then people also know directly that this method can be deleted once JDA adds this method for channels as well

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -141,4 +141,13 @@ public static String abbreviate(String text, int maxLength) {
         return text.substring(0, maxLength - ABBREVIATION.length()) + ABBREVIATION;
     }
 
+    /**
+     * Mentions a guild channel by its id
```
Zabuzard, line 145: * missing dot.
* missing the behavior when given an unknown ID

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -141,4 +141,13 @@ public static String abbreviate(String text, int maxLength) {
...
+     *
+     * @param channelId the channel id to mention
```
Zabuzard, line 147: thats wrong english. instead `@param channelID the ID of the channel to mention`

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -141,4 +141,13 @@ public static String abbreviate(String text, int maxLength) {
...
+     * @return channel mention
```
Zabuzard, line 148: too brief. `the channel as formatted string which Discord interprets as clickable mention`

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
             .isPresent();
     }
 
+    private void sendCooldownResponse(SlashCommandInteractionEvent event) {
+        User user = event.getUser();
+
+        String message =
+                """
+                        Sorry, you can only create a single help thread every 5 minutes. Please use your existing thread %s instead.
```
Zabuzard, line 140: dont hardcode the duration, use the constant

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+                        If you made a typo or similar, you can adjust the title using the command %s and the category with %s :ok_hand:""";
```
Zabuzard, line 141: put the smiley as plain utf-8 symbol in there, like, put 👌 in there. java source code supports utf-8

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+
+        RestAction<String> changeTitle = MessageUtils.mentionGuildSlashCommand(event.getGuild(),
+                HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                HelpThreadCommand.CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory = MessageUtils.mentionGuildSlashCommand(event.getGuild(),
+                HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
+                HelpThreadCommand.CHANGE_CATEGORY_OPTION);
```
Zabuzard, line 148: duplication. might want to create a helper method `mentionHelpChangeCommand(String subcommand)` right here in this class.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+        long lastCreatedThreadId = database
```
Zabuzard, line 149: missing `byAuthor`, maybe just `lastThreadByAuthor` or `lastThreadByAuthorId`

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+            .read(context -> context.selectFrom(HELP_THREADS)
+                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
+                .orderBy(HELP_THREADS.CREATED_AT.desc())
+                .fetch())
+            .get(0)
```
Zabuzard, line 154: use `fetchOne()` instead.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+            .getChannelId();
+
+        if (lastCreatedThreadId == 0) {
+            logger.warn("Can't find the last help thread created by the user with id ({})",
+                    user.getId());
+        }
```
Zabuzard, line 160: this is incorrect. feels like u didnt test it.

if there is no thread, it will fail on your `get(0)` already, not down here.

u have to test code, people trust u on this...

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+        long lastCreatedThreadId = database
+            .read(context -> context.selectFrom(HELP_THREADS)
+                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
+                .orderBy(HELP_THREADS.CREATED_AT.desc())
+                .fetch())
+            .get(0)
+            .getChannelId();
+
+        if (lastCreatedThreadId == 0) {
+            logger.warn("Can't find the last help thread created by the user with id ({})",
```
Zabuzard, line 158: give some context, add what you attempted to do.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard, line 160: why do u continue from here? why doesnt it abort the flow or send some specific message instead? right now you will continue and mention the channel `0`

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -146,6 +132,40 @@ private boolean isUserOnCooldown(User user) {
...
+                    user.getId());
+        }
+
+        RestAction.allOf(changeCategory, changeTitle)
+            .map(mentions -> message.formatted(MessageUtils.mentionChannelById(lastCreatedThreadId),
```
Zabuzard, line 163: what mentions? maybe `commandMentions` or `helpChangeMentions`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: @Zabuzard will it still be the case if i do .fetchOne and not get(0) is there a way to get 0 by default?

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: it already does i said can't find the last help thread and gave the user id

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: fetchOne returns an `Optional`

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: yeah but i meant the context of the command. when u read that message, u dont know that its:
* ask command
* on cooldown

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: this should rather be a comment than part of the javadoc. javadoc is to document the method, not to give internal implementation notes.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: (not done yet)

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -133,35 +133,36 @@ private boolean isUserOnCooldown(User user) {
...
-                        If you made a typo or similar, you can adjust the title using the command %s and the category with %s :ok_hand:""";
-
-        RestAction<String> changeTitle = MessageUtils.mentionGuildSlashCommand(event.getGuild(),
-                HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
-                HelpThreadCommand.CHANGE_TITLE_SUBCOMMAND);
-        RestAction<String> changeCategory = MessageUtils.mentionGuildSlashCommand(event.getGuild(),
-                HelpThreadCommand.COMMAND_NAME, HelpThreadCommand.CHANGE_SUBCOMMAND_GROUP,
-                HelpThreadCommand.CHANGE_CATEGORY_OPTION);
-        long lastCreatedThreadId = database
+                        Sorry, you can only create a single help thread every %s %s. Please use your existing thread %s instead.
```
Zabuzard, line 141: the text should be closer to where its used. right now its quite far away and then its easy to get confused with the parameters.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: or null or sth. point is, it tells u when it didnt find anything and u should check on that. like
```java
Foo lastThreadByAuthor = database...

if (lastThreadByAuthor == null) {
  ...
  return;
}

String lastThreadByAuthorMention = MessageUtils.mentionChannelById(lastThreadByAuthor.getChannelId();
...
```

Zabuzard: ow it just silently fails. u have to respond sth to the user.

Taz03: primitives can't be null no? it'll get to the default value which is 0 iirc

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: can u give a fully constructed sentence, im bad at these things 😅

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: what should i do? should i just remove the cooldown and let the user make a new thread? coz a normal member can't delete threads, right? so it can be a discord problem or smth

Zabuzard: but it doesnt return a primitive. it returns the record. and u call a method on the record. but its null and crashes.

please for god sake, **test** your code. it is mandatory that you do that. trigger all edge cases at least once to verify that they work. its so easy to make mistakes and u cant always rely on people seeing it in code reviews. we will miss something.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: Sure

> Attempted to find the last help thread of user {} during ask-cooldown, but there is none.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: u can pick whats easier:

* ignore the cooldown, allow the thread creation
* just get rid of the "go to channel x" part of the message in this particular case.

Taz03: second option is easier to implement, but what to write then?

> Sorry, you can only create a single help thread every %s %s.
If you made a typo or similar, you can adjust the title using the command %s and the category with %s 👌

this? i don't think this sounds right i prefer this:-

> Sorry, you can only create a single help thread every %s %s.

Taz03: or i think we should just not handle this case, the odds of this happening is extremely low, like only if discord have some problem

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03:
```
org.togetherjava.tjbot.db.DatabaseException: org.jooq.exception.TooManyRowsException: Cursor returned more than one result
	at org.togetherjava.tjbot.db.Database.read(Database.java:92) ~[database-1.0-SNAPSHOT.jar:?]
	at org.togetherjava.tjbot.commands.help.AskCommand.sendCooldownResponse(AskCommand.java:146) ~[main/:?]
	at org.togetherjava.tjbot.commands.help.AskCommand.onSlashCommand(AskCommand.java:98) ~[main/:?]
```
 getting this error when using fetchOne going back to what it was before

Taz03: @Zabuzard

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: u cant decide to "not handle the case". if u dont do anything extra, the interaction just fails and the user is extremely confused.

they think their thread was created, but it wasnt. and they didnt get any message at all. instead, after waiting 15 min, the bot says "application failed and didnt respond".

Taz03: > if u dont do anything extra, the interaction just fails and the user is extremely confused

can u reproduce this situation?? this is impossible to happen they y handle this?

Zabuzard: > can u reproduce this situation?? this is impossible to happen they y handle this?

it is not impossible. as u said, it just needs someone to delete the thread. or maybe a bot downtime at the wrong moment or ...

this is simply a bug. you have to handle it in _a way_. send the user an error message if you want, but at least do _something_.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: because u have to limit it to 1 in the query.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: please move the comment inside the method, thanks

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: > missing the behavior when given an unknown ID

still not done yet.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: still wrong.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: done

Zabuzard (changes requested):
(see unresolved conversations)

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: > it just needs someone to delete the thread

no that is a different case actually, my bad. for `lastCreatedThreadByAuthor` to be null, the ask command which was used before this needs to somehow break and not put anything in the db, thats the only case in which this can be null

if someone just deleted the thread, this'll say `#deleted-channel`

> or maybe a bot downtime at the wrong moment

not sure if this can be the case either coz i only add the cooldown after the thread is successfully created.

anyways even if this is the case the chances are very very low

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Zabuzard: regardless how long we are discussing this, u have to do something sane in the code and not just NOPE LETS CRASH.
at least respond to the user "there was an error, sorry". like, whats the problem, for real.

Taz03: but this is "impossible" to happen

Taz03: image
see this it first creates the thread and then adds it to db and after that only it adds the cooldown, so there is no way cooldown in added but no thread is saved in db

Zabuzard: it is possible. the database could have a hicup, the database could be manipulated, the bot could have crashed completely or shutdown at a bad timing, ...

im not telling u that u have to handle this case in a graceful manner. im telling u that u have to define a sane behavior. if u want to define "this is a case that i dont want to support", then this is fine, but u still have to define that flow then.

i.e. log an error message and respond to the user that sth failed.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -102,8 +122,44 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory =
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
+        long lastThreadByAuthorId = database
+            .read(context -> context.selectFrom(HELP_THREADS)
+                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
+                .orderBy(HELP_THREADS.CREATED_AT.desc())
+                .limit(1)
+                .fetchOne())
```
Tais993, line 148: Doesn't this just do the same?

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: > the database could have a hicup

pretty sure this is not the only place where the program can have a issue with literally the database not saving something

> the bot could have crashed completely or shutdown at a bad timing

the cooldowns get removed after restart, no? so no cooldown on startup so user will be able to create new threads

> but u still have to define that flow then

is it ok if i just send this reply "Sorry, something went wrong try again after {cooldown duration}"??

to me this still sounds like a 0% chance of happening

Tais993 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: no `.fetchOne()` needs to limit the query to 1 result, if i have multiple results it'll throw an exception

Tais993: I see, my bad

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: done

Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -134,34 +136,40 @@ private boolean isUserOnCooldown(User user) {
...
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory =
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
+
+        HelpThreadsRecord lastThreadByAuthor =
+                database.read(context -> context.selectFrom(HELP_THREADS)
+                    .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
+                    .orderBy(HELP_THREADS.CREATED_AT.desc())
+                    .limit(1)
+                    .fetchOne());
```
Zabuzard, line 150: can u move this into a helper method please?

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -134,34 +136,40 @@ private boolean isUserOnCooldown(User user) {
...
-                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong()))
-                .orderBy(HELP_THREADS.CREATED_AT.desc())
-                .fetch())
-            .get(0)
-            .getChannelId();
-
-        if (lastCreatedThreadId == 0) {
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory =
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
```
Zabuzard, line 143: move that below to where its needed first.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -142,12 +142,15 @@ public static String abbreviate(String text, int maxLength) {
     }
 
     /**
-     * Mentions a guild channel by its id
+     * Mentions a guild channel by its id. If the given channelId is unknown the formatted text will
+     * say `#deleted-channel` in discord.
```
Zabuzard, line 146: javadoc is html, not markdown. use `<i>#deleted-channel</i>` instead.

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
```diff
@@ -142,12 +142,15 @@ public static String abbreviate(String text, int maxLength) {
     }
 
     /**
-     * Mentions a guild channel by its id
+     * Mentions a guild channel by its id. If the given channelId is unknown the formatted text will
```
Zabuzard, line 145: missing comma, "... is unknown, the formatted text ..."

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Zabuzard, line 146: Discord not discord

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -134,34 +136,40 @@ private boolean isUserOnCooldown(User user) {
...
+                    .fetchOne());
+
+        String cooldownDuration =
+                COOLDOWN_DURATION_VALUE + " " + COOLDOWN_DURATION_UNIT.name().toLowerCase();
+
+        if (lastThreadByAuthor == null) {
             logger.warn("Can't find the last help thread created by the user with id ({})",
                     user.getId());
+            event
+                .reply("Sorry, something went wrong try again after %s".formatted(cooldownDuration))
```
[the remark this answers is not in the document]
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/utils/MessageUtils.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -156,7 +149,8 @@ private void sendCooldownResponse(SlashCommandInteractionEvent event) {
             logger.warn("Can't find the last help thread created by the user with id ({})",
                     user.getId());
             event
-                .reply("Sorry, something went wrong try again after %s".formatted(cooldownDuration))
+                .reply("Sorry, something went wrong. Please try again after %s"
```
Zabuzard, line 152: is this missing a dot at the end of the sentence?

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: done

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -173,4 +248,9 @@ private static void handleFailure(Throwable exception, InteractionHook eventHook
 
         logger.error("Attempted to create a help thread, but failed", exception);
     }
+
+    private static RestAction<String> mentionHelpChangeCommand(Guild guild, String subcommand) {
+        return MessageUtils.mentionGuildSlashCommand(guild, COMMAND_NAME, CHANGE_SUBCOMMAND_GROUP,
+                subcommand);
+    }
```
Nxllpointer, line 255: Oops. This is probably supposed to be returning a mention of `/help-thread change category/title`. It currently returns `/ask change category/title`. This is probably because of you importing all static things of `HelpThreadCommand` and forgot that this class has the same field but with a different value.

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -102,8 +125,60 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+        String cooldownDuration =
+                COOLDOWN_DURATION_VALUE + " " + COOLDOWN_DURATION_UNIT.name().toLowerCase();
+
+        if (lastThreadByAuthor == null) {
+            logger.warn("Can't find the last help thread created by the user with id ({})",
+                    user.getId());
+            event
+                .reply("Sorry, something went wrong. Please try again after %s."
+                    .formatted(cooldownDuration))
+                .queue();
```
Nxllpointer, line 154: Forgot to make this ephemeral

Nxllpointer (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -102,8 +125,60 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+                        If you made a typo or similar, you can adjust the title using the command %s and the category with %s 👌""";
+
+        RestAction<String> changeTitle = mentionHelpChangeCommand(guild, CHANGE_TITLE_SUBCOMMAND);
+        RestAction<String> changeCategory =
+                mentionHelpChangeCommand(guild, CHANGE_CATEGORY_SUBCOMMAND);
+
+        RestAction.allOf(changeCategory, changeTitle)
+            .map(commandMentions -> message.formatted(cooldownDuration,
+                    MessageUtils.mentionChannelById(lastThreadByAuthor.getChannelId()),
+                    commandMentions.get(0), commandMentions.get(1)))
```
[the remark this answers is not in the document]
Taz03: it'll then be same ugly thing but in a function, ig this is fine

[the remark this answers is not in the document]
Nxllpointer: In an extra function you can store the format parameters in extra variables so you dont have to stretch your .format across 3 lines. This is very unreadable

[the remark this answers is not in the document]
Taz03: done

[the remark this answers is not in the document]
Nxllpointer: Way better imo

Nxllpointer (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
```diff
@@ -250,7 +259,7 @@ private static void handleFailure(Throwable exception, InteractionHook eventHook
     }
 
     private static RestAction<String> mentionHelpChangeCommand(Guild guild, String subcommand) {
-        return MessageUtils.mentionGuildSlashCommand(guild, COMMAND_NAME, CHANGE_SUBCOMMAND_GROUP,
-                subcommand);
+        return MessageUtils.mentionGuildSlashCommand(guild, HelpThreadCommand.COMMAND_NAME,
+                CHANGE_SUBCOMMAND_GROUP, subcommand);
```
Zabuzard, line 263: why did u prefix one but not the other? be consistent. either prefix both (and remove ur static import), or prefix none and roll with the static import

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/AskCommand.java
Taz03: @Zabuzard i can't use static import for `COMMAND_NAME` coz this class also have the same field with different value, so it'll use that one

Nxllpointer: Good call by zabuzard. this just complicates things.

Zabuzard: @Taz03 thats fair, and indicates that a static import here just causes confusion. please get rid of it and fully classify both, thanks.

Taz03: done

Zabuzard (approved):

Nxllpointer (approved):

PR #657, opened 2022-10-24
Fixed issue with jooq version

Zabuzard:
We defined the jooq version at multiple points. This lead to multiple mismatches in the past and also confuses dependabot.

Now, we moved it to a single point. That way, issues like that wont happen anymore.


Zabuzard:
fyi @Nxllpointer

Nxllpointer:
Good idea to make it central

PR #667, opened 2022-11-01
Uniform Color Palette

### LINKED ISSUE #662: Uniform color palette
I have noticed that when coloring embeds for example, [everyone just uses whatever color he likes most.](https://github.com/Together-Java/TJ-Bot/search?l=Java&q=Color)

This is not very pretty and consistent so I suggest we introduce a color palette to use for things like embeds.

We should discuss our favorite color palettes and then maybe have the users vote on the best one.

Nxllpointer:
Closes #662 

This introduces a new `Colors` class that currently contains the Catppuccin Macchiato color palette and the individual feature colors consisting of the palettes colors. I also made the features use their colors from the `Colors` class.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
@@ -1,0 +1,81 @@
+package org.togetherjava.tjbot.commands;
+
+import java.awt.Color;
+
+/**
+ * <a href="https://github.com/catppuccin/catppuccin#-palettes">Catppuccin Macchiato</a> color
+ * palette for all features to use.
+ */
+public final class Colors {
+
+    private Colors() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    // Catppuccin Macchiato colors
+    public static final Color COLOR_ROSEWATER = new Color(0xf4dbd6);
+    public static final Color COLOR_FLAMINGO = new Color(0xf0c6c6);
+    public static final Color COLOR_PINK = new Color(0xf5bde6);
+    public static final Color COLOR_MAUVE = new Color(0xc6a0f6);
+    public static final Color COLOR_RED = new Color(0xed8796);
+    public static final Color COLOR_MAROON = new Color(0xee99a0);
+    public static final Color COLOR_PEACH = new Color(0xf5a97f);
+    public static final Color COLOR_YELLOW = new Color(0xeed49f);
+    public static final Color COLOR_GREEN = new Color(0xa6da95);
+    public static final Color COLOR_TEAL = new Color(0x8bd5ca);
+    public static final Color COLOR_SKY = new Color(0x91d7e3);
+    public static final Color COLOR_SAPPHIRE = new Color(0x7dc4e4);
+    public static final Color COLOR_BLUE = new Color(0x8aadf4);
+    public static final Color COLOR_LAVENDER = new Color(0xb7bdf8);
+    public static final Color COLOR_TEXT = new Color(0xcad3f5);
+    public static final Color COLOR_SUBTEXT1 = new Color(0xb8c0e0);
+    public static final Color COLOR_SUBTEXT0 = new Color(0xa5adcb);
+    public static final Color COLOR_OVERLAY2 = new Color(0x939ab7);
+    public static final Color COLOR_OVERLAY1 = new Color(0x8087a2);
+    public static final Color COLOR_OVERLAY0 = new Color(0x6e738d);
+    public static final Color COLOR_SURFACE2 = new Color(0x5b6078);
+    public static final Color COLOR_SURFACE1 = new Color(0x494d64);
+    public static final Color COLOR_SURFACE0 = new Color(0x363a4f);
+    public static final Color COLOR_BASE = new Color(0x24273a);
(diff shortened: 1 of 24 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
```diff
@@ -0,0 +1,81 @@
...
+    public static final Color LOG_TRACE = COLOR_SKY;
+    public static final Color LOG_DEBUG = COLOR_SAPPHIRE;
+    public static final Color LOG_INFO = SUCCESS;
+    public static final Color LOG_WARN = WARNING;
+    public static final Color LOG_ERROR = FAILURE;
+    public static final Color LOG_FATAL = COLOR_RED;
+
+    public static final Color ROLE_SELECT = COLOR_TEAL;
+
+    public static final Color HELP = COLOR_YELLOW;
```
Taz03, line 79: this is not how it was supposed to be, each feature can use a color no need to make a new variable in a util file for each feature

im very against this

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
Nxllpointer: This way you can easily change the color palette in the future

Taz03: y would u want to do that?? make a var at top of each feature for the color it is going to use
this is supposed to be a file which contains color pallet and doesn't get updated unless we decide to switch to a new theme

@Zabuzard whats ur opinion on this??

Tais993: Yea this is too many colours, adding a colour pallette helps against the 5000 different colours, we should limit it to 5 or something

(a kind of) green for success
(a kind of) red for failure
etc

Taz03: @Tais993 vote for one dark then

Tais993: If you're talking about the suggestion you made in the Discord, read my message again, I'm quite saying the opposite of that. I want colours, but not 500 colours, instead like 5

Taz03: one dark have 7 is that ok?

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
```diff
@@ -0,0 +1,81 @@
...
+    public static final Color COLOR_SUBTEXT0 = new Color(0xa5adcb);
+    public static final Color COLOR_OVERLAY2 = new Color(0x939ab7);
+    public static final Color COLOR_OVERLAY1 = new Color(0x8087a2);
+    public static final Color COLOR_OVERLAY0 = new Color(0x6e738d);
+    public static final Color COLOR_SURFACE2 = new Color(0x5b6078);
+    public static final Color COLOR_SURFACE1 = new Color(0x494d64);
+    public static final Color COLOR_SURFACE0 = new Color(0x363a4f);
+    public static final Color COLOR_BASE = new Color(0x24273a);
+    public static final Color COLOR_MANTLE = new Color(0x1e2030);
+    public static final Color COLOR_CRUST = new Color(0x181926);
```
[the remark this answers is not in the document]
Nxllpointer: Yeah that makes sense

application/src/main/java/org/togetherjava/tjbot/commands/Colors.java
Tais993: Didn't know you were talking about the issue, I gave my vote there.

Nxllpointer:
So how do we now actually want to implement this? Use mostly red green yellow, have a color for each feature, store all feature colors in a common place?

Zabuzard:
1. a utility class like u already added, but in the util package
2. the whole color palette as private constants
3. expose some of them as themed constants `public static Color INFO = SKY;` (just an example)

Tais993:
> 1. a utility class like u already added, but in the util package
> 
>     2. the whole color palette as private constants
> 
>     3. expose some of them as themed constants `public static Color INFO = SKY;` (just an example)

Disagree with 2, see issue, I don't want all the colours.

Zabuzard:
im not saying that the whole color palette should be available for usage. but it should be put as private constants into the code, so that we have the whole color palette if needed. or maybe at least link it.

the only available/exposed colors should be the ones we actually use. with well named constants.

Tais993:
Sure!

Taz03:
i'd say make all the colors public and let separate class decide what to use, a util class should not be changed so frequently, every time we want to do something we'll need to make a new var in the utility class, probably for the same color too like say reminder embed and question overview embed want to use the same color then we'll have 2 variables in the util class for same color

Zabuzard:
Nah. We are looking for:
```java
public final class Colors {
  private static final Color ROSEWATER = ...
  private static final Color MAUVE = ...
  ...

  public static final Color INFO = ROSEWATER;
  public static final Color ERROR = MAUVE;
  ...
}
```

PR #668, opened 2022-11-02
Delete scam help threads

### LINKED ISSUE #655: Cleanup scam help threads
## Motivation

Sometimes people flood all channels with scam messages, such as Discord Nitro scam. When that happens, the bot picks up the message in `#ask_here` and creates a new help thread for them.

While we do have Dyno-Automod cleaning up bad messages and further our own `ScamBlocker` doing so, we still have a trailing help thread left over, since its created by the bot.

As of today, mods have to notice and manually **delete** (not just close) such threads.

## Proposal

We should simply delete threads that have been created by people who have been banned. We already have a database table that lists to each thread the author. So we could simply react to the event of someone being banned, check all their recent threads **(<10 minutes)** and just delete them.

Taz03:
fixes #655 

deletes threads created by the banned user in last 30 mins


application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,43 @@
...
+import org.togetherjava.tjbot.commands.EventReceiver;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
+
+import java.time.Duration;
+import java.time.Instant;
+
+import static org.togetherjava.tjbot.db.generated.Tables.HELP_THREADS;
+
+public final class OnGuildBanDeleteSpamThreadListener extends ListenerAdapter
```
Tais993, line 17: I don't like this name

SpamThreadDeletionListener?

Well first of all it shouldn't start with `On`, it's not a method

It should also be shorter

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,43 @@
...
+        implements EventReceiver {
+    private static final Duration recentThreadDuration = Duration.ofMinutes(30);
```
Tais993, line 19: 30m sounds really short

Tais993 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Taz03: the issue said 10m so i at least did it 30m

Taz03: how much do u think it should be?

Tais993: Nvm, this is fine, I re-read the issue and it was purely talking about spam messages our bot detects.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Taz03: another class in the same package have same naming

Tais993: I see, well I don't like it personally, find it *way too* specific, if JDA renames the method the class's name is obsolete already.

Zabuzard: its not supposed to match JDAs method signature. its just english text for us. "when someone leaves a guild, we delete threads".

that said, i dont like the following:
* its about scam, not spam
* why even mention that aspect. it deletes all recently created threads and thats okay
* the class name reads bad, "OnGuildBan", better would be "OnUserBanned"

so im voting for `OnUserBannedDeleteRecentThreadsListener`

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+import java.time.Instant;
+
+import static org.togetherjava.tjbot.db.generated.Tables.HELP_THREADS;
+
+/**
+ * Implements the functionality to delete recent help threads when the user gets banned.
+ */
+public final class OnGuildBanDeleteSpamThreadListener extends ListenerAdapter
+        implements EventReceiver {
+    private static final Duration recentThreadDuration = Duration.ofMinutes(30);
```
Zabuzard, line 22: naming convention. `RECENT_THREAD_DURATION`.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
+
+import java.time.Duration;
+import java.time.Instant;
+
+import static org.togetherjava.tjbot.db.generated.Tables.HELP_THREADS;
+
+/**
+ * Implements the functionality to delete recent help threads when the user gets banned.
+ */
```
Zabuzard, line 19: the "implements" part reads a bit weird. also, it doesnt really add anything on top of the class name. so u missed the opportunity to write good and helpful javadoc. maybe:

> Reacts on users being banned from a guild. It then deletes (not just closes) all their recently created help threads.
> 
> This for example tackles threads created by scammers, which are otherwise not automatically cleaned up on ban.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+public final class OnGuildBanDeleteSpamThreadListener extends ListenerAdapter
+        implements EventReceiver {
+    private static final Duration recentThreadDuration = Duration.ofMinutes(30);
+
+    private final Database database;
+
+    /**
+     * @param database the database to get help threads from
+     */
```
Zabuzard, line 28: Missing description (thats why sonar fails for ur PR right now).

> Creates a new instance.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+ * Implements the functionality to delete recent help threads when the user gets banned.
+ */
+public final class OnGuildBanDeleteSpamThreadListener extends ListenerAdapter
+        implements EventReceiver {
+    private static final Duration recentThreadDuration = Duration.ofMinutes(30);
+
+    private final Database database;
+
+    /**
+     * @param database the database to get help threads from
```
Zabuzard, line 27: id be more specific on the motivation here `@param database to find help threads created by the banned user`

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public OnGuildBanDeleteSpamThreadListener(Database database) {
+        this.database = database;
+    }
+
+    @Override
+    public void onGuildBan(GuildBanEvent event) {
```
Zabuzard, line 34: when an user is banned multiple times (ban, unban, ban), this will fetch the same entries from the DB again. make sure that the code behaves well in such a case.

i suspect it would, right now, send a lot of exceptions.

also, same happens if a thread was already deleted manually by a mod. the thread-database must not be taken as point of truth, its merely a "snapshot" of a single moment. so u must handle it with more care.

i.e. u should add some sort of "if already deleted" check. and also acount for `getThreadChannelById` to not find it etc. and then add maybe a debug-level log to that, but not further exceptions.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+        getRecentHelpThreads(event.getUser()).forEach(helpThreadsRecord -> event.getJDA()
+            .getThreadChannelById(helpThreadsRecord.getChannelId())
+            .delete()
+            .queue());
+    }
+
+    private Result<HelpThreadsRecord> getRecentHelpThreads(User user) {
```
Zabuzard, line 41: u only need the IDs of the threads, so tell the DB query to only select the IDs, not the full records.

that speeds up the DB call.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
```diff
@@ -0,0 +1,49 @@
...
+        return database
+            .read(context -> context.selectFrom(HELP_THREADS)
+                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong())
+                    .and(HELP_THREADS.CREATED_AT
+                        .greaterOrEqual(Instant.now().minus(recentThreadDuration)))))
```
Zabuzard, line 46: to improve readability, id move the math part of that instant stuff out into a variable.

```java
Instant recentThreadThreshold = Instant.now().minus(...);

return ...
    CREATED_AT.greater(recentThreadTreshold)
  ...
```
id also just use `greater`, its shorter to read and its not like we care whether > or >=

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Tais993: I really am not a fan of the "on" pattern, find it pretty disgusting tbh

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: fair. id still use it, but thats just NIT.

the other point is more important for me

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Taz03: how to do that? @Zabuzard

Zabuzard: `context.select(...ID).from(HELP_THREADS).WHERE(...)`

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public void onGuildBan(GuildBanEvent event) {
+        RestAction.allOf(getRecentHelpThreads(event.getUser()).stream()
+            .map(record -> event.getJDA().getThreadChannelById(record.value1()))
+            .filter(Objects::nonNull)
+            .map(ThreadChannel::delete)
+            .toList()).queue();
+    }
+
+    private Result<Record1<Long>> getRecentHelpThreads(User user) {
```
Zabuzard, line 50: i think it would be better if this would return `List<Long>` instead. easier to understand and u essentially just unwrap it into that on the other side anyways.

application/src/main/java/org/togetherjava/tjbot/commands/help/OnGuildBanDeleteSpamThreadListener.java
Zabuzard: done

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -0,0 +1,59 @@
...
+        this.database = database;
+    }
+
+    @Override
+    public void onGuildBan(GuildBanEvent event) {
+        RestAction.allOf(getRecentHelpThreads(event.getUser()).stream()
+            .map(record -> event.getJDA().getThreadChannelById(record.value1()))
+            .filter(Objects::nonNull)
+            .map(ThreadChannel::delete)
+            .toList()).queue();
```
Zabuzard, line 47: might be wrong, but isnt the behavior now that if a single one of them fails, all of the others fail as well?
and we still get exceptions instead of having a debug-only log.

this is still too fragile and noisy.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: not it excludes the threads which are null, so ig no exception

Zabuzard: but u know, delete can fail as well (for various reasons). for example if the thread gets deleted in the meantime or if there are rate limits or if we have discord hicups, ...

we neither want that the other threads fail to get deleted, nor do we want to see exceptions in that case.

this is a cleanup feature and its okay if it fails to cleanup something. it should just log a debug or info but not throw exceptions around

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: is this ok?

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: > there are rate limits

thats y im combining all the actions and performing all at once to have no rate limit

but the problem is if all the threads are already deleted then only then there will be an exception
```
java.lang.IllegalArgumentException: RestActions may not be empty
```

so ig i can do a normal empty check and if it is null then i'll just won't do anything

Taz03: > if the thread gets deleted in the meantime

it'll get filtered out by `.filter(Objects::nonNull)`

Taz03: image

guess i cant do that without .complete()

Taz03: do u have any alternative method? @Zabuzard

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Zabuzard: ur misunderstanding how RestAction.allOf works.
1. it does not send all of them at once to discord. they are still send individually
2. it does not either "fail all or fail nothing". it will attempt all of them individually and probably abort execution of the rest if just one of them failed, i suppose. u still end up in a mixed state where some worked and some not

what im asking for is: u should send all of them individually and have a proper onFailure listener for them to write a log info or debug level log, instead of sending exceptions all over.

failure of one shouldnt affect the others, hence why u should queue all of them individually.

Taz03: what should be the log message?

Nxllpointer:
Wasnt someone else already working on this? Or did the guy fail to do so?

Taz03:
> Wasnt someone else already working on this? Or did the guy fail to do so?

no idea, the issue was unassigned tho

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Zabuzard:
```java
logger.info("Failed to delete thread {} from banned user.", threadId, failure);
```

Zabuzard: with some sort of
```java
...stream()
...
forEach(action -> action.queue(any -> {}, failure -> {
  logger.warn(...);
}));
```

Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -40,11 +42,12 @@ public UserBannedDeleteRecentThreadsListener(Database database) {
...
     @Override
     public void onGuildBan(GuildBanEvent event) {
-        RestAction.allOf(getRecentHelpThreads(event.getUser()).stream()
-            .map(event.getJDA()::getThreadChannelById)
-            .filter(Objects::nonNull)
-            .map(ThreadChannel::delete)
-            .toList()).queue();
+        getRecentHelpThreads(event.getUser()).forEach(channelId -> event.getJDA()
+            .getThreadChannelById(channelId)
+            .delete()
```
Zabuzard, line 47: ur `map(not null)` was quite good, id keep that. otherwise u get a NPE right now still.

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -40,11 +42,12 @@ public UserBannedDeleteRecentThreadsListener(Database database) {
...
+            .queue(onSuccess -> {
+            }, onFailure -> logger.warn("Failed to delete thread {} from banned user.", channelId,
```
Zabuzard, line 49: the naming is incorrect. `onFailure` sounds as if it would be used as high order object, such as `queue(onSuccess, onFailure)`, but u use them as lambdas, so they are the input object. in this case `onFailure` is actually the `failure`/exception itself.

so `.queue(any -> {}, failure -> logger.warn(..., failure));`
or `.queue(any -> {}, Foo::onFailure);` with a corresponding method
or `.queue(any -> {}, onFailure);` where you created a `Consumer<Throwable> onFailure = ...` beforehand

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: i tried to keep it but i dont know how to

```java
        getRecentHelpThreads(event.getUser()).stream()
            .map(event.getJDA()::getThreadChannelById)
            .filter(Objects::nonNull)
            .map(ThreadChannel::delete)
            .forEach(restAction -> restAction.queue(any -> {}, failure -> {}));
```

here like this i wont be able to get the thread id when it fails so no way to log it

Zabuzard: u dont have to log threads that are null. this is an expected situation that will happen frequently.

we only have to log when `delete` failed.

Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -42,12 +43,12 @@ public UserBannedDeleteRecentThreadsListener(Database database) {
...
-            .delete()
-            .queue(onSuccess -> {
-            }, onFailure -> logger.warn("Failed to delete thread {} from banned user.", channelId,
-                    onFailure)));
+        getRecentHelpThreads(event.getUser()).stream()
+            .map(event.getJDA()::getThreadChannelById)
+            .filter(Objects::nonNull)
+            .forEach(threadChannel -> threadChannel.delete().queue(any -> {
+            }, failure -> logger.warn("Failed to delete thread {} from banned user.",
+                    threadChannel.getId(), failure)));
```
Zabuzard, line 51: might also want to log the id of the banned user

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: that'll require me to change the whole thing again

Taz03: done

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -0,0 +1,67 @@
...
+        Instant recentThreadThreshold = Instant.now().minus(RECENT_THREAD_DURATION);
+
+        return database
+            .read(context -> context.select(HELP_THREADS.CHANNEL_ID)
+                .from(HELP_THREADS)
+                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong())
+                    .and(HELP_THREADS.CREATED_AT.greaterThan(recentThreadThreshold)))
+                .fetch())
+            .stream()
+            .map(Record1::value1)
```
Nxllpointer, line 64: I dont like that. You first have to know that selecting a colnum returns a `Record1` and the method `value1` is also not descriptive. Could you just use selectFrom and then map it with `HelpThreadsRecord::getChannelId`?

```java
return database
            .read(context -> context.selectFrom(HELP_THREADS)
                .where(HELP_THREADS.AUTHOR_ID.eq(user.getIdLong())
                    .and(HELP_THREADS.CREATED_AT.greaterThan(recentThreadThreshold)))
                .fetch())
            .stream()
            .map(HelpThreadsRecord::getChannelId)
            .toList();
```
That is a lot less weird that `Record1::value1`

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
```diff
@@ -0,0 +1,67 @@
...
+    public void onGuildBan(GuildBanEvent event) {
+        getRecentHelpThreads(event.getUser()).stream()
+            .map(event.getJDA()::getThreadChannelById)
+            .filter(Objects::nonNull)
+            .forEach(threadChannel -> threadChannel.delete().queue(any -> {
+            }, failure -> logger.warn("Failed to delete thread {} from banned user {}.",
+                    threadChannel.getId(), event.getUser().getId(), failure)));
+    }
+
+    private List<Long> getRecentHelpThreads(User user) {
```
Nxllpointer, line 54: You are not returning help threads, you are returning channel IDs. You are also getting a **users** help threads, not all recent help threads.
```suggestion
    private List<Long> getUsersRecentHelpThreadChannelIDs(User user) {
```

Nxllpointer (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: it was @Zabuzard's idea to select only the channel id in query, im ok with both but the one u gave is more understandable for sure

Nxllpointer: Huh ok. Maybe zabuzards solution is faster idk

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: what about `getUsersRecentHelpThreadIDs`??

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Taz03: yes, https://github.com/Together-Java/TJ-Bot/pull/668#discussion_r1011364161

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Nxllpointer: Also good. A bit shorter

application/src/main/java/org/togetherjava/tjbot/commands/help/UserBannedDeleteRecentThreadsListener.java
Nxllpointer: Hm ok. Unsatisfying but cant help it :/

Nxllpointer (approved):

Zabuzard (approved):

Zabuzard:
feel free to merge 👍

PR #669, opened 2022-11-02
Adding delete button to filesharing response

SquidXTV:
# About
This feature adds a delete button to the filesharing response, so soft mods or the author can delete this gist.

## Examples
Before usage:
image

After usage:
image

Without permission:
image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
@@ -2,16 +2,25 @@ package org.togetherjava.tjbot.commands.filesharing;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.Role;
 import net.dv8tion.jda.api.entities.User;
 import net.dv8tion.jda.api.entities.channel.ChannelType;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
+import net.dv8tion.jda.api.entities.emoji.Emoji;
+import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
+import net.dv8tion.jda.api.events.interaction.component.SelectMenuInteractionEvent;
 import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+import net.dv8tion.jda.api.interactions.components.ActionRow;
 import net.dv8tion.jda.api.interactions.components.buttons.Button;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.togetherjava.tjbot.commands.MessageReceiverAdapter;
+import org.togetherjava.tjbot.commands.UserInteractionType;
+import org.togetherjava.tjbot.commands.UserInteractor;
+import org.togetherjava.tjbot.commands.componentids.ComponentIdGenerator;
+import org.togetherjava.tjbot.commands.componentids.ComponentIdInteractor;
 import org.togetherjava.tjbot.config.Config;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -224,4 +248,74 @@ private boolean isHelpThread(MessageReceivedEvent event) {
...
+        int status = apiResponse.statusCode();
+        if (status == 404) {
+            throw new IllegalStateException("Gist API unexpected response while deleting gist: %s."
+                .formatted(apiResponse.body()));
+        }
+
+        Message message = event.getMessage();
+        List<Button> buttons = message.getButtons();
+        event.editComponents(ActionRow.of(buttons.stream().map(Button::asDisabled).toList()))
+            .queue();
```
Zabuzard, line 308: @Tais993 i might be wrong, but do we need to `deferReply()` in an onButtonClick event? if so, this should be deferred. its doing a blocking request to github api and waits for the result.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -224,4 +248,74 @@ private boolean isHelpThread(MessageReceivedEvent event) {
...
+            Thread.currentThread().interrupt();
+            throw new IllegalStateException(
+                    "Attempting to delete a gist, but the request got interrupted.", e);
+        }
+
+
+        int status = apiResponse.statusCode();
+        if (status == 404) {
+            throw new IllegalStateException("Gist API unexpected response while deleting gist: %s."
+                .formatted(apiResponse.body()));
```
Zabuzard, line 302: im not sure i like the UX in this case. right now it attempts to delete and then crashes with an exception. the user doesnt get any message and will just spam the button again and again. upon which it will constantly attempt again and crash more and more.

maybe it would be better to convert this into a `log.warn(...)`, tell the user that something went wrong and they should ask a moderator to delete it instead and disable the button as well.

Zabuzard: or u just disable the buttons right away, pretend that everything worked, and log the warning afterwards

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: gonna wait for response before changing something

Zabuzard: just checked, yes its needed. so as soon as ur past the basic checks and right before u start the long operations (the api call), u have to do `event.deferReply().queue();` and then, from there on, u must only do interactions through the hook. that is, u cant use `event.editComponents` anymore, but `event.getHook().editComponents(...)` and similar.

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -266,41 +293,18 @@ public UserInteractionType getInteractionType() {
...
-
-        int status = apiResponse.statusCode();
-        if (status == 404) {
-            throw new IllegalStateException("Gist API unexpected response while deleting gist: %s."
-                .formatted(apiResponse.body()));
-        }
+        deleteGist(gistId);
 
         Message message = event.getMessage();
         List<Button> buttons = message.getButtons();
```
Zabuzard, line 310: I think it might be better to disable the buttons right away, before the long api call.

There is no need for it to wait for it to finish. Advantages:
* u dont need the deferReply thing anymore, since u send ur response right away and do the long stuff afterwards
* the user isnt able to click the buttons multiple times while its doing the api call, since the buttons are disabled right away and not 2 seconds later

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
```diff
@@ -249,6 +249,33 @@ private boolean isHelpThread(MessageReceivedEvent event) {
...
+            apiResponse = CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IllegalStateException(
+                    "Attempting to delete a gist, but the request got interrupted.", e);
+        }
+
+
```
Zabuzard, line 271: two newlines is one too many

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
Zabuzard: (actually i have a better idea - see new conversation down below)

application/src/main/java/org/togetherjava/tjbot/commands/filesharing/FileSharingMessageListener.java
SquidXTV: ok I will do this :D

Taz03 (approved):

PR #672, opened 2022-11-05
Improved pagination

Taz03:
- Used emojis in audit command buttons
image

- Improved pagination utilities

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/AuditCommand.java
@@ -3,8 +3,6 @@ package org.togetherjava.tjbot.commands.moderation;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.*;
-import net.dv8tion.jda.api.entities.emoji.Emoji;
-import net.dv8tion.jda.api.entities.emoji.EmojiUnion;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
 import net.dv8tion.jda.api.interactions.callbacks.IReplyCallback;
@@ -26,8 +24,6 @@ import java.time.ZoneOffset;
 import java.util.*;
 import java.util.stream.Collectors;
 
-import static org.togetherjava.tjbot.commands.utils.Pagination.NEXT_BUTTON_EMOJI;
-import static org.togetherjava.tjbot.commands.utils.Pagination.PREVIOUS_BUTTON_EMOJI;
 
 /**
  * This command lists all moderation actions that have been taken against a given user, for example
@@ -79,7 +75,7 @@ public final class AuditCommand extends SlashCommandAdapter {
 
     @Override
     public void onButtonClick(ButtonInteractionEvent event, List<String> args) {
-        long commandUserId = Long.parseLong(args.get(1));
+        long commandUserId = Long.parseLong(args.get(2));
         long buttonUserId = event.getMember().getIdLong();
 
         if (commandUserId != buttonUserId) {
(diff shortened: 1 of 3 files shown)


Nxllpointer:
Stop stop stop. This should be discussed first as these are major changes to the codebase and I do not agree how you want to make pagination easier. Read [Ground Rules](https://github.com/Together-Java/TJ-Bot/wiki/Contributing#ground-rules) again and first create an issue

Zabuzard:
Im missing some nice screenshots that show me the changes in the PR description.

Please put more effort into your PR descriptions in general, thanks.

Taz03:
there are no actual change to the ui/ux other than the emoji buttons in audit command

all i did is moved stuff to util class and used it instead of creating methods in each class

Taz03:
> these are major changes

it is not, i don't think moving common methods to a util class is that much of a major change

Nxllpointer:
> it is not, i don't think moving common methods to a util class is that much of a major change

You litterally have 2 commits named "major changes". This modifies the existing codebase and not just slightly so this should be discussed.

Zabuzard:
> there are no actual change to the ui/ux other than the emoji buttons in audit command

... so u did change something. and id like to see screenshots. please, it will help me.

Taz03:
> You litterally have 2 commits named "major changes". This modifies the existing codebase and not just slightly so this should be discussed.

that is in the context of this pr, first i was only going to do the emoji button thing, but then i also moved methods to util class anyways u can discuss these changes rn its not too late, i have no problem in this pr getting closed its not like i've put months of effort

Nxllpointer (changes requested):
I do not like this approach to "improve pagination". The core problem is still the same, being that pagination stuff has to be rewritten every single time. Due to this everyone does it differently and it makes the feature larger, making the code less readable. I suggest creating a base class for pagination that you can configure and extend to your liking, that handles most annoying stuff by itself, like the command system does, nobody recieves all events and handles them themselves, that is done in the background.

Taz03:
@Zabuzard do u think the same? i feel like this is how it should be y would we extend a class for just a pagination thing, what if u want to extent `ListenerAdapter` and want to use pagination, how u gonna do that then?? @Nxllpointer

Nxllpointer:
@Taz03 You can just create a second class?

Taz03:
just over complicating simple things, anyways this is how we've been doing thins so far and im just extending it. if u want to change things make a separate pr

Nxllpointer:
Ok fine i will create my own framework

Tais993:
> @Zabuzard do u think the same? i feel like this is how it should be y would we extend a class for just a pagination thing, what if u want to extent `ListenerAdapter` and want to use pagination, how u gonna do that then?? @Nxllpointer

I personally disagree with actually extending a class, but having a proper pagination utility which does the pagination just by giving a list of items (and other customizable things) would be a lot better than just creating a class that copies the duplicating methods.

Nxllpointer:
So how would the customization work? Also if you have a base class you can modify the pagination to fit your needs

Taz03:
ok but while someone makes that we can have this, this is better than having methods copy pasted in 2 different classes

Zabuzard:
im holding myself back here cause my current feelings are rather that i personally reject any PR on this topic if it doesnt fit all existing use cases well.

if we want to have pagination utility, it has to be useful on all pagination use cases we currently have. including the ones that have an individual embed per entry and not everything in a single embed.

it has to be more general than all proposals i saw so far. and the more i think about it, the harder it becomes to find an actual utility. cause at the end of the day, the actual pagination part is pretty small. its just adding two buttons with a page number and then simple math to clamp the pages. and maybe an utility to sublist a list of results to that page. but the actual displaying of the page and flow-entry of the button clicks would still be in the mother class.

so if anything, im thinking more of an utility delegate where the mother code looks sth like
```java
PaginationDelegate pagination = new PaginationDelegate();
...

@Override
public void onSlashCommand(...) {
  List<Foo> results = ... // read from database maybe

  pagination.setResults(results);

  ...
  message...addActionRow(pagination.getButtons())...
}

@Override
public void onButtonClick(...) {
  List<Foo> page = pagination.goToPage(event.getButton());
  ...
  message.editButtons(pagination.getButtons())...
}
```
sth along those lines, i suppose

Nxllpointer:
@Zabuzard did you already see my proof of concept?

Nxllpointer:
no its not. you just put a few static methods in a class

Zabuzard:
> @Zabuzard did you already see my proof of concept?

I did not, because your PR description isnt explaining any of it. So it was too much work for me to figure it out on my own for now.

You can help people a lot of putting more effort into your descriptions. Im not expecting that you write detailed PR descriptions like I usually do, but then you have to accept that I will review your PR last and only when I feel like 😬

Nxllpointer:
Uh I think you are mixing things up. I am not the creator of this PR and did not create my own one. My proof of concept was floating around in the discord.

Zabuzard:
Yeah, I'm confused. In any case, I didnt read any pagination solutions yet. Ill rework the bookmark command now and maybe ill implement my pagination idea there

PR #682, opened 2022-11-09
Code-Actions improvements

Zabuzard:
## Motivation

We recently added code-actions with auto-detection on all code messages. The feature is cool, but triggers too often and hence annoys.

This PR attempts to fix most of the problems.

## Overview

The PR essentially adds 3 features:

1. the auto-detection is reduced to
    * only detect messages in help threads
    * only detect code with at least 3 lines of code
2. therefore, add a message context command for helpers to manually register code messages
3. add a delete button

### Auto-detection

The changes on the auto-detection are straightforward. To improve the code a bit, it has been moved out into its own class.

### Command

The main advantages of the context command are:
* allows adding messages outside of help-threads if desired
* allows adding messages in help threads with less than 3 lines, if desired
* allows adding messages without any code-block **(this is huge)**

command
response

### Delete

Further, a delete-button has been added to get rid of the message if it appears albeit not wanted:

delete

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
@@ -1,0 +1,80 @@
+package org.togetherjava.tjbot.commands.code;
+
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.channel.ChannelType;
+import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+
+import org.togetherjava.tjbot.commands.MessageReceiverAdapter;
+import org.togetherjava.tjbot.commands.utils.CodeFence;
+import org.togetherjava.tjbot.commands.utils.MessageUtils;
+import org.togetherjava.tjbot.config.Config;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Automatically detects messages that contain code and registers them at {@link CodeMessageHandler}
+ * for further processing.
+ */
+public final class CodeMessageAutoDetection extends MessageReceiverAdapter {
+    private static final long MINIMUM_LINES_OF_CODE = 3;
+
+    private final CodeMessageHandler codeMessageHandler;
+
+    private final Predicate<String> isStagingChannelName;
+    private final Predicate<String> isOverviewChannelName;
+
+    /**
+     * Creates a new instance.
+     * 
+     * @param config to figure out whether a message is from a help thread
+     * @param codeMessageHandler to register detected code messages at for further handling
+     */
+    public CodeMessageAutoDetection(Config config, CodeMessageHandler codeMessageHandler) {
+        super(Pattern.compile(".*"));
+
+        this.codeMessageHandler = codeMessageHandler;
+
+        isStagingChannelName = Pattern.compile(config.getHelpSystem().getStagingChannelPattern())
+            .asMatchPredicate();
+        isOverviewChannelName = Pattern.compile(config.getHelpSystem().getOverviewChannelPattern())
+            .asMatchPredicate();
+    }
+
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event) {
+        if (event.isWebhookMessage() || event.getAuthor().isBot() || !isHelpThread(event)) {
+            return;
+        }
+
+        Message originalMessage = event.getMessage();
+        String content = originalMessage.getContentRaw();
+
+        Optional<CodeFence> maybeCode = MessageUtils.extractCode(content);
+        if (maybeCode.isEmpty()) {
+            // There is no code in the message, ignore it
+            return;
+        }
+
+        long amountOfCodeLines =
+                maybeCode.orElseThrow().code().lines().limit(MINIMUM_LINES_OF_CODE + 1).count();
+        if (amountOfCodeLines < MINIMUM_LINES_OF_CODE) {
+            return;
+        }
+
+        codeMessageHandler.addAndHandleCodeMessage(event.getMessage());
+    }
+
+    private boolean isHelpThread(MessageReceivedEvent event) {
+        if (event.getChannelType() != ChannelType.GUILD_PUBLIC_THREAD) {
+            return false;
+        }
+
+        ThreadChannel thread = event.getChannel().asThreadChannel();
+        String rootChannelName = thread.getParentChannel().getName();
+        return isStagingChannelName.test(rootChannelName)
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event) {
+        if (event.isWebhookMessage() || event.getAuthor().isBot() || !isHelpThread(event)) {
+            return;
+        }
+
+        Message originalMessage = event.getMessage();
+        String content = originalMessage.getContentRaw();
```
Tais993, line 53:
```suggestion
        String content = event.getMessage().getContentRaw();
```
You only use this variable, at the bottom of the method you use `event.getMessage()` again, I don't see why a variable is needed.

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Optional<CodeFence> maybeCode = MessageUtils.extractCode(content);
+        if (maybeCode.isEmpty()) {
+            // There is no code in the message, ignore it
+            return;
+        }
+
+        long amountOfCodeLines =
+                maybeCode.orElseThrow().code().lines().limit(MINIMUM_LINES_OF_CODE + 1).count();
```
Tais993, line 62:
```suggestion
                maybeCode.orElseThrow().code().lines().limit(MINIMUM_LINES_OF_CODE).count();
```

4 < 3 is always true
So why not if (3 < 3)

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
```diff
@@ -0,0 +1,80 @@
...
+    private boolean isHelpThread(MessageReceivedEvent event) {
+        if (event.getChannelType() != ChannelType.GUILD_PUBLIC_THREAD) {
+            return false;
+        }
+
+        ThreadChannel thread = event.getChannel().asThreadChannel();
+        String rootChannelName = thread.getParentChannel().getName();
+        return isStagingChannelName.test(rootChannelName)
+                || isOverviewChannelName.test(rootChannelName);
+    }
```
Tais993, line 79: This method is on a multitude of places. We might want to add a HelpUtil, doesn't have to be in this PR but feel free to

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageHandler.java
```diff
@@ -146,8 +154,12 @@ private Button createButtonForAction(CodeAction action, long originalMessageId)
     @Override
     public void onButtonClick(ButtonInteractionEvent event, List<String> args) {
         long originalMessageId = Long.parseLong(args.get(0));
-        CodeAction codeAction = getActionOfEvent(event);
+        if (args.size() >= 3 && DELETE_CUE.equals(args.get(2))) {
```
Tais993, line 157: What does this mean?

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: right. originally i wanted to have `<=`, but with `<` ur right, yes

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: yeah saw that as well. its duplicated 3 times...

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageHandler.java
Zabuzard: yeah its a bit annoying.

right now, the first arg is the id. the second arg is the label of the code action. i could potentially put "delete" in there, but delete is not a code action, so that would mess up the logic a bit.

so i thought ill just add a third optional argument that indicates any further stuff for extra buttons, deletion in this case.

a comment is definitely the easiest solution out. im out of better ideas

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: its a bit difficult to properly extract it, since the method is primarly based on two config patterns.

so either it gets them via arg, but then barely provides any real utility (it essentially just tests the two patterns and thats it), or the utility is not static (so that it can read and hold the config itself).

id probably vote for moving this to a different PR, i dont have a good solution right out of the pocket for today

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageHandler.java
Zabuzard: comment ✔️

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: done ✔️

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Zabuzard: done ✔️

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageAutoDetection.java
Tais993: Sure

application/src/main/java/org/togetherjava/tjbot/commands/code/CodeMessageHandler.java
Tais993: If we have more non-code functionalities it can become a switch which would be a lot more readable. This is fine for now!

Tais993 (approved):

Zabuzard:
Speeding this up a bit since its annoying people lol. Tais review is enough, cheers

PR #684, opened 2022-11-09
Some tweaks to the helper prune system

Zabuzard:
We often get this warning:

warning

Seems that our current settigns are slightly above the sweetspot. This PR adjusts the settings a bit to hopefully reduce this warning.

* pruning earlier, but therefore less

should hopefully lead to being able to prune the desired amount of people each time, getting less warnings.


Zabuzard:
Not that anyone has to be really interested in this, merging

PR #688, opened 2022-11-14
Added suggestion command

### LINKED ISSUE #408: Suggestion command
## See the "Suggestion command" thread for more info

Vote manipulation by blocking people, so they can't vote (react). Or by blocking the bot, so the bot can't react on your messages
Also, low effort suggestions, and possible "preferences" because a well known person made the suggestion.

A command, which creates the suggestion for you based on a filled in "form"

Example layout might be:

> Short description of suggestion:
> ....
> Motivation
> ....
> Drawbacks
> ....
> Benefits
> .... 
> Additional context
> ....

Based on this information, the bot creates an readable embed in an effort to improve the quality of the suggestion.

Whenever the bot mentions the original author of the suggestion can be discussed, not sure or we should / whatever.

**Additional context**
Blocked be #364 (#363) and #382 (#368)

Taz03:
closes #408 
- Added suggest command for suggestions
- Added a stick command to keep a message at the bottom of the channel
- Added unstick command to removed the sticked message from a channel

### Demo
https://user-images.githubusercontent.com/73871477/201829164-45ee5871-199c-4211-9b85-8cca17cb0be1.mp4


Nxllpointer:
Sooo before i take a look at the code, your adding a new feature (sticky messages) that is not really related to the sugestion thing. But why? One can just delete the messages sent by users in the suggestion channel and send a DM or just outright disabling sending messages in the channel. I can see you implementing this to hint other users how to suggest something but i think a better solution would be to 
A: Write it into the channel description (which noone reads because who reads those lets be real)
B: Add the info to the embed footers
and C, my suggestion: Add a big green button below the latest suggestion to open the modal

Taz03:
I feel like adding the text "Use /suggest command" to the embed footer will be a good idea, but the idea to keep the message at the bottom of the channel is suggested by @Tais993, whats ur thoughts on this?

Taz03:
if anyone want they can take over this pr

PR #696, opened 2022-11-21
Fixed a small OOB bug with extracting code

Zabuzard:
Fix for this bug:

bug

Happens in rare conditions when the message has more than three backticks before the language. Such as:

issue


Zabuzard:
7 days. merging

PR #703, opened 2022-11-30
Fixed bug with linkpreview OG twitter broken

Zabuzard:
Fixed a small syntax issue in the JSoup query selection for the OpenGraph Twitter title property.

Unfortunately, JSoup didnt just crash and gave us a proper exception, but instead searched for the element forever...

This is the reason we did a rollback on the latest release. Once its merged, we can rerelease.


Tais993 (approved):
skill issue

PR #705, opened 2022-11-30
Migrated help system to forums

### LINKED ISSUE #585: Migrate help system to forums
JDA released forum support with **v5.0.0-alpha.20**:

https://github.com/DV8FromTheWorld/JDA/releases/tag/v5.0.0-alpha.20

https://github.com/DV8FromTheWorld/JDA/pull/2184/files

Its time to move our help system to forums.

Most of the functionality can remain as it is, since posts are still just threads. What has to be adjusted is mostly:
* channel localization (its a `ForumChannel` now, not a `TextChannel`)
* the `overviewChannel` stays, but questions will not be listed under it anymore, but instead move into the forum channel. the `stagingChannel` itself wont exist in its current form anymore (it will become the forum)
  * the `ImplicitAskListener` functionality can be removed
* thread activity should be indicated by tags, not thread titles
* helper categories should be indicated by tags, not thread titles

Zabuzard:
## Overview

Implements and closes #585.

Migrates our help system to a Discord forum channel.

meme

## Details

Most of the migration is actually fairly simple and straightforward, because forum posts are still just regular `ThreadChannel`s. So most of our code works out of the box.

The first change that had to be done is to adjust the config to go from 2 channels to just one `"helpForumPattern": "question"`. Obviously, this yields a couple of minor renaming changes in many classes that observe messages in threads of this channel (e.g. TopHelper, Bookmarks, ...).

In terms of the actual logic, the only real change are the semantics of thread **categories** and **activity**. In the past, they have been attached to a thread in its title, e.g. `🔻 [Java] How to write Hello World?`. Forums actually have dedicated **tags** which can be used for that. So we moved the categories and activity into forum-tags.

The second bigger change was that we do not have a controlled entry-point into the logic anymore (no `/ask` command or implicit ask listener). So we had to move some logic (explanation message, ...) into an event handler.

overview

single thread

### Removed features

Forums bring a lot of functionality out of the box, making the following features obsolete:
* `/ask`
* `ImplicitAskListener`
* `BotMessageCleanup` (since we do not have a staging channel with implicit posts anymore)
* validation for valid titles and title extraction
* question overview `HelpThreadOverviewUpdater`

### Tags

Forum tags are fairly straightforward. We simply add a tag for each category and for each activity:

tag settings

The code can read the applied tags and find/apply them by name, easy. The code changes for this are all in `HelpSystemHelper`, where we could remove all the renaming/title extraction logic and replace it by using forum tags.

Discord actually has a limit of 5 tags per thread and also allows the user to select multiple. This creates a couple of edge cases that we have to handle. For example, the user can now select multiple categories, i.e. `Java, Frameworks, Spring, JavaFX/Swing`.

We probably dont want to ping all of them, so the system will only look at the first selected category and identify as primary tag, only pinging those helpers. The `/help-thread change category` command has to correctly respect that, deleting the first tag, possibly deleting the desired tag if its somewhere at the end of the list, and then prepend it (see `HelpSystemHelper#changeChannelCategory`). The activity tags are handled in the same way.

If the user already selected 5 tags (the max) and the bot then wants to add the activity tag, it will delete the last tag in the list first.

### Entry point

Since we do not have `/ask` and the `ImplicitAskListener` anymore, we created `HelpThreadCreatedListener`, reacting to the event of threads being created in the forum channel.

Essentially, we just copy pasted the logic from `/ask` over to here, sending the explanation message and pinging helpers.

The text had some minor adjustments, due to the change of the flow. For example, we do not have to write "`@John` has a question and will send the details now" anymore, since the user created the thread themselves (no attention-ping needed) and they already entered their details, since Discord forces writing a description.

### Config

Replaced `stagingChannelPattern` and `overviewChannelPattern` by
```json
"helpForumPattern": "questions"
```

### Discord

1. Create a forum called `questions` (according to the config).
2. Give it a tag for each of the `categories`, selectable by normal users.
3. Give it tags for the activities, they should only be selectable by mods. Names:
  * `Active`
  * `Needs attention`
  * `Nobody helped yet`
4. Require that users select a tag when posting

tag settings

activity tag setting

## Checklist

* [x] config
* [x] migrate misc users (bookmarks, top helper, ...)
* [x] categories based on tags
* [x] activity based on tags
* [x] AskCommand -> on channel created listener
* [x] multiple selected tags, ping all?
* [x] polish


java-coding-prodigy:
If people make the forums themselves? What's the use of the bot then?
Is it just going to update the tags when answers are given and give the asker advice?

application/src/main/java/org/togetherjava/tjbot/commands/bookmarks/BookmarksSystem.java
```diff
@@ -39,7 +39,7 @@ public final class BookmarksSystem {
     static final Color COLOR_FAILURE = new Color(238, 153, 160);
 
     private final Database database;
-    private final Predicate<String> isOverviewChannelName;
+    private final Predicate<String> isHelpForumName;
```
isvora, line 42: Usually when variables are named `isXXX` it means the variable is a boolean. Should we maybe change the naming here to make it more clear it's not?

Tais993: What would you advise then? Reason this is named like this is because a Predicate returns a boolean.
The predicate checks or it "is the help forum name", which is why we use this naming

Zabuzard:
> If people make the forums themselves, what's the use of the bot then?

* explanation message
* soft-ping corresponding helpers
* activity indicators
* auto-close after 12h
* probably some other stuff i forgot

application/src/main/java/org/togetherjava/tjbot/commands/bookmarks/BookmarksSystem.java
Zabuzard: A predicate essentially represents a method. You would also call the method `isFoo()`, so it makes sense (and afaik is idiomatic) to also name the predicate like this.

`isFoo.test(x)`

Zabuzard:
(just rebased)

PR #706, opened 2022-12-01
Fixed bugs with link preview

Zabuzard:
## Overview

This fixes a couple of bugs with link previews.

Major priority, because the feature is currently live and many tags dont work anymore due to it.

## Fixed bugs

1. Sometimes included trailing punctation, like `http://www.example.com/,` or `http://www.example.com/.`
2. Did not respect escaped URLs, like `<http://www.example.com/>`
3. The link detector has a lot of false positives, such as `System.in` being detected as `http://System.in/`. Now filters out everything that is not `http` on the original URL.
4. Some websites use `<meta name="...">` instead of `<meta property="...">`
5. When an individual query failed with an exception, it crashed the whole query instead of being skipped. Solved by using `.exceptionally(...)` with a proper `log.warn` on each task individually.
6. When the full query failed with an exception, it lead to not sending the actual `/tag` anymore (stuck forever). Solved by using `.exceptionally(...)` with a proper `log.error` on the surrounding all-task
7. Forgot a proper timeout for bad websites. Now is **10 seconds** per website.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/utils/LinkPreviews.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/utils/LinkPreviews.java
@@ -52,20 +52,31 @@ public final class LinkPreviews {
     public static List<String> extractLinks(String content) {
         return new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect()
             .stream()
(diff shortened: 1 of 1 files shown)


Zabuzard:
gonna ship this today. merging with one approval

PR #391, opened 2022-02-20
Utility for in-app URL's

Tais993:
This implements a class which eases the developer to generate URL's to items within the Discord app.

A feature should **NOT** rely on this class, Discord can change something anytime, and iOS + Android aren't supported. They're there to enhance the functionality without making it worse.

If anyone has ideas for screens I could look at to find or there's URL(s) for, you can always send a message and I'll take a look or I can find anything.

Closes #387

PR #709, opened 2022-12-02
Fix link for report's 'go to message' button

Taz03:
The current weird link doesn't seem to work on linux systems.

This is a simple link and used more frequently and works on linux systems too

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
@@ -87,6 +87,7 @@ public final class ReportCommand extends BotCommandAdapter implements MessageCon
 
         String reportedMessage = event.getTarget().getContentRaw();
         String reportedMessageID = event.getTarget().getId();
+        String reportedMessageJumpUrl = event.getTarget().getJumpUrl();
         String reportedMessageChannel = event.getTarget().getChannel().getId();
         String reportedAuthorName = event.getTarget().getAuthor().getName();
         String reportedAuthorAvatarURL = event.getTarget().getAuthor().getAvatarUrl();
@@ -100,8 +101,8 @@ public final class ReportCommand extends BotCommandAdapter implements MessageCon
             .build();
 
         String reportModalComponentID = generateComponentId(reportedMessage, reportedMessageID,
-                reportedMessageChannel, reportedMessageTimestamp, reportedAuthorName,
-                reportedAuthorAvatarURL, reportedAuthorID);
+                reportedMessageJumpUrl, reportedMessageChannel, reportedMessageTimestamp,
+                reportedAuthorName, reportedAuthorAvatarURL, reportedAuthorID);
         Modal reportModal = Modal.create(reportModalComponentID, "Report this to a moderator")
             .addActionRow(modalTextInput)
             .build();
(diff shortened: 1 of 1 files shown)


marko-radosavljevic:
A bit of a background on this PR.

Link buttons with `discord:\\` URI scheme/protocol do not work on Linux (confirmed on multiple systems). Noticed on report's 'go to message' button, which is critical for my ability to effectively moderate.

Not sure if those ever worked.

* * *

This PR looks fine, but it does intentionally sidestep our own utility for handling URIs, which should be noted. For example, all link buttons generated with it have the same issue.

Which raises the question, what's the appropriate way to deal with this?

Discord URI scheme doesn't seem to be documented at all? How can other systems even support it? How can we even rely on it? How can I learn about it? Am I missing something? :weary: 

`DiscordClientAction` isn't that well documented either, and probably outdated:
> A few notes;
>- and Android are NOT supported
>- It opens the LAST installed Discord version (Discord, Canary, PTB)

It works on my android, so I guess that's not true anymore.
In my confusion, and in pursuit of answers, I stumbled upon the original PR #391 that states:

> A feature should **NOT** rely on this class, Discord can change something anytime, and iOS + Android aren't supported. They're there to enhance the functionality without making it worse.

So I guess we should offer regular links whenever possible. For example, regular message link for report's 'go to message', and mention for 'author profile'.

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
```diff
@@ -185,9 +183,10 @@ private MessageCreateAction createModMessage(String reportReason,
 
         MessageCreateAction message =
                 modMailAuditLog.sendMessageEmbeds(reportedMessageEmbed, reportReasonEmbed)
-                    .addActionRow(DiscordClientAction.Channels.GUILD_CHANNEL_MESSAGE.asLinkButton(
-                            "Go to Message", guild.getId(), reportedMessage.channelID,
-                            reportedMessage.id));
+                    .addActionRow(Button.link(
+                            "https://discord.com/channels/%s/%s/%s".formatted(guild.getId(),
```
Zabuzard, line 187: dont hardcode magic links in the middle of the code like that. thats bad, fragile and doesnt scale at all. if anything, at least create a proper helper method in a helper class. For example somewhere at `DiscordUtil` or whatever the class was called.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
```diff
@@ -183,20 +177,18 @@ private MessageCreateAction createModMessage(String reportReason,
...
-        moderatorRole.ifPresent(role -> message.setContent(role.getAsMention()));
-
-        return message;
+        return guild.getTextChannelById(reportedMessage.channelID)
+            .retrieveMessageById(reportedMessage.id)
+            .map(Message::getJumpUrl)
+            .flatMap(messageUrl -> modMailAuditLog
+                .sendMessageEmbeds(reportedMessageEmbed, reportReasonEmbed)
+                .setContent(moderatorRole.map(Role::getAsMention).orElse(""))
+                .addActionRow(Button.link(messageUrl, "Go to Message")));
```
Zabuzard, line 191: When the message cant be found (retrieveMessageById fails), the whole thing fails and doesnt send the report anymore. Thats bad. Instead, it should maybe just leave away the button in that case.

Please test your edge cases more careful.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/moderation/ReportCommand.java
Taz03: if that'll happen the report message won't be filed, it is handled already

Zabuzard: > if that'll happen the report message won't be filed, it is handled already

but it should, thats my point. we want to have the report, even if the message ID cant be found anymore.

the old code also handled that gracefully by just having the button point to nowhere.

with ur code, the whole command just silently goes down at this stage.

Zabuzard: like here:

modal
extraction

you could just have the jump URL there. then u dont need that retrieve anymore and all those problems go away

Taz03: done

Zabuzard (approved):

marko-radosavljevic (approved):
Looks good. :relaxed: :heart:

PR #726, opened 2022-12-14
Optimized help thread activity updater

Taz03:
Currently the system uses some complex methods to apply the activity, but all we need is already built into jda. This pr replaces the complex methods with JDA methods

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
@@ -2,23 +2,16 @@ package org.togetherjava.tjbot.commands.help;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
-import net.dv8tion.jda.api.entities.Message;
-import net.dv8tion.jda.api.entities.User;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
-import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
-import net.dv8tion.jda.api.requests.RestAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.commands.Routine;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 /**
  * Routine that periodically checks all help threads and updates their activity based on heuristics.
@@ -28,8 +21,7 @@ import java.util.stream.Collectors;
  */
 public final class HelpThreadActivityUpdater implements Routine {
     private static final Logger logger = LoggerFactory.getLogger(HelpThreadActivityUpdater.class);
-    private static final int SCHEDULE_MINUTES = 30;
-    private static final int ACTIVITY_DETERMINE_MESSAGE_LIMIT = 11;
+    private static final int SCHEDULE_MINUTES = 1;
 
     private final HelpSystemHelper helper;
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
```diff
@@ -28,8 +21,7 @@
  */
 public final class HelpThreadActivityUpdater implements Routine {
     private static final Logger logger = LoggerFactory.getLogger(HelpThreadActivityUpdater.class);
-    private static final int SCHEDULE_MINUTES = 30;
-    private static final int ACTIVITY_DETERMINE_MESSAGE_LIMIT = 11;
+    private static final int SCHEDULE_MINUTES = 1;
```
Zabuzard, line 24: u reduced the number here. did u check the rate limits on changing tags? previously, where it was actually renaming channels, this was subject to heavy rate limits that we must not hit, hence 30mins

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
```diff
@@ -71,33 +63,18 @@ private void updateActivityForGuild(Guild guild) {
...
+            return HelpSystemHelper.ThreadActivity.LOW;
+        }
+
+        if (threadChannel.getMessageCount() > 15) {
+            return HelpSystemHelper.ThreadActivity.HIGH;
+        }
 
-    private static boolean isBotMessage(Message message) {
-        return message.getAuthor().equals(message.getJDA().getSelfUser());
+        return HelpSystemHelper.ThreadActivity.MEDIUM;
```
Zabuzard, line 78: Interesting. How does this behave with threads that are not "cached"? For example if you take the bot offline, then create a thread and start the bot afterwards. Will it be able to pick the correct numbers up, without queuing an action?

(@Tais993)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Taz03: sorry i did that to test, 30 mins was too long to wait. I'll change it back

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Taz03: > For example if you take the bot offline, then create a thread and start the bot afterwards
i tested that and it worked

> Will it be able to pick the correct numbers up, without queuing an action?
no idea

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Zabuzard: after talking with @Tais993 we agreed on 15 minutes. can u change it to 15mins please? thanks :)

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Zabuzard: okay. ill wait for a confirmation by @Tais993 that this approach works. then im good with it :)

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Taz03: done

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Tais993: Iirc it works, message count is included by Discord

Taz03: it works i tested, but ig what zabu wants to ask is "without queuing an acting?" whatever that means

Zabuzard: im just surprised this method comes along without a `RestAction`. so i suppose it works with the cache. but i could imagine cases where the cache doesnt know the thread yet. so its important to understand whether it would work in these cases (@Tais993 ?)

Taz03: i mean u r sending the thread object so it should work without any issue

Tais993: If the cache doesn't know about the thread, this wouldn't run to begin with, and that has always been the same.
So no this shouldn't cause issues, to make sure, we can test it in the develop server.

But from my knowledge, this shouldn't cause any new issues we didn't have before.

Zabuzard: if u say so. to me that isnt necessarily super obvious. just because we have a cached thread, doesnt necessarily mean that it also knows about all its message history. hence why im careful.
but alright

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/commands/help/HelpThreadActivityUpdater.java
Tais993: Well, it doesn't know about the history, it only knows that it contains ... Messages (if above 50, it gives 50)

Tais993 (changes requested):
You mention in the description that this replaces the code to use JDA methods.

While in actuality the behaviour changes, think that's an important notice to make you didn't mentioned.


I don't think that the member count matters, as people get automatically added when you use a tag, so like, thread activity will never be low.

On thread creation, assuming a tag is used, a total of 4 message get send already.
So if there a convo exists of 11 messages, you read the 16 message count for high thread activitity.


So now, 2 days later, activity is still on high, but 0 messages got send?


I think this implementation (and the previous) both lack something very important, neither define the actual "activity"

PR #731, opened 2022-12-15
Prefer uncommon categories first

### LINKED ISSUE #725: Order help categories, uncommon first
## Problem

Often users tend to select `Java` as the first category, just because it has to do with Java.

This leads to the Java helpers often being pinged, despite the question is about something more specific. And the actual helpers for this topic not being alerted at all.

## Solution

The bot should ignore the tag-order given by the user and sort the tags by how common they are, putting the most specific and uncommon tag first:

example

From there on, it should continue with its existing logic to select the first tag and ping those helpers. So in above example, it should:
* reorder to put `Build Tools` before `Java`
* ping `Build Tools` helpers

The order can be taken from the order they are listed in the config (the config, as of today, lists them with common first).

So we can just take the order of the config list and sort by that, reversed.

Zabuzard:
## Overview

Implements and closes #725.

Fixes the problem that often users tend to select Java as the first category, just because it has to do with Java.

This leads to the Java helpers often being pinged, despite the question is about something more specific. And the actual helpers for this topic not being alerted at all.

## Solution

The categories are now selected in reverse-order of the config. Assuming that the config lists them in commonness descending (in particular, Java first).

Now, when an user picks multiple categories:

example

It will sort the categories by commonness ascending (uncommon first) and pick the first, i.e. the most uncommon, pinging the most specific category:

pings correct

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/help/HelpSystemHelper.java
@@ -32,9 +32,11 @@ import java.awt.Color;
 import java.io.InputStream;
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Helper class offering certain methods used by the help system.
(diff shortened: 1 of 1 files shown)


Taz03 (approved):

Zabuzard:
merging today, so that it makes it into the last release for this year, tomorrow :)

PR #740, opened 2023-01-04
Fixed leave event listener

### LINKED ISSUE #619: OnGuildLeaveCloseThreadListener warnings on unknown threads
Sometimes the `OnGuildLeaveCloseThreadListener` logs a warning that it can not find certain threads:

stacktrace

How is this possible? Were the threads deleted, by whom? Are they just closed/archived and JDA doesnt find them anymore? Why does this happen so often?

This is not supposed to happen, the reason should be found and then either fixed or this log has to be adjusted to debug level, if this is an expected event.

Taz03:
2023-01-04_201747-848x1031_scrot

closes #619

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/commands/Features.java
@@ -106,7 +106,7 @@ public class Features {
 
         // Event receivers
         features.add(new RejoinModerationRoleListener(actionsStore, config));
-        features.add(new OnGuildLeaveCloseThreadListener(database));
+        features.add(new OnGuildLeaveCloseThreadListener(config));
         features.add(new UserBannedDeleteRecentThreadsListener(database));
         features.add(new LeftoverBookmarksListener(bookmarksSystem));
         features.add(new HelpThreadCreatedListener(helpSystemHelper));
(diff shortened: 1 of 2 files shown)


Zabuzard:
what is the actual diff? since you renamed, github doesnt show a proper diff... so its hard to review.

Taz03:
> what is the actual diff? since you renamed, github doesnt show a proper diff... so its hard to review.

previously it used db to store authors and thread id, which is good we have all the threads in storage, but on jda side it used cache to search for the threads, which is not ideal coz not all threads are always cached, hence the errors.

now i use retrieveActiveThreads method to get the threads which is not cache based and then i filter it and close the ones made by leaver.

Zabuzard (changes requested):
works, but is there a way to retrieve active threads from a channel? i would prefer an approach like `helpForumChannel.retrieveActiveThreads()...` to avoid retrieving **all** and then matching the pattern afterwards.

Taz03:
> works, but is there a way to retrieve active threads from a channel? i would prefer an approach like `helpForumChannel.retrieveActiveThreads()...` to avoid retrieving **all** and then matching the pattern afterwards.

actually there is a method but it is a get method not a retrieve method https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/entities/channel/attribute/IThreadContainer.html#getThreadChannels()

Zabuzard:
Probably via cache again, since it doesnt return a RestAction, ...?

Taz03:
> Probably via cache again, since it doesnt return a RestAction, ...?

yea most probably its cache, tho the description doesn't talk about that

Zabuzard (approved):

marko-radosavljevic (approved):

PR #757, opened 2023-01-27
Rename packages (esp. commands -> features)

### LINKED ISSUE #418: `commands` package should be named `features`

Zabuzard:
The long awaited PR to rename/reorganize packages properly. No code changes have been done, but due to the file moving, this is a **breaking change**.

All current work (fortunately, almost none), has to do some refactoring when this is merged.

Changes:
* renamed `commands` -> `features`
* got rid of extra packages like `routines` and moved their contents to the topic-package (`features.moderation`) instead
* moved the `/audit` related code (3 files) into dedicated package `features.moderation.audit`

Implements and closes #418.


Taz03 (approved):

Zabuzard:
kicking this out fast to reduce likelyhood of new PRs conflicting with it

PR #759, opened 2023-01-30
Educate about /foo when ?!.foo

### LINKED ISSUE #654: React on ?foo !foo .foo to educate about slash commands
Sometimes people with experience from other bots attempt to use message-commands, such as `?foo`, `!foo` or `.foo`.

We should create a small message listener that reacts to that and educates the user about the fact that we only use slash commands and also how a slash command can be used, little tutorial with a picture maybe.

(to include pictures, see `HelpSystemHelper#108` which includes `CODE_SYNTAX_EXAMPLE_PATH` as picture in an embed).

Zabuzard:
## Overview

Implements and closes #654.

Adds a slash command educator that sends an advice when it sees someone posting a message command `?foo`, `!foo`, `.foo`, telling them how to use slash commands `/foo` instead. Looks like this:

example

## Details

The actual pattern used to match messages is:

```java
[.!?] // Start of message command
[a-zA-Z]{2,15} // Name of message command, e.g. 'close'
.*
```

So it will not react on stuff like `??` or `? whats this`.

## Other

Also added **unit tests**. And got rid of a bigger code duplication with `MessageReceiverAdapter` where nearly every implementation wrote `super(Pattern.compile(".*"));`.


Zabuzard:
1 week, lets go

PR #763, opened 2023-02-03
Top helpers not askers

### LINKED ISSUE #762: Exclude OP (askers) from TopHelpersMessageListener
# About
Should be a good idea to not add messages to the database if its the asker that writes in his question.
Implementation would just be an if statement checking if a user is OP in a forum post.

Taz03:
closes #762

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/tophelper/TopHelpersMessageListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/tophelper/TopHelpersMessageListener.java
@@ -83,7 +83,7 @@ public final class TopHelpersMessageListener extends MessageReceiverAdapter {
         return isHelpForumName.test(rootChannelName);
     }
 
-    boolean isSentByOp(MessageReceivedEvent event) {
+    private boolean isSentByOp(MessageReceivedEvent event) {
         return event.getChannel().asThreadChannel().getOwnerId().equals(event.getAuthor().getId());
     }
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/tophelper/TopHelpersMessageListener.java
```diff
@@ -83,6 +83,14 @@ boolean isHelpThread(MessageChannelUnion channel) {
         return isHelpForumName.test(rootChannelName);
     }
 
+    boolean isSentByOp(MessageReceivedEvent event) {
```
Zabuzard, line 86: dont think its meaningful to expose this for now. id go for private

Zabuzard (changes requested):

Zabuzard:
TopHelper unit tests crash, because you introduced new methods into the flow that havent been mocked yet. Do you want to fix this urself or should someone else take that part over?

Zabuzard:
Has to be done in the `JdaTestHelper` main class or whatever its called

Taz03:
> TopHelper unit tests crash, because you introduced new methods into the flow that havent been mocked yet. Do you want to fix this urself or should someone else take that part over?

i haven't tried yet, i'll like to try that myself first, will update on it

Taz03:
@Zabuzard for some reason when i use shorthand method it passes the test, is it ok?

Zabuzard:
> 

i suppose that method is mocked already - lol!

fair enough, i guess

Zabuzard (approved):

SquidXTV (approved):

PR #773, opened 2023-02-16
Fix website bug with widget on mobile

### LINKED ISSUE #721: Website, deactivate Discord widget on mobile
On our website, the Discord widget does not properly work on mobile:

Doesnt look good:
first

Goes beyond the size:
second

It even deactivates the animations of parts that come after:
third

Google Search Console also detects flaws with it:
gsc

All in all, we should simply deactivate the whole widget on mobile (but keep it on Desktop).

AthithyaIthayan:
Closes and fixes #721, the problem was that the size for the widget was hardcoded. Now its made relative `100%` and by that scales properly.

### DIFF
--- a/website/src/main/resources/public/index.html
+++ b/website/src/main/resources/public/index.html
@@ -74,7 +74,7 @@
                 Lots of friendly volunteers to help you learn Java or chat about everyday
                 things.</small></p>
             <ul>
-                <li>24.000 members</li>
+                <li>25.000 members</li>
                 <li>300 questions answered weekly</li>
                 <li>50.000 messages per month</li>
                 <li>monthly giveaways</li>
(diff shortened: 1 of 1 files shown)


website/src/main/resources/public/index.html
```diff
@@ -74,16 +74,16 @@ <h1><span class="image right"><img style="width: 50%" src="images/logo_frame.png
                 Lots of friendly volunteers to help you learn Java or chat about everyday
                 things.</small></p>
             <ul>
-                <li>20.000 members</li>
+                <li>24.000 members</li>
```
Taz03, line 77: fix the number as zabu said

Taz03 (changes requested):

website/src/main/resources/public/index.html
AthithyaIthayan: Updated it Rn

Zabuzard (approved):

Taz03 (approved):

PR #779, opened 2023-02-20
Handling /reminder failure cases 

### LINKED ISSUE #722: /remind should retry and log after failure
## Problem

Our reminder system (`/reminder create ...`) sometimes fails to send the reminder. This is totally normal and can happen due to Discord/JDA hicups and whatever.

But, when it happens, we completely lose the message and it does not retry. That is a problem.

## Fix

Instead, the system should just retry failed reminders (maybe 3 times), by putting them into the DB again with an `Instant` that is maybe set 1 minute into the future. If it ultimately failed, the log message should log the actual reminder-text and the target user, so that the reminder can be rerouted manually by an admin at least.

## Details

For the failure count (~3 times), we likely need to add an additional column in the reminder DB table. Reminders are send from `RemindRoutine.java`, the table was initially created in `V8__Add_Pending_Reminders.sql`, the migration script `V10__Alter_Top_Helper_Message_Length.sql` shows how one can add a new column to an existing table.

ankitsmt211:
Added a file to alter `PENDING_REMINDER_TABLE` to add another column `REMINDER_FAILURE_COUNTER` which keeps track of attempts made when `/reminder` fails. 

Also modified `RemindRoutine` class to handle failures

Closes and fixes #722


ankitsmt211:
what's happening, is it something wrong with code? not familiar with these automated checks

Taz03:
> what's happening, is it something wrong with code? not familiar with these automated checks

u can click on details to know what each are about

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -1,5 +1,6 @@
 package org.togetherjava.tjbot.features.reminder;
 
+import com.sun.source.tree.OpensTree;
```
Zabuzard, line 3: whats that import? looks odd

application/src/main/resources/db/V13__Add_Reminder_Failure_Counter.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pending_reminders ADD reminder_failure_counter INTEGER DEFAULT 0;
```
Zabuzard, line 1: * its clear from the context that its a reminder-thing, so id rename it to just `failure_counter`. or, to make it clearer what it actually is, `failed_attempts`
* its missing a `NOT NULL` at the end

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -55,25 +58,27 @@ public Schedule createSchedule() {
...
+                            pendingReminder.getAuthorId(), pendingReminder.getContent(),
+                            pendingReminder.getCreatedAt(), pendingReminder.getGuildId());
+
+                    pendingReminder.delete();
+                }));
     }
 
-    private static void sendReminder(JDA jda, long id, long channelId, long authorId,
-            CharSequence content, TemporalAccessor createdAt) {
+    private void sendReminder(JDA jda, int id, long channelId, long authorId,
```
Zabuzard, line 72: why did u change the id to `int`? isnt it a `long`? what does `pendingReminder.getId()` return?

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                            user has disabled DMs or has been deleted."""+ "contents: "+content,
+                    authorId);
+        }
+    }
+
+        private Integer getReminderFailureCounter(int id){
+            return database.read(context -> context.selectFrom(PENDING_REMINDERS).where(PENDING_REMINDERS.ID.eq(id))
+                    .stream().map(PendingRemindersRecord::getReminderFailureCounter).findFirst().get()
+            );
+        }
```
Zabuzard, line 178: it is unecessary that u read the database again. u can get hands on the failure counter where u read it originally, here:

original

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -85,43 +90,42 @@ private static RestAction<ReminderRoute> computeReminderRoute(JDA jda, long chan
...
-        Consumer<Throwable> logFailure = failure -> logger.warn(
-                """
-                        Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue, \
-                        but also happen if the bot disconnected from the target guild and the \
-                        user has disabled DMs or has been deleted.""",
-                id);
+        Consumer<Throwable> logFailure = failure -> {
+            //
+            reminderFailureAction(id, channelId, authorId, content, guildId, reminderFailureCount);
+        };
```
Zabuzard, line 113: the name `logFailure` is kinda incorrect now. that said, i would just put the lambda without defining it in advance now. so `.queue(doNothing(), failure -> reminderFailureAction(...))`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
                             + " the original channel you wanted it to be send to)");
         }
     }
-}
+
+    private void reminderFailureAction(int id, long channelId, long authorId,
```
Zabuzard, line 150: the name is kinda bad. what about `attemptRetryReminder`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                                       CharSequence content, long guildId, int reminderFailureCounter) {
+        if (reminderFailureCounter <= 3) {
```
Zabuzard, line 152: 3 is a magic number. put it into a well named constant `MAX_RETRY_ATTEMPTS` instead

Zabuzard, line 152: prefer to not do:
```java
if (...) {
  some long code
} else {
  sth short
}
```
instead, use early-return:
```java
if (!...) {
  sth short
  return
}

sth long
```
that way, u have no nesting for the long part and its more readable

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+            Instant current = Instant.now();
```
Zabuzard, line 153: instead of putting current here and computing the remindAt at the bottom, compute it up here. `Instant remindAt = Instant.now().plusMinutes(1)`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+            database.write(context -> context.newRecord(PENDING_REMINDERS)
+                    .setId(id)
+                    .setCreatedAt(Instant.now())
```
Zabuzard, line 156: please use the original "createdAt", not a new one

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +146,35 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                    .setContent(content.toString())
+                    .setReminderFailureCounter(reminderFailureCounter + 1)
+                    .insert());
+        } else {
+            logger.warn(
+                    """
+                            Failed to send a reminder with (authorID '{}'), skipping it. This can be due to a network issue, \
+                            but also happen if the bot disconnected from the target guild and the \
+                            user has disabled DMs or has been deleted."""+ "contents: "+content,
+                    authorId);
```
Zabuzard, line 170: dont do string concatenation on a logger. instead use its string interpolation mechanism with `{}`.

```java
logger.warn("Failed to send a reminder after {} retry attempts (id '{}', from author '{}'), skipping it. ... The content was: {}",
  id, authorId, MAX_RETRY_ATTEMPTS, content)
```

Zabuzard (changes requested):

Zabuzard:
@ankitsmt211 the issues are mostly related to your code style. u seem to have bypassed the pre-commit hook when u commited and by that skipped **spotless**. either setup ur pre-commit hooks or run `gradle spotless:apply`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: It wasn't supposed to be there, since the imports were minimized it slipped through. Fixed it after installing sonarLint plugin.

application/src/main/resources/db/V13__Add_Reminder_Failure_Counter.sql
ankitsmt211: done

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: resolved this as well

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: i had a feeling that was incorrect way of doing it, makes it hard to look for it and change in future.

ankitsmt211: done

Zabuzard: its usually easier for people to adjust constants when they are in the expected place (at the beginning of the file) and not in the middle of the code flow

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: image
The table's ID field was Integer, so i think that's why i changed it.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -85,43 +90,42 @@ private static RestAction<ReminderRoute> computeReminderRoute(JDA jda, long chan
...
+                .setContent(route.description());
 
-        Consumer<Throwable> logFailure = failure -> logger.warn(
-                """
-                        Failed to send a reminder (id '{}'), skipping it. This can be due to a network issue, \
-                        but also happen if the bot disconnected from the target guild and the \
-                        user has disabled DMs or has been deleted.""",
-                id);
+        Consumer<Throwable> logFailure = failure -> {
+            //
```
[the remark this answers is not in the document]
ankitsmt211: fixed it

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: fixed it

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: just saw this and updated to `attemptRetryReminder`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: ah. please note that sqlites `INTEGER` type is also for `long`. its a variable size integer :)

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -147,34 +134,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
-                    .setGuildId(guildId)
-                    .setChannelId(channelId)
-                    .setAuthorId(authorId)
-                    .setRemindAt(current.plusSeconds(60))
-                    .setContent(content.toString())
-                    .setReminderFailureCounter(reminderFailureCounter + 1)
-                    .insert());
-        } else {
+    private void attemptRetryReminder(Reminder reminder) {
+        if (!(reminder.failureAttempts() <= MAX_FAILURE_RETRY)) {
```
Zabuzard, line 138: kinda complex to read. instead do `foo > bar`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -147,34 +134,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                    reminder.authorId(),reminder.content());
+            return;
 
-        private Integer getReminderFailureCounter(int id){
-            return database.read(context -> context.selectFrom(PENDING_REMINDERS).where(PENDING_REMINDERS.ID.eq(id))
-                    .stream().map(PendingRemindersRecord::getReminderFailureCounter).findFirst().get()
-            );
         }
+
+        Instant current = Instant.now().plusSeconds(60);
```
Zabuzard, line 149: the naming is now incorrect. go for `remindAt`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -147,34 +134,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
-                    .setContent(content.toString())
-                    .setReminderFailureCounter(reminderFailureCounter + 1)
-                    .insert());
-        } else {
+    private void attemptRetryReminder(Reminder reminder) {
+        if (!(reminder.failureAttempts() <= MAX_FAILURE_RETRY)) {
             logger.warn(
                     """
-                            Failed to send a reminder with (authorID '{}'), skipping it. This can be due to a network issue, \
+                            Failed to send a reminder with (authorID '{}') and (content '{}'), skipping it. This can be due to a network issue, \
```
Zabuzard, line 141: personally, id put the content at the very end of the log message. the content can be really long text and its hard to read the log message if its put right into the sentence

Zabuzard: also, it would be helpful to have the reminder ID, not just the author ID

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -11,17 +10,13 @@
...
 import org.slf4j.LoggerFactory;
-
 import org.togetherjava.tjbot.db.Database;
-import org.togetherjava.tjbot.db.generated.tables.records.PendingRemindersRecord;
 import org.togetherjava.tjbot.features.Routine;
 
 import javax.annotation.Nullable;
-
-import java.awt.Color;
+import java.awt.*;
```
Zabuzard, line 17: star-import on `java.awt` is "forbidden", since it quickly leads to bugs with `java.awt.List` vs `java.util.List`. please configure ur IDE import-settings to disallow star import on java.awt, thanks

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -61,20 +57,18 @@ public void runRoutine(JDA jda) {
                 .where(PENDING_REMINDERS.REMIND_AT.lessOrEqual(now))
                 .stream()
                 .forEach(pendingReminder -> {
-                    sendReminder(jda, pendingReminder.getId(), pendingReminder.getChannelId(),
-                            pendingReminder.getAuthorId(), pendingReminder.getContent(),
-                            pendingReminder.getCreatedAt(), pendingReminder.getGuildId());
+                    Reminder reminder = new Reminder(pendingReminder.getId(),pendingReminder.getCreatedAt(),
+                            pendingReminder.getGuildId(),pendingReminder.getChannelId(),pendingReminder.getAuthorId(),
+                            pendingReminder.getRemindAt(),pendingReminder.getContent(),pendingReminder.getFailureAttempts());
```
Zabuzard, line 62: u can improve the code if u give ur `Reminder` record a method:
```java
public static Reminder from(PendingReminder pendingReminder) {
  return new Reminder(...);
}
```
then u can write `sendReminder(jda, Reminder.from(pendingReminder));` here

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -147,34 +134,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
                     """
-                            Failed to send a reminder with (authorID '{}'), skipping it. This can be due to a network issue, \
+                            Failed to send a reminder with (authorID '{}') and (content '{}'), skipping it. This can be due to a network issue, \
                             but also happen if the bot disconnected from the target guild and the \
-                            user has disabled DMs or has been deleted."""+ "contents: "+content,
-                    authorId);
-        }
-    }
+                            user has disabled DMs or has been deleted.""",
+                    reminder.authorId(),reminder.content());
```
Zabuzard, line 144: imo the log should also log the failure exception. otherwise we cant investigate why it failed

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -147,34 +134,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+        Instant current = Instant.now().plusSeconds(60);
+        database.write(context -> context.newRecord(PENDING_REMINDERS)
+                .setId(reminder.id())
+                        .setCreatedAt(reminder.createdAt())
+                .setGuildId(reminder.guildId())
+                .setChannelId(reminder.channelId())
+                .setAuthorId(reminder.authorId())
+                .setRemindAt(current)
+                .setContent(reminder.content())
+                .setReminderFailureCounter(reminder.failureAttempts()+1)
```
Zabuzard, line 158: to make it visually standing out better, id put this crucial line into a seperate variable beforehand:
```java
int failureAttempts = reminder.failureAttempts() + 1;

...

  .setReminderFailureCounter(failureAttempts)
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: yep just fixed this one, idk why i went with that without thinking why you suggested what you suggested earlier. Over complicated it for no reasons.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: fixed this

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: yea kinda makes sense, didn't think like that because never really used logging.

ankitsmt211: fixed this

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: fixed it for this one, not sure how to change import settings but ill look into it

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: can i actually do that? couldn't find anything related to from keyword in `records`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: do i need to create a custom exception?

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: so im not sure whether to keep it as `int` type or switch back to `long`?

Zabuzard: `long` please :)

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: `intellij settings > code style > imports > dont use * for ...` sth like that

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: records are just classes, u can give them methods:
```java
record Person(String name, int age) {
  public static Person from(String line) {
    ...
  }
}
```

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: no. just carry the `failure` through the method param and then attach it to the log as last parameter `logger.warn(..., failure)`. the logger will auto-attach the stacktrace if u put it as last argument

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: image
 is this really better? kinda makes `getters` longer than how it used to be
for ex: earlier i could access any field using `reminder.field`, but now it has to be `reminder.pendingReminders().field()`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: i should prolly use `failure.getMessage()`? to get proper message from failure throwable

ankitsmt211:
Please don't run the checks yet, this still needs some work and i had to update the recent changes to my remote repo atm.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -58,18 +59,17 @@ public void runRoutine(JDA jda) {
             .where(PENDING_REMINDERS.REMIND_AT.lessOrEqual(now))
             .stream()
             .forEach(pendingReminder -> {
-                sendReminder(jda, pendingReminder.getId(), pendingReminder.getChannelId(),
-                        pendingReminder.getAuthorId(), pendingReminder.getContent(),
-                        pendingReminder.getCreatedAt());
+                sendReminder(jda, Reminder.from(pendingReminder));
```
Taz03, line 62: pass the pendingReminder all along

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +136,32 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+        database.write(context -> context.newRecord(PENDING_REMINDERS)
+            .setId(reminder.pendingReminders().getId())
+            .setCreatedAt(Instant.from(reminder.pendingReminders().getCreatedAt()))
+            .setGuildId(reminder.pendingReminders().getGuildId())
+            .setChannelId(reminder.pendingReminders().getChannelId())
+            .setAuthorId(reminder.pendingReminders().getAuthorId())
+            .setRemindAt(remindAt)
+            .setContent(reminder.pendingReminders().getContent())
+            .setFailureAttempts(failureAttempts)
+            .insert());
```
Taz03, line 165: y u need to add it to the db?? just attempt 3 times in ur code

application/src/main/java/org/togetherjava/tjbot/features/reminder/Reminder.java
```diff
@@ -0,0 +1,12 @@
+package org.togetherjava.tjbot.features.reminder;
+
+import org.togetherjava.tjbot.db.generated.tables.records.PendingRemindersRecord;
+
+public record Reminder(PendingRemindersRecord pendingReminders) {
+
+    public static Reminder from(PendingRemindersRecord pendingReminders) {
+        return new Reminder(pendingReminders);
+    }
```
Taz03, line 9: makes no sense to have a `from` method, just do `new Reminder(pendingReminder)` y the name is plural btw?

application/src/main/resources/db/V13__Add_Reminder_Failure_attempts.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pending_reminders ADD failure_attempts INTEGER DEFAULT 0 NOT NULL ;
```
Taz03, line 1: i believe not null should be before default

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: @Taz03 we decided on that early on. its the easiest approach and also works if the bot might shutdown or crash in between.

if u retry in-code, u have to add the delay (+1 minute) manually and create a proper retry-loop. thats not trivial at all. its much easier to just let the system pick the reminder up again naturally by simply reinserting 🤷

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: gonna double check if thats possible. the `pendingReminder` itself might still be tied to the DB and hold the resource open.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: that's what issue suggested, adding a separate column to keep track of `failure_attempts`

application/src/main/java/org/togetherjava/tjbot/features/reminder/Reminder.java
Zabuzard: then u need to create a second constructor that chains to the original per-value constructor? also possible

application/src/main/resources/db/V13__Add_Reminder_Failure_attempts.sql
Zabuzard: checking other examples in our code base, it doesnt seem to matter. the only odd thing is the space before the semicolon `;`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: if that's not possible i was thinking instead of using `from` method inside record i think it's better to just use `record` otherwise it ends up with really longer calls like `reminder.pendingReminder().getId()`, earlier it was something like `reminder.Id()` with just records.

Taz03: i can't see a reason y it is not possible, do u get an error or something? I believe i've that in this repo somewhere

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: ah. okay. if the autogenerated code has `int`, then im okay with having `int`.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +136,32 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
         }
     }
+
+    private void attemptRetryReminder(Reminder reminder, Throwable failure) {
+        if (reminder.pendingReminders().getFailureAttempts() > MAX_FAILURE_RETRY) {
+            logger
+                .warn("""
+                        Failed to send a reminder with (authorID '{}') skipping it. This can be due to a network issue, \
+                        but also happen if the bot disconnected from the target guild and the \
+                        user has disabled DMs or has been deleted. It's (content includes '{}' with reminderId '{}'). The reason for (failure '{}')""",
```
Zabuzard, line 146: The content-sentence looks and reads weird:

> It's (content includes 'hello world please remind me about foo bar blablabla haha' with reminderId '13').

Instead, go for:

> Reminder id 13, content is: hello world please remind me about foo bar blablabla haha

Zabuzard: having the `failure` in the message itself is incorrect. the logger will auto-attach it, dont add it to the string. just drop the last sentence, but keep it as last parameter in the method call

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: no. u misunderstood me here. i meant like:

```java
record Reminder(String name, int age) {
  public static Reminder from(PendingReminder other) {
    return new Reminder(other.getName(), other.getAge())
  }
}
```
the `Reminder` was not supposed to wrap the `PendingReminder`, but to extract all its content into its own fields.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: (u did it wrong, see comment in newest CR)

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +136,32 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                        reminder.pendingReminders().getAuthorId(),
+                        reminder.pendingReminders().getContent(),
+                        reminder.pendingReminders().getId(), failure);
+            return;
+
+        }
+
+        int failureAttempts = reminder.pendingReminders().getFailureAttempts() + 1;
+        Instant remindAt = Instant.now().plusSeconds(60);
```
Zabuzard, line 155: isnt there a `plusMinutes(1)`?

ankitsmt211: just `nanos`, `milis` and `seconds`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: ok will update this

ankitsmt211: image
is this better now?

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: should i hold onto this one for now ? taz mentioned to pass on pendingReminder itself so

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: not sure why u want to put it into `(...)`, reads weird imo. the apostrophe should be `Its´` afaik, missing space before the `(` and missing dot at the end of the sentence.

all in all, id go for

> Reminder id 13, content is: hello world please remind me about foo bar blablabla haha

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: if its possible to use `pendingReminder` itself, then i would prefer that and delete your `Reminder` record

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: wow been using Its wrong all along, turns out It's and Its are two different things with a slight variations. Apparently It's short for `It is` or `It has`

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: so should i do this instead? i didn't know it was for mins

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: for the record, just passed `pendingReminder` and it works

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: yeah. and if the apostrophe is put at the end, its means that u possess something, as in "that thing belongs to me".

in any case, those shortcuts generally should not be used in code. intellij also has a warning for it actually

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: oops, i meant to have 1 **minute** of delay

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -137,30 +139,29 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+            .setId(pendingReminder.getId())
+            .setCreatedAt(Instant.from(pendingReminder.getCreatedAt()))
+            .setGuildId(pendingReminder.getGuildId())
+            .setChannelId(pendingReminder.getChannelId())
+            .setAuthorId(pendingReminder.getAuthorId())
             .setRemindAt(remindAt)
-            .setContent(reminder.pendingReminders().getContent())
+            .setContent(pendingReminder.getContent())
             .setFailureAttempts(failureAttempts)
             .insert());
```
Zabuzard, line 166: Could you try the following instead:

```java
pendingReminder.setRemindAt(remindAt);
pendingReminder.setFailureAttempts(failureAttempts);
pendingReminder.insert();
```

maybe that works as well, then you could get rid of everything else here

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: that should work and definitely would clean up the code, ill look into it.

ankitsmt211: updated as requested and pushed the local repo

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -154,15 +154,13 @@ Failed to send a reminder with (authorID '{}') skipping it. This can be due to a
...
-            .setAuthorId(pendingReminder.getAuthorId())
-            .setRemindAt(remindAt)
-            .setContent(pendingReminder.getContent())
-            .setFailureAttempts(failureAttempts)
-            .insert());
+        database.write(context -> {
+            pendingReminder.setRemindAt(remindAt);
+            pendingReminder.setFailureAttempts(failureAttempts);
+            pendingReminder.insert();
+        });
```
Zabuzard, line 161: to signal that you dont use `context`, you can call it `any` instead of `context` (in other languages people also like to use `_`, but thats not possible in java)

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -154,15 +154,13 @@ Failed to send a reminder with (authorID '{}') skipping it. This can be due to a
...
+
+
+
```
Zabuzard, line 164: 3 newlines seem a bit too much :P

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: that's not spotless?

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: Is it just an FYI or a suggestion to be added?

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Taz03: no

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Taz03: that is a suggestion to be added, when i read the name any, ik that it is not used inside. So u should always have good naming

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: i don't understand, it's a pre-commit hook and it did what it was supposed to do `format the code` and i thought maybe its Spotless's doing.

Taz03: spotless isn't supposed to format ur code 100%, u can always use some edge case exploit, so thats y the review process is there, just remove the empty lines

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -142,4 +138,27 @@ static ReminderRoute toPrivate(PrivateChannel channel) {
...
+                        but also happen if the bot disconnected from the target guild and the \
+                        user has disabled DMs or has been deleted. It's reminderId '{}' content includes '{}'.""",
+                        pendingReminder.getAuthorId(), pendingReminder.getContent(),
+                        pendingReminder.getId(), failure);
+            return;
+
+        }
+
+        int failureAttempts = pendingReminder.getFailureAttempts() + 1;
+        Instant remindAt = Instant.now().plus(1, ChronoUnit.MINUTES);
```
Taz03, line 156: y not set it to 1 min after the initial reminding time?

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: Not sure, followed what zabu suggested.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
Zabuzard: spotless didnt add those newlines, that must have been u. spotless isnt perfect, it cant remove the newlines unfortunately.

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
```diff
@@ -57,7 +57,7 @@ public Schedule createSchedule() {
     @Override
     public void runRoutine(JDA jda) {
         Instant now = Instant.now();
-        database.write(context -> context.selectFrom(PENDING_REMINDERS)
+        database.write(any -> any.selectFrom(PENDING_REMINDERS)
```
Zabuzard, line 60: this was a mistake. `any` is supposed to indicate **i dont care what or who this parameter is, since i never use it**. but here, you are using it.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: i had a feeling that since `.queue()` was used on each api call then it could be delayed a bit as well

application/src/main/java/org/togetherjava/tjbot/features/reminder/RemindRoutine.java
ankitsmt211: but does it matter? isn't it supposed to be just a parameter being passed into the the function? be it `context` or `any`?

Zabuzard: but the reader needs to understand what it is, if u are using it. it is the database-context.

i mean, its clear that its important to have a good variable naming. `dogs` instead of `foos` or `objects`. but when you do not actually use it, and the parameter is just forced upon you - but you are not interested, then its idiomatic to call it `any` (or in other languages `_`).

for example, if you have a `List<Pair<String, Int>> persons` int kotlin and want to print all the strings, then u would write `persons.forEach { (name, _) -> println(name) }`. See how it is important to clearly name what the left parameter is, since we didnt use a well named class `Person` but `Pair` instead. But we dont care what the second thing is (`age`), so naming it would only distract the reader. Thats kinda the idea.

ankitsmt211: from what i can understand, think it's safe to  conclude if im gonna use something, i should care about the name of variable. Thanks for the explanation.

Taz03: "when i read any, i know that it is never used inside"

Taz03 (approved):

Issue #645, opened 2022-10-20
/modmail modal support


Zabuzard:
## Overview

Right now, our `/modmail` command is a simple slash-command. Messages are entered within the command as argument:

![example](https://i.imgur.com/hiRxuQn.png)

This works well, but is suboptimal. A better UX would be achieved by using **modal**s. Similar to this:

![modal](https://i.imgur.com/Yc2DqsE.png)

One would simply use the command `/modmail`, without arguments, and it would popup a modal with options to input all the 3 arguments:
* the message to forward
* which server to contact
* anonymous yes/no

(unsure whether one can add the latter two to a modal. if not, then simply keep them in the slashcommand and have the modal only have the message)

To get started with modals, see our tutorial https://github.com/Together-Java/TJ-Bot/wiki/Create-and-use-modals
The example command created there should be very similar to what we want for `/modmail`.

Taz03:
what about the stay-anonymous option tho? will the user write true or false in an modal text??

what about the server option? how u gonna show options

kavink98:
@Zabuzard I can pick this up

marko-radosavljevic:
@kavink98 awesome!

Welcome to the project by the way :relaxed: :heart:

hamzanaciri99:
Can I work on this one?

Taz03:
sure enjoy

PR #800, opened 2023-03-09
add modal support to /modmail command

hamzanaciri99:

## Pull-request

### Changes

- [x] Existing code
- [ ] New feature

Closes Issue: #645

## Description

Add modal for users to write their message, the command still accepts guild and "Wants to stay anonymous" as parameters, only message has been moved into the modal.

The new command looks like this:


Taz03:
i'd say keep the anonymous field in the modal as well, what do u guys think?

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
```diff
@@ -104,22 +112,47 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
         authorToLastModMailInvocation.put(userId, Instant.now());
-        event.deferReply().setEphemeral(true).queue();
 
-        long userGuildId = event.getOption(OPTION_GUILD).getAsLong();
-        Optional<TextChannel> modMailAuditLog = getModMailChannel(event.getJDA(), userGuildId);
+        userToSlashCommandEvent.put(userId, event);
+        createMessageModal(event);
+    }
+
+    private void createMessageModal(SlashCommandInteractionEvent event) {
```
Taz03, line 120: sendMessageModal

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
```diff
@@ -104,22 +112,47 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+        event.replyModal(modal).queue();
+    }
+
+    @Override
+    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
+        String userMessage = event.getValue(OPTION_MESSAGE).getAsString();
+        long userId = event.getUser().getIdLong();
+
+        SlashCommandInteractionEvent slashCommandEvent =
+                userToSlashCommandEvent.getIfPresent(userId);
```
Taz03, line 140: no, send the guild id as modal args instead

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
```diff
@@ -78,7 +82,7 @@ public ModMailCommand(JDA jda, Config config) {
 
         guildOption.addChoices(choices);
 
-        getData().addOptions(messageOption, guildOption, anonymousOption);
+        getData().addOptions(/* messageOption, */ guildOption, anonymousOption);
```
Zabuzard, line 85: this commented out section looks like a debugging left-over

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
```diff
@@ -45,10 +49,12 @@ public final class ModMailCommand extends SlashCommandAdapter {
     private static final String OPTION_MESSAGE = "message";
     private static final String OPTION_STAY_ANONYMOUS = "stay-anonymous";
     private static final String OPTION_GUILD = "server";
-    private static final int COOLDOWN_DURATION_VALUE = 30;
+    private static final int COOLDOWN_DURATION_VALUE = 0; // 30
     private static final ChronoUnit COOLDOWN_DURATION_UNIT = ChronoUnit.MINUTES;
     private static final Color AMBIENT_COLOR = Color.BLACK;
     private final Cache<Long, Instant> authorToLastModMailInvocation = createCooldownCache();
+    private final Cache<Long, SlashCommandInteractionEvent> userToSlashCommandEvent =
+            createSlashCommandCache();
```
Zabuzard, line 57: this cache is problematic. the main issue is that ur saving the `...Event`, which is a JDA class that must not be persisted for too long. Ideally, it should not be persisted for longer than its own event at all. Now, you could ofc adjust this to save a `record` instead, with only the values you actually need.

However, there is a more idiomatic approach to this. You are using the cache to get hands on the previously selected slash-command-options, to remember them. This is actually a perfect application for the component-id.

You may have noticed that the modal has a component-id, which you created with `generateComponentId()`. Now, you can give this method arbitrary data, for example `generateComponentId(guildId, stayAnonymous)` and it will remember it and give it back to you `onModalSubmitted`. You may have noticed the second parameter there, `List<String> args`. That is exactly the list of arguments that you have put into `generateComponentId`. So with above example, you could then do:

```java
String guildId = args.get(0);
String stayAnonymous = args.get(1);
```
(its ofc string, so you might have to do some trivial parsing, like `Boolean.parseBoolean(args.get(1))` and similar)

So that way you can completely get rid of the cache and have a more idiomatic and elegant solution - which even works when the bot is shutdown and restarted in the meantime (since we persist the component-id data in the database)

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
```diff
@@ -104,22 +112,47 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
-        Optional<TextChannel> modMailAuditLog = getModMailChannel(event.getJDA(), userGuildId);
+        userToSlashCommandEvent.put(userId, event);
+        createMessageModal(event);
+    }
+
+    private void createMessageModal(SlashCommandInteractionEvent event) {
+        TextInput message =
+                TextInput.create(OPTION_MESSAGE, "Your message", TextInputStyle.PARAGRAPH)
+                    .setPlaceholder("What do you want to tell them?")
+                    .setMinLength(10)
```
Zabuzard, line 124: i guess a min length of 3 would be more realistic. looking through the latest mod-mails, most of them are like "scam" or "crypto" or similar comments

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/modmail/ModMailCommand.java
hamzanaciri99: Ops!

Taz03:
> i'd say keep the anonymous field in the modal as well, what do u guys think?

@Zabuzard what about this?

hamzanaciri99:
> i'd say keep the anonymous field in the modal as well, what do u guys think?

Wouldn't it be ambiguous for users? the current one gives options, they would have to type yes or no, but users who used it before may type true and false instead.
It would be nice if we could validate user input before submitting, and show a message if the input in invalid (!= yes/no)

Taz03:
> the current one gives options, they would have to type yes or no, but users who used it before may type true and false instead.

we can accept everything from `y/n, yes/no, t/f, true/false`

> It would be nice if we could validate user input before submitting

unfortunately that is not possible

Zabuzard:
if modals support dropdowns, i would want to move as much as possible into it. if thats not an option, i would rather want to keep it as slash-command. dont want to have input-parsing y/n/yes/no/... on that honestly. in particular if we have a slash-command input already anyways due to the guild-dropdown (unless modals have dropdowns)

Zabuzard (approved):

Taz03 (approved):

Zabuzard:
Awesome, thanks for your contribution ❤️ Are you on our Discord? If so, could you tell me your account name, then I can give you the contributor badge 👍

hamzanaciri99:
Thank you @Zabuzard for your kind words.
Yes, I am on your Discord, my name is Rimuru.

Taz03:
i can see 2 Rimuru in discord, pls tell ur tag as well

hamzanaciri99:
Rimuru#2995

PR #802, opened 2023-03-13
remove bootstrap launcher

Taz03:


application/src/main/java/org/togetherjava/tjbot/BootstrapLauncher.java
```diff
@@ -1,43 +0,0 @@
...
-            // And 0 means no workers, so JDA cannot function, no Callback's on REST-Requests
-            // are executed
-            // NOTE This will likely be fixed with Java 18 or newer, remove afterwards (see
-            // https://bugs.openjdk.java.net/browse/JDK-8274349 and
-            // https://github.com/openjdk/jdk/pull/5784)
-            // noinspection UseOfSystemOutOrSystemErr
-            System.out.println("Available Cores \"" + cores + "\", setting Parallelism Flag");
-            // noinspection AccessOfSystemProperties
-            System.setProperty("java.util.concurrent.ForkJoinPool.common.parallelism", "1");
-        }
```
Zabuzard, line 41: Has this been deployed to the Termurin JDK that we use already? Otherwise our bot will be totally unusable, as its thread pools would be allowed to use 0 threads then.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/BootstrapLauncher.java
marko-radosavljevic: Yup, it's been part of openjdk for a year and a half, from 17.0.2 onwards.

Checked the temurin 19 source, and the fix is there as well. :relaxed:

Zabuzard (approved):

marko-radosavljevic (approved):

PR #808, opened 2023-03-18
Update to SlashCommandEducator to exclude strings that end with ")" o…

### LINKED ISSUE #804: SlashCommandEducator should ignore method calls
Sometimes people write messages like

> .toString()

Currently, `SlashCommandEducator` picks that up and thinks it is a command, for example `.close`. In this case, it is obviously a false positive.

To prevent this, we should additionally check whether the message ends with `)`. If so, we can just ignore it. Just adding the check here in `SlashCommandEducator.java`:

code

And additionally adding `".toString()"` to the tests in `SlashCommandEducatorTest.java` here:

test

tmcdonnell2:
…r ";" from potential incorrect commands.  Also, updates to SlashCommandEducatorTest to include new tests.

Closes #804 .

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/basic/SlashCommandEducator.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/basic/SlashCommandEducator.java
@@ -24,7 +24,7 @@ public final class SlashCommandEducator extends MessageReceiverAdapter {
     private static final Predicate<String> IS_MESSAGE_COMMAND = Pattern.compile("""
             [.!?] #Start of message command
             [a-zA-Z]{2,15} #Name of message command, e.g. 'close'
-            .* #Rest of the message
+            .*[^);] #Rest of the message (don't end with code stuff)
             """, Pattern.COMMENTS).asMatchPredicate();
 
     @Override
@@ -36,8 +36,7 @@ public final class SlashCommandEducator extends MessageReceiverAdapter {
         String content = event.getMessage().getContentRaw();
 
         if (IS_MESSAGE_COMMAND.test(content) &&
-            content.length() < MAX_COMMAND_LENGTH &&
-            (!content.endsWith(";") && !content.endsWith(")"))) {
+            content.length() < MAX_COMMAND_LENGTH) {
             sendAdvice(event.getMessage());
         }
     }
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/features/basic/SlashCommandEducator.java
```diff
@@ -35,7 +35,9 @@ public void onMessageReceived(MessageReceivedEvent event) {
 
         String content = event.getMessage().getContentRaw();
 
-        if (IS_MESSAGE_COMMAND.test(content) && content.length() < MAX_COMMAND_LENGTH) {
+        if (IS_MESSAGE_COMMAND.test(content) &&
+            content.length() < MAX_COMMAND_LENGTH &&
+            (!content.endsWith(";") && !content.endsWith(")"))) {
```
Zabuzard, line 40: the extra `(...)` is obsolete

Zabuzard (changes requested):
Good stuff. Seems that you have to run spotless for auto-formatting though :)

application/src/main/java/org/togetherjava/tjbot/features/basic/SlashCommandEducator.java
Taz03: do this `!(content.endsWith(";") || content.endsWith(")"))`

Zabuzard: Disagree. The two checks are separate checks. To me, the checks read like:
* is message command?
* message small?
* message doesnt end with ;?
* message doesnt end with )?

That said, at this point the better approach would likely be to just add all those checks into the regex `IS_MESSAGE_COMMAND`.

Zabuzard: Thats the current pattern:
```java
    private static final Predicate<String> IS_MESSAGE_COMMAND = Pattern.compile("""
            [.!?] #Start of message command
            [a-zA-Z]{2,15} #Name of message command, e.g. 'close'
            .* #Rest of the message
""", Pattern.COMMENTS).asMatchPredicate();
```
Thats how to add this part:
```java
    private static final Predicate<String> IS_MESSAGE_COMMAND = Pattern.compile("""
            [.!?] #Start of message command
            [a-zA-Z]{2,15} #Name of message command, e.g. 'close'
            .*[^;)] #Rest of the message
""", Pattern.COMMENTS).asMatchPredicate();
```

tmcdonnell2: Hi guys! Thanks for the feedback.

I did see the regex - but felt that it was best left alone. I did initially  try to do regex for the test but I couldn't get it to work. But this change makes way more sense. I also was just following the instructions left for the issue. 

What do you mean by:

> the extra (...) is obsolete

tmcdonnell2:
I got the merge going again. I got Spotless working and should now pass. I am unsure about the other workflow programs. 

Spotless suggested changing a method in SlashCommandEducator because the parameter being passed in was already a final in the class and thus didn't need to be passed in and could be accessed freely. 

One other thing I couldn't figure out (and wasn't sure if it made sense) was to try and make the response ephemeral? I think the classes involved can't do this and I don't know if its what you want.

application/src/main/java/org/togetherjava/tjbot/features/basic/SlashCommandEducator.java
Zabuzard: Thanks 👍 You wrote
```java
if (foo ... && (!bar && !baz))
               ^            ^
``` 
Which is the same as just writing
```java
if (foo ... && !bar && !baz)
```

Zabuzard (approved):

Zabuzard:
Thanks, I like it 👍 

> One other thing I couldn't figure out (and wasn't sure if it made sense) was to try and make the response ephemeral? I think the classes involved can't do this and I don't know if its what you want.

Yeah, thats not possible. Ephemeral only works if you have a client interaction to interact with. For example if the user invoked a slash command, you can respond ephemeral to that. But you can not send ephemeral messages _"out of nowhere"_.

Zabuzard:
Seems that the unit tests fail for one of the strings:

> SlashCommandEducatorTest > sendsAdviceOnMessageCommand(String) > SlashCommandEducatorTest.sendsAdviceOnMessageCommand(String)[7] FAILED
>
>     WantedButNotInvoked at SlashCommandEducatorTest.java:49

application/src/test/java/org/togetherjava/tjbot/features/reminder/RawReminderTestHelper.java
```diff
@@ -14,8 +14,8 @@
 import static org.togetherjava.tjbot.db.generated.tables.PendingReminders.PENDING_REMINDERS;
 
 final class RawReminderTestHelper {
-    private Database database;
-    private JdaTester jdaTester;
+    private final Database database;
+    private final JdaTester jdaTester;
```
Taz03, line 18: revert this (out of scope of this pr)

Taz03 (changes requested):

Taz03 (approved):

PR #811, opened 2023-03-25
Major bugfix with broken Bootstrap

Zabuzard:
Bot is chain crashing bc it tries to find the wrong class after it got removed.

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -27,7 +27,7 @@ jib {
         }
     }
     container {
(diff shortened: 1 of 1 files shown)


marko-radosavljevic (approved):

Taz03 (approved):
sry :c

Zabuzard:
@Taz03 no worries. happens to everyone. thats why we have a test server, so we noticed early enough :)

PR #820, opened 2023-04-10
ChatGPT attempts new questions from the questions channel

tmcdonnell2:
Code execution from user perspective/birds-eye view:

# ROSY PATH
Step 1) Enter your question into the questions forum.  The question is formed from the title and question message. The title is always include and the message takes up the remaining space we have set for messages.
Additionally, the tags selected by the user are appended to the front of the question to provide more context to ChatGPT. From just a quick look at questions in the forum on the main discord, people ask questions without proper context, much like how titles doesn't include anything about Java since its implied by the asking in this discord. However, users generally seem good at adding the tags for the question. 
Step 1  Ask question

Step 2) If a suitable question is found, it is sent to ChatGPT. Once the response returns it is checked to see if it was a bad answer.  The answer is sent as an embed of an explanation message: it explains that the next message is AI generated, someone should be long soon, and you can use /chatgpt to ask the AI more questions. All this is sent AFTER the original help thread messages (format your code like this, provide details, close the thread when done).
Screenshot 2023-04-22 155707

Screenshot 2023-04-22 154953

# ERROR PATH
Should at any point an error be produced, then all is returned is a message about how you have use /chatgpt to ask questions while you wait for a human. It is like the normal path except for an answer from ChatGPT.
image

### REASONS FOR ERRORS
- Timeout (most common)
- ChatGPT unable to respond (a basic check for "As an AI language model" since this _usually_ means a non-response)
- Combined the question title and question message are both combined < 20 characters since it is unlikely that this will be enough context for ChatGPT to work with and it either won't respond or ask for more context.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/Features.java
@@ -72,9 +72,9 @@ public class Features {
         ModerationActionsStore actionsStore = new ModerationActionsStore(database);
         ModAuditLogWriter modAuditLogWriter = new ModAuditLogWriter(config);
         ScamHistoryStore scamHistoryStore = new ScamHistoryStore(database);
-        HelpSystemHelper helpSystemHelper = new HelpSystemHelper(config, database);
-        CodeMessageHandler codeMessageHandler = new CodeMessageHandler();
         ChatGptService chatGptService = new ChatGptService(config);
+        HelpSystemHelper helpSystemHelper = new HelpSystemHelper(config, database, chatGptService);
+        CodeMessageHandler codeMessageHandler = new CodeMessageHandler();
 
         // NOTE The system can add special system relevant commands also by itself,
         // hence this list may not necessarily represent the full list of all commands actually
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -183,15 +183,16 @@ public final class HelpSystemHelper {
                 %s.
                 """::formatted;
 
-        List<MessageEmbed> aiResponseEmbeds = new ArrayList<>();
+        RestAction<Message> message =
+                mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
+                    .map(preambleResponse)
+                    .flatMap(threadChannel::sendMessage);
+
         for (String aiResponse : chatGPTAnswer.get()) {
-            aiResponseEmbeds.add(HelpSystemHelper.embedWith(aiResponse));
+            message = message.map(aiResponse::formatted).flatMap(threadChannel::sendMessage);
         }
 
-        return mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
-            .map(preambleResponse)
-            .flatMap(threadChannel::sendMessage)
-            .flatMap(embed -> threadChannel.sendMessageEmbeds(aiResponseEmbeds));
+        return message;
     }
 
     private Optional<String> prepareChatGptQuestion(ThreadChannel threadChannel,
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -138,31 +138,13 @@ public final class HelpSystemHelper {
         return action.setEmbeds(embeds);
     }
 
-    /**
-     * Gather first message from thread and pass into asynchronous sendChatGptAttempt.
-     * 
-     * @param threadChannel - The forum thread channel where the question originated
-     * @return A response from the AI service indicating success or failure.
-     * @see HelpSystemHelper#sendChatGptAttempt
-     */
-    RestAction<Message> prepareChatGptAttempt(ThreadChannel threadChannel) {
+    public RestAction<Message> prepareChatGptAttempt(ThreadChannel threadChannel) {
+
         return threadChannel.retrieveMessageById(threadChannel.getIdLong())
             .flatMap(message -> sendChatGptAttempt(message.getContentRaw(), threadChannel));
     }
 
-    /**
-     * Determine between the title of the thread and the first message which to send to the AI. It
-     * uses a simple heuristic of length to determine if enough context exists in a question. If the
-     * title is used, it must also include a question mark since the title is often used more as an
-     * indicator of topic versus a question.
-     * 
-     * @param questionFirstMessage - The first message of the thread which originates from the
-     *        question asker.
-     * @param threadChannel - The thread in which the question was asked.
-     * @return An answer for the user from the AI service or a message indicating either an error or
-     *         why the message wasn't used.
-     */
-    private RestAction<Message> sendChatGptAttempt(String questionFirstMessage,
+    public RestAction<Message> sendChatGptAttempt(String questionFirstMessage,
             ThreadChannel threadChannel) {
         String questionTitle = threadChannel.getName();
         StringBuilder stringBuilder = new StringBuilder(questionFirstMessage);
@@ -176,7 +158,7 @@ public final class HelpSystemHelper {
                                 Your first message or title did not fit the format to be sent to ChatGPT. \
                                 Your message needs to be between %s and %s characters. If your question is longer than %s \
                                 then your title needs to include a question mark."
-                                    """,
+                                """,
                         MIN_QUESTION_LENGTH, MAX_QUESTION_LENGTH, MAX_QUESTION_LENGTH));
             }
             stringBuilder.replace(0, stringBuilder.length(), questionTitle);
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -30,9 +30,10 @@ import org.togetherjava.tjbot.features.utils.MessageUtils;
 
 import javax.annotation.Nullable;
 
-import java.awt.Color;
+import java.awt.*;
 import java.io.InputStream;
 import java.util.*;
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -68,7 +69,7 @@ public final class HelpSystemHelper {
     private static final int MAX_QUESTION_LENGTH = 200;
     private static final int MIN_QUESTION_LENGTH = 10;
     private static final String CHATGPT_FAILURE_MESSAGE =
-            "You can use %s to ask ChatGPT about your question while you wait for a member to respond.";
+            "You can use %s to ask ChatGPT about your question while you wait for a human to respond.";
 
     /**
      * Creates a new instance.
@@ -149,74 +150,78 @@ public final class HelpSystemHelper {
      * title is used, it must also include a question mark since the title is often used more as an
      * indicator of topic versus a question.
      * 
-     * @param questionFirstMessage - The first message of the thread which originates from the
-     *        question asker.
-     * @param threadChannel - The thread in which the question was asked.
+     * @param originalQuestion The first message of the thread which originates from the question
+     *        asker.
+     * @param threadChannel The thread in which the question was asked.
      * @return An answer for the user from the AI service or a message indicating either an error or
      *         why the message wasn't used.
      */
     RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
-            String questionFirstMessage) {
+            String originalQuestion) {
 
-        Optional<String> questionOptional =
-                prepareChatGptQuestion(threadChannel, questionFirstMessage);
+        Optional<String> questionOptional = prepareChatGptQuestion(threadChannel, originalQuestion);
         Optional<String> chatGPTAnswer;
 
-        try {
-            String question = questionOptional.orElseThrow();
-            logger.debug("The final question sent to chatGPT: {}", question);
-            chatGPTAnswer = chatGptService.ask(question);
+        String question = questionOptional.orElseThrow();
+        logger.debug("The final question sent to chatGPT: {}", question);
+        chatGPTAnswer = chatGptService.ask(question);
 
-            String aiResponse = chatGPTAnswer.orElseThrow();
-            String lowercaseAiResponse = aiResponse.toLowerCase(Locale.US);
-            if (lowercaseAiResponse.contains("as an ai language model")) {
-                logger.debug("Response from ChatGPT, lacks context or unable to respond?: {}",
-                        aiResponse);
-                return useChatGptFallbackMessage(threadChannel);
-            }
-
-            UnaryOperator<String> response =
-                    """
-                            Here is an AI assisted attempt to answer your question 🤖. Maybe it helps! \
-                            In any case, a human is on the way 👍. To continue talking to the AI, you can use \
-                            %s.
-                            """::formatted;
-
-            return mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
-                .map(response)
-                .flatMap(threadChannel::sendMessage)
-                .flatMap(embed -> threadChannel
-                    .sendMessageEmbeds(HelpSystemHelper.embedWith(aiResponse)));
-        } catch (NoSuchElementException e) {
-            logger.warn("Something went wrong while trying to communicate with the ChatGpt API: {}",
-                    e.getMessage());
+        if (chatGPTAnswer.isEmpty()) {
+            logger.warn("Something went wrong while trying to communicate with the ChatGpt API");
             return useChatGptFallbackMessage(threadChannel);
         }
+
+        String aiResponse = chatGPTAnswer.get();
+        String lowercaseAiResponse = aiResponse.toLowerCase(Locale.US);
+        if (lowercaseAiResponse.contains("as an ai language model")) {
+            logger.debug("Response from ChatGPT, lacks context or unable to respond?: {}",
+                    aiResponse);
+            return useChatGptFallbackMessage(threadChannel);
+        }
+
+        UnaryOperator<String> response = """
+                Here is an AI assisted attempt to answer your question 🤖. Maybe it helps! \
+                In any case, a human is on the way 👍. To continue talking to the AI, you can use \
+                %s.
+                """::formatted;
+
+        return mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
+            .map(response)
+            .flatMap(threadChannel::sendMessage)
+            .flatMap(embed -> threadChannel
+                .sendMessageEmbeds(HelpSystemHelper.embedWith(aiResponse)));
     }
 
     private Optional<String> prepareChatGptQuestion(ThreadChannel threadChannel,
-            String questionFirstMessage) {
+            String originalQuestion) {
         String questionTitle = threadChannel.getName();
         StringBuilder questionBuilder = new StringBuilder(MAX_QUESTION_LENGTH);
 
-        if (questionFirstMessage.length() < MIN_QUESTION_LENGTH
+        if (originalQuestion.length() < MIN_QUESTION_LENGTH
                 && questionTitle.length() < MIN_QUESTION_LENGTH) {
             return Optional.empty();
         }
 
-        if (questionTitle.length() + questionFirstMessage.length() > MAX_QUESTION_LENGTH) {
-            // - 1 for the space that will be added afterward.
-            questionFirstMessage = questionFirstMessage.substring(0,
-                    (MAX_QUESTION_LENGTH - 1) - questionTitle.length());
+        questionBuilder.append(questionTitle).append(" ");
+        if (originalQuestion.length() > MAX_QUESTION_LENGTH - questionBuilder.length()) {
+            originalQuestion =
+                    originalQuestion.substring(0, MAX_QUESTION_LENGTH - questionBuilder.length());
         }
 
-        questionBuilder.append(questionTitle).append(" ").append(questionFirstMessage);
+        questionBuilder.append(originalQuestion);
+
+        StringBuilder tagBuilder = new StringBuilder();
+        int stringLength = questionBuilder.length();
         for (ForumTag tag : threadChannel.getAppliedTags()) {
-            if (questionBuilder.length() > MAX_QUESTION_LENGTH) {
+            String tagName = tag.getName();
+            stringLength += tagName.length();
+            if (stringLength > MAX_QUESTION_LENGTH) {
                 break;
             }
-            questionBuilder.insert(0, String.format("%s ", tag.getName()));
+            tagBuilder.append(String.format("%s ", tagName));
         }
+
+        questionBuilder.insert(0, tagBuilder);
 
         return Optional.of(questionBuilder.toString());
     }
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -24,13 +24,15 @@ import org.togetherjava.tjbot.config.HelpSystemConfig;
 import org.togetherjava.tjbot.db.Database;
 import org.togetherjava.tjbot.db.generated.tables.HelpThreads;
 import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
+import org.togetherjava.tjbot.features.chaptgpt.ChatGptService;
 import org.togetherjava.tjbot.features.utils.MessageUtils;
 
 import javax.annotation.Nullable;
 
-import java.awt.Color;
+import java.awt.*;
 import java.io.InputStream;
 import java.util.*;
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -59,15 +61,18 @@ public final class HelpSystemHelper {
     private final Set<String> threadActivityTagNames;
     private final String categoryRoleSuffix;
     private final Database database;
+    private final ChatGptService chatGptService;
 
     /**
      * Creates a new instance.
      *
      * @param config the config to use
      * @param database the database to store help thread metadata in
+     * @param chatGptService the access point to the ChatGPT API
      */
-    public HelpSystemHelper(Config config, Database database) {
+    public HelpSystemHelper(Config config, Database database, ChatGptService chatGptService) {
         HelpSystemConfig helpConfig = config.getHelpSystem();
+        this.chatGptService = chatGptService;
         this.database = database;
 
         helpForumPattern = helpConfig.getHelpForumPattern();
@@ -129,6 +134,19 @@ public final class HelpSystemHelper {
                 .addFiles(FileUpload.fromData(codeSyntaxExampleData, CODE_SYNTAX_EXAMPLE_PATH));
         }
         return action.setEmbeds(embeds);
+    }
+
+    public RestAction<Message> sendChatGPTAttempt(ThreadChannel threadChannel) {
+        Optional<String> chatGPTAnswer = chatGptService.ask(threadChannel.getName());
+
+        String response = "While you are waiting, ChatGPT can give your question shot: ";
+        MessageCreateAction action = threadChannel.sendMessage(response);
+
+        List<MessageEmbed> embedding = List.of(HelpSystemHelper
+            .embedWith(chatGPTAnswer.orElse("We attempted to answer your question with ChatGPT but "
+                    + ChatGptService.ERROR_MESSAGE)));
+
+        return action.setEmbeds(embedding);
     }
 
     void writeHelpThreadToDatabase(long authorId, ThreadChannel threadChannel) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -155,38 +155,20 @@ public final class HelpSystemHelper {
      * @return An answer for the user from the AI service or a message indicating either an error or
      *         why the message wasn't used.
      */
-    RestAction<Message> constructChatGptAttempt(String questionFirstMessage,
-            ThreadChannel threadChannel) {
-        // Will be at most 100 characters.
-        String questionTitle = threadChannel.getName();
-        StringBuilder stringBuilder = new StringBuilder();
+    RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
+            String questionFirstMessage) {
 
-        if (questionFirstMessage.length() < MIN_QUESTION_LENGTH
-                && questionTitle.length() < MIN_QUESTION_LENGTH) {
-            return useChatGptFallbackMessage(threadChannel);
-        }
+        Optional<String> questionOptional =
+                prepareChatGptQuestion(threadChannel, questionFirstMessage);
+        Optional<String> chatGPTAnswer;
 
-        if (questionTitle.length() + questionFirstMessage.length() > MAX_QUESTION_LENGTH) {
-            // - 1 for the space that will be added afterward.
-            questionFirstMessage = questionFirstMessage.substring(0,
-                    (MAX_QUESTION_LENGTH - 1) - questionTitle.length());
-        }
+        try {
+            String question = questionOptional.orElseThrow();
+            logger.debug("The final question sent to chatGPT: {}", question);
+            chatGPTAnswer = chatGptService.ask(question);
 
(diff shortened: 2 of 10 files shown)


Zabuzard:
Mind sharing some screenshots of the flow? Much easier to look at than starring at the code only.

Also, you gotta resolve that merge conflict :)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +138,71 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            }
+            stringBuilder.replace(0, stringBuilder.length(), questionTitle);
+        }
+
+        for (ForumTag tag : tags) {
+            stringBuilder.insert(0, String.format("%s ", tag.getName()));
+        }
+
+        String question = stringBuilder.toString();
+        String response = """
```
Zabuzard, line 190: might be helpful to add sth like `To continue talking to the AI, you can use /chatgpt.` (see our `mentionSlashCommand(...)` helper method)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +138,71 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+                It will also the tags you selected to provide more context. \
+                In any case, a human is on the way \uD83D\uDC4D
+                """;
+        Optional<String> chatGPTAnswer = chatGptService.ask(question);
+        MessageCreateAction action = threadChannel.sendMessage(response);
+
+        List<MessageEmbed> embedding = List.of(HelpSystemHelper.embedWith(chatGPTAnswer.orElse("""
+                An attempt was made to answer your question with ChatGPT but \
+                an error has occurred while trying to communicate with ChatGPT.
+                """)));
```
Zabuzard, line 201: i think it would be better to not post anything if it fails. like, not even the intial message.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -24,13 +24,15 @@
 import org.togetherjava.tjbot.db.Database;
 import org.togetherjava.tjbot.db.generated.tables.HelpThreads;
 import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
+import org.togetherjava.tjbot.features.chaptgpt.ChatGptService;
 import org.togetherjava.tjbot.features.utils.MessageUtils;
 
 import javax.annotation.Nullable;
 
-import java.awt.Color;
+import java.awt.*;
```
Zabuzard, line 32: `java.awt.*` is a "forbidden" import, since it hides `java.util.List` with `java.awt.List`. Please change ur IDE settings to disallow star-import on awt.

Zabuzard (changes requested):

tmcdonnell2:
> Mind sharing some screenshots of the flow? Much easier to look at than starring at the code only.
> 
> Also, you gotta resolve that merge conflict :) Seems you worked from the wrong branch and now you have the other unrelated changes in this branch as well, which causes the merge conflict. You have to discard those changes and correctly rebase on `develop`.

Oh I thought I did revert the branch back to the correct state? or is it better to discard and rebase? I am guessing the latter since this is an extension of another branch and THAT branch ultimately changed as well.

I will get screenshots shortly.

tmcdonnell2:
Code execution from user perspective:/birds-eye view:

Step 1) Enter your question into the questions forum. Although unlikely robust enough, it does some basic logic to determine where to grab the question from. In this case, the message is skipped over since it is too short (< 20 characters). The title is used since it contains a '?' AND is long enough (>20 characters). 
Step 1  Ask question

Step 2) If a suitable question is found, the AI responds. It is sent first. 
Step  2  Get AI response back

Step 3) Finally, another message is sent explaining that is an AI generated response, someone should be long soon, and that you can use /chatgpt to ask the AI more questions.
Step 3  Explaination and SlashCommand

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2: I am having some trouble with this change given that mentionSlashCommand requires a message. My main issue is twofold: designing the response (should the AI explanation message come first or second? Should it be an embed?). My other issue is just working with JDA and the RestMessage class in general and getting messages sent. Is there any documentation on how to do this? If how I have just changed it isn't passable (and I suspect it isn't), I will have to ask in questions (I can taste the irony) on how to grok this out.

Zabuzard: The Tj-Bot explanation can be a message, the chat-gpt response an attached embed, i guess. But you are free to design it as you like, if it looks bad otherwise 🤷

Yeah, we can discuss it in #questions :)

Zabuzard: The way I had it in mind originally, was like:

"Here is an AI assisted attempt blablabla use /chatgpt to follow up blablabla:"
> I am the chatgpt answer, in an embed, hello

tmcdonnell2:
Code execution from user perspective/birds-eye view:

Step 1) Enter your question into the questions forum. Although unlikely robust enough, it does some basic logic to determine where to grab the question from. In this case, the message is skipped over since it is too short (< 20 characters). The title is used since it contains a '?' AND is long enough (>20 characters). 
Additionally, the tags selected by the user are appended to the front of the question to provide more context to (in this case) ChatGPT. This is done because, from just a quick look at questions in the forum on the main discord, people ask questions without proper context, much like how this Title doesn't include Java since its implied by the asking in this discord channel. However, users generally seem good at adding the tags for the question. So adding the tags will help get fuller answers. 
Step 1  Ask question

Step 2) If a suitable question is found, an explanation message is sent. It is sent AFTER the original help thread messages but it is the first of the two new messages. It explains that the next message is AI generated, someone should be long soon, and you can use /chatgpt to ask the AI more questions. Finally, ChatGPT's response is then given as an embed.
Screenshot 2023-04-18 100914

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,92 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+     * title is used, it must also include a question mark since the title is often used more as an
+     * indicator of topic versus a question.
+     * 
+     * @param questionFirstMessage - The first message of the thread which originates from the
+     *        question asker.
+     * @param threadChannel - The thread in which the question was asked.
+     * @return An answer for the user from the AI service or a message indicating either an error or
+     *         why the message wasn't used.
+     */
+    private RestAction<Message> sendChatGptAttempt(String questionFirstMessage,
```
Taz03, line 170: it is not sending anything, might consider a method rename

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2: How do you mean? This method sends either the AI generated message or a fallback message to the thread depending on the outcome.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,80 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            }
+            stringBuilder.insert(0, String.format("%s ", tag.getName()));
+        }
+
+        String question = stringBuilder.toString();
+        logger.debug("The final question sent to chatGPT: {}", question);
+        Optional<String> chatGPTAnswer = chatGptService.ask(question);
+
+        if (chatGPTAnswer.isPresent()) {
+            String aiResponse = chatGPTAnswer.get();
```
Zabuzard, line 188: `Optional#get` is superseded by `orElseThrow()`. does the same, but the name is better.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,80 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            String lowercaseAiResponse = aiResponse.toLowerCase();
+            if (lowercaseAiResponse.contains("as an ai language model")) {
+                logger.debug("Response from ChatGPT, lacks context or unable to respond?: {}",
+                        aiResponse);
+                return useChatGptFallbackMessage(threadChannel);
+            }
```
Zabuzard, line 194: the method does too much. consider putting this preparational stuff into a helper method

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,80 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+     * 
+     * @param questionFirstMessage - The first message of the thread which originates from the
+     *        question asker.
+     * @param threadChannel - The thread in which the question was asked.
+     * @return An answer for the user from the AI service or a message indicating either an error or
+     *         why the message wasn't used.
+     */
+    RestAction<Message> constructChatGptAttempt(String questionFirstMessage,
+            ThreadChannel threadChannel) {
+        // Will be at most 100 characters.
```
Zabuzard, line 160: where is this 100 coming from? this comment will likely get out of date fast

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,80 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+        String questionTitle = threadChannel.getName();
+        StringBuilder stringBuilder = new StringBuilder();
```
Zabuzard, line 162: builder for what? the name should include what ur building here, the response?
also, consider giving it a size-hint. any hint is better than none.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,80 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            stringBuilder.insert(0, String.format("%s ", tag.getName()));
+        }
+
+        String question = stringBuilder.toString();
+        logger.debug("The final question sent to chatGPT: {}", question);
+        Optional<String> chatGPTAnswer = chatGptService.ask(question);
+
+        if (chatGPTAnswer.isPresent()) {
+            String aiResponse = chatGPTAnswer.get();
+            String lowercaseAiResponse = aiResponse.toLowerCase();
```
Zabuzard, line 189: `toLowerCase` must be used with specifying a locale. `toLowerCase(Locale.US)`, otherwise it can result in nasty bugs when we move the bot to a different server.

Zabuzard (changes requested):
mind sharing up2date pictures of the conversation and the error cases? ideally put them into the description of the PR.

not sure if u did that, but i wanted to have no message at all (just silence) if chatgpt has an issue - did u do that? or does it still respond with some fallback "there was an error blablabla"

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2: It comes from the allowable length of titles for threads on discord (atleast when I was generating a question via the discord). Is that number adjustable? I know discord could change it at anytime. I can remove the comment but I thought it was some nice context.

tmcdonnell2:
> 

I have updated the description with how questions in the forum answered by ChatGPT- with the rosy and the error path.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
Zabuzard: im pretty sure that there would be a constant in JDA for this. anyways, my point is more that u have to write stuff like that or link it. otherwise people will be confused

tmcdonnell2: No that makes sense - If I need to make another change I will add it back.

tmcdonnell2:
> mind sharing up2date pictures of the conversation and the error cases? ideally put them into the description of the PR.
> 
> not sure if u did that, but i wanted to have no message at all (just silence) if chatgpt has an issue - did u do that? or does it still respond with some fallback "there was an error blablabla"

I updated the description to show these paths. This shows what messages are sent and you can see if you like it or not.

application/src/main/java/org/togetherjava/tjbot/features/Features.java
```diff
@@ -72,9 +72,9 @@ public static Collection<Feature> createFeatures(JDA jda, Database database, Con
         ModerationActionsStore actionsStore = new ModerationActionsStore(database);
         ModAuditLogWriter modAuditLogWriter = new ModAuditLogWriter(config);
         ScamHistoryStore scamHistoryStore = new ScamHistoryStore(database);
-        HelpSystemHelper helpSystemHelper = new HelpSystemHelper(config, database);
-        CodeMessageHandler codeMessageHandler = new CodeMessageHandler();
         ChatGptService chatGptService = new ChatGptService(config);
+        HelpSystemHelper helpSystemHelper = new HelpSystemHelper(config, database, chatGptService);
+        CodeMessageHandler codeMessageHandler = new CodeMessageHandler();
```
Zabuzard, line 77: Why this change? Prefer to not change `Features` unless needed (this file causes a lot of merge conflicts with everyone)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -59,15 +64,22 @@ public final class HelpSystemHelper {
     private final Set<String> threadActivityTagNames;
     private final String categoryRoleSuffix;
     private final Database database;
+    private final ChatGptService chatGptService;
+    private static final int MAX_QUESTION_LENGTH = 200;
+    private static final int MIN_QUESTION_LENGTH = 10;
+    private static final String CHATGPT_FAILURE_MESSAGE =
+            "You can use %s to ask ChatGPT about your question while you wait for a member to respond.";
```
Zabuzard, line 71: _... for a human to respond_

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
 
+    /**
+     * Determine between the title of the thread and the first message which to send to the AI. It
+     * uses a simple heuristic of length to determine if enough context exists in a question. If the
+     * title is used, it must also include a question mark since the title is often used more as an
+     * indicator of topic versus a question.
+     * 
+     * @param questionFirstMessage - The first message of the thread which originates from the
+     *        question asker.
+     * @param threadChannel - The thread in which the question was asked.
```
Zabuzard, line 154: dont write `-` to separate it, just space. that would be more idiomatic and is also whats used everywhere else in the code base.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+     * @return An answer for the user from the AI service or a message indicating either an error or
+     *         why the message wasn't used.
+     */
+    RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
+            String questionFirstMessage) {
```
Zabuzard, line 159: `questionFirstMessage` is too specific. what if we ever change the order? then this name is wrong.

maybe `originalQuestion` or just `question`.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+                            In any case, a human is on the way 👍. To continue talking to the AI, you can use \
+                            %s.
+                            """::formatted;
+
+            return mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
+                .map(response)
+                .flatMap(threadChannel::sendMessage)
+                .flatMap(embed -> threadChannel
+                    .sendMessageEmbeds(HelpSystemHelper.embedWith(aiResponse)));
+        } catch (NoSuchElementException e) {
```
Zabuzard, line 190: never use exceptions for logic that u could also check with simple if-else. in particular here, the Optional API has a very rich `orElse` API, make use of it.

```java
answer.orElseGet(() -> {
  logger.warn(...);
  return useChatGptFallbackMessage(...);
});
```

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            logger.warn("Something went wrong while trying to communicate with the ChatGpt API: {}",
+                    e.getMessage());
```
Zabuzard, line 192: ur surpressing the whole stacktrace, which is the most valuable part of the error message. just give it `e`, not `e.getMessage()`

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            return useChatGptFallbackMessage(threadChannel);
+        }
+    }
+
+    private Optional<String> prepareChatGptQuestion(ThreadChannel threadChannel,
+            String questionFirstMessage) {
+        String questionTitle = threadChannel.getName();
+        StringBuilder questionBuilder = new StringBuilder(MAX_QUESTION_LENGTH);
```
Zabuzard, line 200: i suppose questions are rarely max length, are they? try to go for something that is more common in practice, maybe `max / 2`?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+            return Optional.empty();
+        }
+
+        if (questionTitle.length() + questionFirstMessage.length() > MAX_QUESTION_LENGTH) {
+            // - 1 for the space that will be added afterward.
+            questionFirstMessage = questionFirstMessage.substring(0,
+                    (MAX_QUESTION_LENGTH - 1) - questionTitle.length());
+        }
+
+        questionBuilder.append(questionTitle).append(" ").append(questionFirstMessage);
```
Zabuzard, line 213: i think it would be easier to read if the order would be changed here. like:
```java
questionBuilder.append(title).append(" ");
String shortQuestion = question.substring(0, questionBuilder.length());
questionBuilder.append(shortQuestion);
```
Then the comment is also not needed anymore, since its understandable.

U have to see how that `if` fits into the mix though.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +143,90 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+        for (ForumTag tag : threadChannel.getAppliedTags()) {
+            if (questionBuilder.length() > MAX_QUESTION_LENGTH) {
+                break;
+            }
+            questionBuilder.insert(0, String.format("%s ", tag.getName()));
```
Zabuzard, line 218: inserts are really painful, performance-wise. id suggest u create another stringbuilder to first construct the string and then have a single `insert` into ur target, instead of multiple.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -92,7 +92,10 @@ private void handleHelpThreadCreated(ThreadChannel threadChannel) {
 
     private RestAction<Message> createMessages(ThreadChannel threadChannel) {
         return sendHelperHeadsUp(threadChannel).flatMap(Message::pin)
-            .flatMap(any -> helper.sendExplanationMessage(threadChannel));
+            .flatMap(any -> helper.sendExplanationMessage(threadChannel))
+            .flatMap(any -> threadChannel.retrieveMessageById(threadChannel.getIdLong())
+                .flatMap(message -> helper.constructChatGptAttempt(threadChannel,
+                        message.getContentRaw())));
```
Zabuzard, line 98: in the past, this helper method returned the explanation message, now it returns the chat gpt message.
please double check all its current users to make sure thats okay with them

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/Features.java
tmcdonnell2: My understanding is that this allows for only one ChatGPTService object to be created but be passed around the project. The class is not statically accessible. I could refactor it to be or if there is another workaround?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2: While I appreciate what the Optional API can do, I think how I have this setup will require some reworking to fix. I use ```useChatGptFallBackMessage```  to create a ```RestAction<Message>```. This is required since inside that function it used ```mentionGuildSlashCommand``` which creates a ```RestAction<String>```. From looking at the RestAction documentation, this is no way to get the content of the message out (beyond reflection and blocking the thread). If I were able to get it out as a String then the other method would have to be rewritten. For now, I have replaced it with a simple check if the optiona is empty and then logging and returning the fallback message. Let me know if there is some other way I should be doing this.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2:
```
questionBuilder.append(questionTitle).append(" ");
        if (originalQuestion.length() > MAX_QUESTION_LENGTH - questionBuilder.length()) {
            originalQuestion = originalQuestion.substring(0, MAX_QUESTION_LENGTH - questionBuilder.length());
        }

        questionBuilder.append(originalQuestion);
```

This is what I got after some refactoring. More or less readable?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
tmcdonnell2: You would be surprised just how fast 200 characters gets filled up. Just taking some sample questions that are on the forum right now they all are around 200 or exceed 200 characters (MAX_LENGTH as defined of now). If the person includes almost any code (and with Java being verbose) the characters are destroyed. It might even be good if we carve out code in the future and just try to get to the human language bits of the question.

The other bit is that I add the tags to the question if a question is too short thus pushing up question lengths.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
tmcdonnell2: It still returns the explanation message.  It only has one usage which is in HelpThreadCreatedListener:
```
private void handleHelpThreadCreated(ThreadChannel threadChannel) {
        helper.writeHelpThreadToDatabase(threadChannel.getOwnerIdLong(), threadChannel);

        Runnable createMessages = () -> {
            try {
                createMessages(threadChannel).queue();
            } catch (Exception e) {
                logger.error(
                        "Unknown error while creating messages after help-thread ({}) creation",
                        threadChannel.getId(), e);
            }
        };

        // The creation is delayed, because otherwise it could be too fast and be executed
        // after Discord created the thread, but before Discord send OPs initial message.
        // Sending messages at that moment is not allowed.
        SERVICE.schedule(createMessages, 5, TimeUnit.SECONDS);
}
```

This is called by another function ```onChannelCreate``` which checks to ensure that what is being created is a help thread.
Is this what you mean? I think this works since we want to send this along with each new helper thread created. The messages created also appears to be the ping to the helpers and the explanation messages. Seems like a good place to extend the functionality. But I can appreciate that it might need a new place to live. But it seems like it would still be part of the ```queue()``` mentioned in the above code. Let me know what you think.

application/src/main/java/org/togetherjava/tjbot/features/Features.java
Zabuzard: The code is fine, but all u did was changing the order:

foo

And that seems unecessary.

Zabuzard: Ah, now I see it. Fair.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +144,94 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+        StringBuilder tagBuilder = new StringBuilder();
+        int stringLength = questionBuilder.length();
+        for (ForumTag tag : threadChannel.getAppliedTags()) {
+            String tagName = tag.getName();
+            stringLength += tagName.length();
+            if (stringLength > MAX_QUESTION_LENGTH) {
+                break;
+            }
+            tagBuilder.append(String.format("%s ", tagName));
+        }
```
Tais993, line 222: replace with a stream?

```suggestion
        String tagsJoined = threadChannel.getAppliedTags().stream()
           .map(ForumTag::getName)
           .filter(tag -> tag.length() > MAX_QUESTION_LENGTH)
           .collect(Collectors.joining(" "));
```

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -210,20 +202,19 @@ private Optional<String> prepareChatGptQuestion(ThreadChannel threadChannel,
...
+            .map(ForumTag::getName)
+            .collect(Collectors.joining(" "))
+            .concat(" ");
 
-        questionBuilder.insert(0, tagBuilder);
+        String question = questionBuilder.insert(0, tagsJoined)
+            .toString()
+            .lines()
+            .limit(MAX_QUESTION_LENGTH)
+            .collect(Collectors.joining());
```
Tais993, line 215: Is this what you want, you changed the behaviour.

This limits it so there cant be more than 200 lines in the reply.
While the enhanced for checked or the ForumTag name's length was shorter than MAX_QUESTION_LENGTH

I'm not sure what's correct, but both don't make entirely sense to me, could you clarify what its meant to do?

tmcdonnell2: So this does change the behaviour slightly but I think the proposed stream missed what was going on. The length that was being checked in the old for loop was the length of the overall question, not the forum tags. We want questions to be under MAX_QUESTION_LENGTH since we don't want to burn through our ChatGPT credits.

This does change the behaviour slightly though. Ideally, I want the tags not to be added if the question exceeds the length limit. In this case, all tags are added and the length is shrunk to accommodate the tags. With this, we lose more of the context of the question (slowly by a few characters per tag). Is there a way to do this with Streams? Perhaps the old for loop may be better?

Tais993:
```java
        for (ForumTag tag : threadChannel.getAppliedTags()) {
            String tagName = tag.getName();
            stringLength += tagName.length();
            if (stringLength > MAX_QUESTION_LENGTH) {
                break;
            }
            tagBuilder.append(String.format("%s ", tagName));
        }
```

The old for loop you only checked or the tag was longer than MAX_QUESTION_LENGTH

So the old for was also incorrect? I was confused about the check already either way.

If you wanna keep the stream, the filter has to change.
`tagBuilder + tagLength <=  MAX_QUESTION_LENGTH`


Since you'd want to check whenever the tag *can be added* to the tagBuilder right? Since the tagBuilder string can't be longer than MAX_QUESTION_LENGTH

tmcdonnell2: Sorry a little bit more context - if you look before the loop you see that I declare the stringLength variable to be the length of the questionBuilder (which includes the title and first message question). That is the stringLength I am checking against. So long as that original question plus the appended tag doesn't go over MAX_QUESTION_LENGTH then it should it would add it. (Sorry on mobile so it's hard to quote things)

Tais993: You're absolutely right, I missed the "+"

I'm on phone as well atm, writing the stream nicely with this would require me to think for a second.

I could take a look Tuesday.

Might end up getting returned to a for loop tbh, just from quick thinking I think a stream would be too cumbersome.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -131,6 +151,86 @@ private RestAction<Message> sendExplanationMessage(GuildMessageChannel threadCha
...
+     */
+    RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
+            String originalQuestion) {
+
+        Optional<String> questionOptional = prepareChatGptQuestion(threadChannel, originalQuestion);
+        Optional<String> chatGPTAnswer;
+
+        String question = questionOptional.orElseThrow();
+        logger.debug("The final question sent to chatGPT: {}", question);
+        chatGPTAnswer = chatGptService.ask(question);
```
Tais993, line 174: What if the response is over the character limit? ChatGPT often gives really big responses, so I can't imagine this to not happen.

Tais993 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
marko-radosavljevic: Yup, that's already an issue with `/chatgpt` that is on master. It silently crashes if it's over the character limit.

And just cutting the answer to fit the limit is a bad solution, obviously. What's the point of getting 80% of the answer, if it's a code example or a step-by-step guide? Code won't work, guide is incomplete and useless.

So one solution is to prompt it to fit the answer within limits, and test it to make sure it works well. With a hard cutoff if a limit is reached.

Another would be to post lengthier answers as a file.

marko-radosavljevic: Yeah, that's one of my proposed solutions.

We either tell it to fit the answer withing the limit (`limit - 500` for optimal results), with a fallback. Or we let it post the full answer as a file.

I like both of them. First one is short and neat, no files. My only worry is that the quality of the answers might lack. For example, can you fit a coherent guide how to install and setup something, with examples, within the limit.

marko-radosavljevic: Tested few examples, it seems embed length is enough for most guides with examples.

My worry with embeds is that they are too narrow for code, it might ruin the formatting. Since code would be needed in almost every answer, it's something to consider.

And 2000 char limit or regular messages is likely not enough for bigger answers. Close, but not quite there.

tmcdonnell2: > Tested few examples, it seems embed length is enough for most guides with examples.
>My worry with embeds is that they are too narrow for code, it might ruin the formatting. Since code would be needed in almost every answer, it's something to consider.
>
>And 2000 char limit or regular messages is likely not enough for bigger answers. Close, but not quite there.

Thus, we should fit answers in a file?

marko-radosavljevic: Try embeds first, for cosistency. Post a picture of a more complicated answer with code examples.

Files are the last restort, since they are kinda ugly and weird UI wise. You know, getting a file in response to a bot command.

Tais993: A file is fine for PC, you have good UX then, not as good as just a message but it's fine.

On phone, yeah uh, have fun opening notepad on your phone to read it.

Issue #812, opened 2023-03-25
ChatGPT - Slash command improvements


Zabuzard:
We recently added a `/chatgpt` slashcommand. It is now time to fine-tune it:

* limit message length to prevent people from sending a huge ass message that rips all our money (maybe 200?)
* limit usages per user to maybe once per 10 seconds (can use a Caffeine cache)
* make the message optional and if its left out, a modal will popup - allowing users to enter multi-line messages, for example including code (same limits apply)

Until this issue is implemented, the slash command has to remain locked for regular users, to prevent abuse.

PR #821, opened 2023-04-11
made some changes according to issue #812

lebedevhhh:
made the max message length 200.
made it impossible to user to spam the gpt command (10 seconds delay) using Caffein cache.


lebedevhhh:
> someone else is already assigned for that work(and they also have an open pr)
tyvm ok

PR #885, opened 2023-09-17
Add transfer question feature

### LINKED ISSUE #868: Transfer question Context Command
## Motivation

## Feature

transfer

## Details

ankitsmt211:
* add command in message context
* can be used to transfer questions 
* resolves #868

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -74,7 +74,7 @@ dependencies {
 
     implementation 'com.github.ben-manes.caffeine:caffeine:3.1.1'
     
-    implementation 'org.kohsuke:github-api:1.315'
+    implementation 'org.kohsuke:github-api:1.316'
 
     testImplementation 'org.mockito:mockito-core:5.3.1'
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.10.0'


application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.regex.Pattern;
+
+public final class TransferQuestionCommand extends BotCommandAdapter
+        implements MessageContextCommand {
+    private static final Logger logger = LoggerFactory.getLogger(TransferQuestionCommand.class);
+    private static final String COMMAND_NAME = "transfer-question";
+    private static final String TRANSFER_QUESTION_TITLE_ID = "transferID";
+    private static final String TRANSFER_QUESTION_INPUT_ID = "tranferQuestion";
+    private static final String TRANSFER_QUESTION_TAG = "tags";
+    private static final String QUESTIONS_CHANNEL_ID = "1051826284008853505";
```
Taz03, line 35: don't hard code this and use longs for channel id

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /**
+     * Verifies if user has role higher or equal to Community Ambassador in terms of hierarchy.
+     */
+    public boolean hasRequiredRoles(MessageContextInteractionEvent event) {
+
+        List<Role> userRoles = event.getMember().getRoles();
+        List<Role> communityAmbassador =
+                event.getGuild().getRolesByName("COMMUNITY AMBASSADOR", true);
```
Taz03, line 62: don't hard code this either

Taz03 (changes requested):
cal u explain what this transfer feature is all about?

ankitsmt211:
> cal u explain what this transfer feature is all about?

Any one with role higher or equal to Community Ambassador, can transfer questions from any non-question channel, it will create a modal and prefill all the values. Once submitted, the question will be posted in question forums, original message will be deleted.  A dm will be sent to user as a soft warning.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Any suggestions? Get channel by name maybe?

SquidXTV:
Thanks for the work done ❤️

Zabuzard:
please share screenshots of every part of the flow. otherwise its very hard to review the UX part of things

ankitsmt211:
1. context menu of any text

transfer-feature-1

2. Sometimes modal creation fails, but you can either click on reload icon on fail message or do it again.

transfer-feature-2

3. Prefill modal menu, values can be tweaked.

transfer-feature-3

4. on Submit, new post is created in questions forum with "Java" tag as default.


transfer-feature-4

5.  Link and info about action is pasted in original channel, also you can see the original message is deleted as well.

transfer-feature-5

6. DM'd user

Screenshot 2023-09-18 151433

7. Unauthorized user tries to use this command

transfer-feature-6

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: yes, get channel name from config

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+            onSuccess.sendMessage(
+                    "Hi, you are receiving this dm because you tried to ask question in wrong chat, please use #questions for asking questions, more info #rules -Together Java")
+                .queue(onNestedSuccess -> deleteOriginalMessage(event, originalMessageId),
+                        onNestedFailure -> {
+                            logger.warn("Failed to open private channel for user of message : "
+                                    + originalMessageId, onNestedFailure);
+                            logger.info("sending message to source channel", onNestedFailure);
+                            sendMessageToSource(event, originalUser);
+                        });
+        }, onFailure -> {
```
Taz03, line 186: this is blocking

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+            .addActionRow(transferQuestionTag)
+            .build();
+
+        event.replyModal(reportModal).queue();
+    }
+
+
+    @Override
+    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
+        User author = event.getJDA().getUserById(args.get(0));
```
Taz03, line 123: `getUserById` uses cache so its never guaranteed that u'll get the user

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+        if (message.length() >= TITLE_MAX_LENGTH) {
+            int lastWordEnd = message.lastIndexOf(' ', TITLE_MAX_LENGTH);
+
+            if (lastWordEnd == -1) {
+                lastWordEnd = TITLE_MAX_LENGTH;
+            }
+
+            message = message.substring(0, lastWordEnd);
+        }
+
```
Taz03, line 138: use util  method

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+        return isTitleValid(message) ? message : "Untitled";
+    }
+
+    public static boolean isTitleValid(CharSequence title) {
+        String titleCompact = TITLE_COMPACT_REMOVAL_PATTERN.matcher(title).replaceAll("");
+
+        return titleCompact.length() >= TITLE_COMPACT_LENGTH_MIN
+                && titleCompact.length() <= TITLE_COMPACT_LENGTH_MAX
+                && !titleCompact.toLowerCase(Locale.US).contains("help");
```
Taz03, line 147:
```suggestion
                && !titleCompact.toLowerCase(Locale.US);
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+        String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
+
+        ForumChannel questionsForum = event.getJDA().getForumChannelById(QUESTIONS_CHANNEL_ID);
+
+        List<ForumTag> tags = questionsForum.getAvailableTagsByName("JAVA", true);
+
+        ForumTag javaTag = tags.get(0);
+
+        event.getJDA()
+            .getForumChannelById(QUESTIONS_CHANNEL_ID)
```
Taz03, line 163: u already have a var created  for this `quesitonsForum`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+            .createForumPost(forumTitle, forumMessage)
+            .setTags(ForumTagSnowflake.fromId(javaTag.getId()))
```
Taz03, line 165: y always javaTag

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+        TextInput transferQuestionInput = TextInput
+            .create(TRANSFER_QUESTION_INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
+            .setValue(originalMessage)
+            .setRequiredRange(3, 1000)
+            .build();
+
+        TextInput transferQuestionTag = TextInput
+            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
+            .setValue("Java")
+            .build();
```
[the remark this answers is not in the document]
Zabuzard: @Taz03 modals only support text input according to JDA doc

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: that suggestion doesnt make sense to me

Taz03: ohh sorry i meant to remove  the whole thing, coz we don't have the rule to discard  questions with "help" in their title  anymore right?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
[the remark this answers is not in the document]
Taz03: so theres no better way to do  that?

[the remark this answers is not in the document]
Taz03: Theres this [StringSelectMenu](https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/interactions/components/selections/StringSelectMenu.html) maybe we can send  it before or after the main modal

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+        event.replyModal(reportModal).queue();
+    }
+
+
+    @Override
+    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
+        User author = event.getJDA().getUserById(args.get(0));
+        createForumPost(event, author);
+
+    }
```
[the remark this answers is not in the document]
ankitsmt211: Why though? i already have message inside `createForumPost` which will link post after successful creation in source channel.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: yeah. shouldnt be blocked, i guess

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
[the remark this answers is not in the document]
Zabuzard: im not sure if u can make this an ephemeral dialog though. perhaps sth to explore, but i would also be okay if we have that in a second iteration of the feature instead, cause it sounds complex

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+            .build();
+
+        TextInput transferQuestionTag = TextInput
+            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
+            .setValue("Java")
+            .build();
+
+
+
+        String transferQuestionModalComponentID = generateComponentId(author);
```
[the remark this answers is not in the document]
ankitsmt211: also this makes sense, now the operations looks more generic with doSomethingWithMessage(args.get(i)) rather than doSomethingWithMessage(originalMessageId), which it's easy to tell what arguement is being passed on.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: removed `!titleCompact.toLowerCase(Locale.US).contains("help")`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: which one? This is something i used from earlier help system to generate title

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Earlier suggestion was to include a selection menu but didn't seam feasible, used java as it's default and most used tag. Plus this is just a feature to be used as "worst case scenario" clean up not to be used primarily. If any post do not belong to certain category, any helper can see that and change category.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+public final class TransferQuestionCommand extends BotCommandAdapter
+        implements MessageContextCommand {
+    private static final Logger logger = LoggerFactory.getLogger(TransferQuestionCommand.class);
+    private static final String COMMAND_NAME = "transfer-question";
+    private static final String TRANSFER_QUESTION_TITLE_ID = "transferID";
+    private static final String TRANSFER_QUESTION_INPUT_ID = "tranferQuestion";
+    private static final String TRANSFER_QUESTION_TAG = "tags";
+    private static final String QUESTIONS_CHANNEL_ID = "1051826284008853505";
+    private String originalMessageId;
+    private String originalChannelId;
```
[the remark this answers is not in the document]
Taz03: > why is this a bad idea? curious but ye since im already using modal args i can do that.

it'll mess up the data if 2 or more users are using the feature at the same time

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: [MessageUtils#abbreviate](https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/features/utils/MessageUtils.java#L148)

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: hmm, but this  won't work  anyways coz u can't just hard code the java tag, it is a config thing. The best option I can think of is having a selection menu

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: that method is a different one with a different behavior and purpose

Zabuzard: abbreviate cuts in the middle of a word and then adds an elipsis. like `hello wo...`.
the one bee is using here is the old code that we used for the old help system, for guessing thread titles. it cuts at word boundaries instead.
subtle difference, but better UX.

it would make sense though to offer it as helper somewhere

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Was it because of "not being able to pull user?" or the nested calls? , How about now?
image

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,213 @@
...
+    public void sendMessageToSource(ModalInteractionEvent event, User user) {
+        event.getChannel()
+            .sendMessage(user.getAsMention()
+                    + "Was unable to send DM, please use #questions for asking questions, more info #rules")
+            .queue(onSuccess -> {
+            },
+
+
+
+                    onFailure -> logger.warn("Failed to send Warning in source channel"));
```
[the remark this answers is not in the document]
ankitsmt211: is there a problem with a warn message? an action failed shouln't it be in logs

ankitsmt211:
@Taz03 you mentioned something about getting channel from config, are you suggesting "regex pattern" for helper forum? Maybe get a list of all channels and match with that to get the target channel?

ankitsmt211:
Not a fan of "three" back to back mentions are being triggered here for OP
* on creation of new post
* private DM
* link to newly created post in source channel

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: its because of nested calls

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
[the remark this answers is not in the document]
Taz03: ehh.. we're not logging on every failure

Taz03:
> @Taz03 you mentioned something about getting channel from config, are you suggesting "regex pattern" for helper forum? Maybe get a list of all channels and match with that to get the target channel?

yes

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: u could read the config. the tags there are sorted by relevance, the first one is always considered the main mega tag

config

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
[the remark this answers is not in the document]
Taz03: we're only logging on common failures, on what scenario do u think sending message will fail in the source channel? maybe discord hiccup? maybe if someone deleted the channel within a few milliseconds. those are rare and we're not logging it in other features

marko-radosavljevic:
> Not a fan of "three" back to back mentions are being triggered here for OP
> 
>     * on creation of new post
> 
>     * private DM
> 
>     * link to newly created post in source channel

I agree, let's get rid of the mention in source channel.

That way we don't annoy and confuse the newcomer with mentions all over the place. They only get mentioned where we want them to go, to a correct help thread. Additionally, we keep the source channel clean, and don't interrupt ongoing discussions.

Seems like a cleaner and more streamlined approach. :relaxed:

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: this one is still not implemented, ill do this one hopefully soon.

PR #886, opened 2023-09-17
Dynamically enable features with config file

Alathreon:
Also contain fix to disable jshell code action, but not auto chatgpt

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -101,5 +101,11 @@
         "pruneMemberAmount": 7,
         "inactivateAfterDays": 90,
         "recentlyJoinedDays": 4
+    },
+    "feature_blacklist": {
+        "normal": [
+        ],
+        "special": [
+        ]
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
@@ -1,0 +1,35 @@
+package org.togetherjava.tjbot.config;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import java.util.Set;
+
+/**
+ * Blacklist of features, use {@link FeatureBlacklist#isEnabled(T)} to test if a feature is enabled.
+ * If a feature is blacklisted, it won't be enabled by the bot, and so will be ignored.
+ * 
+ * @param <T> the type of the feature identifier
+ */
+public class FeatureBlacklist<T> {
+    private final Set<T> featureIdentifierBlacklist;
+
+    /**
+     * Creates a feature blacklist
+     * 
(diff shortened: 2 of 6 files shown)


application/config.json.template
```diff
@@ -89,10 +89,16 @@
...
     "jshell": {
         "baseUrl": "<put_jshell_rest_api_url_here>",
         "rateLimitWindowSeconds": 10,
         "rateLimitRequestsInWindow": 3
+    },
+    "feature_blacklist": {
+        "normal_feature_blacklist": [
+        ],
+        "special_feature_blacklist": [
+        ]
```
Zabuzard, line 102: redudancy in the names. its clear from context already that this is a feature blacklist. so just `normal` and `special` would be good enough at this point.

that said, from a user point of view, its kinda annoying that this is split - would it be possible to only have a single list? id prefer that then.

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
```diff
@@ -0,0 +1,34 @@
+package org.togetherjava.tjbot.config;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import java.util.Set;
+
+/**
+ * Blacklist of features, use {@link FeatureBlacklist#isEnabled(T)} to test if a feature is enabled.
```
Zabuzard, line 8: javadoc should briefly explain what a blacklisted feature is.

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
```diff
@@ -0,0 +1,34 @@
...
+ * 
+ * @param <T> the type of the feature identifier
```
Zabuzard, line 10: what is this generic? i dont think its meaningful to expose it, just set it directly to `Feature`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
Alathreon: There are two reasons:
- First is because there are Feature implementations, which are classes and so the type is Class, while special blacklist is custom and so can only use String
- Second, just in case, in the future if for example we want enums or something else

application/config.json.template
Alathreon: Fixed, tell me what you think.

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
Alathreon: Done, tell me what you think.

Zabuzard (approved):

PR #888, opened 2023-09-18
Fixed some ChatGPT issues

Zabuzard:
* disabling ChatGPT did not work properly
* ChatGPT send too much logs (leftover `info` log and double error logs)
* answer-attempt must be after the explanation messages and also in the same failure/dependency queue
* (the source code URL must not be a fake by default)


Zabuzard:
(@tmcdonnell2 for visibility)

tmcdonnell2:
Thanks for the tag!

PR #895, opened 2023-09-21
reopening transfer feature PR

### LINKED ISSUE #868: Transfer question Context Command

ankitsmt211:
Adds transfer feature
* implemented suggested changes
* fixes #868

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
@@ -9,15 +9,11 @@ import net.dv8tion.jda.api.entities.channel.forums.ForumTag;
 import net.dv8tion.jda.api.events.channel.ChannelCreateEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.dv8tion.jda.api.requests.RestAction;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.features.EventReceiver;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -27,8 +23,6 @@ import java.util.concurrent.TimeUnit;
  * user.
  */
 public final class HelpThreadCreatedListener extends ListenerAdapter implements EventReceiver {
-    private static final Logger logger = LoggerFactory.getLogger(HelpThreadCreatedListener.class);
-    private static final ScheduledExecutorService SERVICE = Executors.newScheduledThreadPool(2);
 
     private final HelpSystemHelper helper;
     private final Cache<Long, Instant> threadIdToCreatedAtCache = Caffeine.newBuilder()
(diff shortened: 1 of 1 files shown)


ankitsmt211:
i believe the it's failing because of 
image

Zabuzard:
rebase please

ankitsmt211:
Linking screenshots from the original PR, all the steps are intact except step 7 and no more pings when linking new post in source channel, only username is mentioned(since a DM and a mention in new post is already done).
 https://github.com/Together-Java/TJ-Bot/pull/885#issuecomment-1723088379

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        List<ForumTag> tags = questionsForum.getAvailableTagsByName("JAVA", true);
+        ForumTag javaTag = tags.get(0);
+
+        questionsForum.createForumPost(forumTitle, forumMessage)
+            .setTags(ForumTagSnowflake.fromId(javaTag.getId()))
+            .map(ForumPost::getMessage)
+            .onSuccess(message -> message.reply("original post from" + originalUser.getAsMention())
+                .queue(onSuccess -> {
+                }, onFailure -> logger.warn("Failed to mention user in transferred post",
+                        onFailure)))
```
Taz03, line 135: don't log here

Taz03 (changes requested):

ankitsmt211:
I'm thinking to refactor the process, 
Earlier
* Create post
* Mention OP
* DM OP
* link to new post with OP's name in source channel

Suggested
* Create Post
* Mention OP in post
* DM OP

Reasons: Since Op is mentioned twice during the process, they are already redirected to new post. There's no benefit to posting link to new post apart from possibily adding unwanted text source channel along with promotion of newly created post.

Taz03:
Agree, but i'd say give the creator an ephemeral message in source channel linking the forum post

ankitsmt211:
> Agree, but i'd say give the creator an ephemeral message in source channel linking the forum post

Seems like ephermal messages can be only sent to interactions, but OP is not the one interacting. So any ideas with that?

ankitsmt211:
> Agree, but i'd say give the creator an ephemeral message in source channel linking the forum post

Do you mean maybe whoever is doing the transfer? Since that makes sense if they were able to make the transfer complete?

marko-radosavljevic:
> I'm thinking to refactor the process

Agree, will copy my response from previous PR:
> Not a fan of "three" back to back mentions are being triggered here for OP
> 
>     * on creation of new post
> 
>     * private DM
> 
>     * link to newly created post in source channel

I agree, let's get rid of the mention in source channel.

That way we don't annoy and confuse the newcomer with mentions all over the place. They only get mentioned where we want them to go, to a correct help thread. Additionally, we keep the source channel clean, and don't interrupt ongoing discussions.

Seems like a cleaner and more streamlined approach. :relaxed:

marko-radosavljevic:
> Seems like ephermal messages can be only sent to interactions, but OP is not the one interacting. So any ideas with that?

Yeah, as far as I know, it's not possible to do something like that.

> Do you mean maybe whoever is doing the transfer? Since that makes sense if they were able to make the transfer complete?

An ephemeral message to the person doing the transfer is probably not needed, they already know what happened and why. There isn't much to notify them about, either.

ankitsmt211:
I was able to squeeze all of it into "one" chain, and make functional. Failed DM's are handled inside `dmUser` using `ErrorFlatMap`.
Apart from that, should i log anything else from this chain? 

  ```java  
@Override
    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
        event.getJDA()
            .retrieveUserById(args.get(0))
          .map(user->user)
          .flatMap(fetchedUser->createForumPost(event, fetchedUser))
          .flatMap(user->dmUser(event.getChannel(), user))
          .flatMap(deleted->event.reply("transferred to"+deleted.getAsMention()).setEphemeral(true))
          .flatMap(info->deleteOriginalMessage(event.getJDA(),args.get(2),args.get(1)))
            .queue();
    }
```
>  .flatMap(deleted->event.reply("transferred to"+deleted.getAsMention()).setEphemeral(true))

Needed to reply to event, otherwise modal menu wouldn't close and showed error at the top. 

if this is good to go, i can commit and push for any further tweaks.

PR #899, opened 2023-09-23
added required changes for logging blacklisted features

adiChoudhary:
added a boolean check to log blacklisted features

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
@@ -35,11 +35,7 @@ public class FeatureBlacklist<T> {
      * @return true if a feature is enabled, false otherwise
      */
     public boolean isEnabled(T featureId) {
-        boolean isBlackListed = featureIdentifierBlacklist.contains(featureId);
-        if (isBlackListed) {
-            logger.info("Feature {} is disabled", featureId);
-        }
-        return !isBlackListed;
+        return !featureIdentifierBlacklist.contains(featureId);
     }
 
     public <F> Stream<F> disableMatching(Stream<F> features, Function<F, T> idExtractor) {
--- a/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
@@ -1,6 +1,8 @@
 package org.togetherjava.tjbot.config;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Set;
 
@@ -12,6 +14,7 @@ import java.util.Set;
  */
 public class FeatureBlacklist<T> {
     private final Set<T> featureIdentifierBlacklist;
+    private static final Logger logger = LoggerFactory.getLogger(FeatureBlacklist.class);
 
     /**
      * Creates a feature blacklist
@@ -30,6 +33,10 @@ public class FeatureBlacklist<T> {
      * @return true if a feature is enabled, false otherwise
      */
     public boolean isEnabled(T featureId) {
-        return !featureIdentifierBlacklist.contains(featureId);
+        boolean isBlackListed = featureIdentifierBlacklist.contains(featureId);
+        if (isBlackListed) {
+            logger.info(String.format("%s is blacklisted 😥", featureId));
+        }
+        return !isBlackListed;
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
@@ -5,6 +5,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
```diff
@@ -30,6 +33,10 @@ public FeatureBlacklist(Set<T> featureIdentifierBlacklist) {
      * @return true if a feature is enabled, false otherwise
      */
     public boolean isEnabled(T featureId) {
-        return !featureIdentifierBlacklist.contains(featureId);
+        boolean isBlackListed = featureIdentifierBlacklist.contains(featureId);
+        if (isBlackListed) {
+            logger.info(String.format("%s is blacklisted 😥", featureId));
```
Zabuzard, line 38: * use the loggers own formatting api, not urs (otherwise the string is also computed if info is disabled) - `logger.info("{} is blacklisted", featureId)`
* id change the text slightly for better UX into `Feature {} is disabled.`

Zabuzard (changes requested):
good stuff, thanks 😄 and welcome to the project 👍

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
adiChoudhary: 👍

Alathreon (changes requested):
An idea  would be make so the blacklist itself filter the features, so logging would be easier:
isEnabled would be left untouched, and a new method would be added
```java
class FeatureBlacklist<T> {
  ...
  public Stream<F> disableMacthing(Stream<F> features, Function<T, F> idExtractor) {
    return features.filter(f -> {
      T id = idExtractor.apply(f);
      if(this.isEnabled(id)) {
        return true;
      } else {
        logger.info("{} is blacklisted 😥", id);
        return false;
      }
    });
  }
}
```

Features.java use case:
```java
return blacklist.disableMacthing(features.stream(), Feature::getClass).toList();
```

Code actions use case:
```java
List<CodeAction> codeActions =
        blacklist.disableMacthing(
          Stream.of(new FormatCodeCommand(), new EvalCodeCommand(jshellEval),
          f -> f.getClass().getName()
        ).toList();
```

adiChoudhary:
@Alathreon incorporating the above change would introduce changes in class
image
This would require refactoring of connected classes. Should we do that
image
was thinking of creating a simple decorator to add logging functionality
image
We can do something like above also, this would allow us to easily supply any filtration logic, keep our logging intact

Alathreon:
@adiChoudhary 
> incorporating the above change would introduce changes in class

F isn't a generic that you put on the class, but I indeed forgot to declafre it correctly, I modified my review message, it should be declared on the method directly.

> This would require refactoring of connected classes. Should we do that

No

> We can do something like above also, this would allow us to easily supply any filtration logic, keep our logging intact
This code is the same as the previous isEnabled idea, which we decided to not follow, so no

adiChoudhary:
@Alathreon, @Zabuzard  added suggested changes, please review

Alathreon:
@adiChoudhary you didn't reverse the changes you did to isEnabled, and did you verify that it works ?

adiChoudhary:
sorry thats why i was seeing 2 log statements, will do that

adiChoudhary:
Reverted it back now

application/src/main/java/org/togetherjava/tjbot/config/FeatureBlacklist.java
```diff
@@ -32,4 +37,15 @@ public FeatureBlacklist(Set<T> featureIdentifierBlacklist) {
     public boolean isEnabled(T featureId) {
         return !featureIdentifierBlacklist.contains(featureId);
     }
+
```
java-coding-prodigy, line 40: Please add Java doc here.

application/src/main/java/org/togetherjava/tjbot/features/Features.java
```diff
@@ -159,6 +159,6 @@ public static Collection<Feature> createFeatures(JDA jda, Database database, Con
         features.add(new JShellCommand(jshellEval));
 
         FeatureBlacklist<Class<?>> blacklist = blacklistConfig.normal();
-        return features.stream().filter(f -> blacklist.isEnabled(f.getClass())).toList();
+        return blacklist.disableMatching(features.stream(), Feature::getClass).toList();
```
java-coding-prodigy, line 162: IMO this should be made `Object::getClass`

java-coding-prodigy (changes requested):
Since we won’t be using `isEnabled` outside of the `disableMatching` method, maybe we could in-line and remove it?

ankitsmt211:
There's a new PR #980 that seems to be targetting the same issue, i assume this is not active anymore? Perhaps a polite reminder to author of new PR to check issues with maintainers beforehand. Otherwise there work might be a waste of effort.

marko-radosavljevic:
Is this PR nearly done, or it should be closed in favour of #980? Both seem to be inactive, and this one is at least passing all checks.

I mean, this looks fine, maybe slightly overcomplicated since project should be readable by beginners, but I think we are long  past that point hah. It's just a few lines anyway, seems like there is not a good reason to drag this out for 2 more months. ^^

Taz03:
merged #1048

PR #901, opened 2023-09-24
ChatGPT response Dismiss button

### LINKED ISSUE #897: Dismiss Button for ChatGPT answer
Sometimes ChatGPTs automatic-answer-attempt is not helpful.

We should add a button attached to the end of his message(s) that allows certain people to `Dismiss` it (delete the message).
The message is created somewhere around [HelpSystemHelper.java#L163C25-L193](https://github.com/Together-Java/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java#L163C25-L193)

Code-wise, the challenge is that ChatGPT can post multiple messages. To delete all of them, we can simply remember all those message IDs within the buttons argument-list. See the existing GIST Dismiss button for details (`FileSharingMessageListener.java`).

People allows to click dismiss would include:
- OP
- anyone with the `"tagManageRolePattern": "Moderator|Community Ambassador|Top Helpers .+",` role

SquidXTV:
# About
closes #897 

Adds the dismiss button to the chatgpt response:
image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -188,19 +188,18 @@ public final class HelpSystemHelper {
                             """::formatted)
                     .flatMap(threadChannel::sendMessage)
                     .onSuccess(m -> ids.add(m.getId()));
-        String[] answers = chatGPTAnswer.get();
+        String[] answers = chatGPTAnswer.orElseThrow();
 
         for (int i = 0; i < answers.length; i++) {
-            MessageCreateAction messageCreateAction = threadChannel.sendMessage(answers[i]);
+            MessageCreateAction answer = threadChannel.sendMessage(answers[i]);
 
             if (i == answers.length - 1) {
-                message = message.flatMap(ignored -> messageCreateAction
+                message = message.flatMap(any -> answer
                     .addActionRow(generateDismissButton(componentIdInteractor, ids)));
                 continue;
             }
 
-            message = message
-                .flatMap(ignored -> messageCreateAction.onSuccess(m -> ids.add(m.getId())));
+            message = message.flatMap(ignored -> answer.onSuccess(m -> ids.add(m.getId())));
         }
 
         return message;
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -188,22 +188,28 @@ public final class HelpSystemHelper {
                             """::formatted)
                     .flatMap(threadChannel::sendMessage)
                     .onSuccess(m -> ids.add(m.getId()));
-
         String[] answers = chatGPTAnswer.get();
+
         for (int i = 0; i < answers.length; i++) {
+            MessageCreateAction messageCreateAction = threadChannel.sendMessage(answers[i]);
+
             if (i == answers.length - 1) {
-                message = message.map(answers[i]::formatted)
-                    .flatMap(m -> threadChannel.sendMessage(m)
-                        .addActionRow(Button.danger(componentIdInteractor
-                            .generateComponentId(ids.toArray(String[]::new)), "Dismiss")));
+                message = message.flatMap(ignored -> messageCreateAction
+                    .addActionRow(generateDismissButton(componentIdInteractor, ids)));
                 continue;
             }
-            message = message.map(answers[i]::formatted)
-                .flatMap(threadChannel::sendMessage)
-                .onSuccess(m -> ids.add(m.getId()));
+
+            message = message
+                .flatMap(ignored -> messageCreateAction.onSuccess(m -> ids.add(m.getId())));
         }
 
         return message;
+    }
+
+    private Button generateDismissButton(ComponentIdInteractor componentIdInteractor,
+            List<String> ids) {
+        String buttonId = componentIdInteractor.generateComponentId(ids.toArray(String[]::new));
+        return Button.danger(buttonId, "Dismiss");
     }
 
     private Optional<String> prepareChatGptQuestion(ThreadChannel threadChannel,
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -1,10 +1,7 @@
 package org.togetherjava.tjbot.features.help;
 
 import net.dv8tion.jda.api.EmbedBuilder;
-import net.dv8tion.jda.api.entities.Guild;
-import net.dv8tion.jda.api.entities.Message;
-import net.dv8tion.jda.api.entities.MessageEmbed;
-import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.entities.channel.attribute.IThreadContainer;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
@@ -12,6 +9,7 @@ import net.dv8tion.jda.api.entities.channel.forums.ForumTag;
 import net.dv8tion.jda.api.entities.channel.forums.ForumTagSnowflake;
 import net.dv8tion.jda.api.entities.channel.middleman.GuildChannel;
 import net.dv8tion.jda.api.entities.channel.middleman.GuildMessageChannel;
+import net.dv8tion.jda.api.interactions.components.buttons.Button;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
 import net.dv8tion.jda.api.utils.FileUpload;
@@ -26,6 +24,7 @@ import org.togetherjava.tjbot.db.generated.tables.HelpThreads;
 import org.togetherjava.tjbot.db.generated.tables.records.HelpThreadsRecord;
 import org.togetherjava.tjbot.features.chatgpt.ChatGptCommand;
 import org.togetherjava.tjbot.features.chatgpt.ChatGptService;
+import org.togetherjava.tjbot.features.componentids.ComponentIdInteractor;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


SquidXTV:
Two questions:
1. Is the concurrent arraylist needed in `HelpSystemHelper` or would a normal one work as well?
2. The `event.getMember` seems to be `@Nullable`, should I care? because the filesharing feature didnt care as well

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -176,17 +178,29 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
+
+        String[] answers = chatGPTAnswer.get();
+        for (int i = 0; i < answers.length; i++) {
+            if (i == answers.length - 1) {
+                message = message.map(answers[i]::formatted)
+                    .flatMap(m -> threadChannel.sendMessage(m)
+                        .addActionRow(Button.danger(componentIdInteractor
+                            .generateComponentId(ids.toArray(String[]::new)), "Dismiss")));
+                continue;
+            }
```
Zabuzard, line 200: the order of that code section tripped me hard. please put a comment to explain what the block is about

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -123,4 +135,47 @@ private RestAction<Message> sendHelperHeadsUp(ThreadChannel threadChannel) {
...
+    @Override
+    public void acceptComponentIdGenerator(ComponentIdGenerator generator) {
+        componentIdInteractor.acceptComponentIdGenerator(generator);
+    }
+
+    @Override
+    public void onButtonClick(ButtonInteractionEvent event, List<String> args) {
+        ThreadChannel channel = event.getChannel().asThreadChannel();
+        Member interactionUser = event.getMember();
+        if (channel.getOwnerIdLong() != interactionUser.getIdLong()
```
Zabuzard, line 158: please add a comment to make it easier to understand which button this is all about. this class is not a chatgpt class after all, but the general help thread thing.

tbh, im really not a fan of how the chatgpt thing is mixed into the helpthread class, instead of having some sort of extra class (not ur fault though)

Zabuzard (changes requested):
1. Is the concurrent arraylist needed in HelpSystemHelper or would a normal one work as well?

Yes, the lambda can be invoked from multiple different threads and also concurrently.

2. The event.getMember seems to be @Nullable, should I care? because the filesharing feature didnt care as well

What does the javadoc (or source code) say? When will it be null? Once you checked that, is the situation a situation u want to support, or do you consider the situation a bug/impossible?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
SquidXTV: I'm also not a fan of this, would it be possible to split it into a different class?

Zabuzard: sure. but its totally okay if you dont want to do this in this PR. just make sure to add comments instead then

SquidXTV:
It says this:
> This is null if the interaction is not from a guild.

What is a guild in this context?

Zabuzard:
> It says this:
> 
> > This is null if the interaction is not from a guild.
> 
> What is a guild in this context?

Guild = Server

Think about buttons in a DM only conversation. Then you have no server-context.

Member = User + in a Server

There is also `event.getUser()` to get it without the server-context.

ankitsmt211:
> What is a guild in this context?
Guild refers to "server" from what I know.

SquidXTV:
Though I need a Member to check the roles.
Can I ignore it because it will not be null

Zabuzard:
> Though I need a Member to check the roles. Can I ignore it because it will not be null

You are wording this wrong. It is not about ignoring it. It is about explicitly demanding that it must not be null.

I.e. if it is null, you consider it as fatal error. And what should happen for fatal errors? Crash! So implicitly invoking a NPE is a deliberate choice and the correct design.

SquidXTV:
> > Though I need a Member to check the roles. Can I ignore it because it will not be null
> 
> You are wording this wrong. It is not about ignoring it. It is about explicitly demanding that it must not be null.
> 
> I.e. if it is null, you consider it as fatal error / programmer bug. And what should happen for bugs? Crash! So implicitly invoking a NPE is a deliberate choice and the correct design.
> 
> Although unecessary, you can make this even more explicit by saying `Objects.requireNonNull(foo).bar()` instead of `foo.bar()`.

yeah sorry I was in school lol
didnt have much time thinking about how to write it, thanks

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
SquidXTV: tbh I would encapsulate all the ChatGPT stuff out of the HelpSystem classes, it doesn't really fit in there ngl
but if moving all of it, it makes more sense to make a separate pr for that

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -176,17 +178,29 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
+            message = message.map(answers[i]::formatted)
+                .flatMap(threadChannel::sendMessage)
+                .onSuccess(m -> ids.add(m.getId()));
```
Zabuzard, line 203: 2 remarks:
* i dont see `ids.add(..)` on the last message, intentional?
* i dont like the code duplication of the "map formatted, send" part. and it kinda trips me when reading the code for the first time, since i wouldnt know what that block is about (adding the dismiss button)

u can get rid of the code duplication by refactoring it sth like this:
```java
message = message.map(answers[i]::formatted)
  .flatMap(threadChannel.sendMessage);

if (i == answers.length - 1) {
  message = message.addActionRow(...);
}

message = message.onSuccess(...);
```

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
SquidXTV: ah ok yeah that makes more sense

SquidXTV: and yeah the missing onSuccess on the last message is intentional, I get it through the Button event later

Zabuzard: fair. with my changes, u would likely get it in the list as well. but shouldnt hurt :)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -176,22 +178,40 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
                 mentionGuildSlashCommand(threadChannel.getGuild(), ChatGptCommand.COMMAND_NAME)
                     .map("""
                             Here is an AI assisted attempt to answer your question 🤖. Maybe it helps! \
                             In any case, a human is on the way 👍. To continue talking to the AI, you can use \
                             %s.
                             """::formatted)
-                    .flatMap(threadChannel::sendMessage);
+                    .flatMap(threadChannel::sendMessage)
+                    .onSuccess(m -> ids.add(m.getId()));
+        String[] answers = chatGPTAnswer.get();
```
Zabuzard, line 191: `get()` is deprecated in favor of `orElseThrow()`

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -176,22 +178,40 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
+
+        for (int i = 0; i < answers.length; i++) {
+            MessageCreateAction messageCreateAction = threadChannel.sendMessage(answers[i]);
```
Zabuzard, line 194: better name might be `partialMessage` or `messagePart`

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -176,22 +178,40 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
+
+            if (i == answers.length - 1) {
+                message = message.flatMap(ignored -> messageCreateAction
```
Zabuzard, line 197: more idiomatic for `ignored` is usually `_` (or if that doesnt work), `any`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
SquidXTV: not my code, but will fix it

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
SquidXTV: true forgot to change it after testing if it even works

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
SquidXTV: why though if there is a `isEmpty` check before

Tais993: `Get` does precisely the same, only difference is that `orElseThrow` has an objectively better naming.

If you read the documentation, it also states that `orElseThrow` should be used instead.

SquidXTV: >  only difference is that `orElseThrow` has an objectively better naming.
why though lol, it will never throw so this name is kinda misleading imo

Zabuzard: its not misleading. it indicates that u consider the fact that it cant be empty at this stage a critical bug. so its precisely what u want to indicate exactly that.

Zabuzard: like, consider it a `getOrCrash()`. meaning: if its not there, theres a bad bug in the code. in other terms: hey reader, i expect this to be filled at this point.

Zabuzard (approved):

Zabuzard:
has been approved since 2 weeks. will be merged now.

PR #902, opened 2023-09-24
Bump Java Version to 20

java-coding-prodigy:
Bump Java Version of the bot.
Please reflect the same on the website too


java-coding-prodigy:
Also had to upgrade gradle to `8.3` so there are a few changes from the upgrade as well

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/mathcommands/wolframalpha/WolframAlphaImages.java
```diff
@@ -66,10 +65,10 @@ static BufferedImage renderSubPod(SubPod subPod) {
...
         try {
-            graphics.drawImage(ImageIO.read(new URL(sourceImage.getSource())), IMAGE_MARGIN_PX,
-                    IMAGE_MARGIN_PX, null);
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+            graphics.drawImage(ImageIO.read(new URI(sourceImage.getSource()).toURL()),
+                    IMAGE_MARGIN_PX, IMAGE_MARGIN_PX, null);
+        } catch (IOException | URISyntaxException e) {
+            throw new RuntimeException(e);
         }
```
Taz03, line 72: ig u'll have to catch and throw both separately

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/mathcommands/wolframalpha/WolframAlphaImages.java
Zabuzard: why are these code changes in this PR? are they required for java 20?

Zabuzard (changes requested):

java-coding-prodigy:
The changes in `WolframAlphaImages.java` are required because `URL#<init>(String` is deprecated in Java 20 and the suggested alternative is to use `URI#toURL`.
This also changes the exception handling a little bit but functionally it is the same as what we were doing previously.

Zabuzard:
make spotless and codeql happy please

java-coding-prodigy:
> make spotless and codeql happy please

For CodeQL to be fixed we need #903 
Spotless is also having some weird error while trying to build, I’ll have to check that.

java-coding-prodigy:
> > make spotless and codeql happy please
> 
> For CodeQL to be fixed we need #903 Spotless is also having some weird error while trying to build, I’ll have to check that.

Apparently that isn’t the issue.
For some reason Gradle 8.3 is not able to run files compiled by Java 20 even though it is supposed to support Java 20
image

Taz03 (approved):

Zabuzard (approved):

PR #905, opened 2023-09-28
Transfer Feature reopen

### LINKED ISSUE #868: Transfer question Context Command

ankitsmt211:
* Adds a transfer feature for moderation team
* when used, creates a new post in helper forum
* DM user about this infraction
* Delete original message from source channel
* Sends original message to source on dm fail
* resolves #868 

 on a sidenote, third time is the charm

**UPDATED**
**Step 1: Text in chat**

text-sent

**Step 2: Transfer Command**

select-transfer-command

**Step 3: Modal prefilled with values**
image

**Step 4: selecting transfer-question creates a new post in helper forum**

image

**Step 5: followed by DMing User (contains guildName at last)**

image

**Step 6:  deletes original Message**

**Incase of failed DM:**

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -173,22 +173,18 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, ForumPost forumPost,
             Guild guild) {
 
+        String messageTemplate =
+                """
+                        Hello %s 👋 You have asked a question in the wrong channel %s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
+
+                        Your question has been automatically transferred to %s, please continue there, thank you 👍
+                        """;
+
         return forumPost.author.openPrivateChannel()
-            .flatMap(channel -> channel.sendMessage(
-                    """
-                            Hello 👋 You have asked a question on %s in the wrong channel. Not a big deal, but none of the experts who could help you are reading your question there 🙁.
-
-                            Your question has been automatically transferred to %s, please continue there. You might also want to give #welcome a quick read, thank you 👍.
-                            """
-                        .formatted(guild.getName(), forumPost.message.getJumpUrl())))
-            .onErrorFlatMap(error -> sourceChannel.sendMessage(
-                    """
-                            Hello %s 👋 You have asked a question in the wrong channel. Not a big deal, but none of the experts who could help you are reading your question there 🙁.
-
-                            Your question has been automatically transferred to %s, please continue there. You might also want to give #welcome a quick read, thank you 👍.
-                            """
-                        .formatted(forumPost.author.getAsMention(),
-                                forumPost.message.getJumpUrl())));
+            .flatMap(channel -> channel.sendMessage(messageTemplate.formatted("",
+                    "on " + guild.getName(), forumPost.message.getJumpUrl())))
+            .onErrorFlatMap(error -> sourceChannel.sendMessage(messageTemplate
+                .formatted(forumPost.author.getAsMention(), "", forumPost.message.getJumpUrl())));
     }
 
     private RestAction<Void> deleteOriginalMessage(JDA jda, String channelId, String messageId) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -156,12 +156,13 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String transferQuestionTag = event.getValue(TRANSFER_QUESTION_TAG).getAsString();
 
         ForumChannel questionsForum = getHelperForum(event.getJDA());
+        String mostCommonTag = defaultTags.get(0);
 
         String queryTag = StringDistances.closestMatch(transferQuestionTag, defaultTags)
-            .orElse(defaultTags.get(0));
+            .orElse(mostCommonTag);
 
         ForumTag defaultTag = getDefaultTagOr(questionsForum.getAvailableTagsByName(queryTag, true),
-                () -> questionsForum.getAvailableTagsByName(defaultTags.get(0), true).get(0));
+                () -> questionsForum.getAvailableTagsByName(mostCommonTag, true).get(0));
 
         return questionsForum.createForumPost(forumTitle, forumMessage)
             .setTags(ForumTagSnowflake.fromId(defaultTag.getId()))
@@ -199,8 +200,8 @@ public final class TransferQuestionCommand extends BotCommandAdapter
             .filter(forumChannel -> isHelpForumName.test(forumChannel.getName()))
             .findFirst();
 
-        return forumChannelOptional
-            .orElseThrow(() -> new RuntimeException("Helper Forum Not found"));
+        return forumChannelOptional.orElseThrow(() -> new IllegalStateException(
+                "Did not find the helper-forum while trying to transfer a question. Make sure the config is setup properly."));
     }
 
     private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -24,10 +24,12 @@ import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.features.BotCommandAdapter;
 import org.togetherjava.tjbot.features.CommandVisibility;
 import org.togetherjava.tjbot.features.MessageContextCommand;
+import org.togetherjava.tjbot.features.utils.StringDistances;
 
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
 public final class TransferQuestionCommand extends BotCommandAdapter
@@ -43,6 +45,8 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final int TITLE_COMPACT_LENGTH_MAX = 30;
     private final Predicate<String> isHelpForumName;
 
+    private final List<String> defaultTags;
+
 
     /**
      * Creates a new instance.
@@ -52,6 +56,8 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
         isHelpForumName =
                 Pattern.compile(config.getHelpSystem().getHelpForumPattern()).asMatchPredicate();
+
+        defaultTags = config.getHelpSystem().getCategories();
     }
 
     @Override
@@ -135,15 +141,19 @@ public final class TransferQuestionCommand extends BotCommandAdapter
             .fromContent(event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString());
         String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
         ForumChannel questionsForum = getHelperForum(event.getJDA());
-        List<ForumTag> tags = questionsForum.getAvailableTagsByName("JAVA", true);
-        ForumTag javaTag = tags.get(0);
+
+        String transferQuestionTag = event.getValue(TRANSFER_QUESTION_TAG).getAsString();
+
+        String queryTag = StringDistances.closestMatch(transferQuestionTag, defaultTags).orElse(defaultTags.get(0));
+
+        ForumTag defaultTag = getDefaultTagOr(questionsForum.getAvailableTagsByName(queryTag, true),
+                () -> questionsForum.getAvailableTagsByName(defaultTags.get(0), true).get(0));
 
         return questionsForum.createForumPost(forumTitle, forumMessage)
-            .setTags(ForumTagSnowflake.fromId(javaTag.getId()))
+            .setTags(ForumTagSnowflake.fromId(defaultTag.getId()))
             .map(ForumPost::getMessage)
             .flatMap(message -> message.reply("Original Post from " + originalUser.getAsMention()))
             .map(sent -> originalUser);
-
     }
 
     private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, User originalUser,
@@ -163,8 +173,6 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         return jda.getTextChannelById(channelId).deleteMessageById(messageId);
     }
 
-
-
     private ForumChannel getHelperForum(JDA jda) {
         Optional<ForumChannel> forumChannelOptional = jda.getForumChannels()
             .stream()
@@ -173,5 +181,10 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
         return forumChannelOptional
             .orElseThrow(() -> new RuntimeException("Helper Forum Not found"));
+    }
+
+    private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
+            Supplier<ForumTag> defaultTag) {
+        return tagsFoundOnForum.isEmpty() ? defaultTag.get() : tagsFoundOnForum.get(0);
     }
 }
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -100,10 +100,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
             .addActionRow(modalTag)
             .build();
 
-        event.replyModal(transferModal)
-            .queue(success -> logger.debug(
-                    "{} with id: {}  triggered the transfer action on forumPost with id: {}",
-                    event.getUser().getName(), event.getUser().getId(), originalMessageId));
+        event.replyModal(transferModal).queue();
     }
 
     @Override
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -39,9 +39,9 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         implements MessageContextCommand {
     private static final Logger logger = LoggerFactory.getLogger(TransferQuestionCommand.class);
     private static final String COMMAND_NAME = "transfer-question";
-    private static final String TRANSFER_QUESTION_TITLE_ID = "transferID";
-    private static final String TRANSFER_QUESTION_INPUT_ID = "transferQuestion";
-    private static final String TRANSFER_QUESTION_TAG = "tags";
+    private static final String TITLE_ID = "transferID";
+    private static final String INPUT_ID = "transferQuestion";
+    private static final String TAG = "tags";
     private static final int TITLE_MAX_LENGTH = 50;
     private static final Pattern TITLE_COMPACT_REMOVAL_PATTERN = Pattern.compile("\\W");
     private static final int TITLE_COMPACT_LENGTH_MIN = 2;
@@ -71,32 +71,29 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = defaultTags.get(0);
 
-        TextInput transferQuestionTitle =
-                TextInput.create(TRANSFER_QUESTION_TITLE_ID, "Title", TextInputStyle.SHORT)
-                    .setMaxLength(70)
-                    .setMinLength(8)
-                    .setValue(createTitle(originalMessage))
-                    .build();
-
-        TextInput transferQuestionInput = TextInput
-            .create(TRANSFER_QUESTION_INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
-            .setValue(originalMessage)
-            .setRequiredRange(3, 2000)
+        TextInput title = TextInput.create(TITLE_ID, "Title", TextInputStyle.SHORT)
+            .setMaxLength(70)
+            .setMinLength(4)
+            .setValue(createTitle(originalMessage))
             .build();
 
-        TextInput transferQuestionTag = TextInput
-            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
+        TextInput input =
+                TextInput.create(INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
+                    .setValue(originalMessage)
+                    .setRequiredRange(3, 2000)
+                    .build();
+
+        TextInput tag = TextInput.create(TAG, "Transfer question tags", TextInputStyle.SHORT)
             .setValue(mostCommonTag)
             .build();
 
-        String transferQuestionModalComponentID =
+        String modalComponentID =
                 generateComponentId(authorId, originalMessageId, originalChannelId);
-        Modal transferModal =
-                Modal.create(transferQuestionModalComponentID, "transfer question menu")
-                    .addActionRow(transferQuestionTitle)
-                    .addActionRow(transferQuestionInput)
-                    .addActionRow(transferQuestionTag)
-                    .build();
+        Modal transferModal = Modal.create(modalComponentID, "transfer question menu")
+            .addActionRow(title)
+            .addActionRow(input)
+            .addActionRow(tag)
+            .build();
 
         event.replyModal(transferModal)
             .queue(success -> logger.debug(
@@ -144,7 +141,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
     private RestAction<ForumPost> createForumPost(ModalInteractionEvent event, User originalUser) {
 
-        String originalMessage = event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString();
+        String originalMessage = event.getValue(INPUT_ID).getAsString();
 
         MessageEmbed embedForPost = makeEmbedForPost(originalUser, originalMessage);
 
@@ -153,8 +150,8 @@ public final class TransferQuestionCommand extends BotCommandAdapter
             .setEmbeds(embedForPost)
             .build();
 
-        String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
-        String transferQuestionTag = event.getValue(TRANSFER_QUESTION_TAG).getAsString();
+        String forumTitle = event.getValue(TITLE_ID).getAsString();
+        String transferQuestionTag = event.getValue(TAG).getAsString();
 
         ForumChannel questionsForum = getHelperForum(event.getJDA());
         String mostCommonTag = defaultTags.get(0);
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -42,10 +42,10 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final String TITLE_ID = "transferID";
     private static final String INPUT_ID = "transferQuestion";
     private static final String TAG = "tags";
-    private static final int TITLE_MAX_LENGTH = 50;
-    private static final Pattern TITLE_COMPACT_REMOVAL_PATTERN = Pattern.compile("\\W");
-    private static final int TITLE_COMPACT_LENGTH_MIN = 2;
-    private static final int TITLE_COMPACT_LENGTH_MAX = 30;
+    private static final int TITLE_GUESS_MAX_LENGTH = 50;
+    private static final Pattern TITLE_GUESS_COMPACT_REMOVAL_PATTERN = Pattern.compile("\\W");
+    private static final int TITLE_GUESS_COMPACT_LENGTH_MIN = 2;
+    private static final int TITLE_GUESS_COMPACT_LENGTH_MAX = 30;
     private final Predicate<String> isHelpForumName;
     private final List<String> defaultTags;
     private static final Color EMBED_COLOR = new Color(50, 164, 168);
@@ -119,11 +119,11 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     }
 
     private static String createTitle(String message) {
-        if (message.length() >= TITLE_MAX_LENGTH) {
-            int lastWordEnd = message.lastIndexOf(' ', TITLE_MAX_LENGTH);
+        if (message.length() >= TITLE_GUESS_MAX_LENGTH) {
+            int lastWordEnd = message.lastIndexOf(' ', TITLE_GUESS_MAX_LENGTH);
 
             if (lastWordEnd == -1) {
-                lastWordEnd = TITLE_MAX_LENGTH;
+                lastWordEnd = TITLE_GUESS_MAX_LENGTH;
             }
 
             message = message.substring(0, lastWordEnd);
@@ -133,10 +133,10 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     }
 
     private static boolean isTitleValid(CharSequence title) {
-        String titleCompact = TITLE_COMPACT_REMOVAL_PATTERN.matcher(title).replaceAll("");
+        String titleCompact = TITLE_GUESS_COMPACT_REMOVAL_PATTERN.matcher(title).replaceAll("");
 
-        return titleCompact.length() >= TITLE_COMPACT_LENGTH_MIN
-                && titleCompact.length() <= TITLE_COMPACT_LENGTH_MAX;
+        return titleCompact.length() >= TITLE_GUESS_COMPACT_LENGTH_MIN
+                && titleCompact.length() <= TITLE_GUESS_COMPACT_LENGTH_MAX;
     }
 
     private RestAction<ForumPost> createForumPost(ModalInteractionEvent event, User originalUser) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -48,7 +48,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final int TITLE_GUESS_COMPACT_LENGTH_MAX = 30;
     private static final Color EMBED_COLOR = new Color(50, 164, 168);
     private final Predicate<String> isHelpForumName;
-    private final List<String> defaultTags;
+    private final List<String> tags;
 
 
     /**
@@ -62,7 +62,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         isHelpForumName =
                 Pattern.compile(config.getHelpSystem().getHelpForumPattern()).asMatchPredicate();
 
-        defaultTags = config.getHelpSystem().getCategories();
+        tags = config.getHelpSystem().getCategories();
     }
 
     @Override
@@ -71,7 +71,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String originalMessageId = event.getTarget().getId();
         String originalChannelId = event.getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
-        String mostCommonTag = defaultTags.get(0);
+        String mostCommonTag = tags.get(0);
 
         TextInput modalTitle = TextInput.create(MODAL_TITLE_ID, "Title", TextInputStyle.SHORT)
             .setMaxLength(70)
@@ -159,16 +159,16 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String transferQuestionTag = event.getValue(MODAL_TAG).getAsString();
 
         ForumChannel questionsForum = getHelperForum(event.getJDA());
-        String mostCommonTag = defaultTags.get(0);
+        String mostCommonTag = tags.get(0);
 
-        String queryTag = StringDistances.closestMatch(transferQuestionTag, defaultTags)
-            .orElse(mostCommonTag);
+        String queryTag =
+                StringDistances.closestMatch(transferQuestionTag, tags).orElse(mostCommonTag);
 
-        ForumTag defaultTag = getDefaultTagOr(questionsForum.getAvailableTagsByName(queryTag, true),
+        ForumTag tag = getTagOr(questionsForum.getAvailableTagsByName(queryTag, true),
                 () -> questionsForum.getAvailableTagsByName(mostCommonTag, true).get(0));
 
         return questionsForum.createForumPost(forumTitle, forumMessage)
-            .setTags(ForumTagSnowflake.fromId(defaultTag.getId()))
+            .setTags(ForumTagSnowflake.fromId(tag.getId()))
             .map(createdPost -> new ForumPost(originalUser, createdPost.getMessage()));
     }
 
@@ -207,7 +207,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
                 "Did not find the helper-forum while trying to transfer a question. Make sure the config is setup properly."));
     }
 
-    private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
+    private static ForumTag getTagOr(List<ForumTag> tagsFoundOnForum,
             Supplier<ForumTag> defaultTag) {
         return tagsFoundOnForum.isEmpty() ? defaultTag.get() : tagsFoundOnForum.get(0);
     }
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -46,13 +46,15 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final Pattern TITLE_GUESS_COMPACT_REMOVAL_PATTERN = Pattern.compile("\\W");
     private static final int TITLE_GUESS_COMPACT_LENGTH_MIN = 2;
     private static final int TITLE_GUESS_COMPACT_LENGTH_MAX = 30;
+    private static final Color EMBED_COLOR = new Color(50, 164, 168);
     private final Predicate<String> isHelpForumName;
     private final List<String> defaultTags;
-    private static final Color EMBED_COLOR = new Color(50, 164, 168);
 
 
     /**
      * Creates a new instance.
+     *
+     * @param config to get the helper forum and tags
      */
     public TransferQuestionCommand(Config config) {
         super(Commands.message(COMMAND_NAME), CommandVisibility.GUILD);
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -48,6 +48,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final int TITLE_COMPACT_LENGTH_MAX = 30;
     private final Predicate<String> isHelpForumName;
     private final List<String> defaultTags;
+    private static final Color TURQUOISE = new Color(50, 164, 168);
 
 
     /**
@@ -98,7 +99,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
         event.replyModal(transferModal)
             .queue(success -> logger.info(
-                    "{} with id: {}  triggered the transfer action on message with id: {}",
+                    "{} with id: {}  triggered the transfer action on forumPost with id: {}",
                     event.getUser().getName(), event.getUser().getId(), originalMessageId),
                     failed -> {
                     });
@@ -106,16 +107,13 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
     @Override
     public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
-        event.deferReply().queue();
+        event.deferEdit().queue();
 
         event.getJDA()
             .retrieveUserById(args.get(0))
             .flatMap(fetchedUser -> createForumPost(event, fetchedUser))
-            .flatMap(user -> dmUser(event.getChannel(), user, event.getGuild()))
+            .flatMap(postEntity -> dmUser(event.getChannel(), postEntity, event.getGuild()))
             .flatMap(dmSent -> deleteOriginalMessage(event.getJDA(), args.get(2), args.get(1)))
-            .flatMap(deletedOriginalMessage -> event.getHook()
-                .sendMessage(String.format("Question transferred to %s",
-                        getHelperForum(event.getJDA()))))
             .queue();
     }
 
@@ -140,7 +138,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
                 && titleCompact.length() <= TITLE_COMPACT_LENGTH_MAX;
     }
 
-    private RestAction<User> createForumPost(ModalInteractionEvent event, User originalUser) {
+    private RestAction<PostEntity> createForumPost(ModalInteractionEvent event, User originalUser) {
 
         String originalMessage = event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString();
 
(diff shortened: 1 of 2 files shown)


ankitsmt211:
Including this comment from issue, related to this PR
image

Zabuzard:
in ur PR description, please add pictures of the entire flow and all edge cases, so that i can review UX, thanks.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+    private RestAction<User> createForumPost(ModalInteractionEvent event, User originalUser) {
+        MessageCreateData forumMessage = MessageCreateData
+            .fromContent(event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString());
+        String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
+        ForumChannel questionsForum = getHelperForum(event.getJDA());
+        List<ForumTag> tags = questionsForum.getAvailableTagsByName("JAVA", true);
+        ForumTag javaTag = tags.get(0);
+
+        return questionsForum.createForumPost(forumTitle, forumMessage)
+            .setTags(ForumTagSnowflake.fromId(javaTag.getId()))
```
Taz03, line 140: if not select menu, i'll suggest to make the post without any tags (if possible)

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+            .map(sent -> originalUser);
+
+    }
+
+    private RestAction<ForumChannel> dmUser(MessageChannelUnion sourceChannel, User originalUser) {
+
+        return originalUser.openPrivateChannel()
+            .flatMap(channel -> channel.sendMessage(
+                    "Hi, TJ here. You are getting this message " + "because you tried to ask a "
+                            + "question in wrong chat, please read rules - Together Java"))
```
Taz03, line 152: y the random break in string? also u can't just hard code the server name

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+            .onErrorFlatMap(error -> sourceChannel
+                .sendMessage("Due to failed dm interaction, you are getting "
+                        + "this message here. Please read rules about asking questions - Together Java"
+                        + originalUser.getAsMention()))
+            .onSuccess(failed -> logger.warn("failed to dm user {} with id: {}",
+                    originalUser.getName(), originalUser.getId()))
```
Taz03, line 158: what is this

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+        String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
+        ForumChannel questionsForum = getHelperForum(event.getJDA());
+        List<ForumTag> tags = questionsForum.getAvailableTagsByName("JAVA", true);
+        ForumTag javaTag = tags.get(0);
+
+        return questionsForum.createForumPost(forumTitle, forumMessage)
+            .setTags(ForumTagSnowflake.fromId(javaTag.getId()))
+            .map(ForumPost::getMessage)
+            .flatMap(message -> message.reply("Original Post from " + originalUser.getAsMention()))
+            .map(sent -> originalUser);
```
Taz03, line 143: don't do this here

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+            .flatMap(channel -> channel.sendMessage(
+                    "Hi, TJ here. You are getting this message " + "because you tried to ask a "
+                            + "question in wrong chat, please read rules - Together Java"))
+            .onErrorFlatMap(error -> sourceChannel
+                .sendMessage("Due to failed dm interaction, you are getting "
+                        + "this message here. Please read rules about asking questions - Together Java"
+                        + originalUser.getAsMention()))
+            .onSuccess(failed -> logger.warn("failed to dm user {} with id: {}",
+                    originalUser.getName(), originalUser.getId()))
+            .map(dm -> getHelperForum(dm.getJDA()));
```
Taz03, line 159: don't do this here

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+
+    }
+
+    private RestAction<Void> deleteOriginalMessage(JDA jda, String channelId, String messageId) {
+        return jda.getTextChannelById(channelId)
+            .deleteMessageById(messageId)
+            .onSuccess(deleted -> logger.warn(
+                    "message with id: {} has been deleted due to transfer action being successful",
+                    messageId));
```
Taz03, line 168: no need to log

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: My bad, i thought all the lines of code should split up to not go beyond a certain limit horizontally, but i think it's bad for logger ?

Taz03: u can just leave that on  spotless

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: if you mean "logging the deleted message", isn't it a good idea to store a deleted message. It's an important event? no?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: if you mean "logging the deleted message", isn't it a good idea to store a deleted message. It's an important event? no?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: mapping `sent` to `originalUser`, if it is required do it in the main chain in calling method

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: no i mean the mapping

ankitsmt211: removed "onSuccess"

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Didn't realise, onSuccess triggered right after the chain it follows. It's removed.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Questions forum madates the need for a  tag when creating new ones, im not sure if i can bypass that without a tag.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: ohh now i see what ur doing, nvm mb

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: looks like it is required here

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,182 @@
...
+                    event.getUser().getName(), event.getUser().getId(), originalMessageId),
+                    failed -> {
+                    });
+    }
+
+    @Override
+    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
+        event.getJDA()
+            .retrieveUserById(args.get(0))
+            .map(user -> user)
```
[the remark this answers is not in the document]
ankitsmt211: wasn't necessary, removed it.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: is this better?
image

ankitsmt211:
> in ur PR description, please add pictures of the entire flow and all edge cases, so that i can review UX, thanks.

**Step 1: Text in chat**

text-sent

**Step 2: Transfer Command**

select-transfer-command

**Step 3: selecting transfer-question creates a new post in helper forum**

image

**Step 4: followed by DMing User (contains guildName at last)**

image


**Step 5:  deletes original Message**



**step 6: confirmation**

image

**Incase of failed DM:**

image

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: message deletes are auto-logged in discords own log system. dont worry :)

note for the future: if at all, it wouldnt be a warning, but rather info or debug or trace

ankitsmt211:
Messed up with the last merge for "java version bump", wasted couple of hours fixing that. Really need to get a grip of git. Did most of the requested changes.

Would appreciate a review, hopefully didn't break anything.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Idk why i put it to "warn", it's just part of the process🤦‍♂️

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: so, what other options do u have? i can think of putting all the tags by default separated by a comma and then the user can delete the unnecessary ones, what do u think?

ankitsmt211: is it possible to do select menu in modal? i don't like putting all tags in, gonna clutter the field. We could also rely on `/help-thread-change-category`.

Taz03: > is it possible to do select menu in modal?

no, not possible

> i don't like putting all tags in, gonna clutter the field. We could also rely on /help-thread-change-category

yes, but we'll have to put some tag in the beginning as u said its not possible to create the thread without a tag

ankitsmt211: `Java` tag is put by default, altho it's impossible that this will be gone in future(being a java oriented server) but i can also pull first tag from config.

Taz03: > altho it's impossible that this will be gone in future(being a java oriented server)

this bot is designed to be deployed for any server

> i can also pull first tag from config

i'd suggest to add the last tag, coz that'll add the least number of members (probably), and then it can be customized after creation

ankitsmt211: > this bot is designed to be deployed for any server

Interesting, do we support any server atm?
 
> i'd suggest to add the last tag, coz that'll add the least number of members (probably), and then it can be customized after creation

ill start looking into that

ankitsmt211:
@Taz03 
image

Pulling tags by id works only if forum has that tag, so is it safe to assume that there will always be a match? if not should i handle that maybe with try catch or something and a log maybe?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: > Interesting, do we support any server atm?

idk, maybe someone somewhere is hosting this bot for their server as well, but that is very unlikely. We don't give any special support if someone wants to, if that's what ur asking

Taz03:
> if not should i handle that maybe with try catch or something and a log maybe?

yes just log it

ankitsmt211:
Changes made to createForumPost, since it mandates presence of a tag but if forum doesnot have said tag it's a problem.
 
`defaultTags` is a list of tags pulled from `config`

```java 
 private RestAction<User> createForumPost(ModalInteractionEvent event, User originalUser) throws IndexOutOfBoundsException {
        MessageCreateData forumMessage = MessageCreateData
            .fromContent(event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString());
        String forumTitle = event.getValue(TRANSFER_QUESTION_TITLE_ID).getAsString();
        ForumChannel questionsForum = getHelperForum(event.getJDA());

        List<ForumTag> tags = questionsForum.getAvailableTagsByName(defaultTags.get(defaultTags.size()-1), true);
        
        try{

           ForumTag defaultTag = tags.get(0);
            return questionsForum.createForumPost(forumTitle, forumMessage)
              .setTags(ForumTagSnowflake.fromId(defaultTag.getId()))
              .map(ForumPost::getMessage)
              .flatMap(message -> message.reply("Original Post from " + originalUser.getAsMention()))
              .map(sent -> originalUser);
        }
        catch(IndexOutOfBoundsException indexOutOfBoundsException){
            logger.warn("No tags are present with given value",indexOutOfBoundsException);
            throw new IndexOutOfBoundsException("No tags were found in forum channel matching given tag");
        }
        
    }
```


Method being called inside flatMap chain

```java
   public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
        event.deferReply().queue();

        event.getJDA()
            .retrieveUserById(args.get(0))
            .flatMap(fetchedUser -> {
                try {
                    return createForumPost(event, fetchedUser);
                } catch (IndexOutOfBoundsException indexOutOfBoundsException) {

                    event.getHook().sendMessage("transfer action failed, no tags found").queue();
                    throw new RuntimeException(indexOutOfBoundsException);
                }
            })
            .flatMap(user -> dmUser(event.getChannel(), user, event.getGuild()))
            .flatMap(dmSent -> deleteOriginalMessage(event.getJDA(), args.get(2), args.get(1)))
            .flatMap(deletedOriginalMessage -> event.getHook().sendMessage("Question Transferred"))
            .queue();
    }
```


I used indexoutbounds because that was thrown when no tags were present and i was doing `get(0)`, let me know if should create a custom exception or this works.



console(incase i remove the tag forum) 

image

discord

image

Zabuzard (changes requested):
UX:
1. In the question thread, change the order of the two messages, rephrase and add embed:
`@John has a question:`
`embed with authors profile picture, name and the message content`

2. DM message, rephrase:
> Hello 👋 You have asked a question on Together Java in the **wrong channel**
> Not a big deal, but none of the experts who could help you are reading your question there 🙁
>
> Your question has been **automatically transferred** to #questions-foo-bar, please continue there. You might also want to give #welcome a **quick read**, thank you 👍

3. rephase confirmation:
> Transferred to #questions-foo-bar

4. rephrase DM failure:
> Hello @John 👋 You have asked a question in the **wrong channel**
> Not a big deal, but none of the experts who could help you are reading your question there 🙁
>
> Your question has been **automatically transferred** to #questions-foo-bar, please continue there. You might also want to give #welcome a **quick read**, thank you 👍

ankitsmt211:
UX improvements
1. Forum post now has embed(Had to mention OP seperately because the ping inside embed does not work)
image


2. DM sent to user
image

3. Confirmation
image

ankitsmt211:
@Zabuzard You didn't update if perms will be handled from server or not.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -188,4 +198,12 @@ private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
             Supplier<ForumTag> defaultTag) {
         return tagsFoundOnForum.isEmpty() ? defaultTag.get() : tagsFoundOnForum.get(0);
     }
+
+    private MessageEmbed makeEmbedForPost(User originalUser, String modifiedMessage) {
+        return new EmbedBuilder().setAuthor(originalUser.getName())
+            .setThumbnail(originalUser.getAvatarUrl())
+            .setDescription(modifiedMessage)
+            .setColor(1752220)
```
Zabuzard, line 206: unreadable color. use `Color.decodeHex(...)` or whatever its called. that also has IDE support. and put it into a constant

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -153,21 +164,20 @@ private RestAction<User> createForumPost(ModalInteractionEvent event, User origi
...
-            .onErrorFlatMap(error -> sourceChannel.sendMessage(
-                    "Due to failed dm interaction, you are getting this message here. Please read rules about asking questions -%s%s"
-                        .formatted(guild.getName(), originalUser.getAsMention())));
-
+            .flatMap(channel -> channel.sendMessage(String.format(
+                    "Hello 👋 You have asked a question on %s in the wrong channel %n Not a big deal, but none of the experts who could help you are reading your question there 🙁 %n Your question has been automatically transferred to %s, please continue there. %n You might also want to give #welcome a quick read, thank you \uD83D\uDC4D",
+                    guild.getName(), getHelperForum(guild.getJDA()))))
+            .onErrorFlatMap(error -> sourceChannel.sendMessage(String.format(
+                    "Hello %s 👋 You have asked a question in the wrong channel %n Not a big deal, but none of the experts who could help you are reading your question there 🙁 %n Your question has been automatically transferred to %s, please continue there. %n You might also want to give #welcome a quick read, thank you \uD83D\uDC4D",
```
Zabuzard, line 179: * Whats that `\uD83D\uDC4D`? Put smileys directly in the code, no unique-escapes please
* prefer text-blocks instead of all that `%n`. so `"""`
* the first sentence lacks a dot (wrong channel[.] Not a big)

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -137,8 +141,15 @@ private static boolean isTitleValid(CharSequence title) {
     }
 
     private RestAction<User> createForumPost(ModalInteractionEvent event, User originalUser) {
-        MessageCreateData forumMessage = MessageCreateData
-            .fromContent(event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString());
+
+        String originalMessage = event.getValue(TRANSFER_QUESTION_INPUT_ID).getAsString();
+
+        String modifiedMessageForPost = String.format("Original Post from: %s %n %n %s",
+                originalUser.getAsMention(), originalMessage);
```
Zabuzard, line 148: (thats not what i suggested, reads weird like that - and why are there newlines?)

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -111,7 +113,9 @@ public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
             .flatMap(fetchedUser -> createForumPost(event, fetchedUser))
             .flatMap(user -> dmUser(event.getChannel(), user, event.getGuild()))
             .flatMap(dmSent -> deleteOriginalMessage(event.getJDA(), args.get(2), args.get(1)))
-            .flatMap(deletedOriginalMessage -> event.getHook().sendMessage("Question Transferred"))
+            .flatMap(deletedOriginalMessage -> event.getHook()
+                .sendMessage(String.format("Question transferred to %s",
+                        getHelperForum(event.getJDA()))))
```
Zabuzard, line 118: its sorta crucial for good UX that its the link to the actual thread, not just #questions

Zabuzard (changes requested):
btw, prefer `"...".formatted(...)` over `String.format("...", ...)`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: please adjust the embed to look more natural. see for example this:

example

Zabuzard: so use the proper way of setting the embed author (with avatar and name). then put the message as message into it. u dont need any title or footer.

Zabuzard:
> @Zabuzard You didn't update if perms will be handled from server or not.

Just checked, its configurable in Discord 👍

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: I did not do that to avoid promoting new questions, in chat. This is only meant for moderator/user of command that action has been completed. OP already gets a ping in thread and a DM about it. So there's no problem navigating to it.

ankitsmt211:
> btw, prefer `"...".formatted(...)` over `String.format("...", ...)`

is there any specific reason to use one over the other?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: > (thats not what i suggested, reads weird like that - and why are there newlines?)

wanted to seperate the question from other stuff so it's easier to read.

ankitsmt211:
Embed after changes
image

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: thats fair, but there wont be any other stuff to separate from if you go with what i said. Essentially, you want the normal message content to be `@Foo has a question:` and then the embedd that should look like the original message without any extras.

The ping must be outside the embed, otherwise it does not ping. This is similar to `/tag ... reply:@John` which looks like this:
demo

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: Yeah but the moderator perhaps also wants to navigate there? Or would it be hard to add it in the code?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -199,11 +204,12 @@ private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
...
-        return new EmbedBuilder().setAuthor(originalUser.getName())
-            .setThumbnail(originalUser.getAvatarUrl())
-            .setDescription(modifiedMessage)
-            .setColor(1752220)
+    private MessageEmbed makeEmbedForPost(User originalUser, String originalMessage) {
+        return new EmbedBuilder()
+            .setAuthor(originalUser.getName(), originalUser.getAvatarUrl(),
+                    originalUser.getAvatar().getUrl())
+            .setDescription(originalMessage)
+            .setColor(new Color(50, 164, 168))
```
Zabuzard, line 212: create the color as constant and reuse

Zabuzard (changes requested):

Zabuzard:
Can you keep the pics in the PR description up2date? For example, how does the `@John has a question: <embed>` look like?

christolis:
> > btw, prefer `"...".formatted(...)` over `String.format("...", ...)`
> 
> is there any specific reason to use one over the other?

Both methods are identical ([source](https://stackoverflow.com/a/70997302)), although the former option looks cleaner.

ankitsmt211:
> Can you keep the pics in the PR description up2date? For example, how does the `@John has a question: <embed>` look like?

I don't understand, do you want me to get rid of older pictures and add new ones? or just add updated embed here. Coz i did add the updated embed, it features OP's username, small icon and a question.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -48,6 +48,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final int TITLE_COMPACT_LENGTH_MAX = 30;
     private final Predicate<String> isHelpForumName;
     private final List<String> defaultTags;
+    private static final Color TURQUOISE = new Color(50, 164, 168);
```
Zabuzard, line 51: bad name. instead of saying what color it is (we can all see that in intellij), say what its used for. in this case, it is the color for ur embed. so `EMBED_COLOR` or similar would be better, thanks

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -98,24 +99,21 @@ public void onMessageContext(MessageContextInteractionEvent event) {
 
         event.replyModal(transferModal)
             .queue(success -> logger.info(
```
Zabuzard, line 101: `logger.info` is spammy. use `debug` for this

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -162,27 +160,28 @@ private RestAction<User> createForumPost(ModalInteractionEvent event, User origi
             .setTags(ForumTagSnowflake.fromId(defaultTag.getId()))
             .map(ForumPost::getMessage)
             .flatMap(message -> message.reply(originalUser.getAsMention()))
-            .map(sent -> originalUser);
+            .map(sent -> new PostEntity(originalUser, sent));
     }
 
-    private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, User originalUser,
+    private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, PostEntity postEntity,
```
Zabuzard, line 166: i dont like the variable name, to which post is it referring? the original message in chit_chat? the forum post message? sth else?

Zabuzard (changes requested):
also:

> Step 3: selecting transfer-question creates a new post in helper forum followed by mention in reply

this doesnt look the way i would like to have it. It should be the following:

> `@John` has a question:
> `<embed>`

Note that its a single message, not two. You can send a message with normal message-content **and** an embed. See `/tag ... reply:` as an example.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Post is only referred to posts created in forum channels, text messages should be referred as messages.

Zabuzard: yeah, but which one is it? u could rename the variable to make it a bit clearer imo. `forumPost` perhaps 🤷 
dunno, it tripped me.

SquidXTV:
Does this still work with chatgpt's automatic answer?

ankitsmt211:
> Does this still work with chatgpt's automatic answer?

Do you mean if chatgpt responds to it with an answer? Yes it does.

SquidXTV:
> > Does this still work with chatgpt's automatic answer?
> 
> 
> 
> Do you mean if chatgpt responds to it with an answer? Yes it does.

Yeah ok nice, I was wondering because it's in a embed in that case.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+
+        TextInput transferQuestionInput = TextInput
+            .create(TRANSFER_QUESTION_INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
+            .setValue(originalMessage)
+            .setRequiredRange(3, 2000)
+            .build();
+
+        TextInput transferQuestionTag = TextInput
+            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
+            .setValue("Java")
```
Taz03, line 88: i still see java here

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Because it's supposed to be there, the idea is Java is the most popular tag so 90% of the times you won't need to change the tag but if you do it can be done and will find the closest match regardless.

Taz03: the idea was to pull the first tag from config and the use ped on submit

ankitsmt211: I don't see how it helps, it's a modal field supposed to be prefilled for convenience. It can be changed or not.

Zabuzard: taz is right on this. please take the first entry from the config instead.

u have to think about other servers as well. for example, TJ-Bot is used on my kotlin server as well. and the top tag there is `Kotlin`.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+    public void onMessageContext(MessageContextInteractionEvent event) {
+        String originalMessage = event.getTarget().getContentRaw();
+        String originalMessageId = event.getTarget().getId();
+        String originalChannelId = event.getChannel().getId();
+        String authorId = event.getTarget().getAuthor().getId();
+
+        TextInput transferQuestionTitle =
+                TextInput.create(TRANSFER_QUESTION_TITLE_ID, "Title", TextInputStyle.SHORT)
+                    .setMaxLength(70)
+                    .setMinLength(8)
```
Zabuzard, line 76: min length 4 should be enough

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+                    .setValue(createTitle(originalMessage))
+                    .build();
+
+        TextInput transferQuestionInput = TextInput
+            .create(TRANSFER_QUESTION_INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
```
Zabuzard, line 81: I dont like the name here. The title shouldnt say what element ur looking at, but explain how to use it. Could you show a picture of the entire modal dialog please?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+            .setValue(originalMessage)
+            .setRequiredRange(3, 2000)
+            .build();
+
+        TextInput transferQuestionTag = TextInput
+            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
```
Zabuzard, line 87: Same here. Ill give u more detailed UX changes after seeing the modal picture.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+            .setValue("Java")
+            .build();
+
+        String transferQuestionModalComponentID =
```
Zabuzard, line 91: tip on naming. the context that this is about transferring questions is clear based on the class name already. everyone reading this class knows that we are now transffering a question.

hence, its obsolete and rather distracting to prefix all ur variables with `transferQuestion` (applies to the other variables here as well)

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+                generateComponentId(authorId, originalMessageId, originalChannelId);
+        Modal transferModal =
+                Modal.create(transferQuestionModalComponentID, "transfer question menu")
```
Zabuzard, line 94: whats that string for? where is it displayed?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+                    .addActionRow(transferQuestionInput)
+                    .addActionRow(transferQuestionTag)
+                    .build();
+
+        event.replyModal(transferModal)
+            .queue(success -> logger.debug(
+                    "{} with id: {}  triggered the transfer action on forumPost with id: {}",
+                    event.getUser().getName(), event.getUser().getId(), originalMessageId),
+                    failed -> {
+                    });
```
Zabuzard, line 105: that part is obsolete, considering that `queue` has an overload that only takes a success-callback

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
+        event.deferEdit().queue();
+
+        event.getJDA()
+            .retrieveUserById(args.get(0))
```
Zabuzard, line 113: put `args.get(0)` into a well named variable please.
```java
String userId = args.get(0);

event.getJDA()
  .retrieveUserById(userId)
  ...
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+            .flatMap(fetchedUser -> createForumPost(event, fetchedUser))
+            .flatMap(createdforumPost -> dmUser(event.getChannel(), createdforumPost,
+                    event.getGuild()))
+            .flatMap(dmSent -> deleteOriginalMessage(event.getJDA(), args.get(2), args.get(1)))
+            .queue();
```
Zabuzard, line 118: same with the args here. put them all at the very top of the method.
```java
... = args.get(0);
... = args.get(1);
... = args.get(2);

event.getJda()
...
```
that makes the interface and communication-bridge much clearer

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.regex.Pattern;
+
+public final class TransferQuestionCommand extends BotCommandAdapter
+        implements MessageContextCommand {
+    private static final Logger logger = LoggerFactory.getLogger(TransferQuestionCommand.class);
+    private static final String COMMAND_NAME = "transfer-question";
+    private static final String TRANSFER_QUESTION_TITLE_ID = "transferID";
+    private static final String TRANSFER_QUESTION_INPUT_ID = "transferQuestion";
+    private static final String TRANSFER_QUESTION_TAG = "tags";
+    private static final int TITLE_MAX_LENGTH = 50;
```
Zabuzard, line 45: confusing name. it suggest that its the max length of the title, but it is not. the max length of the title is 70 according to ur modal code.

this is the max length of the `TITLE_GUESS` or `AUTO_TITLE`. please adjust

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,220 @@
...
+
+                            Your question has been automatically transferred to %s, please continue there. You might also want to give #welcome a quick read, thank you 👍.
+                            """
+                        .formatted(guild.getName(), forumPost.message.getJumpUrl())))
+            .onErrorFlatMap(error -> sourceChannel.sendMessage(
+                    """
+                            Hello %s 👋 You have asked a question in the wrong channel. Not a big deal, but none of the experts who could help you are reading your question there 🙁.
+
+                            Your question has been automatically transferred to %s, please continue there. You might also want to give #welcome a quick read, thank you 👍.
+                            """
```
Zabuzard, line 185: duplication of the text. it looks almost identical. whats the difference? the ping?

get rid of the duplication. for example by using the `Hello %s ...` text and then conditionally formatting that either with `""` or the mention.

also, if we cant make `#welcome` clickable (in this iteration), id rather remove the last sentence entirely and write:
```java
... please continue there, thank you 👍
```
(also, if u end on a smiley, dont put a dot)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: ig the title is fine

Taz03: we can add how to use it in the field, as a background text

Zabuzard: i dont quite follow what your saying. anyways, i want to see a picture of the dialog and then ill give feedback on that :)

ankitsmt211: updated modal picture, should have included it dk how i missed.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: At the top of modal menu

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: You sure? Title is auto generated, which gives "Untitled"(length is 8 here) if nothing. So i think it won't accept anything less than 8.

ankitsmt211: lol makes sense, idk what i was thinking.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Should i at least keep "modal" as prefix? otherwise the variables would be "title", "input", "tag".

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: not sure how to do this one, these messages are for two different calls. Could you maybe suggest something?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: u can keep it. it just depends how clear the context is. suppose u have the following:
```java
void createModal() {
  String title = "...";
  ...
}
```
in this case its absolutely clear that the title is the title of the modal, bc the method is literally giving the context. in this case, `modalTitle` would be kinda unecessary.

but it doesnt hurt that much, since its so much shorter. im okay with that

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: So its user-visible? Cause this switch is lower-case, which doesnt look good.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: prepare both texts beforehand with variables:

```java
String textTemplate = """
  Hello %s 👋 You have asked a question on %s in the wrong channel.
  ...
  ...
""";

String dmText = textTemplate.formatted("", guild.getName(), forumPost.message.getJumpUrl());
String channelText = textTemplate.formatted(forumPost.author.getAsMention(), "", forumPost.message.getJumpUrl());
```
And then later:
```java
.flatMap(channel -> channel.sendMessage(channelText))
.onErrorFlatMap(error -> sourceChannel.sendMessage(dmText))
```

Zabuzard (changes requested):
modal review

Zabuzard:
@ankitsmt211 could you share a picture of the modal with every content erased, so that i can read the "background text" on each field please? thanks

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: ah oki this one makes sense, i thought you want to conditionally do both the calls in one line or something.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -69,61 +69,64 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
-                    .build();
-
-        TextInput transferQuestionInput = TextInput
-            .create(TRANSFER_QUESTION_INPUT_ID, "Transfer question menu", TextInputStyle.PARAGRAPH)
-            .setValue(originalMessage)
-            .setRequiredRange(3, 2000)
+        TextInput modalTitle = TextInput.create(MODAL_TITLE_ID, "Title", TextInputStyle.SHORT)
+            .setMaxLength(70)
+            .setMinLength(4)
+            .setPlaceholder("describe question in short")
```
Zabuzard, line 77: start of sentence, capital. `Describe` and please write proper english. `Describe the question in short`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -69,61 +69,64 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
             .build();
 
-        TextInput transferQuestionTag = TextInput
-            .create(TRANSFER_QUESTION_TAG, "Transfer question tags", TextInputStyle.SHORT)
-            .setValue("Java")
+        TextInput modalInput =
+                TextInput.create(MODAL_INPUT_ID, "Question", TextInputStyle.PARAGRAPH)
+                    .setValue(originalMessage)
+                    .setRequiredRange(3, 2000)
+                    .setPlaceholder("contents of question")
```
Zabuzard, line 85: start of sentence, capital. and proper english. `Contents of the question`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -69,61 +69,64 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
+                    .build();
+
+        TextInput modalTag = TextInput.create(MODAL_TAG, "Most fitting tag", TextInputStyle.SHORT)
+            .setValue(mostCommonTag)
+            .setPlaceholder("suitable tag for question")
```
Zabuzard, line 90: same. `Suitable tag for the question`

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -69,61 +69,64 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
             .build();
 
-        String transferQuestionModalComponentID =
+        String modalComponentID =
```
Zabuzard, line 93: `Id` not `ID` please

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -69,61 +69,64 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
-                    failed -> {
-                    });
+                    event.getUser().getName(), event.getUser().getId(), originalMessageId));
     }
 
     @Override
     public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
         event.deferEdit().queue();
 
+        String authorId = args.get(0);
```
Zabuzard, line 111: author of what? the original post or the user who invokes the command? please clarify and adjust the variable accordingly

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -169,22 +173,18 @@ private RestAction<ForumPost> createForumPost(ModalInteractionEvent event, User
...
-                            Hello %s 👋 You have asked a question in the wrong channel. Not a big deal, but none of the experts who could help you are reading your question there 🙁.
-
-                            Your question has been automatically transferred to %s, please continue there. You might also want to give #welcome a quick read, thank you 👍.
-                            """
-                        .formatted(forumPost.author.getAsMention(),
-                                forumPost.message.getJumpUrl())));
+            .flatMap(channel -> channel.sendMessage(messageTemplate.formatted("",
+                    "on " + guild.getName(), forumPost.message.getJumpUrl())))
+            .onErrorFlatMap(error -> sourceChannel.sendMessage(messageTemplate
+                .formatted(forumPost.author.getAsMention(), "", forumPost.message.getJumpUrl())));
```
Zabuzard, line 187: this code is hard to read, cause its super nested and doing too much at the same time.
do the formatting beforehand and put them into variables as shown earlier. then the code becomes much simpler to read

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -169,22 +173,18 @@ private RestAction<ForumPost> createForumPost(ModalInteractionEvent event, User
     private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, ForumPost forumPost,
             Guild guild) {
 
+        String messageTemplate =
+                """
+                        Hello %s 👋 You have asked a question in the wrong channel %s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
```
Zabuzard, line 178: You get odd spaces for formatting either. Example:
```java
Hello[two spaces]👋 You have asked a question i nthe wrong channel[one space].
```
Instead, you have to write it as `Hello %s👋 You have asked a question in the wrong channel%s.`

and then format one as `"[one space]on " + ...` and the other as `" " + forumPost.author.getAsMention()`

Zabuzard (changes requested):

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: perhaps `originalAuthorId`? to give the connection to the original message

Zabuzard (approved):

Zabuzard:
good job, thank you for ur efforts 😃

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -0,0 +1,226 @@
...
+        Modal transferModal = Modal.create(modalComponentId, "Transfer this question")
+            .addActionRow(modalTitle)
+            .addActionRow(modalInput)
+            .addActionRow(modalTag)
+            .build();
+
+        event.replyModal(transferModal)
+            .queue(success -> logger.debug(
+                    "{} with id: {}  triggered the transfer action on forumPost with id: {}",
+                    event.getUser().getName(), event.getUser().getId(), originalMessageId));
```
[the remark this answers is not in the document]
ankitsmt211: figured might do this just to keep a track of "who's doing the transfer".

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -207,7 +204,7 @@ private ForumChannel getHelperForum(JDA jda) {
                 "Did not find the helper-forum while trying to transfer a question. Make sure the config is setup properly."));
     }
 
-    private static ForumTag getDefaultTagOr(List<ForumTag> tagsFoundOnForum,
+    private static ForumTag getTagOr(List<ForumTag> tagsFoundOnForum,
```
Taz03, line 204: `getTagOrDefault`

Taz03 (approved):

Zabuzard (approved):

Zabuzard:
congrats 🎉 lets try it out on the test server

PR #909, opened 2023-10-07
Catch error when unable to send DM

### LINKED ISSUE #797: AutoPruneHelperRoutine crashes on DM when DMs closed
The `AutoPruneHelperRoutine` class crashes when attempting to send a DM to someone who does not accept DMs.

stacktrace

Instead, it should just log the issue on debug level. Therefore, simply add a failure handler `.queue(() -> {}, failure -> logger.debug(..., failure))`.

here

christolis:
Fixes #797. I was unable to test this, but I think this might be a solid solution.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
@@ -163,8 +163,9 @@ public final class AutoPruneHelperRoutine implements Routine {
         guild.removeRoleFromMember(member, role)
             .flatMap(any -> member.getUser().openPrivateChannel())
             .flatMap(channel -> channel.sendMessage(dmMessage))
-            .queue(null,
-                    failure -> logger.debug("Tried to send a DM to a user but failed: ", failure));
+            .queue(null, failure -> logger
+                .debug("Failed sending a DM to user (%s) while pruning them from a helper role."
+                    .formatted(member.getId()), failure));
     }
 
     private void warnModsAbout(String message, Guild guild) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
@@ -163,8 +163,8 @@ public final class AutoPruneHelperRoutine implements Routine {
         guild.removeRoleFromMember(member, role)
             .flatMap(any -> member.getUser().openPrivateChannel())
             .flatMap(channel -> channel.sendMessage(dmMessage))
-            .queue(msg -> {
-            }, failure -> logger.debug("Tried to send a DM to a user but failed: ", failure));
+            .queue(null,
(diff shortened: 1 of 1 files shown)


SquidXTV:
you can maybe instead supply `null` for the success `Consumer`.
not sure if its better/cleaner code then though

> success - The success callback that will be called at a convenient time for the API. (can be null to use default)

https://docs.jda.wiki/net/dv8tion/jda/api/requests/RestAction.html#queue(java.util.function.Consumer,java.util.function.Consumer)

surajkumar:
Maybe we can have the users name in the log message? Though not too important

SquidXTV:
> Maybe we can have the users name in the log message? Though not too important

wont it be in the exception?

SquidXTV (approved):

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
```diff
@@ -163,7 +163,8 @@ private void pruneMemberFromRole(Member member, Role role, ForumChannel helpForu
         guild.removeRoleFromMember(member, role)
             .flatMap(any -> member.getUser().openPrivateChannel())
             .flatMap(channel -> channel.sendMessage(dmMessage))
-            .queue();
+            .queue(null,
+                    failure -> logger.debug("Tried to send a DM to a user but failed: ", failure));
```
Zabuzard, line 167: the log message is missing context. should add sth like
`Failed sending a DM to user (1234567) while pruning them from a helper role.` (with the number being the user ID)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
christolis: Done! 🙌

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
```diff
@@ -163,7 +163,9 @@ private void pruneMemberFromRole(Member member, Role role, ForumChannel helpForu
         guild.removeRoleFromMember(member, role)
             .flatMap(any -> member.getUser().openPrivateChannel())
             .flatMap(channel -> channel.sendMessage(dmMessage))
-            .queue();
+            .queue(null, failure -> logger
+                .debug("Failed sending a DM to user (%s) while pruning them from a helper role."
+                    .formatted(member.getId()), failure));
```
Taz03, line 168: use [logger formatter](https://github.com/apache/logging-log4j2#usage)

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
christolis: It's now done! ✅

Taz03 (approved):

Zabuzard (approved):

PR #915, opened 2023-10-20
fix inability to read embeds in helper threads

### LINKED ISSUE #914: GPT fix to read embeds
GPT is unable to read helper threads with "embeds", created as a result of transfer-question  feature.

ankitsmt211:
resolves #914

Embed:
image
Response:
image

Normal Post:
image

Response:
image


Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -112,4 +112,11 @@ private RestAction<Message> sendHelperHeadsUp(ThreadChannel threadChannel) {
         return threadChannel.sendMessage(headsUpWithoutRole)
             .flatMap(message -> message.editMessage(headsUpWithRole));
     }
+
+    private String getEmbedOrRawMessage(Message message) {
+        if (message.getEmbeds().isEmpty()) {
+            return message.getContentRaw();
+        }
+        return message.getEmbeds().get(0).getDescription();
```
Zabuzard, line 120: could u add together multi-embed content please? just a bit of thinking-ahead, and bc its so easy to do 😃

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -112,4 +112,11 @@ private RestAction<Message> sendHelperHeadsUp(ThreadChannel threadChannel) {
         return threadChannel.sendMessage(headsUpWithoutRole)
             .flatMap(message -> message.editMessage(headsUpWithRole));
     }
+
+    private String getEmbedOrRawMessage(Message message) {
```
Zabuzard, line 116: NIT: the title isnt bad, but i propose changing it regardless.
currently, it describes what the method internally does. but instead, it should reflect the purpose of the method. its supposed to extract the **question content**. so perhaps `getQuestionContent(message)`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
ankitsmt211: Is it possible to have multiple embeds in one post? Also you mean like iterate over the list and concat all the descriptions?

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
Zabuzard: > Also you mean like iterate over the list and concat all the descriptions?

Yeah. instead of only taking the content of the first, iterate all and just `+` it together.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -117,6 +117,11 @@ private String getMessageContent(Message message) {
         if (message.getEmbeds().isEmpty()) {
             return message.getContentRaw();
         }
-        return message.getEmbeds().get(0).getDescription();
+
+        StringBuilder messageContent = new StringBuilder();
+        message.getEmbeds()
+            .forEach(embed -> messageContent.append("%s%n".formatted(embed.getDescription())));
```
Zabuzard, line 123: works. but a bit more idiomatic would be to use streams
```java
message.getEmbeds().stream()
  .collect(Collectors.joining("\n"))
```
(also note that its necessary to hardcode the line separator to `\n` here, since that is what discord uses. if u go platform-dependent, u would get `\r\n` on windows, which discord would be confused about

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

Taz03 (approved):

PR #932, opened 2023-10-24
refactor method for generating title of forum post

ankitsmt211:
`createTitle` method in `TransferCommand` class was generating a lot of "Untitled" titles as most titles were being rejected due to a minor bug. Now the titles are generated properly.

titles generated on last 5 transfered questions:
image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -129,7 +129,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
                 lastWordEnd = TITLE_MAX_LENGTH;
             }
 
-            message = message.substring(0, lastWordEnd).replace("\n", " ");
+            message = message.substring(0, lastWordEnd).replace('\n', ' ');
         }
 
         return isTitleValid(message) ? message : "Untitled";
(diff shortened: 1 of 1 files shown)


Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -123,14 +122,14 @@ public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
...
+        if (message.length() >= TITLE_MAX_LENGTH) {
+            int lastWordEnd = message.lastIndexOf(' ', TITLE_MAX_LENGTH);
 
             if (lastWordEnd == -1) {
-                lastWordEnd = TITLE_GUESS_MAX_LENGTH;
+                lastWordEnd = TITLE_MAX_LENGTH;
             }
 
-            message = message.substring(0, lastWordEnd);
+            message = message.substring(0, lastWordEnd).replace("\n", " ");
```
Zabuzard, line 132:
```suggestion
            message = message.substring(0, lastWordEnd).replace('\n', ' ');
```
when only replacing char by char, prefer the overload that takes char instead of string. so `replace('\n', ' ')`. mini performance boost and just a tad cleaner

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

PR #933, opened 2023-10-24
refactor methods to handle invalid transfers

### LINKED ISSUE #926: Transfer crashes on short messages
When triggering the transfer command on a short message whose auto-title would be short, it runs into an exception without falling back to `Untitled`:

message

stacktrace

ankitsmt211:
* refactor bot transfer checks
* add check for too short message transfers 
* resolves #926

updates:
* transfer on short messages now opens up modal menu with empty question/text field with "untitled" as title.
* longer messages than max length are trimmed.
 
image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -238,7 +238,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
     private void handleQuestionTooShort(MessageContextInteractionEvent event) {
         event
-            .reply("message content should be at least %s characters long"
+            .reply("Message content should be at least %s characters long."
                 .formatted(INPUT_MIN_LENGTH))
             .setEphemeral(true)
             .queue();
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -43,6 +43,8 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final Pattern TITLE_GUESS_COMPACT_REMOVAL_PATTERN = Pattern.compile("\\W");
     private static final int TITLE_MIN_LENGTH = 3;
     private static final Color EMBED_COLOR = new Color(50, 164, 168);
+    private static final int INPUT_MAX_LENGTH = 2000;
+    private static final int INPUT_MIN_LENGTH = 3;
     private final Predicate<String> isHelpForumName;
     private final List<String> tags;
 
@@ -64,7 +66,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     @Override
     public void onMessageContext(MessageContextInteractionEvent event) {
 
-        if (isBotMessageTransfer(event)) {
+        if (isInvalidForTransfer(event)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -89,12 +90,14 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
                 TextInput.create(MODAL_INPUT_ID, "Question", TextInputStyle.PARAGRAPH)
-                    .setValue(originalMessage)
                     .setRequiredRange(INPUT_MIN_LENGTH, INPUT_MAX_LENGTH)
-                    .setPlaceholder("Contents of the question")
-                    .build();
+                    .setPlaceholder("Contents of the question");
+
+        if (!isQuestionTooShort(originalMessage)) {
+            modalInputBuilder.setValue(originalMessage);
+        }
```
Zabuzard, line 100: what if its too long? 😬

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: ofc didn't even think of that. Now it makes sense, we gotta cover the edge cases since discord is relying on us not the user input.

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -228,11 +234,31 @@ private MessageEmbed makeEmbedForPost(User originalUser, String originalMessage)
...
+        return author.isBot();
+    }
+
+    private void handleBotMessageTransfer(MessageContextInteractionEvent event) {
+        event.reply("Cannot transfer messages from a bot.").setEphemeral(true).queue();
+    }
+
+    private boolean isQuestionTooShort(String question) {
+        return question.length() < INPUT_MIN_LENGTH;
+    }
```
Taz03, line 247: do u really need to create a method for these, one liners

ankitsmt211: Originally i started that way but it was a mess inside one single method it's too big already, at least methods abstract away the details to keep it short.

ankitsmt211:
Question, if `onMessageContext` was a private method, would it be better to use `assert` instead of validation methods i did to ensure message is of required length for modal description?

Zabuzard:
> Question, if `onMessageContext` was a private method, would it be better to use `assert` instead of validation methods i did to ensure message is of required length for modal description?

Id prefer to not use `assert`. Its disabled unless u run a specific VM argument. Prefer using ordinary exceptions and the logger with its various logging levels instead. Theres also `throw new AssertionError(...)` if you want a good equivalent to an assert. Note that asserts are not for user-errors but to mark bugs in the code.

ankitsmt211:
> Id prefer to not use `assert`. Its disabled unless u run a specific VM argument. Prefer using ordinary exceptions and the logger with its various logging levels instead. Theres also `throw new AssertionError(...)` if you want a good equivalent to an assert. Note that asserts are not for user-errors but to mark bugs in the code.

True that was just curious, from what i understand assert also throws assertion error by default. But ye it needs to be enabled.

Taz03 (approved):

PR #938, opened 2023-10-30
add pinned announcement listener

### LINKED ISSUE #921: Delete the pinned-message in help threads
We have a feature that auto-pins the first message in each help-thread.

However, we should delete the message that indicates that event in discord, as it creates unecessary noise:

pin

ankitsmt211:
* listens for pinned announcement
* removes if belongs to helper forum
* resolves #921

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
@@ -13,8 +13,7 @@ import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
 /**
- * Listener that receives all pinned announcement messages from helper forum so that it can be
- * removed
+ * Removes notification-announcements about pinned messages from the helper forum.
  */
 public final class PinnedAnnouncementListener extends MessageReceiverAdapter {
 
@@ -32,7 +31,7 @@ public final class PinnedAnnouncementListener extends MessageReceiverAdapter {
 
     @Override
     public void onMessageReceived(MessageReceivedEvent event) {
-        removePinnedAnnouncement(event);
+        handlePinnedAnnouncement(event);
     }
 
     private boolean isHelpThread(MessageChannelUnion channel) {
@@ -48,7 +47,7 @@ public final class PinnedAnnouncementListener extends MessageReceiverAdapter {
         return message.getType() == MessageType.CHANNEL_PINNED_ADD;
     }
 
-    private void removePinnedAnnouncement(MessageReceivedEvent event) {
+    private void handlePinnedAnnouncement(MessageReceivedEvent event) {
         if (isPinnedAnnouncement(event.getMessage()) && isHelpThread(event.getChannel())) {
             event.getMessage().delete().queue();
         }
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
@@ -1,0 +1,56 @@
+package org.togetherjava.tjbot.features.help;
+
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.MessageType;
+import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
```diff
@@ -0,0 +1,56 @@
...
+import org.togetherjava.tjbot.features.MessageReceiverAdapter;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Listener that receives all pinned announcement messages from helper forum so that it can be
+ * removed
+ */
+public final class PinnedAnnouncementListener extends MessageReceiverAdapter {
```
Zabuzard, line 19: im not a fan of the name. id go for something in the direction of `RemoveHelpThreadPinMessageListener`. Perhaps u can find a shorter name in that direction

application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.features.MessageReceiverAdapter;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Listener that receives all pinned announcement messages from helper forum so that it can be
+ * removed
```
Zabuzard, line 17: Sounds a bit weird (and is missing dot at the end). Perhaps just reorder

> Removes notification-announcements about pinned messages from the helper forum.

application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public PinnedAnnouncementListener(Config config) {
+        isHelpForumName =
+                Pattern.compile(config.getHelpSystem().getHelpForumPattern()).asMatchPredicate();
+    }
+
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event) {
+        removePinnedAnnouncement(event);
+    }
```
Zabuzard, line 36: theres a problem with naming and responsibilities here. a method called `removePinnedAnnocuncement` sounds unconditional. but ur method still checks if its the right thread.

id suggest to either move the `if` stuff to this method, but then the question is if an extra method is still justified.
or to rename it to `handlePinnedAnnouncement`, to make it clear that it didnt decide anything yet

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
ankitsmt211: "RemoveHelpThreadPinMessageListener" sounds like it removes "pinned messages", imo it's more like an announcement for everyone to see something is pinned. Wanted to go with "HelpThreadPinnedAnnouncementListener" but it's a mouthful. Then is it necessary to add "help thread"? It's already in help package so that kinda of gives it away.

application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
ankitsmt211: Agreed, you have a solid reasoning.

application/src/main/java/org/togetherjava/tjbot/features/help/PinnedAnnouncementListener.java
Zabuzard: It doesnt just listen to it. It fulfils a very specific purpose, namely deleting them. If you would create a second feature that acts "on pin notifications", it would be in a separate class. hence why i think its crucial to have the actual action/purpose in the name. and not just the "when"

Zabuzard: `PinNotificationRemover`?

ankitsmt211: Much better

Zabuzard (approved):

Taz03 (approved):

PR #943, opened 2023-11-02
bug-fix transfer feature

ankitsmt211:
* default avatar if user do not have one
* remove embed from warning message to user
* minor fixes

if avatar is missing:
image

if present:
image

User warnings:
Before:
image

Now:
image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -7,7 +7,6 @@ import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.MessageEmbed;
 import net.dv8tion.jda.api.entities.User;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
-import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
 import net.dv8tion.jda.api.entities.channel.forums.ForumTag;
 import net.dv8tion.jda.api.entities.channel.forums.ForumTagSnowflake;
 import net.dv8tion.jda.api.entities.channel.unions.MessageChannelUnion;
@@ -30,7 +29,6 @@ import org.togetherjava.tjbot.features.utils.StringDistances;
 
 import java.awt.Color;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
@@ -81,7 +79,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
         String originalMessage = event.getTarget().getContentRaw();
         String originalMessageId = event.getTarget().getId();
-        String originalChannelId = event.getTarget().getChannel().getId();
+        String originalChannelId = event.getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
 
@@ -191,7 +189,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
                 """
                         Hello%s 👋 You have asked a question in the wrong channel%s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
 
-                        Your question has been automatically transferred to %s , please continue there, thank you 👍
+                        Your question has been automatically transferred to %s, please continue there, thank you 👍
                         """;
 
         String messageForDm = messageTemplate.formatted("", " on" + " " + guild.getName(),
@@ -206,10 +204,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     }
 
     private RestAction<Void> deleteOriginalMessage(JDA jda, String channelId, String messageId) {
-        TextChannel sourceChannel = Objects.requireNonNull(jda.getTextChannelById(channelId),
-                "Source channel could not be found for transfer-question feature");
-
-        return sourceChannel.deleteMessageById(messageId);
+        return jda.getTextChannelById(channelId).deleteMessageById(messageId);
     }
 
     private ForumChannel getHelperForum(JDA jda) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -191,7 +191,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
                 """
                         Hello%s 👋 You have asked a question in the wrong channel%s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
 
-                        Your question has been automatically transferred to %s, please continue there, thank you 👍
+                        Your question has been automatically transferred to %s , please continue there, thank you 👍
                         """;
 
         String messageForDm = messageTemplate.formatted("", " on" + " " + guild.getName(),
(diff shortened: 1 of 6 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -223,9 +229,11 @@ private static ForumTag getTagOrDefault(List<ForumTag> tagsFoundOnForum,
     }
 
     private MessageEmbed makeEmbedForPost(User originalUser, String originalMessage) {
+        ImageProxy avatarOrDefault = originalUser.getEffectiveAvatar();
```
Zabuzard, line 232: could u do a quick full-text-search of `getAvatar()`? i suspect that bug might be elsewhere as well, for example in `/audit`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
SquidXTV: https://github.com/search?q=repo%3ATogether-Java%2FTJ-Bot%20getAvatar&type=code

ankitsmt211: ty squid, @Zabuzard  should i update em all?

ankitsmt211:
image
There's one instance like this, where author is nullable so i can't use effectiveUrl. Any suggestions around this? or should i leave em as it is?

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Zabuzard: that would be cool, thanks

Zabuzard:
According to the JDA doc, the avatar url param is `Nullable`:

doc

Why again did our previous code crash? `null` should be supported here by JDA.

Zabuzard:
The bug was not giving `null` to `setAvatar(...)`, it was about us invoking a method on sth that was `null`, see:

code

ankitsmt211:
> The bug was not giving `null` to `setAuthor(...)`, it was about us invoking a method on sth that was `null`, see:
> 
> code

you are right about that, i'll discard other changes but i still think effectiveAvatar is better otherwise we get this 
image

With effectiveAvatar, its either userAvatar or a default one.

Zabuzard:
sure thing

ankitsmt211:
Changes are not significant, i have replaced avatar url only in all files. Few more bug fixes in `TransferFeature` class. I have tested the changes in two commands only though, `transfer-feature` and `report`. Like it should, displays a default avatar if none is present for user. If necessary i can also look into how about testing changes in other than these two features.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -232,10 +232,12 @@ private void reportScamMessage(MessageReceivedEvent event, String reportTitle,
         }
 
         User author = event.getAuthor();
+        String avatarOrDefaultUrl = author.getEffectiveAvatarUrl();
```
Taz03, line 235: the var seems unnecessary

application/src/main/java/org/togetherjava/tjbot/features/moderation/audit/AuditCommand.java
```diff
@@ -136,8 +136,10 @@ private List<List<ActionRecord>> groupActionsByPages(List<ActionRecord> actions)
     }
 
     private static EmbedBuilder createSummaryEmbed(User user, Collection<ActionRecord> actions) {
+        String avatarOrDefaultUrl = user.getEffectiveAvatarUrl();
```
Taz03, line 139: same here

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -204,7 +206,10 @@ private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, ForumPost
     }
 
     private RestAction<Void> deleteOriginalMessage(JDA jda, String channelId, String messageId) {
-        return jda.getTextChannelById(channelId).deleteMessageById(messageId);
+        TextChannel sourceChannel = Objects.requireNonNull(jda.getTextChannelById(channelId),
+                "Source channel could not be found for transfer-question feature");
+
+        return sourceChannel.deleteMessageById(messageId);
```
Taz03, line 212: what was the issue with the previous implementation?

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
ankitsmt211: I disagree, there's three kinds of urls method. `default`, `effective` and `avatar`. Now that you know what `effective` means, it prolly seems obvious but it's not. In the long run `avatarOrDefault` provides more accurate information imo. I actually had to test that `effective` meant `avatarOrDefault`.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: It's nothing too scary, possibly null value in case there's no such ID or if cache isn't initialized(latter isn't the problem i think, zabu mentioned something along the lines). IDE was screaming at me, so just to be on the safer side i chose to change this.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Taz03: just see the docs?

ankitsmt211: image

I have and i still think the variable is better.

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -189,7 +191,7 @@ private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, ForumPost
                 """
                         Hello%s 👋 You have asked a question in the wrong channel%s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
 
-                        Your question has been automatically transferred to %s, please continue there, thank you 👍
+                        Your question has been automatically transferred to %s , please continue there, thank you 👍
                         """;
 
         String messageForDm = messageTemplate.formatted("", " on" + " " + guild.getName(),
```
Zabuzard, line 197: `" on" + " "` is odd, why not just `" on "` in a single string.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -189,7 +191,7 @@ private RestAction<Message> dmUser(MessageChannelUnion sourceChannel, ForumPost
                 """
                         Hello%s 👋 You have asked a question in the wrong channel%s. Not a big deal, but none of the experts who could help you are reading your question there 🙁
 
-                        Your question has been automatically transferred to %s, please continue there, thank you 👍
+                        Your question has been automatically transferred to %s , please continue there, thank you 👍
```
Zabuzard, line 194: the extra space is confusing and the next guy seeing this code will remove it again.

was this to get rid of the auto-preview? then put it into an extra variable with a proper name and perhaps a comment. and move it down to where you put in the jumpurl, not up here.
so:

```java
// Prevents Discord creating a distracting auto-preview for the link
String jumpUrlSuffix = " ";

messageTemplate.formatted(..., forumPost.message.getJumpUrl() + jumpUrlSuffix);
```

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

PR #945, opened 2023-11-05
Feat/delete messages upto 24 hrs

### LINKED ISSUE #912: Delete last X messages from user for moderation
Problem:
Sometimes you have a lot of messages from a user, that might not be appropriate or against the rules. Folks who write texts in smaller lengths, it can pile up to an "annoying no of deletes"

Possible solution:
 Introduction of new command "/delete user nNoOfMessages", which deletes the most n messages from given user.

ankitsmt211:
First draft of feature, delete messages upto 24 hrs in same channel.
* db table to keep track of messages meta data
* routine to clear older records than 24 hrs
* listener to create new records on each new message
* new package for feature
* new command "/purge-in-channel"
* resolves #912 

Additions:
* db entries also has an emergency brake to prevent spamming DB to death.
* a conditional sub routine, when approaching records limit auto deletes N no of oldest records from table to make some space for new entries.

command interface:

image

when there's no records:
image

if there's any records within selected duration:
image

Note: Duration is optional, if skipped it defaults to last hour.


ankitsmt211:
Couple of things to work atm:
* do something with reason for using this command, perhaps store it somewhere.
* How long to keep message data in DB, currently anything older than 2 hrs is removed.

Zabuzard (changes requested):
Please dont forget to also update the `PP.md` file. We must inform users about how we store their data.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageHistoryRoutine.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public Schedule createSchedule() {
+        return new Schedule(ScheduleMode.FIXED_RATE, 0, SCHEDULE_INTERVAL_SECONDS,
+                TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void runRoutine(JDA jda) {
+        Instant now = Instant.now();
+        Instant twoHourBack = now.minus(2, ChronoUnit.HOURS);
```
Zabuzard, line 40: `twoHourBack` is a bad variable name and its bound to rot.

dont describe the what, describe the why. what is this instant used for? it describes all messages that will be purged, so `purgeMessagesBefore` or sth like that

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageHistoryRoutine.java
```diff
@@ -0,0 +1,47 @@
...
+
+        database.write(context -> context.selectFrom(MESSAGE_HISTORY)
+            .where(MESSAGE_HISTORY.SENT_AT.lessThan(twoHourBack))
+            .stream()
+            .forEach(UpdatableRecordImpl::delete));
```
Zabuzard, line 45: streams must be put under `try-with-resources` to ensure they are closed.

also, u should look into a batch-delete. im not sure if this would delete the records one by one or send a single delete request, please check the doc.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageListener.java
```diff
@@ -0,0 +1,46 @@
...
+import java.time.Instant;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.features.MessageReceiverAdapter;
+
+/**
+ * Listens for new message throughout the guild, then stores some metadata for each message in a
+ * database.
+ */
+public class MessageListener extends MessageReceiverAdapter {
```
Zabuzard, line 14: too generic name. adjust it to fit the purpose. this one is for storing messagemetadata. `MessageMetadataListener`.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageListener.java
```diff
@@ -0,0 +1,46 @@
...
+        long authorId = event.getAuthor().getIdLong();
+
+        database.write(context -> context.newRecord(MESSAGE_HISTORY)
+            .setSentAt(Instant.now())
+            .setGuildId(guildId)
+            .setChannelId(channelId)
+            .setMessageId(messageId)
+            .setAuthorId(authorId)
+            .insert());
+    }
```
Zabuzard, line 45: there needs to be a form of cooldown prevention to ensure that a single user cant spam our database/vps to death by posting a million messages in the night.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param database the database to retrieve message records of a target user
+     */
+
+    public PurgeHistoryCommand(Database database) {
+        super(COMMAND_NAME, "delete user message history upto an hour", CommandVisibility.GUILD);
+        getData()
+            .addOption(OptionType.USER, USER_OPTION,
+                    "user who's messages you want to purge within an hour", true)
+            .addOption(OptionType.STRING, REASON_OPTION, "Reason for purging user messages", true);
```
Zabuzard, line 41: here u started the sentence with capital R, the other sentences started small. consistency

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
```diff
@@ -0,0 +1,123 @@
...
+        String reason = Objects.requireNonNull(event.getOption(REASON_OPTION).getAsString(),
+                "reason is null");
+
+        handleHistory(event, author, reason, targetOption, event.getHook());
+    }
+
+    private void handleHistory(SlashCommandInteractionEvent event, Member author, String reason,
+            OptionMapping targetOption, InteractionHook hook) {
+        Instant now = Instant.now();
+        Instant oneHourBack = now.minus(1, ChronoUnit.HOURS);
```
Zabuzard, line 61: same naming issue as before

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
```diff
@@ -0,0 +1,123 @@
...
+
+        User target = targetOption.getAsUser();
+        String sourceChannelId = event.getChannel().getId();
+
+        if (!validateHierarchy(author, targetOption)) {
+            hook.sendMessage("Target user has a higher role than you").setEphemeral(true).queue();
```
Zabuzard, line 67: the message is confusing. it should start with something like "dude, there is an issue. and this is the reason ..." instead of just "reason".
so perhaps
> Cannot purge history of an user with a higher role than you

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
```diff
@@ -0,0 +1,123 @@
...
+            return;
+        }
+
+        List<String> messageIdsForDeletion = new ArrayList<>();
+
+        database.write(context -> context.selectFrom(MESSAGE_HISTORY)
+            .where(MESSAGE_HISTORY.AUTHOR_ID.equal(target.getIdLong())
+                .and(MESSAGE_HISTORY.CHANNEL_ID.equal(Long.valueOf(sourceChannelId)))
+                .and(MESSAGE_HISTORY.SENT_AT.greaterOrEqual(oneHourBack)))
+            .stream()
```
Zabuzard, line 77: try-with-resources

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    private final Database database;
+
+    /**
+     * Creates an instance of the command.
+     *
+     * @param database the database to retrieve message records of a target user
+     */
+
+    public PurgeHistoryCommand(Database database) {
+        super(COMMAND_NAME, "delete user message history upto an hour", CommandVisibility.GUILD);
```
Zabuzard, line 37: the command only purges within the given channel, no? not clear from the description.
also, to prevent confusion, id actually rename the command then. perhaps to `purge-in-channel`

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/package-info.java
```diff
@@ -0,0 +1,9 @@
+/**
+ * This package offers moderation command to delete messages from a user within last hour.
```
Zabuzard, line 2: too specific. the description will rot and be incorrect within no time.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeHistoryCommand.java
ankitsmt211: yes, that's correct. Will do.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageListener.java
ankitsmt211: that's a good idea, 100 messages seems right? or should it be lower.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageHistoryRoutine.java
ankitsmt211: got it.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageHistoryRoutine.java
ankitsmt211: couldn't think of a better one at the time, i think imma go with your suggestion.

ankitsmt211:
> Please dont forget to also update the `PP.md` file. We must inform users about how we store their data.

Is this necessary? `user_id`, `message_id`, `timestamp` of actions, `channel_id` is already mentioned in database section.

source:
image

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/MessageListener.java
Zabuzard: sounds fine

Zabuzard:
> > Please dont forget to also update the `PP.md` file. We must inform users about how we store their data.
> 
> Is this necessary? `user_id`, `message_id`, `timestamp` of actions, `channel_id` is already mentioned in database section.
> 
> source: image

Ah, good. Perhaps tweak this paragraph:

> Furthermore, upon utilization of our help service, `user_ids`, `channel_ids` and `timestamps` are stored to track when/how many questions a user asks. The data may be stored for up to 30 days.

(added `and timestamps`)

ankitsmt211:
Atm, duration is optional (default is 1hr) but if provided can be entered from 1-24 as value. I was thinking perhaps a better way would be to provide choices such as `3 hrs`, `6 hrs`, `12 hrs` and `24 hrs` where 1 hr is default. This makes easier for user to enter duration altho there are already checks in place for value to be between 1-24 if i stick to current way of input.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeMessageListener.java
```diff
@@ -41,16 +47,35 @@ private void updateHistory(MessageReceivedEvent event) {
...
+                .setSentAt(Instant.now())
+                .setGuildId(guildId)
+                .setChannelId(channelId)
+                .setMessageId(messageId)
+                .setAuthorId(authorId)
+                .insert());
+
+            incrementRecordsCounter();
+        } else {
+            logger.warn("purge history reached limit");
```
Zabuzard, line 61: when this happens, u will spawn our logs to death. which will fill the disk even faster, lol.

lower the log level to DEBUG or make it only write this log every 5min or so.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeMessageListener.java
ankitsmt211: Record limit for a day is pretty high so we shouldn't hit the warning in near future. But it's obviously going to hit the limit at one point in future. So debug seems like a better alternative than timer based.

Zabuzard: i just mean, ur deating the purpose of why u created the limit in the first place.

u created it to not spam the DB in order to prevent the disk from running out of space. but instead, u create an entry in the log file on each message, which gets saved in the log files instead. so you shifted the problem from DB to log file, the disk is still running out of space.

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeMessageListener.java
```diff
@@ -58,7 +58,7 @@ private void updateHistory(MessageReceivedEvent event) {
 
             incrementRecordsCounter();
         } else {
-            logger.warn("purge history reached limit");
+            logger.debug("purge history reached limit");
```
Zabuzard, line 61: the log would be more useful if it also gives a suggestion how to fix the problem.

"Purge History reached limit. Either delete some rows of the table or wait for them to expire."

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/history/PurgeMessageListener.java
ankitsmt211: Actually i think i might have a better idea, Imagine the DB is maxed out with 7.5k messages. No more records are being saved, which means any new spam messages would be ignored and using `purge-in-channel` would only delete older messages which might not be the best thing . Ideally i think we should keep the messages from newest to oldest upto the "max limit". 

Possible Solution:
Let's say we have a limit 7.5k, we hit a mark of 7.4k. A routine should be trigged that only clears out last 500 records or so. This way all the new messages are being tracked and command is not useless past the limit. Thoughts ?

ankitsmt211:
@Zabuzard  @Tais993 @Taz03 
I need some suggestions about a possible problem,

Context:
* A routine (`MessageHistoryRoutine`) that clears DB records every 24 hrs(no need to hold records past that)
* Purge command that has duration options such as 3hrs, 6 hrs, 12 hrs, 24 hrs.(since we're only storing records upto 24 hrs, max is that)


It become problematic when someone in future adds a new option in `PurgeHistoryCommand`  for 48 hrs or 72 hrs and forgets to reflect that change in routine class that clears every expired record past 24 hrs(so db will never hold anything past 24 hrs). 

Point is numbers can change with time and i wanna make sure both the classes are on same page when it comes to this duration limit.

I decided to move constants to a different class `Duration`, figured it would be neat to add other constants related to duration as well. I'm still not sure if this helps much, Any alternatives or thoughts or perhaps im doing something wrong?

### Current:
`MessageHistoryRoutine.java`

image

`PurgeHistoryCommand.java`

image

### **Possibly newer changes:**
Add `Duration.java`

image

`MessageHistoryRoutine.java`

image


`PurgeHistoryCommand.java`

image

ankitsmt211:
This feature has reached it's final stage and is now open for reviews.

PR #948, opened 2023-11-12
Bugfix/role back message

### LINKED ISSUE #946: get role back message is wrong
when you are inactive for a while you get your helper role removerd. he then says to go to #⁠questions to get it again, but you need to go to #⁠select-your-roles. The text needs to be changed.

ankitsmt211:
* adds channel pattern for `select-your-role` in `config.java` and `template`
* method to fetch `select-your-role` channel
* replace `help-forum` with `select-your-channel`
resolves #946 

## Release Note

Changes Config, adds:

```json
"selectRolesChannelPattern": "select-your-roles"
```


application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
```diff
@@ -150,15 +150,16 @@ private boolean isMemberInactive(Member member, Instant when) {
                     .and(HELP_CHANNEL_MESSAGES.SENT_AT.greaterThan(latestActiveMoment)))) == 0;
     }
 
-    private void pruneMemberFromRole(Member member, Role role, ForumChannel helpForum) {
+    private void pruneMemberFromRole(Member member, Role role, JDA jda) {
         Guild guild = member.getGuild();
+        TextChannel selectRolesChannel = getSelectRolesChannel(jda);
```
Zabuzard, line 155: ur executing this on each member who u prune, instead of finding the channel beforehand **once** and passing the channel down

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
```diff
@@ -173,4 +174,14 @@ private void warnModsAbout(String message, Guild guild) {
...
     }
+
+    private TextChannel getSelectRolesChannel(JDA jda) {
+        Optional<TextChannel> selectRolesChannelOptional = jda.getTextChannels()
+            .stream()
+            .filter(textChannel -> selectYourRolesChannelNamePredicate.test(textChannel.getName()))
+            .findFirst();
+
+        return selectRolesChannelOptional.orElseThrow(() -> new IllegalStateException(
+                "Could not find role selection channel during AutoPruneHelper routine. Make sure the config is setup properly."));
```
Zabuzard, line 185: the channel isnt necessary for the feature to work. pruning is also more of a background thing, it shouldnt throw exceptions for a missing channel.

have it a warn-log instead and fallback to leaving out the channel part of the DM

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
ankitsmt211: Im not sure if i understand, do you suggest to have a field which use `getRoleSelectChannel` method to get the channel and using that?

ankitsmt211: i think i understand, you mean instead of passing down jda i should get the channel at the top and pass that instead.

Zabuzard: > i think i understand, you mean instead of passing down jda i should get the channel at the top and pass that instead.
yes

Zabuzard (dismissed):

ankitsmt211:
added a minor fix for earlier merge, somehow i managed to sneak  an if block without curlies.

Zabuzard (approved):

PR #955, opened 2023-11-17
refactor onModalSubmit in TransferQuestionCommand to handle multiple users submission

ankitsmt211:
Problem:
When multiple users use `transfer-question` feature at the same time, it can lead to creation of two or more threads in helper forum in rare occasions. To verify this you'd prolly need an alt account to time it correclty.

Solution:
Bug fix handles this problem by retrieving the original message that needs to be transferred on submitting the transfer modal,
if it's already transferred then message shouldn't be present.

Implementation:
* extracted `onModalSubmit` chain to a method `transferFlow` for clarity.
* added a method(`handledAlready`) to notify user if the transfer was already handled.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -32,6 +32,7 @@ import java.awt.Color;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
@@ -128,12 +129,13 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         ForumChannel helperForum = getHelperForum(event.getJDA());
         TextChannel sourceChannel = event.getChannel().asTextChannel();
 
-        // to avoid race condition, when multiple users do the transfer.
-        // message is retrieved, already transferred messages are deleted.
-        event.getChannel()
-            .retrieveMessageById(messageId)
-            .queue(notHandled -> transferFlow(event, channelId, authorId, messageId),
-                    handled -> alreadyHandled(sourceChannel, helperForum));
+        // Has been handled if original message was deleted by now.
+        // Deleted messages cause retrieveMessageById to fail.
+        Consumer<Message> notHandledAction =
+                any -> transferFlow(event, channelId, authorId, messageId);
+        Consumer<Throwable> handledAction = any -> alreadyHandled(sourceChannel, helperForum);
+
+        event.getChannel().retrieveMessageById(messageId).queue(notHandledAction, handledAction);
     }
 
     private void transferFlow(ModalInteractionEvent event, String channelId, String authorId,
@@ -150,7 +152,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
 
     private void alreadyHandled(TextChannel sourceChannel, ForumChannel helperForum) {
         sourceChannel.sendMessage(
-                "The question has been already forwarded to the helper forum for assistance. Kindly review %s for more information."
+                "It appears that someone else has already transferred this question. Kindly see %s for details."
                     .formatted(helperForum.getAsMention()))
             .queue();
     }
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -125,6 +125,19 @@ public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
...
         String channelId = args.get(2);
+        ForumChannel helperForum = getHelperForum(event.getJDA());
+        TextChannel sourceChannel = event.getChannel().asTextChannel();
+
+        // to avoid race condition, when multiple users do the transfer.
+        // message is retrieved, already transferred messages are deleted.
+        event.getChannel()
+            .retrieveMessageById(messageId)
+            .queue(notHandled -> transferFlow(event, channelId, authorId, messageId),
+                    handled -> alreadyHandled(sourceChannel, helperForum));
```
Zabuzard, line 136: u can avoid the comments by writing more self-explanatory code. in this case, u could create a helper method named  `handleIfNotAlreadyHandled` or `handleIfUnhandled`.

whats tripping me, as reader, is the rather unconvential way this detects deletion (im not saying its wrong, but a normal user would expect some simple `if`).
so I think a comment would be helpful that says:
```java
// Has been handled if original message was deleted by now.
// Deleted messages cause retrieveMessageById to fail.
```

The only other comment I have is that right now u react on any failure of `retreiveMessageById` with "oh, so it was handled already". But that might not be the case, the method can also fail bc of other reasons.
The interface should give u the exception (likely that `handled` thing?) and JDA offers the failure-reason as enum somewhere. if u search the code base with `instanceof`, u should find examples.

Also, I find the use of `notHandled` and `handled` confusing. Ur naming these variables to indicate what the lambda is about, but by doing that, one completely doesnt understand what the variable itself is even for. Instead, name the lambda:
```java
Foo notHandledAction = any -> transferFlow(...);
Foo handledAction = any -> alreadyHandled(...);

event.getChannel()
  ...
  queue(notHandledAction, handledAction);
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -135,6 +148,13 @@ public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
             .queue();
     }
 
+    private void alreadyHandled(TextChannel sourceChannel, ForumChannel helperForum) {
+        sourceChannel.sendMessage(
+                "The question has been already forwarded to the helper forum for assistance. Kindly review %s for more information."
```
Zabuzard, line 153: Perhaps rename:
> It appears that someone else has already transferred this question. Kindly see %s for details.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: Hmm i wasn't too happy with the handling either, you're right it could fail due to a lot of reasons, i'll start refactoring thanks.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -133,7 +135,13 @@ public void onModalSubmitted(ModalInteractionEvent event, List<String> args) {
         // Deleted messages cause retrieveMessageById to fail.
         Consumer<Message> notHandledAction =
                 any -> transferFlow(event, channelId, authorId, messageId);
-        Consumer<Throwable> handledAction = any -> alreadyHandled(sourceChannel, helperForum);
+
+        Consumer<Throwable> handledAction = any -> {
+            if (any instanceof ErrorResponseException errorResponseException
+                    && errorResponseException.getErrorResponse() == ErrorResponse.UNKNOWN_MESSAGE) {
+                alreadyHandled(sourceChannel, helperForum);
+            }
```
Zabuzard, line 143: ... and `if` not, perhaps log the issue as error. `log.error("Unknown error while ...", any)`.
And since u now actually use `any`, it shouldnt be named like that anymore. Perhaps `error` or `failure`.

Zabuzard (changes requested):

Zabuzard (approved):

Taz03 (approved):

PR #971, opened 2023-11-30
upgrade gradle wrapper to 8.5

Taz03:


SquidXTV:
Could it be that you need to reload gradle once in irder to install new gradle wrapper jar? Other Gradle upgrades always got changes in the wrapper file

Taz03:
tried doing `./gradlew build` but nothing happened

ankitsmt211:
> tried doing `./gradlew build` but nothing happened

what's the issue here? This is what's recommended if you want the latest version, for specific version you can mention the version.

image

src: https://docs.gradle.org/current/userguide/gradle_wrapper.html#:~:text=Upgrading%20the%20Gradle%20Wrapper,-Projects%20typically%20want&text=One%20way%20to%20upgrade%20the,in%20Adding%20the%20Gradle%20Wrapper.

SquidXTV:
> tried doing ./gradlew build but nothing happened

You sure you use the wrapper installation instead of your local installation? Check IntelliJ settings, iirc the jar must update too

Zabuzard:
yeah. best is to dig out an old PR where we updated gradle already and then follow along

Taz03:
i tried fixing deprecations but these are still  left, might want to create an issue for it
image

SquidXTV (approved):

Zabuzard (approved):

PR #972, opened 2023-12-01
upgrade java to 21

Taz03:


marko-radosavljevic (approved):

marko-radosavljevic:
I will do the toolchains as a separate PR, so it can be tested if it works predictably well on different environments. It's a pretty mature tech now, and there is decent IDE support. It will be an imporvement in any case tho.

PR #973, opened 2023-12-01
handle invalid channel transfers

ankitsmt211:
using `transfer-feature` from either a thread or forum thread would result into an error. Since it's not meant to be used outside of regular chat channels.

image

Pr refactors `isInvalidTransfer` to handle it with an ephermal response.

image


ankitsmt211:
#976 will be resolving the same issue.

PR #966, opened 2023-11-24
Continue/feature/reference gh

ankitsmt211:
continuing #599 

### Note: This merge will require `Config` to be changed on vps as well, otherwise bot crashes.

brief:
* adds `/github-search`, which auto suggests a list of issues and PR to choose. It also accepts a title if you want to manually enter and select.
* if a new message is of type `#IssueNumberHere`, it will reply back with an embed containing information about issue.
* This is currently set to channel `#general` using `config`, can be tweaked to a particular channel.
* More repositoriy Id's can be added to `config` to allow searching across multiple repositories.

Changes:
* resolve merge conflicts
* Java doc fixes
* add avatar of user to embed
* add date of creation to footer of embed
* refactor embed builder for response

Embed on search/issue reference:

image

PR #981, opened 2023-12-10
Continue/feature/reference gh

ankitsmt211:
### **For earlier context related to PR #966** 

continuing #599 

### Note: This merge will require `Config` to be changed on vps as well, otherwise bot crashes.

brief:
* adds `/github-search`, which auto suggests a list of issues and PR to choose. It also accepts a title if you want to manually enter and select.
* if a new message is of type `#IssueNumberHere`, it will reply back with an embed containing information about issue.
* This is currently set to channel `#general` using `config`, can be tweaked to a particular channel.
* More repositoriy Id's can be added to `config` to allow searching across multiple repositories.

Changes:
* resolve merge conflicts
* Java doc fixes
* add avatar of user to embed
* add date of creation to footer of embed
* refactor embed builder for response

Embed on search/issue reference:

image


Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
```diff
@@ -0,0 +1,122 @@
...
+            try {
+                return repo.getIssues(GHIssueState.ALL);
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        })
+            .flatMap(List::stream)
+            .sorted(GITHUB_ISSUE_TIME_COMPARATOR)
+            .map(issue -> "[#%d] %s".formatted(issue.getNumber(), issue.getTitle()))
+            .toList();
```
SquidXTV, line 118: what happened to this formatting?

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubCommand.java
SquidXTV: image

ankitsmt211: what's wrong?

SquidXTV: The formatting is weird, and why is it using parallelStream?

ankitsmt211: sequence of repositories shouldn't be a problem, prolly just a preference by original contributor. We're following google style sheet for formatting, everything is formatted accordingly otherwise it wouldn't pass pre merge test cases.

SquidXTV: ok I understand the formatting, but the `#parallelStream` is probably slower, why using it then?

ankitsmt211: from my understanding, parallel Streams are faster since we're performing each operation parallely although the difference here won't matter much since dataset is not significant enough. But it comes at a cost of sequence, so if you need order to be maintained then you should stick to normal ones.

SquidXTV: The problem is also that using parallel streams adds overhead for managing multiple threads, memory locality, ...
See [Baeldung: When to use parallel stream](https://www.baeldung.com/java-when-to-use-parallel-stream)
especially the NQ model, do we have more than 10000 elements in the stream?
> Parallelism can bring performance benefits in certain use cases. But parallel streams cannot be considered as a magical performance booster.

In this case, I think sequential stream is better

ankitsmt211: fair, there's no justification for going beyond the normal route. I'll swap it with sequential stream.

SquidXTV (dismissed):

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,209 @@
...
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    @Override
+    public void onMessageReceived(MessageReceivedEvent event) {
+        if (event.getAuthor().isBot() || !isAllowedChannelOrChildThread(event)) {
+            return;
+        }
```
ankitsmt211, line 90: currently `#issueNumber` is only allowed in certain channels, but it's possible people might use em in other channels and then without any indication they won't know where to use this command. Imo we should respond with an ephermal response here something along the lines, "Hey, you used github reference feature in wrong channel. Please use `#allowedChannel`"

On a second note, why block this in other channels.
image
These are allowed channels currently.
Thoughts ?

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
```diff
@@ -0,0 +1,209 @@
...
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    /**
+     * Looks through all of the given repositories for an issue/pr with the given id.
+     */
+    Optional<GHIssue> findIssue(int id) {
+        return repositories.stream().map(repository -> {
+            try {
```
ankitsmt211, line 182: changing to sequential stream might have caused some problems, searches are inaccurate for some reasons. Switching back to parallel stream resolves the issue for `/github-search` but it shows inconsistent behaviour for `#issueNumber` based searches since we have multiple repsitories, so sometimes it will link issue number 11 from tjbot repo, sometimes from tjplays repo if there's a match. I'm trying to find a way to make it consistent.

PR is not up for merge yet.

SquidXTV: > changing to sequential stream might have caused some problems, searches are inaccurate for some reasons. Switching back to parallel stream resolves the issue for `/github-search`

This sounds very wrong to me, the error must lie somewhere else

> but it shows inconsistent behaviour for #issueNumber based searches since we have multiple repsitories, so sometimes it will link issue number 11 from tjbot repo, sometimes from tjplays repo if there's a match. I'm trying to find a way to make it consistent. Although im not sure if this is a concern, since issue numbers won't usually match for current number of issues across repos.
TJ-Bot is way ahead in issue numbers than modernjava or tjplays.

This probably comes from using `findAny`, you might want to use the config for the order, which to prefer.

marko-radosavljevic:
Haha, you didn't need to dismiss 2 passing reviews just to add repos to the template, it's just a template anyway. ^^

No worries tho, I will count them as passing. :heart:

ankitsmt211:
> Haha, you didn't need to dismiss 2 passing reviews just to add repos to the template, it's just a template anyway. ^^
> 
> No worries tho, I will count them as passing. ❤️

apologies if that comes off as rude, but i didnt want someone to merge it(it was one click away ) by any chance since the results were inconsistent with when parallel stream was swapped with sequential, also the config changes ofc. It's better to fix issues when it's already active then making a new PR😉.

application/src/main/java/org/togetherjava/tjbot/commands/github/GitHubReference.java
ankitsmt211: i'm investing the underlying problem, but this was just an observation of mine due to a recent change.

ankitsmt211: found out the cause, since both features were finding issues by id so it was defaulting to first repo aka tj-bot. Now github search also uses title along with id to match whereas github reference uses default repo id to search results accordingly. New changes resolves both issues.

SquidXTV (approved):

Zabuzard:
im ready to merge it, but theres still one open conversation @ankitsmt211

ankitsmt211:
> im ready to merge it, but theres still one open conversation @ankitsmt211 

Go ahead, marked it resolved.

Zabuzard (approved):

PR #986, opened 2023-12-11
handling error when member left guild

ankitsmt211:
During auto-archive routine for help threads, bot sends an embed educating OP why the thread was archived, it also pings OP in the process. This whole archiving routine depends on whether member was retrieved, thus it fails consecutively until discords built in archive feature kicks in.

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -32,8 +32,8 @@ import java.util.function.Function;
  */
 public final class HelpThreadAutoArchiver implements Routine {
     private static final Logger logger = LoggerFactory.getLogger(HelpThreadAutoArchiver.class);
-    private static final int SCHEDULE_MINUTES = 6;
-    private static final Duration ARCHIVE_AFTER_INACTIVITY_OF = Duration.ofSeconds(12);
+    private static final int SCHEDULE_MINUTES = 60;
+    private static final Duration ARCHIVE_AFTER_INACTIVITY_OF = Duration.ofHours(12);
 
     private final HelpSystemHelper helper;
 
@@ -48,7 +48,7 @@ public final class HelpThreadAutoArchiver implements Routine {
 
     @Override
     public Schedule createSchedule() {
-        return new Schedule(ScheduleMode.FIXED_RATE, 0, SCHEDULE_MINUTES, TimeUnit.SECONDS);
+        return new Schedule(ScheduleMode.FIXED_RATE, 0, SCHEDULE_MINUTES, TimeUnit.MINUTES);
     }
 
     @Override
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -159,10 +159,10 @@ public final class HelpThreadAutoArchiver implements Routine {
             })
             .mapToResult()
             .flatMap(sentEmbed -> {
-                if (sentEmbed.isFailure()) {
-                    return handleFailure.apply(sentEmbed.getFailure());
+                if (sentEmbed.isSuccess()) {
+                    return archiveThread.get();
                 }
-                return archiveThread.get();
+                return handleFailure.apply(sentEmbed.getFailure());
             })
             .queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -140,8 +140,10 @@ public final class HelpThreadAutoArchiver implements Routine {
                 if (member.isSuccess()) {
                     return sendEmbedWithMention.apply(member);
                 }
+                LOGGER.info(
+                        "Owner of thread with id: {} left the server, sending embed without mention",
+                        threadChannel.getId(), member.getFailure());
 
-                LOGGER.debug("Unable to mention user", member.getFailure());
                 return sendEmbedWithoutMention.apply(member);
             })
             .mapToResult()
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -162,6 +162,7 @@ public final class HelpThreadAutoArchiver implements Routine {
                 if (sentEmbed.isSuccess()) {
                     return archiveThread.get();
                 }
+
                 return handleFailure.apply(sentEmbed.getFailure());
             })
             .queue();
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -9,7 +9,6 @@ import net.dv8tion.jda.api.entities.MessageEmbed;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
-import net.dv8tion.jda.api.exceptions.ErrorResponseException;
 import net.dv8tion.jda.api.requests.RestAction;
 import net.dv8tion.jda.api.utils.Result;
 import net.dv8tion.jda.api.utils.TimeUtil;
@@ -148,7 +147,7 @@ public final class HelpThreadAutoArchiver implements Routine {
             })
             .mapToResult()
             .flatMap(sentEmbed -> {
-                if (sentEmbed.getFailure() instanceof ErrorResponseException) {
+                if (sentEmbed.isFailure()) {
                     LOGGER.warn(
                             "Unknown error occurred during help thread auto archive routine, archiving thread",
                             sentEmbed.getFailure());
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -23,7 +23,6 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -129,16 +128,17 @@ public final class HelpThreadAutoArchiver implements Routine {
 
     private void handleArchiveFlow(ThreadChannel threadChannel, MessageEmbed embed) {
 
-        Consumer<Throwable> logFailure = error -> {
+        Supplier<RestAction<Void>> archiveThread =
+                () -> threadChannel.getManager().setArchived(true);
+
+        Function<Throwable, RestAction<Void>> handleFailure = (error) -> {
             if (error instanceof ErrorResponseException) {
                 logger.warn(
                         "Unknown error occurred during help thread auto archive routine, archiving thread",
                         error);
             }
+            return archiveThread.get();
         };
-
-        Supplier<RestAction<Void>> archiveThread =
-                () -> threadChannel.getManager().setArchived(true);
 
         Function<Result<Member>, RestAction<Message>> sendEmbedWithMention =
                 member -> threadChannel.sendMessage(member.get().getAsMention()).addEmbeds(embed);
@@ -157,9 +157,11 @@ public final class HelpThreadAutoArchiver implements Routine {
                 logger.debug("Unable to mention user", member.getFailure());
                 return sendEmbedWithoutMention.apply(member);
             })
-            .flatMap(any -> archiveThread.get())
-            .onErrorFlatMap(error -> {
-                logFailure.accept(error);
+            .mapToResult()
+            .flatMap(sentEmbed -> {
+                if (sentEmbed.isFailure()) {
+                    return handleFailure.apply(sentEmbed.getFailure());
+                }
                 return archiveThread.get();
             })
             .queue();
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -3,11 +3,15 @@ package org.togetherjava.tjbot.features.help;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.MessageEmbed;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.exceptions.ErrorResponseException;
+import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.api.utils.Result;
 import net.dv8tion.jda.api.utils.TimeUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -20,6 +24,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 /**
  * Routine, which periodically checks all help threads and archives them if there has not been any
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,16 +127,30 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
     }
 
     private void handleArchiveFlow(ThreadChannel threadChannel, MessageEmbed embed) {
+
         Consumer<Throwable> handleFailure = error -> {
             if (error instanceof ErrorResponseException) {
                 logger.warn("Unknown error occurred during help thread auto archive routine",
                         error);
             }
```
ankitsmt211, line 135: if something breaks, should we archive anyway?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,16 +127,30 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+
         threadChannel.getGuild()
             .retrieveMemberById(threadChannel.getOwnerIdLong())
-            .flatMap(author -> threadChannel.sendMessage(author.getAsMention()).addEmbeds(embed))
+            .mapToResult()
+            .flatMap(member -> {
+                if (member.isSuccess()) {
+                    return sendEmbedWithMention.apply(member);
+                }
+                logger.debug("Member already left server, archiving thread without mention");
```
Taz03, line 151: just say cannot mention user

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
Taz03: yes ig

Zabuzard (approved):

Zabuzard (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,16 +127,34 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+                if (member.isSuccess()) {
+                    return sendEmbedWithMention.apply(member);
+                }
+
+                logger.debug("Unable to mention user", member.getFailure());
+                return sendEmbedWithoutMention.apply(member);
+            })
             .flatMap(any -> threadChannel.getManager().setArchived(true))
             .queue(any -> {
             }, handleFailure);
```
Taz03, line 160: do an `onErrorMap` so we'll always only do 1 queue

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: not sure if i understand wym, could you show an example? Thanks

Taz03:
```java
        .flatMap(...)
        .onErrorMap(any -> {
            log...
            return threadChannel.archive();
        })
        .queue();
```
Something like  this

Zabuzard: (or onErrorFlatMap, if that results in another rest action)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +127,37 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+                return sendEmbedWithoutMention.apply(member);
+            })
             .flatMap(any -> threadChannel.getManager().setArchived(true))
-            .queue(any -> {
-            }, handleFailure);
+            .onErrorFlatMap(error -> {
+                handleFailure.accept(error);
+                return threadChannel.getManager().setArchived(true);
+            })
+            .queue();
```
ankitsmt211, line 161: too ugly?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +128,40 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
         };
 
+        Supplier<RestAction<Void>> archiveThread =
+                () -> threadChannel.getManager().setArchived(true);
+
+        Function<Result<Member>, RestAction<Message>> sendEmbedWithMention =
+                member -> threadChannel.sendMessage(member.get().getAsMention()).addEmbeds(embed);
+
+        Function<Result<Member>, RestAction<Message>> sendEmbedWithoutMention =
+                member -> threadChannel.sendMessageEmbeds(embed);
```
ankitsmt211, line 147: lol whole gang is here

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +128,40 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+                if (member.isSuccess()) {
+                    return sendEmbedWithMention.apply(member);
+                }
+
+                logger.debug("Unable to mention user", member.getFailure());
+                return sendEmbedWithoutMention.apply(member);
+            })
+            .flatMap(any -> archiveThread.get())
+            .onErrorFlatMap(error -> {
+                logFailure.accept(error);
```
Taz03, line 162: better to  make put the code here instead of making  consumer

Taz03 (changes requested):

Zabuzard (changes requested):
that duplication here isnt good:
dup

also, ur technically also attempting to archive when archiving itself failed, which isnt good.
the "try this, but always archive" shouldnt span across archiving itself.

instead, try to write it like:
```java
...
.flatMap(member -> ...)
.onError(handleFailure)
.flatMap(any -> threadChannel.getManager().setArchived(true))
.queue();
```

ankitsmt211:
@Zabuzard im blank here, it requires me to return a `RestAction<Message>`, i don't think i can do that with `OnError....` since we already handled with or without mention in `mapToResult`, there's no point in sending a new message. The only reason i was think to handle archiving as backup was if something breaks and `AutoArchiver` retries again and again so archive anyway. On a second note, do we even need it? `Threads` cant be deleted, only other variable was member leaving. Now if somehow it breaks then during second try, should be archived anyway. Thoughts?

image

Zabuzard:
might misremember the name. onFailure(...) perhaps.
personally, i dont mind u removing the error handling on that part. i dont think it's necessary

Zabuzard:
just checked. was looking for the equivalent to my onSuccess(...) method. seems it doesnt exist.
so it would either be:

.onErrorMap(any -> {
  do the log thing;
  return any;
}

or via mapToResults

ankitsmt211:
seems much better, at least chain is pretty readable now.

Zabuzard (dismissed):

ankitsmt211:
Don't merge yet.

ankitsmt211:
@Taz03 It's good to go when you can.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +127,44 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+                logger.debug("Unable to mention user", member.getFailure());
+                return sendEmbedWithoutMention.apply(member);
+            })
+            .mapToResult()
+            .flatMap(sentEmbed -> {
+                if (sentEmbed.isSuccess()) {
+                    return archiveThread.get();
+                }
+
+                return handleFailure.apply(sentEmbed.getFailure());
```
Taz03, line 166:
```suggestion
                if (sentEmbed.getFailure() instanceof ErrorResponseException) {
                    // log
                }

                return threadChannel.getManager().setArchived(true);
```

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: what's the issue with earlier implementation? It tells a beautiful story, if embed is a success archive the thread otherwise handle failure..which logs and then regardless archives it. It looks like you're trying to squeeze in the failure part at the cost of readability.

Taz03: if its failure log, and archive anyways. doesn't need to be more complex than that

ankitsmt211: and it's doing that exactly, what im trynna say is you're squeezing in the failure part. Im trying to clean the chain cleaner, like i said it's already doing too much.

ankitsmt211: The functional implementation is local to method, so you dont have to jump around. It has a name as well, which abstracts away the details when reading pipeline. So something bad happened, oki it's calling a function `handleFailure` which consumes a `throwable` so maybe ill look at later but ik it handles failures. 

I don't wanna do the if check inline because, it just adds more brackets then you're keeping track of what's what inside chain. Ill do the changes but i still feel the same way.

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +126,33 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+                if (member.isSuccess()) {
+                    return sendEmbedWithMention.apply(member);
+                }
+
+                logger.debug("Unable to mention user", member.getFailure());
+                return sendEmbedWithoutMention.apply(member);
+            })
+            .mapToResult()
+            .flatMap(sentEmbed -> {
+                if (sentEmbed.getFailure() instanceof ErrorResponseException) {
```
surajkumar, line 149: Do we want to only do this on `ErrorResponseException`? What about any other type of error, they should be logged too.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +126,33 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
-            .flatMap(any -> threadChannel.getManager().setArchived(true))
-            .queue(any -> {
-            }, handleFailure);
+            .mapToResult()
+            .flatMap(member -> {
+                if (member.isSuccess()) {
+                    return sendEmbedWithMention.apply(member);
+                }
+
+                logger.debug("Unable to mention user", member.getFailure());
```
surajkumar, line 144: As discussed on the server already, we should add more context to this. I would consider this a `warning` but if this happens all the time, can we change this bit of code so it's clear what's going on e.g. adding a check so see if the user is in server, if they are, error/warn log this message. Otherwise, info log that they have left.

surajkumar: Also, constants should be uppercase. As should the `Logger`, so make it `LOGGER` while you're here please. It's not ugly to make it LOGGER.

surajkumar (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: I suppose we can log it, goal was to handle errors inside JDA anything over network or external is logged by JDA no(correct me if i'm wrong)?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: since we already handled this with `sendEmbedWithoutMention`, it just had to be there could be in debug or info as long as it's taken care of.

surajkumar: The problem is that you're needing to explain this. At the minimum add documentation to anybody else without context will know what's going on. But I'd much rather prefer the code to be rewritten in a way that can express it. You can choose the solution here either way.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
surajkumar: Maybe but if we're being given a `getFailure()` we should use it.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: I have made some changes in pipeline to display intent more clearly, lmk what you think.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -130,24 +130,26 @@ private void handleArchiveFlow(ThreadChannel threadChannel, MessageEmbed embed)
...
+        Supplier<RestAction<Message>> sendEmbedWithoutMention =
+                () -> threadChannel.sendMessageEmbeds(embed);
 
         threadChannel.getGuild()
             .retrieveMemberById(threadChannel.getOwnerIdLong())
             .mapToResult()
-            .flatMap(member -> {
-                if (member.isSuccess()) {
-                    return sendEmbedWithMention.apply(member);
+            .flatMap(foundMember -> {
```
surajkumar, line 139: Nice good readabilty change

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
Taz03: this is valid but everywhere else in the code, it's `logger`, so for consistency, keep it `logger`. And if you want to  change it,  change it everywhere else as well in a separate pr

ankitsmt211: i was wondering why i used `logger` twice in both PRs, prolly followed from rest of code. Can we start from here? It prolly should be done anyway.

surajkumar: If it's lowercase logger everywhere then yeah it should be changed. Surprised the linter doesn't flag it. But changing it everywhere is out of the scope of this PR. 

We have 2 solutions:

1. Create a PR to address this 
2. Update whenever we touch a file

I'd go for the second option, if we see such cases, then they can be fixed when next touched. 

We can raise a request for somebody to do it in a big PR though for those who want to contribute but don't understand enough. 

In any case, it's good how it is now in this PR with the uppcase.

Taz03: > In any case, it's good how it is now in this PR with the uppcase.

its not, no one has agreed to the change + it'll only make the code more inconsistent

ankitsmt211: changed back to `logger`, new PR can be raised for refactoring later.

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -122,18 +126,35 @@ private static boolean shouldBeArchived(MessageChannel channel, Instant archiveA
...
+
+                return sendEmbedWithoutMention.get();
+            })
+            .mapToResult()
+            .flatMap(sentEmbed -> {
+                if (sentEmbed.isFailure()) {
+                    logger.warn(
+                            "Unknown error occurred during help thread auto archive routine, archiving thread",
+                            sentEmbed.getFailure());
+                }
```
Taz03, line 155: well, in what case can embed sending fail? now that u handle user not fetchable separately. Ig we can safely remove this part

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
ankitsmt211: it's supposed to be catch it all, just to be at least aware if something fails on sending message that can be handled in future. Are you sure?

Taz03: well, we never check if a message sending failed so... remove it

ankitsmt211: fair

Taz03 (approved):

marko-radosavljevic (approved):
Thanks. :heart:

PR #989, opened 2023-12-17
collect meta data from threads

### LINKED ISSUE #660: Collecting analytics for #active_questions
Recently, a question was asked by a member of the server, along the lines of "How long does it take for a question to get answered?". This information is unknown. 

We could gather analytics based on the data in #active_questions. 

**Data such as**:

- Time from when a question is opened to the time a question is closed.
- The it takes for somebody to respond to a question
- The number of questions asked per day
- The number of questions asked by a specific discord
- The category questions are asked under

.. and any more that you can think of ..

Then we can then calculate averages based on this information to answer questions like "on average, how long does it take for a question to be answered". 

The information could be displayed in a specific analytics channel and for the purpose of marketing, a website (or as a dynamic image to throw onto advert boards, should you wish). 

To get the initial dataset, a quick scrub should take no time at all. 
Going forward, it might require a lot more effort to keep the data up-to-date after the initial scrub. 

Opinions, please.

ankitsmt211:
resolves #660 

gonna close this, there's a better way to do this instead of wriitng seperate listener should be done in already existing one.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/MetadataMiner.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/MetadataMiner.java
@@ -1,0 +1,123 @@
+package org.togetherjava.tjbot.features.help;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
+import net.dv8tion.jda.api.events.channel.ChannelCreateEvent;
+import net.dv8tion.jda.api.events.channel.update.ChannelUpdateArchivedEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.features.EventReceiver;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/MetadataMiner.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private boolean wasThreadAlreadyHandled(long threadChannelId) {
+        // NOTE Discord/JDA fires this event twice per thread (bug?), we work around by remembering
+        // the threads we already handled
+        Instant now = Instant.now();
+        // NOTE It is necessary to do the "check if exists, otherwise insert" atomic
+        Instant createdAt = threadIdToCreatedAtCache.get(threadChannelId, any -> now);
+        return createdAt != now;
+    }
```
ankitsmt211, line 112: copied method for checking already handled thread, should i instead just increase visibility of one inside `HelpThreadCreatedListener`?

application/src/main/java/org/togetherjava/tjbot/features/help/MetadataMiner.java
```diff
@@ -0,0 +1,123 @@
...
+
+        logger.warn("ticket tags applied: "+tag);
+
+            database.write(context -> context.newRecord(HELP_THREADS)
+                .setChannelId(channelId)
+                .setCreatedAt(createdAt)
+                .setAuthorId(authorId)
+                .setTicketStatus(TicketStatus.ACTIVE.val)
+                .setTag(tag)
+                .insert());
```
ankitsmt211, line 90: im struggling with this bit, somehow it's doing writes twice. Error pops up with integrity violation for duplicate `threadId`.
I can't find what's going on though, thoughts?

PR #1041, opened 2024-03-02
Fix: Pre-commit hook includes untracked files in commits

### LINKED ISSUE #1039: Pre-commit hook includes untracked files in commits
**Describe the bug**
From the looks of it, whenever one attempts to commit any changes while having untracked files in their working tree, the pre-commit hook also includes these files in the commit. This is somewhat severe because I accidentally shared some untracked configuration files that included a Discord bot token for my testing bot, which was thankfully revoked by Discord itself as soon as I pushed.

**Expected behavior**
The pre-commit hooks should not include untracked files.

**To Reproduce**
1. Make a new file in the working directory that is not tracked by `.gitignore`.
2. Ensure that the file is untracked by using `git restore --staged <file_name>`.
3. Change a different file's contents, such as a Java source code file that is tracked.
4. Add the Java source code file to the index and make a commit from it.
5. Assuming the pre-commit hook was executed with your commit, you should now see the untracked file included in the commit.

**Screenshots**
image
image
image

**Additional context**
macOS version: Sonoma 14.3.1 (23D60)
git version: 2.39.3 (Apple Git-145)

christolis:
Fixes #1039.

With this pull request, untracked files do not get included in commits handled by the pre-commit hook.

**Screenshots**
image
image
image

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -44,8 +44,8 @@ if [ "$spotlessExitCode" -ne 0 ]; then
   exit "$spotlessExitCode"
 fi
 
-# Spotless possibly found changes, apply them
-git add -A
+# Spotless possibly found changes, apply them, excluding untracked files
+git add -u
 
 # Restore back the unstaged changes
 pop_stash


Taz03 (approved):

ankitsmt211 (approved):

Zabuzard:
I dont quite get the point of this. Untracked files should NOT be automatically included in commits. Otherwise you can not keep files untracked, which can be on purpose (I do this quite often).
@marko-radosavljevic for review please

marko-radosavljevic (dismissed):
Blocked until I review and test it, per Daniel's request.

christolis:
For reference the `-A` option, found in the original code, works like this ([source](https://git-scm.com/docs/git-add#Documentation/git-add.txt--A), emphasis mine):

> Update the index not only where the working tree has a file matching <pathspec> but also where the index already has an entry. This adds, modifies, and removes index entries to match the working tree.
> 
> If no <pathspec> is given when -A option is used, **all files in the entire working tree are updated** (old versions of Git used to limit the update to the current directory and its subdirectories).

The `-u` option in the PR works like this ([source](https://git-scm.com/docs/git-add#Documentation/git-add.txt--u), emphasis mine):

> Update the index just where it already has an entry matching <pathspec>. This removes as well as modifies index entries to match the working tree, **but adds no new files**.
> 
> If no <pathspec> is given when -u option is used, all **tracked** files in the entire working tree are updated (old versions of Git used to limit the update to the current directory and its subdirectories).

tj-wazei (approved):
I have tested and reproduced the issue. It is indeed a bug.  

In my test, the following output was provided when running `git status`:
```
On branch bug-test
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   application/src/main/java/org/togetherjava/tjbot/Application.java

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        application/src/main/java/org/togetherjava/tjbot/BugTest.java
```
After running `git commit -m "test"`

This is the result of `git status`:
```
On branch bug-test
nothing to commit, working tree clean
```

The output from the pre-commit hook shows: 

```
****Done pre-commit hooks****
[bug-test 96a2cb1] test
 2 files changed, 6 insertions(+)
 create mode 100644 application/src/main/java/org/togetherjava/tjbot/BugTest.java
```

Clearly, we can see it's commiting untracked files.

PR #1042, opened 2024-03-03
Feature: Cake Day

### LINKED ISSUE #1035: cake day for TJ server
 if you're a reddit user, cake day shouldn't be a new concept to you. Basically when a user registered on the website, that day is considered cake day for that user. Each year user gets a custom icon attached to their username on their cake day. This is what it looks like 
image
It's just anniversary since you joined in a nutshell.

based on that idea, im proposing too add same feature on TJ. Once a year a custom role is given to user which will give them an icon for 24 hours to celebrate their cake day(if you will).

**Additional context**

Roles will not have any special priviledges, only that custom icon with maybe saying "cake day" when any other member clicks on that icon. If user has any other roles that has icon, it should override with cake_day role for 24 hrs maybe?

christolis:
*I know that this is not a certain feature to be added, but I still wanted to experiment and see what I could come up with :)*

# Description
This pull request introduces the addition of the cake day feature to the server, allowing members to commemorate their annual membership anniversary with a unique "Cake day" role, with which they can celebrate within the community.

This feature uses the database and makes a `cake_days` table in order to store each member's joined date, as well the guild in which they joined (in case the bot ever gets to handle multiple guilds). The date stored in the database is split into two columns: the month and the day in one (`joined_month_day`), and the year in an other (`joined_year`). The reason the join date is split into these two columns is so that an index can be added for the `joined_month_day` column in order to make reading from the database faster. Therefore, it will be easier for the bot to find all cake days by the month and day, a search that would occur daily by this feature's routine.

If the bot finds that the `cake_days` table has no records, then it attempts to populate the table from all members of all guilds in a separate thread with batched insert queries in order to optimize the table population task.

# Configuration changes
|---	|---	|---	|---	|
| `cakeDayConfig.rolePattern` 	| A pattern of the cake day<br>role to give to members. 	| String 	| `"Cake Day"` 	|

# Diagram
image

# TODO
- [x] Write JavaDocs for all new code.
- [x] Implement cake-day role re-assignment daily. First it should revoke the role from all members, then based on the day's month and day, it should find all matching cake days and for the matches where the year difference is bigger than `0`, assign them the role and whatever cosmetic stuff necessary.
- [x] Talk with the team about whether changing the member cache policy as well as the chunking filter to all is a valid approach for fetching all members of the server *(is it worth storing all member references in memory just for one bulk database insert that will not even happen more than once?)*

Closes #1035.


application/src/main/java/org/togetherjava/tjbot/config/Config.java
```diff
@@ -123,6 +125,7 @@ private Config(@JsonProperty(value = "token", required = true) String token,
         this.helperPruneConfig = Objects.requireNonNull(helperPruneConfig);
         this.featureBlacklistConfig = Objects.requireNonNull(featureBlacklistConfig);
         this.selectRolesChannelPattern = Objects.requireNonNull(selectRolesChannelPattern);
+        this.cakeDayConfig = cakeDayConfig;
```
SquidXTV, line 128: Missing `Objects.requireNonNull`

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,311 @@
...
+     *
+     * @param guild the guild from which to remove the role from members
+     * @param role the role to be removed from the members
+     * @param members the list of members from which the role will be removed
+     */
+    private void removeRoleFromMembers(Guild guild, Role role, List<Member> members) {
+        members.forEach(member -> {
+            UserSnowflake snowflake = UserSnowflake.fromId(member.getIdLong());
+            guild.removeRoleFromMember(snowflake, role).complete();
+        });
```
SquidXTV, line 162: Can't you chain these RestActions?

application/config.json.template
```diff
@@ -110,5 +110,8 @@
         "special": [
         ]
     },
-    "selectRolesChannelPattern": "select-your-roles"
+    "selectRolesChannelPattern": "select-your-roles",
+    "cakeDayConfig": {
+        "rolePattern": "cakeDayRolePattern"
```
SquidXTV, line 115: give this a proper default value imo

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/config/Config.java
christolis: Added it! ✅

application/config.json.template
christolis: Done! 👍

application/src/main/java/org/togetherjava/tjbot/config/Config.java
SquidXTV: > Added it! ✅

Instead of always writing "Added"/"Done" or smth just simply press the `Resolve conversation` button <3

christolis: > > Added it! ✅
> 
> Instead of always writing "Added"/"Done" or smth just simply press the `Resolve conversation` button <3

Okay! 🚀

Jokes aside, I always let the OP resolve the conversations so that they don't unresolve it later to add more stuff

SquidXTV: > Jokes aside, I always let the OP resolve the conversations so that they don't unresolve it later to add more stuff. I don't know if that's the right way to do it though.

I prefer if you just do it

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: `RestActions` are now handled better! :)

application/src/main/java/org/togetherjava/tjbot/config/CakeDayConfig.java
```diff
@@ -0,0 +1,10 @@
+package org.togetherjava.tjbot.config;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Configuration record for the Cake Day feature.
+ */
+public record CakeDayConfig(
+        @JsonProperty(value = "rolePattern", required = true) String rolePattern) {
+}
```
SquidXTV, line 10: Add `Objects.requireNonNull` checks in constructor

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+     * @param guild the guild for which to reassign the cake day role
+     */
+    protected void reassignCakeDayRole(Guild guild) {
+        Role cakeDayRole = getCakeDayRole(guild).orElse(null);
+
+        if (cakeDayRole == null) {
+            return;
+        }
+
+        refreshMembersCakeDayRoles(cakeDayRole, guild);
```
SquidXTV, line 84: why not using proper `Optional` usage here?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * Refreshes the Cake Day roles for members in the specified guild.
+     *
+     * @param cakeDayRole the Cake Day role to refresh
+     * @param guild the guild in which to refresh Cake Day roles
+     */
```
SquidXTV, line 92: Are these docs needed for a private method?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+            removeRoleFromMembers(guild, cakeDayRole, members);
+            addTodayMembersCakeDayRole(guild);
+        });
+    }
+
+    /**
+     * Asynchronously adds the specified cake day role to guild members who are celebrating their
+     * cake day today.
+     * <p>
+     * The cake day role is added to members who have been in the guild for at least one year.
```
SquidXTV, line 104: Shouldnt it add the role if the current day is the anniversary day?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+     *
+     * @param snowflake the snowflake ID of the user to whom the cake day role will be added
+     * @param guild the guild in which the cake day role will be added to the user
+     */
+    protected void addCakeDayRole(UserSnowflake snowflake, Guild guild) {
+        Role cakeDayRole = getCakeDayRole(guild).orElse(null);
+
+        if (cakeDayRole == null) {
+            return;
+        }
```
SquidXTV, line 131: Missing proper `Optional` usage

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+     * @param member the {@link Member} to whom the cake day role will be added
+     */
+    protected void addCakeDayRole(Member member) {
+        Guild guild = member.getGuild();
+        UserSnowflake snowflake = UserSnowflake.fromId(member.getId());
+        Role cakeDayRole = getCakeDayRole(guild).orElse(null);
+
+        if (cakeDayRole == null) {
+            return;
+        }
```
SquidXTV, line 148: Missing proper `Optional` usage

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+        guild.addRoleToMember(snowflake, cakeDayRole).complete();
+    }
+
+    /**
+     * Removes a specified role from a list of members in a {@link Guild}.
+     *
+     * @param guild the {@link Guild} from which to remove the role from members
+     * @param role the {@link Role} to be removed from the members
+     * @param members the {@link List} of members from which the {@link Role} will be removed
+     */
```
SquidXTV, line 159: These docs for private method needed?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,318 @@
...
+            .and(CAKE_DAYS.GUILD_ID.eq(guildId))
+            .execute());
+    }
+
+    /**
+     * Retrieves the cake day {@link Role} from the specified guild.
+     *
+     * @param guild the {@link Guild} from which to retrieve the cake day role
+     * @return an {@link Optional} containing the cake day role if found, otherwise empty
+     */
```
SquidXTV, line 232: These docs for private method needed?

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: They could be removed if you insist! Although I don't see how having docs for it would affect things.

SquidXTV: it doesnt really, the thing is in this case the method name is enough documentation

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: It is now used properly! ✅

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: It is now used properly! ✅

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: It is now used properly!

application/src/main/java/org/togetherjava/tjbot/config/CakeDayConfig.java
christolis: That is now done! 🙌

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Solved on `#tjbot-discussion`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Resolved on `#tjbot-discussion`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Resolved on `#tjbot-discussion`.

SquidXTV (dismissed):

ankitsmt211:
It would be nice if one of the @Together-Java/moderators can review some of the critical parts of this feature when you can.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Handles the event of a message being received in a guild.
+     * <p>
+     * It caches the user's cake day and inserts the member's cake day into the database if not
+     * already present.
+     *
+     * @param event the {@link MessageReceivedEvent} representing the message received
+     */
+    @Override
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
```
tj-wazei, line 41:
```java
LocalDate currentDate = LocalDate.now();
                List<Member> anniversaries = guild
                        .loadMembers()
                        .get()
                        .stream()
                        .filter(member -> member.getTimeJoined().getDayOfMonth() == currentDate.getDayOfMonth() && member.getTimeJoined().getMonth() == currentDate.getMonth())
                        .toList();
```

Why can't we run this as a routine regularly?

SquidXTV: Wouldn't this include "0-day" anniversaries?

tj-wazei: You can update the filter to include `member.getTimeJoined().getYear() != currentDate.getYear()`

christolis: > ```java
> LocalDate currentDate = LocalDate.now();
> List<Member> anniversaries = guild
>   .loadMembers()
>   .get()
>   .stream()
>   .filter(member -> member.getTimeJoined().getDayOfMonth() == currentDate.getDayOfMonth() && member.getTimeJoined().getMonth() == currentDate.getMonth())
>   .toList();
> ```
> 
> Why can't we run this as a routine regularly?

Wouldn't this be expensive to run once every day? It has to go through all members every time and there's been some hurdles with Discord's rate limiter, forcing us to use a database for this.

tj-wazei: That snippet will be expensive but we also know that JDA caches on startup. So use the cached members when doing the loop.

SquidXTV: This resolved?

application/src/main/java/org/togetherjava/tjbot/config/CakeDayConfig.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Configuration record for the Cake Day feature.
+ */
+public record CakeDayConfig(
+        @JsonProperty(value = "rolePattern", required = true) String rolePattern) {
+
+    /**
+     * Configuration constructor for the Cake Day feature.
+     */
+    public CakeDayConfig {
```
SquidXTV, line 18: This comment really needed? It would maybe make sense if it links to the corresponding classes. Though even that should be done as some `@see` in the records comment instead.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+    private final Database database;
+
+    /**
+     * Constructs a {@link CakeDayService} with the given configuration and database.
+     *
+     * @param config the configuration for cake day management
+     * @param database the database for storing cake day information
+     */
+    public CakeDayService(Config config, Database database) {
+        this.config = config.getCakeDayConfig();
```
SquidXTV: I really dont like the naming here. Both CakeDayConfig and Config variables are called `config`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     * @param guild the guild for which to reassign the cake day role
+     */
+    protected void reassignCakeDayRole(Guild guild) {
+        Optional<Role> cakeDayRole = getCakeDayRole(guild);
+
+        if (cakeDayRole.isEmpty()) {
+            return;
+        }
+
+        refreshMembersCakeDayRoles(cakeDayRole.get(), guild);
```
SquidXTV: could use `Optional#ifPresent` here

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+            Member member = guild.getMemberById(cakeDayRecord.getUserId());
+
+            if (member == null) {
+                return;
+            }
+
+            boolean isAnniversaryDay = hasMemberCakeDayToday(member);
+            int yearsSinceJoin = OffsetDateTime.now().getYear() - cakeDayRecord.getJoinedYear();
+
+            if (yearsSinceJoin > 0 && isAnniversaryDay) {
```
SquidXTV: Hm so this is confusing me a bit. `hasMemberCakeDayToday` checks if today is the cake day using the `Member member` **which includes** checking if the anniversary year is ` > 0`. This looks like checking the same thing twice (`yearsSinceJoin` calculation uses the db record though). Also is the member query needed for this check? Doesn't the database also store the join date (make sure it is up2date though)? You would only need the member instance fo adding the role later on, so you could move that `getMemberById` later in the chain and thus making it more efficient.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+    protected void addCakeDayRole(Member member) {
+        Guild guild = member.getGuild();
+        UserSnowflake snowflake = UserSnowflake.fromId(member.getId());
+        Optional<Role> cakeDayRole = getCakeDayRole(guild);
+
+        if (cakeDayRole.isEmpty()) {
+            return;
+        }
+
+        guild.addRoleToMember(snowflake, cakeDayRole.get()).complete();
```
SquidXTV: A `member` **is** a `UserSnowflake`, so it makes no sense that you get the snowflake from the member.

SquidXTV: Does the `RestAction` from `addRoleToMember` have to be blocking with `complete`? Otherwise use `queue` etc instead.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     * @param member the {@link Member} to whom the cake day role will be added
+     */
+    protected void addCakeDayRole(Member member) {
+        Guild guild = member.getGuild();
+        UserSnowflake snowflake = UserSnowflake.fromId(member.getId());
+        Optional<Role> cakeDayRole = getCakeDayRole(guild);
+
+        if (cakeDayRole.isEmpty()) {
+            return;
+        }
```
SquidXTV, line 126: If you look up the method chain then you do this: `refreshMembersCakeDayRoles` -> `addTodayMembersCakeDayRole` -> `addCakeDayRole`. So why do you query the `cakeDayRole` here again instead of passing the instance of `Role` from `refreshMembersCakeDayRoles`?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+    private void removeRoleFromMembers(Guild guild, Role role, List<Member> members) {
+        List<RestAction<Result<Void>>> chain = members.stream()
+            .map(member -> guild.removeRoleFromMember(member, role).mapToResult())
+            .toList();
+
+        if (chain.isEmpty()) {
+            return;
+        }
+
+        RestAction.allOf(chain).queue();
```
SquidXTV: Probably good if this got logging if it fails for some member. Otherwise we might miss if remove role failed for a member and then that member walking around with the role for a year lol

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     *
+     * @param member the member whose cake day information is to be inserted
+     * @param guildId the ID of the guild to which the member belongs
+     * @return an Optional containing the query to insert cake day information if the member has a
+     *         join time; empty Optional otherwise
+     */
+    private Optional<Query> createMemberCakeDayQuery(Member member, long guildId) {
+        if (!member.hasTimeJoined()) {
+            return Optional.empty();
+        }
```
SquidXTV: The `Member#hasTimeJoined` mentions that you can load the join time using `member = guild.retrieveMemberById(member.getId())...`

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     * @param member the member whose cake day is to be inserted into the database
+     * @param guildId the ID of the guild to which the member belongs
+     */
+    protected void insertMemberCakeDayToDatabase(Member member, long guildId) {
+        Query insertQuery = createMemberCakeDayQuery(member, guildId).orElse(null);
+
+        if (insertQuery == null) {
+            logger.warn("Tried to add member {} to database but found no time joined",
+                    member.getId());
+        }
```
SquidXTV: This warning message using internal details of `createMemberCakeDayQuery` because it knows that the optional is empty **only** if there was no accurate time joined. I would question if this is then the correct place to have this warning. You can just put the warning into the `if` in the `createMemberCakeDayQuery`. In that cake you can even get rid of the `orElse(null)` and use `.ifPresent` for the database write.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+
+        database.write(context -> context.batch(insertQuery).execute());
```
SquidXTV: Not really into JOOQ and our database stuff but why are you using `#batch` here? Is there no `#execute` method or is it different?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * Creates a query to insert a member's cake day information into the database.
+     *
+     * @param member the member whose cake day information is to be inserted
+     * @param guildId the ID of the guild to which the member belongs
+     * @return an Optional containing the query to insert cake day information if the member has a
+     *         join time; empty Optional otherwise
+     */
+    private Optional<Query> createMemberCakeDayQuery(Member member, long guildId) {
```
SquidXTV: This can be `static`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     * Retrieves the cake day {@link Role} from the specified guild.
+     *
+     * @param guild the {@link Guild} from which to retrieve the cake day role
+     * @return an {@link Optional} containing the cake day role if found, otherwise empty
+     */
+    private Optional<Role> getCakeDayRoleFromGuild(Guild guild) {
+        return guild.getRoles()
+            .stream()
+            .filter(role -> cakeDayRolePredicate.test(role.getName()))
+            .findFirst();
```
SquidXTV: Is there no `Guild#getRoleByName` or something?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Removes the cake day information of the specified user from the database and clears the cache
+     * for the guild.
+     *
+     * @param user the {@link User} who left the guild
+     * @param guild the {@link Guild} from which the user left
+     */
+    protected void handleUserLeft(User user, Guild guild) {
```
SquidXTV: This method name doesn't really represent **what** it does but for what its used. Generally this method simply removes the cake day information from the db (and cache).

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    private List<CakeDaysRecord> findCakeDaysTodayFromDatabase(Guild guild) {
+        String todayMonthDay = OffsetDateTime.now().format(MONTH_DAY_FORMATTER);
+
+        return database
+            .read(context -> context.selectFrom(CAKE_DAYS)
+                .where(CAKE_DAYS.JOINED_MONTH_DAY.eq(todayMonthDay))
+                .and(CAKE_DAYS.GUILD_ID.eq(guild.getIdLong()))
+                .fetch())
+            .collect(Collectors.toList());
```
SquidXTV: Doesn't `database.read` already return a list?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,309 @@
...
+     *         {@link Optional} if no record is found
+     */
+    protected Optional<CakeDaysRecord> findUserCakeDayFromDatabase(long userId) {
+        return database
+            .read(context -> context.selectFrom(CAKE_DAYS)
+                .where(CAKE_DAYS.USER_ID.eq(userId))
+                .fetch())
+            .collect(Collectors.toList())
+            .stream()
+            .findFirst();
```
SquidXTV: You sure the `.collect(...).stream().findFirst()` is really needed? Is there no other method to use here?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayRoutine.java
```diff
@@ -0,0 +1,39 @@
...
+public class CakeDayRoutine implements Routine {
+
+    private final CakeDayService cakeDayService;
+
+    /**
+     * Constructs a new {@link CakeDayRoutine} instance.
+     *
+     * @param cakeDayService an instance of the cake day service
+     */
+    public CakeDayRoutine(CakeDayService cakeDayService) {
```
SquidXTV, line 20: This comment seems pretty useless.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
```diff
@@ -0,0 +1,84 @@
...
+public class CakeDayListener extends ListenerAdapter implements EventReceiver {
+
+    private final CakeDayService cakeDayService;
+
+    /**
+     * Constructs a new CakeDayListener with the given {@link CakeDayService}.
+     *
+     * @param cakeDayService the {@link CakeDayService} to be used by this listener
+     */
+    public CakeDayListener(CakeDayService cakeDayService) {
```
SquidXTV, line 23: This comment seems pretty useless.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        cakeDayService.addToCache(author);
+        Optional<CakeDaysRecord> cakeDaysRecord =
+                cakeDayService.findUserCakeDayFromDatabase(authorId);
+        if (cakeDaysRecord.isPresent()) {
+            return;
+        }
+
+        cakeDayService.insertMemberCakeDayToDatabase(member, guildId);
```
SquidXTV, line 63: You could use `Optional#ifPresent` here.

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
Zabuzard: `List#getFirst()`

Zabuzard: if all u need is the first hit (or any hit, bc there is no sorting), best is to modify the query to only return that result on DB level and not everything. im pretty sure jooq has a method for that and i recall us using it somewhere already

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
Zabuzard: id rename the parameter then though. dunno, i think for the pov of the class and where its being used in the methods it makes sense that `config` is the config for this feature. so perhaps the parameter `Config` can then be called `fullConfig` or something instead here.

application/src/main/java/org/togetherjava/tjbot/config/CakeDayConfig.java
Zabuzard: Possible that we require javadoc on this one (public class). But yeah, we can add some details or a `@see`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
ankitsmt211, line 36: we should update our table on user join/leave events instead of onMessage, like i mentioned earlier onMessage only triggers for active users which means after few weeks. It's being triggered for 1000 messages from same users without adding any value(updating our table with joined dates). 

User join event will add new entry to our table, user leaves will remove that entry.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayRoutine.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    @NotNull
+    public Schedule createSchedule() {
+        return new Schedule(ScheduleMode.FIXED_RATE, 0, 1, TimeUnit.DAYS);
+    }
+
+    @Override
+    public void runRoutine(@NotNull JDA jda) {
+        jda.getGuilds().forEach(cakeDayService::reassignCakeDayRole);
```
ankitsmt211, line 32: the routine should start a seperate thread that will collect two different list of users
* users with already cake day role, that needs to be removed on next day
* users that are eligible for cake day role 
it will then perform these updates at a delay of 1-1.5 seconds to be cautious with API rate limits.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
christolis: I don't think we can do it in a single statement. The only alternative to `Optional#ifPresent` for cleaner code is `Optional#ifPresentOrElse`, which will result in having to pass the first argument as null, which is not recommended as both parameters in the aforementioned method have a `@NotNull` annotation.

Let me know if you were thinking of something else other than `Optional#ifPresent`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
christolis: Good point, it does not add any value in this case. It will be removed.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayRoutine.java
christolis: Comment will be removed as it indeed adds no value. Thank you for pointing it out.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: You guys are right. This issue has been addressed. Please let me know if you find the new version to be better.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: You are right, we actually don't need `.collect(Collectors.toList());`.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: You are correct on that. I have changed the method name so that it reflects on what it does.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: There is a `Guild#getRolesByName` that I have decided to use, which allows for us to filter by role name from the API itself instead of doing it ourselves and overfetching as a result. I will use that method instead. Good catch.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Good catch. We should use `execute()` as we are only dealing with one insertion.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Has been done.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: I decided to use that instead.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: You are right. This has been addressed. Thank you for clarifying that.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: This has been fixed, thank you for letting me know.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: This has been fixed. Thank you for addressing.

christolis: Good catch! You are right, it has been updated.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: You are right, I removed that extra redundant check.

application/src/main/java/org/togetherjava/tjbot/config/CakeDayConfig.java
christolis: I added a `@see` for more clarity.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayListener.java
suryatejess: I would like to work on this

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayRoutine.java
suryatejess: I would like to work on this.

suryatejess: By 2 threads, do you mean 2 different functions or literal java threads?

christolis: > By 2 threads, do you mean 2 different functions or literal java threads?

I think he means literally a new Java thread.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * Checks if the provided {@link Member} has their "cake day" today.
+     * 
+     * @param member the {@link Member} whose cake day is to be checked
+     * @return true if the member has their cake day today; otherwise, false
+     */
+    protected boolean hasMemberCakeDayToday(Member member) {
+        OffsetDateTime now = OffsetDateTime.now();
+        OffsetDateTime joinMonthDate = member.getTimeJoined();
```
tj-wazei, line 270: We store the join date in the database, there's no need to re-query this information again. Primary reasons is because members joining/leaving will desync against the database.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,285 @@
...
+     * The Discord retrieval is performed asynchronously; on success the database write is executed.
+     * If the Discord retrieval fails (e.g., member not found, missing permissions, network issues),
+     * the insert is skipped and a warning is logged.
+     *
+     * @param member the member whose cake day information should be inserted
+     * @param guildId the id of the guild associated with the cake day record
+     */
+    protected void insertMemberCakeDayToDatabase(Member member, long guildId) {
+        member.getGuild().retrieveMemberById(member.getIdLong()).queue(retrievedMember -> {
+            OffsetDateTime timeJoined = retrievedMember.getTimeJoined();
```
tj-wazei, line 176: can we use `Instant` please? `retrievedMember.getTimeJoined().toInstant()`

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,285 @@
...
+            String joinedMonthDay = timeJoined.format(MONTH_DAY_FORMATTER);
+
+            Query query = DSL.insertInto(CAKE_DAYS)
+                .set(CAKE_DAYS.JOINED_MONTH_DAY, joinedMonthDay)
```
tj-wazei, line 180: If we use Instant we can then simplify the date storage as `instant.getEpochSecond()`

tj-wazei (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Quick thought about this: will we be able to easily and cleanly get the specific day, hour and second aside? Asking because at the end, that's the data we are aiming to store.

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
```diff
@@ -0,0 +1,285 @@
...
+import static org.togetherjava.tjbot.db.generated.tables.CakeDays.CAKE_DAYS;
+
+/**
+ * Service for managing the Cake Day feature.
+ */
+public class CakeDayService {
+    private static final Logger logger = LoggerFactory.getLogger(CakeDayService.class);
+    private static final DateTimeFormatter MONTH_DAY_FORMATTER =
+            DateTimeFormatter.ofPattern("MM-dd");
+    private final Set<String> cakeDaysCache = new HashSet<>();
```
tj-wazei, line 34: Let's remove this entirely. We don't need in-memory caching and this could become really big given how good our up-time is. I don't even think this cache is used.

tj-wazei (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
tj-wazei:
```java
LocalDate cakeday = instance.toLocalDate();
LocalDate today = LocalDate.now();

boolean isCakeDay = cakeday.getMonth() == today.getMonth() 
  && cakeday.getDayOfMonth() == today.getDayOfMonth();
```

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
christolis: Perhaps use something like Caffeine with an eviction policy to keep its size manageable. What are your thoughts on that?

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
tj-wazei: Actually, I think we can do it even simpler:

```java
boolean isCakeDay = MonthDay.from(instance.toLocalDate()).equals(MonthDay.from(LocalDate.now()));
```

application/src/main/java/org/togetherjava/tjbot/features/cakeday/CakeDayService.java
tj-wazei: What's the cache for? I don't think we need this at all.

application/src/main/resources/db/V16__Add_Cake_Days.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE cake_days
+(
+    id                  INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    joined_month_day 	TEXT NOT NULL,
+    joined_year         INT NOT NULL,
```
Zabuzard, line 5: please just `TIMESTAMP` (`Instant` in the code then)
```sql
joined_at TIMESTAMPT NOT NULL,
```

application/src/main/resources/db/V16__Add_Cake_Days.sql
```diff
@@ -0,0 +1,10 @@
...
+    guild_id            BIGINT NOT NULL,
+    user_id             BIGINT NOT NULL
+);
+
+CREATE INDEX cake_day_idx ON cake_days(joined_month_day);
```
Zabuzard, line 10: an index isnt needed, the amount of entries in the DB are laughable, remove that

Zabuzard (changes requested):
quick review on the database stuff (as that is the most important to get right)

application/src/main/resources/db/V16__Add_Cake_Days.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE cake_days
+(
+    id                  INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
```
Zabuzard, line 3: this was only added to have a proper `PRIMARY_KEY`? remove it and instead write `PRIMARY KEY (user_id, guild_id)`

See example:
```sql
CREATE TABLE bookmarks
(
    created_at TIMESTAMP NOT NULL,
    note       TEXT,
    delete_at  TIMESTAMP,

    PRIMARY KEY (author_id, channel_id)
)
```

Zabuzard (changes requested):

application/src/main/resources/db/V16__Add_Cake_Days.sql
Zabuzard: also, sqlite doesnt even use indices on such lookups, so it wouldnt help anyways

Issue #935, opened 2023-10-25
Make prompt of chatgpt visible


Taz03:
Right now if we use chatgpt command it only gives us the answer, and we have no way to see what we prompted.

It'd be nice to have the option to see the prompt, and as a public server, everyone should be able to see everyones prompt.

My Implementation idea would be to have a button on each answer which when clicked sends a model with the prompt, and when the model is submitted it works as a normal chatgpt command, to allow quick edits as well.

Zabuzard:
What about posting 2 embeds?
the original question, with the author picture and then the chatgpt answer as embed, also with a author picture.
That way it will look like a discord-native conversation.

Note that the slash command (was) just for internal dev-testing. Hence the "bad" UX.

Taz03:
Embeds have 2 big problems
1. This is how code looks like
![image](https://github.com/Together-Java/TJ-Bot/assets/73871477/3a676b88-0042-4fda-8de6-f266e990dd04)

2. No image and link embeds

Zabuzard:
1. im aware, the benefits outweight the disadvantages though. we will also introduce it for chatgpts auto-answers soon
2. we have `LinkPreview` which automatically adds images and links.

ankitsmt211:
this issue no longer persists after gpt response are done via embed, closing this issue feel free to re open if anyone feels otherwise.

PR #1046, opened 2024-03-06
AI response should be in embed

### LINKED ISSUE #917: ChatGPT Auto-answer should be in an embed
Its necessary for better UX that ChatGPT answers are posted in embeds with a proper author and icon that clearly marks the content as AI answer.

Right now, its too confusing for users since it looks like a regular message.

The disadvantage is that code snippets are typically harder to read in embeds. We will try it out and see how bad it is. Its likely that we can live with it.

Also, when posting embeds, the text limit is boosted from 2k to 4k characters (? check embed max size). So that would reduce the amount of messages that need to be posted, which is good.

### LINKED ISSUE #928: ChatGpt Auto-Answer posts multiple mini messages

ankitsmt211:
* helper method to generate embed
* remove AI parser class that partitioned responses & related tests
* refactor dismiss button on generated responses
* update docs
* resolves #917 and possibly #935 (since now response embed contains title)

* resolves #928 

Here's how new AI generated responses look like: 

gpt-response-embed


SquidXTV (approved):
Looks good to me :D
Btw you want to change the user icon to chatgpt icon?
Are we even allowed to?

ankitsmt211:
>  Btw you want to change the user icon to chatgpt icon? Are we even allowed to?

that would be really nice if we could do that, i can't think of a way to do that. Prolly would need store random asset somewhere and get a url perhaps that we can pass for author.

SquidXTV:
> > Btw you want to change the user icon to chatgpt icon? Are we even allowed to?
> 
> that would be really nice if we could do that, i can't think of a way to do that. Prolly would need store random asset somewhere and get a url perhaps that we can pass for author.

https://en.m.wikipedia.org/wiki/File:ChatGPT_logo.svg
https://chat.openai.com/apple-touch-icon.png
https://chat.openai.com/favicon-32x32.png

ankitsmt211:
I will give these a try, also i linked a new issue #928 since it will be resolved as well so we can touch 3 of the issues with this merge.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -143,27 +143,43 @@ RestAction<Message> constructChatGptAttempt(ThreadChannel threadChannel,
...
+
+        int embedTitleLimit = MessageEmbed.TITLE_MAX_LENGTH;
+        String capitalizedTitle = Character.toUpperCase(title.charAt(0)) + title.substring(1);
+
+        String titleForEmbed = capitalizedTitle.length() > embedTitleLimit
+                ? capitalizedTitle.substring(0, embedTitleLimit)
+                : capitalizedTitle;
+
+        return new EmbedBuilder()
+            .setAuthor(selfUser.getName(), null, selfUser.getEffectiveAvatarUrl())
```
Taz03, line 173: maybe add gpt as author?

Taz03 (changes requested):

Taz03 (approved):

PR #1049, opened 2024-03-08
Feat: Add slash command to generate application form for various community roles

### LINKED ISSUE #1024: Add a form to apply to certain community roles
We want members to know that we appreciate and want their help and contributions for various tasks like organizing events, outreach, community management etc. Additionally, we want to make it easy to apply for these roles.

So a bot message like "We are always looking for community members that want to contribute to our community, and take charge, if you are interested you can apply for various positions here!" and a button. The channel where this is posted would be in the config. Roles you can apply to should also be in a config.

Member clicks a button, selects a role they want to apply to (optionally, a brief description of the role, if the UI permits), and fill a form answering the question "What makes you a valuable addition to this team" (but a more friendly and less corporate version with a happy emoji).

Their application is posted in a channel defined in the config. With information about the applicant, their picture and discord name, similar to modmail/report.

christolis:
Closes #1024.
# Screenshots
ScreenRecording2024-03-11at18 29 35-ezgif com-video-to-gif-converter
image

# Configuration changes
|---	|---	|---	|---	|
| `applicationForm.applicationChannelPattern` 	| Where submitted <br>applications end up. 	| String 	| `"applications-log"` 	|

# TODO
- [x] Add a 5 minute cooldown per member for sending applications to prevent spam (feel free to recommend alternatives).
  - This could be done by having a `HashMap<Member, OffsetDateTime>` where the key is the `Member` who sent an application and the value is the date and time that they sent it. There should be a condition every time a `Member` attempts to send *any* application which utilizes the aforementioned `HashMap` to prevent spam.
- [x] Add JavaDocs for every method where it's necessary.
- [x] ~~Switch to `EntitySelectInteractionEvent` for the dropdown menu.~~ (**EDIT**: Impossible since this would forcibly include all existing roles)
- [x] Move selectable roles from the config to the command parameters in a similar fashion with previous commands.
  - This eliminates emoji/description metadata as well that was previously defined in the configuration file.

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -199,6 +199,10 @@
         "channel": "quotes",
         "reactionEmoji": "⭐"
     },
+    "roleApplicationSystem": {
+        "submissionsChannelPattern": "staff-applications",
+        "defaultQuestion": "What makes you a good addition to the team?"
+    },
     "memberCountCategoryPattern": "Info",
     "topHelpers": {
         "rolePattern": "Top Helper.*",
--- a/application/src/main/java/org/togetherjava/tjbot/config/RoleApplicationSystemConfig.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/RoleApplicationSystemConfig.java
@@ -1,0 +1,38 @@
+package org.togetherjava.tjbot.config;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import net.dv8tion.jda.api.interactions.components.text.TextInput;
+
+import java.util.Objects;
+
+/**
+ * Represents the configuration for an application form, including roles and application channel
+ * pattern.
+ *
+ * @param submissionsChannelPattern the pattern used to identify the submissions channel where
+ *        applications are sent
+ * @param defaultQuestion the default question that will be asked in the role application form
+ */
+public record RoleApplicationSystemConfig(
+        @JsonProperty(value = "submissionsChannelPattern",
+                required = true) String submissionsChannelPattern,
+        @JsonProperty(value = "defaultQuestion", required = true) String defaultQuestion) {
+
+    /**
+     * Constructs an instance of {@link RoleApplicationSystemConfig} with the provided parameters.
+     * <p>
+     * This constructor ensures that {@code submissionsChannelPattern} and {@code defaultQuestion}
+     * are not null and that the length of the {@code defaultQuestion} does not exceed the maximum
+     * allowed length.
+     */
+    public RoleApplicationSystemConfig {
+        Objects.requireNonNull(submissionsChannelPattern);
+        Objects.requireNonNull(defaultQuestion);
+
+        if (defaultQuestion.length() > TextInput.MAX_LABEL_LENGTH) {
+            throw new IllegalArgumentException(
+                    "defaultQuestion length is too long! Cannot be greater than %d"
+                        .formatted(TextInput.MAX_LABEL_LENGTH));
+        }
+    }
+}
--- a/application/src/main/java/org/togetherjava/tjbot/config/Config.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/Config.java
@@ -48,6 +48,7 @@ public final class Config {
     private final RSSFeedsConfig rssFeedsConfig;
     private final String selectRolesChannelPattern;
     private final String memberCountCategoryPattern;
+    private final RoleApplicationSystemConfig roleApplicationSystemConfig;
     private final QuoteBoardConfig quoteBoardConfig;
     private final TopHelpersConfig topHelpers;
     private final DynamicVoiceChatConfig dynamicVoiceChatConfig;
@@ -106,6 +107,8 @@ public final class Config {
                     required = true) String selectRolesChannelPattern,
             @JsonProperty(value = "quoteBoardConfig",
                     required = true) QuoteBoardConfig quoteBoardConfig,
+            @JsonProperty(value = "roleApplicationSystem",
+                    required = true) RoleApplicationSystemConfig roleApplicationSystemConfig,
             @JsonProperty(value = "topHelpers", required = true) TopHelpersConfig topHelpers,
             @JsonProperty(value = "dynamicVoiceChatConfig",
                     required = true) DynamicVoiceChatConfig dynamicVoiceChatConfig) {
@@ -144,6 +147,7 @@ public final class Config {
         this.rssFeedsConfig = Objects.requireNonNull(rssFeedsConfig);
         this.selectRolesChannelPattern = Objects.requireNonNull(selectRolesChannelPattern);
         this.quoteBoardConfig = Objects.requireNonNull(quoteBoardConfig);
+        this.roleApplicationSystemConfig = roleApplicationSystemConfig;
         this.topHelpers = Objects.requireNonNull(topHelpers);
         this.dynamicVoiceChatConfig = Objects.requireNonNull(dynamicVoiceChatConfig);
     }
@@ -458,6 +462,15 @@ public final class Config {
      */
     public String getMemberCountCategoryPattern() {
         return memberCountCategoryPattern;
+    }
+
+    /**
+     * The configuration related to the application form.
+     *
+     * @return the application form config
+     */
+    public RoleApplicationSystemConfig getRoleApplicationSystemConfig() {
+        return roleApplicationSystemConfig;
     }
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/features/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/Features.java
@@ -69,6 +69,7 @@ import org.togetherjava.tjbot.features.moderation.temp.TemporaryModerationRoutin
 import org.togetherjava.tjbot.features.projects.ProjectsThreadCreatedListener;
 import org.togetherjava.tjbot.features.reminder.RemindRoutine;
 import org.togetherjava.tjbot.features.reminder.ReminderCommand;
+import org.togetherjava.tjbot.features.roleapplication.CreateRoleApplicationCommand;
 import org.togetherjava.tjbot.features.rss.RSSHandlerRoutine;
 import org.togetherjava.tjbot.features.system.BotCore;
 import org.togetherjava.tjbot.features.system.LogLevelCommand;
@@ -219,6 +220,7 @@ public class Features {
         features.add(new JShellCommand(jshellEval));
         features.add(new MessageCommand());
         features.add(new RewriteCommand(chatGptService));
+        features.add(new CreateRoleApplicationCommand(config));
 
         FeatureBlacklist<Class<?>> blacklist = blacklistConfig.normal();
         return blacklist.filterStream(features.stream(), Object::getClass).toList();
--- a/application/src/main/java/org/togetherjava/tjbot/features/roleapplication/CreateRoleApplicationCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/roleapplication/CreateRoleApplicationCommand.java
@@ -1,0 +1,286 @@
+package org.togetherjava.tjbot.features.roleapplication;
+
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.Permission;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.entities.emoji.Emoji;
+import net.dv8tion.jda.api.entities.emoji.EmojiUnion;
+import net.dv8tion.jda.api.events.interaction.ModalInteractionEvent;
+import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
+import net.dv8tion.jda.api.events.interaction.component.StringSelectInteractionEvent;
+import net.dv8tion.jda.api.interactions.commands.CommandInteraction;
+import net.dv8tion.jda.api.interactions.commands.OptionMapping;
+import net.dv8tion.jda.api.interactions.commands.OptionType;
+import net.dv8tion.jda.api.interactions.commands.build.SlashCommandData;
+import net.dv8tion.jda.api.interactions.components.ActionRow;
+import net.dv8tion.jda.api.interactions.components.selections.SelectOption;
+import net.dv8tion.jda.api.interactions.components.selections.StringSelectMenu;
+import net.dv8tion.jda.api.interactions.components.text.TextInput;
+import net.dv8tion.jda.api.interactions.components.text.TextInputStyle;
+import net.dv8tion.jda.api.interactions.modals.Modal;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.config.RoleApplicationSystemConfig;
+import org.togetherjava.tjbot.features.CommandVisibility;
+import org.togetherjava.tjbot.features.SlashCommandAdapter;
+import org.togetherjava.tjbot.features.componentids.Lifespan;
+
+import javax.annotation.Nullable;
+
+import java.awt.Color;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.IntStream;
+
+/**
+ * Represents a command to create an application form for members to apply for roles.
+ * <p>
+ * This command is designed to generate an application form for members to apply for roles within a
+ * guild.
+ */
+public class CreateRoleApplicationCommand extends SlashCommandAdapter {
+    protected static final Color AMBIENT_COLOR = new Color(24, 221, 136, 255);
+    private static final int OPTIONAL_ROLES_AMOUNT = 5;
+    private static final String ROLE_COMPONENT_ID_HEADER = "application-create";
+    private static final String OPTION_PARAM_ID_DELIMITER = "_";
+    private static final int OPTIONS_PER_ROLE = 3;
+    private static final int MINIMUM_ANSWER_LENGTH = 50;
+    private static final int MAXIMUM_ANSWER_LENGTH = 500;
+
+    private final RoleApplicationHandler handler;
+    private final RoleApplicationSystemConfig config;
+
+    private static final Logger logger =
+            LoggerFactory.getLogger(CreateRoleApplicationCommand.class);
+
+    /**
+     * Constructs a new {@link CreateRoleApplicationCommand} with the specified configuration.
+     * <p>
+     * This command is designed to generate an application form for members to apply for roles.
+     *
+     * @param config the configuration containing the settings for the application form
+     */
+    public CreateRoleApplicationCommand(Config config) {
+        super("application-form", "Generates an application form for members to apply for roles.",
+                CommandVisibility.GUILD);
+
+        this.config = config.getRoleApplicationSystemConfig();
+
+        generateRoleOptions();
+        handler = new RoleApplicationHandler(this.config);
+    }
+
+    /**
+     * Populates this command's instance {@link SlashCommandData} object with the proper arguments
+     * for this command.
+     */
+    private void generateRoleOptions() {
+        final SlashCommandData data = getData();
+
+        IntStream.range(1, OPTIONAL_ROLES_AMOUNT + 1).forEach(index -> {
+            data.addOption(OptionType.STRING, generateOptionId("title", index),
+                    "The title of the role");
+            data.addOption(OptionType.STRING, generateOptionId("description", index),
+                    "The description of the role");
+            data.addOption(OptionType.STRING, generateOptionId("emoji", index),
+                    "The emoji of the role");
+        });
+    }
+
+    private static String generateOptionId(String name, int id) {
+        return "%s%s%d".formatted(name, OPTION_PARAM_ID_DELIMITER, id);
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        if (!handleHasPermissions(event)) {
+            return;
+        }
+
+        final List<OptionMapping> optionMappings = event.getInteraction().getOptions();
+        if (optionMappings.isEmpty()) {
+            event.reply("You have to select at least one role.").setEphemeral(true).queue();
+            return;
+        }
+
+        long incorrectArgsCount = getIncorrectRoleArgsCount(optionMappings);
+        if (incorrectArgsCount > 0) {
+            event.reply("Missing information for %d roles.".formatted(incorrectArgsCount))
+                .setEphemeral(true)
+                .queue();
+            return;
+        }
+
+        sendMenu(event);
+    }
+
+    @Override
+    public void onStringSelectSelection(StringSelectInteractionEvent event, List<String> args) {
+        SelectOption selectOption = event.getSelectedOptions().getFirst();
+        Member member = event.getMember();
+
+        if (member == null) {
+            logger.error("Member was null during onStringSelectSelection()");
+            return;
+        }
+
+        if (selectOption == null) {
+            logger.error("selectOption was null during onStringSelectSelection()");
+            return;
+        }
+
+        long remainingMinutes = handler.getMemberCooldownMinutes(member);
+        if (remainingMinutes > 0) {
+            String correctMinutesWord = remainingMinutes == 1 ? "minute" : "minutes";
+
+            event
+                .reply("Please wait %d %s before sending a new application form."
+                    .formatted(remainingMinutes, correctMinutesWord))
+                .setEphemeral(true)
+                .queue();
+            return;
+        }
+
+        TextInput body = TextInput
+            .create(generateComponentId(event.getUser().getId()), config.defaultQuestion(),
+                    TextInputStyle.PARAGRAPH)
+            .setRequired(true)
+            .setRequiredRange(MINIMUM_ANSWER_LENGTH, MAXIMUM_ANSWER_LENGTH)
+            .setPlaceholder("Enter your answer here")
+            .build();
+
+        EmojiUnion emoji = selectOption.getEmoji();
+        String roleDisplayName;
+
+        if (emoji == null) {
+            roleDisplayName = selectOption.getLabel();
+        } else {
+            roleDisplayName = "%s %s".formatted(emoji.getFormatted(), selectOption.getLabel());
+        }
+
+        Modal modal = Modal
+            .create(generateComponentId(event.getUser().getId(), roleDisplayName),
+                    String.format("Application form - %s", selectOption.getLabel()))
+            .addActionRow(ActionRow.of(body).getComponents())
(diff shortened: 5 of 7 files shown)


Suleman70:
Can i be assigned to this?

Taz03 (changes requested):
branch update is required, it is using removed APIs

christolis:
Marked PR as draft since there needs to be some rework on certain things which will be specified in the TODO.

ankitsmt211:
@christolis what happens to the generated form after x user applies through form? Does the form need to be re-created? Or is it only unavailable for x user to a certain time period? 

Also how do we update the form if need be? 
Delete message and recreate?

christolis:
> @christolis what happens to the generated form after x user applies through form? Does the form need to be re-created? Or is it only unavailable for x user to a certain time period?
> 
> Also how do we update the form if need be? Delete message and recreate?

The generated form is persistent and used by everybody. It does not need to be recreated by whoever has the `MANAGE_ROLES` permission. As for updating it, that's right. Currently, the way it is developed, the message has to be deleted and get the command executed again, which is not that big of a problem considering the rarity of updating such form.

application/src/main/java/org/togetherjava/tjbot/features/basic/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,411 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.IntStream;
+
+/**
+ * Represents a command to create an application form for members to apply for roles.
+ * <p>
+ * This command is designed to generate an application form for members to apply for roles within a
+ * guild.
+ */
+public class ApplicationCreateCommand extends SlashCommandAdapter {
```
surajkumar, line 56: This is a pretty big class file. Can we seperate it's concerns so it's easier to follow? Primarily, keeping the slash command distinct to the modal interaction.

application/src/main/java/org/togetherjava/tjbot/features/basic/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    /**
+     * Wrapper class which represents a menu role for the application create command.
+     * <p>
+     * The reason this exists is due to the fact that {@link StringSelectMenu.Builder} does not have
+     * a method which takes emojis as input as of writing this, so we have to elegantly pass in
+     * custom data from this POJO.
+     */
+    private static class MenuRole {
```
surajkumar, line 373: This can be a `record`

SquidXTV (changes requested):

application/config.json.template
```diff
@@ -115,5 +115,8 @@
         "fallbackChannelPattern": "java-news-and-changes",
         "pollIntervalInMinutes": 10
     },
+    "applicationForm": {
+        "applicationChannelPattern": "applications-log"
```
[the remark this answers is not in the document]
christolis: What would be a recommended default application channel?

surajkumar (changes requested):
Great PR and the thought put into the development of the feature is evident. 

Please can you look into the comments, additionally, there are some duplicate checks happening e.g. multiple `guild == null` checks. This can be done once much earlier and then you don't have to worry about it.

application/config.json.template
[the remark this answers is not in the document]
SquidXTV: Well maybe just `applications`? But thats maybe not concrete enough. Just the `-log` being a bit weird. With that I need to think about the error-log.. etc channels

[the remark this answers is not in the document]
surajkumar: `staff-applications`

application/src/main/java/org/togetherjava/tjbot/features/basic/ApplicationCreateCommand.java
christolis: `MenuRole` can not be a record the way it is currently used is due to the fact that its fields are set based on a few conditions located in `addRolesToMenu()` that's found in the same file. Moreover, there wouldn't be a lot to gain by converting it into a record, unless there's something I have overlooked.

SquidXTV: You could make it a `record` and put the conditions into a custom constructor

tj-wazei: Worked with Chris, came to a good solution.

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        if (!member.hasPermission(Permission.MANAGE_ROLES)) {
+            event.reply("You do not have the required manage role permission to use this command")
+                .setEphemeral(true)
+                .queue();
+            return false;
+        }
+
+        Member selfMember = guild.getSelfMember();
```
tj-wazei, line 237: Can remove this entire block as the feature is not creating/managing roles

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,297 @@
...
+
+    @Override
+    public void onStringSelectSelection(StringSelectInteractionEvent event, List<String> args) {
+        SelectOption selectOption = event.getSelectedOptions().getFirst();
+
+        if (selectOption == null) {
+            return;
+        }
+
+        OffsetDateTime timeSentCache = applicationApplyHandler.getApplicationSubmitCooldown()
```
tj-wazei, line 128: Let's extract this out into it's own method for checking if the user is able to invoke the command. Keep's this method a bit more tidy and the code more readable.

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,297 @@
...
+                    .reply("Please wait %d minutes before sending a new application form."
+                        .formatted(remainingMinutes))
+                    .setEphemeral(true)
+                    .queue();
+                return;
+            }
+        }
+
+        String questionLabel = roleApplicationSystemConfig.defaultQuestion();
+        if (questionLabel.length() > TextInput.MAX_LABEL_LENGTH) {
```
tj-wazei, line 146: We can move this entire check into the record. Up to you, not sure if we'd want an exception on start or a concatenated message.

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
tj-wazei, line 237: I don't think selfMember is correct, it should be event.member() or something. Self member is the bot itself. Double check.

tj-wazei (changes requested):

application/src/main/java/org/togetherjava/tjbot/config/RoleApplicationSystemConfig.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * Represents the configuration for an application form, including roles and application channel
+ * pattern.
+ */
+public record RoleApplicationSystemConfig(
```
tj-wazei, line 11: Please could you document these fields. There's no mention anywhere else unfortunantly on what these do.

tj-wazei (dismissed):
LGTM!

tj-wazei (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,271 @@
...
+        Member member = event.getMember();
+
+        if (selectOption == null || member == null) {
+            return;
+        }
+
+        long remainingMinutes = applicationApplyHandler.getMemberCooldownMinutes(member);
+        if (remainingMinutes > 0) {
+            event
+                .reply("Please wait %d minutes before sending a new application form."
```
SquidXTV, line 126: This says 
> "Please wait 1 minutes before sending a new application form" 

in the case of `remainingMinutes` being 1.

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
Taz03:
```suggestion
                .reply("Please wait %d minute(s) before sending a new application form."
```

christolis: Thank you, now fixed! 🚀

SquidXTV (dismissed):
Thanks for your effort <3

application/config.json.template
```diff
@@ -115,5 +115,12 @@
         "fallbackChannelPattern": "java-news-and-changes",
         "pollIntervalInMinutes": 10
     },
+    "roleApplicationSystem": {
+        "submissionsChannelPattern": "staff-applications",
+        "defaultQuestion": "What makes you a good addition to the team?",
+        "minimumAnswerLength": 50,
+        "maximumAnswerLength": 500,
+        "applicationSubmitCooldownMinutes": 5
```
Zabuzard: these 3 can be hardcoded. not needed in the config

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.Map;
+import java.util.stream.IntStream;
+
+/**
+ * Represents a command to create an application form for members to apply for roles.
+ * <p>
+ * This command is designed to generate an application form for members to apply for roles within a
+ * guild.
+ */
+public class ApplicationCreateCommand extends SlashCommandAdapter {
```
Zabuzard: the feature should be renamed to match the package name. `CreateRoleApplicationCommand`

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Handles the actual process of submitting role applications.
+ * <p>
+ * This class is responsible for managing application submissions via modal interactions, ensuring
+ * that submissions are sent to the appropriate application channel, and enforcing cooldowns for
+ * users to prevent spamming.
+ */
+public class ApplicationApplyHandler {
```
Zabuzard: the feature should be renamed to match the package name. `RoleApplicationHandler`

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+    protected static final Color AMBIENT_COLOR = new Color(24, 221, 136, 255);
+    private static final int OPTIONAL_ROLES_AMOUNT = 5;
+    private static final String ROLE_COMPONENT_ID_HEADER = "application-create";
+    private static final String VALUE_DELIMITER = "_";
+    private static final int ARG_COUNT = 3;
+
+    private final ApplicationApplyHandler applicationApplyHandler;
+    private final RoleApplicationSystemConfig roleApplicationSystemConfig;
```
Zabuzard: u can just call it `config` for simplicity

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+ * guild.
+ */
+public class ApplicationCreateCommand extends SlashCommandAdapter {
+    protected static final Color AMBIENT_COLOR = new Color(24, 221, 136, 255);
+    private static final int OPTIONAL_ROLES_AMOUNT = 5;
+    private static final String ROLE_COMPONENT_ID_HEADER = "application-create";
+    private static final String VALUE_DELIMITER = "_";
+    private static final int ARG_COUNT = 3;
+
+    private final ApplicationApplyHandler applicationApplyHandler;
```
Zabuzard: u could call it just `handler` for simplicity. context is clear within this class

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+     *
+     * @param config the configuration containing the settings for the application form
+     */
+    public ApplicationCreateCommand(Config config) {
+        super("application-form", "Generates an application form for members to apply for roles.",
+                CommandVisibility.GUILD);
+
+        this.roleApplicationSystemConfig = config.getRoleApplicationSystemConfig();
+
+        generateRoleOptions(getData());
```
Zabuzard: i dont think its a good design that ur private helper method takes this argument that it has access to itself already. so id remove the argument and let the method call it by itself

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    private void generateRoleOptions(SlashCommandData data) {
+        IntStream.range(1, OPTIONAL_ROLES_AMOUNT + 1).forEach(index -> {
+            data.addOption(OptionType.STRING, generateOptionId("title", index),
+                    "The title of the role");
+            data.addOption(OptionType.STRING, generateOptionId("description", index),
+                    "The description of the role");
+            data.addOption(OptionType.STRING, generateOptionId("emoji", index),
+                    "The emoji of the role");
+        });
```
Zabuzard, line 92: what is this `OPTIONAL_ROLES_AMOUNT`? does it mean the amount of roles is hardcoded? does it mean its limited?
why is there a loop here, why is there a `+1`.

this part is questionable to me and i dont quite get how it looks for the slash command user

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+ * Represents a command to create an application form for members to apply for roles.
+ * <p>
+ * This command is designed to generate an application form for members to apply for roles within a
+ * guild.
+ */
+public class ApplicationCreateCommand extends SlashCommandAdapter {
+    protected static final Color AMBIENT_COLOR = new Color(24, 221, 136, 255);
+    private static final int OPTIONAL_ROLES_AMOUNT = 5;
+    private static final String ROLE_COMPONENT_ID_HEADER = "application-create";
+    private static final String VALUE_DELIMITER = "_";
```
Zabuzard: that name needs improvement. what does it delimit? `OPTION_NAME_ID_DELIMITER` or something

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+    private static final int ARG_COUNT = 3;
```
Zabuzard: unclear what it refers to, needs a rename. for example `OPTIONS_PER_ROLE`

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    private static long getIncorrectRoleArgsCount(final List<OptionMapping> args) {
+        final Map<String, Integer> frequencyMap = new HashMap<>();
+
+        args.stream()
+            .map(OptionMapping::getName)
+            .map(name -> name.split(VALUE_DELIMITER)[1])
+            .forEach(number -> frequencyMap.merge(number, 1, Integer::sum));
+
+        return frequencyMap.values().stream().filter(value -> value != 3).count();
```
Zabuzard: hardcoded 3. u probably meant to use the constant u defined above

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+
+        event.replyEmbeds(embed).addActionRow(menuBuilder.build()).queue();
+    }
+
+    private static MessageEmbed createApplicationEmbed() {
+        return new EmbedBuilder().setTitle("Apply for roles")
+            .setDescription(
+                    """
+                            We are always looking for community members that want to contribute to our community \
+                            and take charge. If you are interested, you can apply for various positions here!""")
```
Zabuzard: needs some smiley to soften up the text a bit. for example a 👍

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @Override
+    public void onStringSelectSelection(StringSelectInteractionEvent event, List<String> args) {
+        SelectOption selectOption = event.getSelectedOptions().getFirst();
+        Member member = event.getMember();
+
+        if (selectOption == null || member == null) {
+            return;
+        }
```
Zabuzard: can this happen in practice or does it represent an error case that should trigger a warning?

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
```diff
@@ -0,0 +1,290 @@
...
+     * @param count the number used to determine whether to return the singular or plural form
+     * @return the singular form if count equals 1, otherwise the plural form
+     */
+    private String selectWordFromCount(final Number count, final String singularForm,
+            final String pluralForm) {
+        if (count.intValue() == 1) {
+            return singularForm;
+        }
+        return pluralForm;
+    }
```
Zabuzard: i think this is a bit overengineered. can remove that helper (as its also only used once) and instead make the call-site a bit easier to read by inlining:
```java
String minutesText = remainingMinutes == 1 ? "minute", "minutes";

File: application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java [hunk 22]
```diff
@@ -0,0 +1,290 @@
...
+    public void onStringSelectSelection(StringSelectInteractionEvent event, List<String> args) {
+        SelectOption selectOption = event.getSelectedOptions().getFirst();
+        Member member = event.getMember();
+
+        if (selectOption == null || member == null) {
+            return;
+        }
+
+        long remainingMinutes = applicationApplyHandler.getMemberCooldownMinutes(member);
+        String correctMinutesWord = selectWordFromCount(remainingMinutes, "minute", "minutes");
```
Zabuzard: move into the `if`. not needed outside

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        User applicant = event.getUser();
+        EmbedBuilder embed =
+                new EmbedBuilder().setAuthor(applicant.getName(), null, applicant.getAvatarUrl())
+                    .setColor(ApplicationCreateCommand.AMBIENT_COLOR)
+                    .setTimestamp(Instant.now())
+                    .setFooter("Submitted at");
+
+        String roleString = args.getLast();
```
Zabuzard: prefer extracting the args at the beginning, right after ur length check. and id prefer explicit indices here to make sure the thing is at the index u expected. so `args.get(1)` in this case

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param event the modal interaction event triggering the application submission
+     * @param args the arguments provided in the application submission
+     * @param answer the answer provided by the applicant to the default question
+     */
+    protected void sendApplicationResult(final ModalInteractionEvent event, List<String> args,
+            String answer) {
+        Guild guild = event.getGuild();
+        if (args.size() != 2 || guild == null) {
+            return;
+        }
```
Zabuzard: sounds like this would be an error case that should emit a log message instead of silent failure?

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+        Optional<TextChannel> applicationChannel = getApplicationChannel(guild);
+        if (applicationChannel.isEmpty()) {
+            return;
+        }
```
Zabuzard, line 91: warning instead of silent failure?

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+        User applicant = event.getUser();
+        EmbedBuilder embed =
+                new EmbedBuilder().setAuthor(applicant.getName(), null, applicant.getAvatarUrl())
+                    .setColor(ApplicationCreateCommand.AMBIENT_COLOR)
+                    .setTimestamp(Instant.now())
+                    .setFooter("Submitted at");
```
Zabuzard: prefer reordering these two calls so it tells the story better "Submitted at ... Instant.now()"

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationApplyHandler.java
```diff
@@ -0,0 +1,142 @@
...
+            .filter(channel -> channel.getType().isMessage())
+            .map(TextChannel.class::cast)
+            .findFirst();
+    }
+
+    public Cache<Member, OffsetDateTime> getApplicationSubmitCooldown() {
+        return applicationSubmitCooldown;
+    }
+
+    protected void submitApplicationFromModalInteraction(ModalInteractionEvent event,
```
Zabuzard: why protected? this class doesnt make use of inheritance. id remove all the protected and make the class `final` instead

Zabuzard (changes requested):

LovinLifee:
Commenting so I can be assigned

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
christolis: We are dealing with a limitation in Discord's slash command design where the amount of parameters a slash command can take is limited. From some [research](https://github.com/discord/discord-api-docs/issues/2312), I can observe talks about it being `50`, which effectively means that no more than 50 slash command parameters are allowed to be registered.

As of writing this, the roles defined by this slash command are not direct references to actual Discord guild roles, but rather *custom* roles that the slash command user creates at the same time as typing the command.

For instance, for a user to create an application form with 2 roles, namely one called "Moderator" and one called "Expert", they could have to execute this command:

image

```
/application-form title_1:Moderator description_1:Enforce the community's rules and punish those who don't! emoji_1::star:  title_2:Expert description_2:A Java veteran with plenty of experience on their hands. emoji_2::gear: 
```

Running the above command leads to the following:

image

As I mentioned previously, these roles are totally customizable from the executor's side. The `IntStream` you see is to generate the slash command parameters, namely `description_1`, `description_2` all the way until the value that `OPTIONAL_ROLES_AMOUNT` holds.

To answer your `+1` question, it is due to the fact that we want to manipulate the parameters range so that we avoid parameters named `description_0`.

Zabuzard: oh. now that i see the UX in action im not sure im a fan of that design. its super weird seeing all these `_1...5` parameters and building some sort of super-slash-command to fit them all.

from user perspective it would probably be better to create some form of dialog instead. i.e. a slash command that starts the interaction, the bot opening a popup modal window where u enter the details for the first entry. and then a checkbox or button or anything where u can tell the bot that u want to add another one. and then it opens another popup modal window. eventually u select that ur done and now the bot wraps it up and creates the form or something

Zabuzard: im thinking sth like: `/application-form` without params. then a modal opens, u enter the details for the first entry, submit it.
the bot now writes the message with the application form, with this single entry to apply for.
and adds two buttons to it: `Add More`, `Finalize`.
u click `Add More`, get the same modal popup again, fill in details, submit. the bot edits the application-form to also include the other choice (and still the two buttons).
eventually u click on `Finalize` and it will remove the two buttons, done.

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
christolis: Virtually speaking, the specific event can never be triggered with the member being `null`, unless the interaction happens outside of the guild.

`selectedOption` is always an immutable nonnull `List`, albeit it could theoretically be empty if a slash command user somehow creates an application form with no roles *and* someone happens to select something from that empty list. Pretty much impossible if you ask me.

I added these null checks predominantly for the defensive programming aspect, not expecting these to be `null` at a frequent rate and there is no needed reaction from the bot's side in case either of these end up being `null` at some point.

Zabuzard: i see. we prefer fail-fast, not defensive programming.
when there is an error, we want to detect and log/crash asap so it doesnt linger unnoticed 👍

application/src/main/java/org/togetherjava/tjbot/features/roleapplication/ApplicationCreateCommand.java
christolis: Seems like a better alternative to the original method and it doesn't limit the user on how many roles they could add. About to implement this right now and see how it turns out 🙌

Just currently thinking about editing existing application forms after they have already been finalized. Hopefully there is a way to get existing data from existing application forms, but I will do some research.

Taz03 (dismissed):
LGTM

surajkumar (approved):
LGTM

PR #1052, opened 2024-03-12
Feature: ChatGPT - Code formatter

### LINKED ISSUE #813: ChatGPT - Code formatter
We have a code formatting feature, somewhere around `CodeAction` (`FormatterCodeAction` maybe?).

The existing formatter is totally fine, but we should add a ChatGPT powered layer before it. If the ChatGPT service (`OpenAiService`) returns `Optional.empty()`, the code should fall back to our existing formatter instead.

To save money, the feature should not consider the ChatGPT layer for really long code.

Care must be taken that ChatGPT only formats the code and doesnt otherwise change it.

christolis:
Closes #813.


application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
```diff
@@ -96,12 +97,58 @@ public Optional<String> ask(String question, String context) {
...
+
+                                --- BEGIN CODE ---
+                                %s
+                                --- END CODE ---
+                        """,
+                code);
+
+        Optional<String> response = getMessageResponse(payload);
+
+        if (response.isEmpty() || response.get().equalsIgnoreCase("empty")) {
```
ankitsmt211, line 135: Shouldn't just checking optional is empty enough here? Why prompt GPT to respond with "empty" 🧐?

application/src/main/java/org/togetherjava/tjbot/features/basic/ChatGPTFormatter.java
```diff
@@ -0,0 +1,31 @@
+package org.togetherjava.tjbot.features.basic;
+
+import org.togetherjava.tjbot.features.chatgpt.ChatGptService;
+
+import java.util.Optional;
+
+/**
+ * A utility class for formatting code using ChatGPT.
```
ankitsmt211, line 8: this class should be in `code` package instead since `basic` is for example commands only imo

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
```diff
@@ -96,12 +97,58 @@ public Optional<String> ask(String question, String context) {
...
 
+        String instructions = "KEEP IT CONCISE, NOT MORE THAN 280 WORDS";
+        String questionWithContext = "context: Category %s on a Java Q&A discord server. %s %s"
+            .formatted(context, instructions, question);
+
+        return getMessageResponse(questionWithContext);
+    }
+
+    /**
+     * Prompt ChatGPT with code for it to format it.
```
ankitsmt211, line 108: maybe  `Provide ChatGPT with code to format.` instead

ankitsmt211 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
christolis: What prompt should we change it to so that it responds with literally no answer instead of instructing it to respond with "empty"? And even in that case, won't the OpenAI API return an empty string?

ankitsmt211: why would it respond with no answer? if for some reason API fails we'll get empty optional. Was there any instance of it returning empty string?

We won't even need to check and directly do 
`return getMessageResponse(payload);`

ankitsmt211 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
christolis: But what if the API receives non-valid code or code that's too long? It's instructed to respond with "empty" if any of the predefined conditions are not met in order to correctly decide whether to delegate the code formatting task to the original code or not.

Feel free to share any better methods that would solve this.

ankitsmt211: Let me try and address some of the concerns

1. This will be applied by someone when they use context menu options for code action, it won't be automatically formatted.
2. For formatting there's no such thing as invalid code, it would apply the formatting rules automatically of a certain language based on input even if it's a diff langauges then what it interprets.
3. As for length, you're already checking length for code greater than 2k chars to return optional at start of the method.

the method is itself is solid, but this if check ` if (response.isEmpty() || response.get().equalsIgnoreCase("empty"))` seems unnecessary imo

you can just do `return getMessageResponse(payload)` instead.

and also can safely get rid of prompt for "empty" string as well

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
Taz03: i kinda agree, just let gpt do whatever it can,  and if it fails  to  respond  then only use the other formatter

application/src/main/java/org/togetherjava/tjbot/features/code/FormatCodeCommand.java
```diff
@@ -13,6 +14,11 @@
  */
 final class FormatCodeCommand implements CodeAction {
     private final Formatter formatter = new Formatter();
+    private final ChatGPTFormatter chatGPTFormatter;
+
+    public FormatCodeCommand(ChatGptService service) {
+        this.chatGPTFormatter = new ChatGPTFormatter(service);
+    }
```
ankitsmt211, line 21: documentation please

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
```diff
@@ -96,12 +97,51 @@ public Optional<String> ask(String question, String context) {
...
+     */
+    private Optional<String> getMessageResponse(String message) {
         try {
-            String instructions = "KEEP IT CONCISE, NOT MORE THAN 280 WORDS";
-            String questionWithContext = "context: Category %s on a Java Q&A discord server. %s %s"
-                .formatted(context, instructions, question);
-            ChatMessage chatMessage = new ChatMessage(ChatMessageRole.USER.value(),
-                    Objects.requireNonNull(questionWithContext));
+            ChatMessage chatMessage =
+                    new ChatMessage(ChatMessageRole.USER.value(), Objects.requireNonNull(message));
```
Taz03, line 144: make sure to add context in help thread chat gpt feature so that doesn't break

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
ankitsmt211: `ChatGptService` contains ask method which also has context and is used for both help threads and `/chatgpt` feature

Taz03: yes, and this pr is removing that to make it more generic. So the context should be moved to the using sites

ankitsmt211: image

I'm not sure what you mean by removing it, i still see context here in local copy of this PR.

Taz03 (approved):

ankitsmt211 (dismissed):
looks good 👍

ankitsmt211:
i tried running a random snippet but results were unexpected, 
```
@Configuration
@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {

    // this you get for free when you configure the db connection in application.properties file
    @Autowired
    private DataSource dataSource;

    // this bean is created in the application starter class if you're looking for it
    @Autowired
    private PasswordEncoder encoder;

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
        .csrf().disable()
        .authorizeRequests()
        .antMatchers(HttpMethod.OPTIONS, "/api/**").permitAll() // For CORS, the preflight request
        .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()     // will hit the OPTIONS on the route
        .antMatchers("/api/**").authenticated() // Requests for our REST API must be authorized.
        .anyRequest().permitAll()               // All other requests are allowed without authorization.
        .and()
        .httpBasic();                           // Use HTTP Basic Authentication

        http
        .sessionManagement()
        .sessionCreationPolicy(SessionCreationPolicy.STATELESS);
    }

    @Override
    protected void configure(AuthenticationManagerBuilder auth) throws Exception {
        String userQuery = "SELECT username, password, enabled FROM User WHERE username=?";
        String authQuery = "SELECT username, role FROM User WHERE username=?";
        auth
        .jdbcAuthentication()
        .dataSource(dataSource)
        .usersByUsernameQuery(userQuery)
        .authoritiesByUsernameQuery(authQuery)
        .passwordEncoder(encoder);
    }
}@Configuration
@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {

    // this you get for free when you configure the db connection in application.properties file
    @Autowired
    private DataSource dataSource;

    // this bean is created in the application starter class if you're looking for it
    @Autowired
    private PasswordEncod
```
OUTPUT:
```
er encoder;

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .csrf().disable()
            .authorizeRequests()
                .antMatchers(HttpMethod.OPTIONS, "/api/**").permitAll() // For CORS, the preflight request
                .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()     // will hit the OPTIONS on the route
                .antMatchers("/api/**").authenticated() // Requests for our REST API must be authorized.
                .anyRequest().permitAll()               // All other requests are allowed without authorization.
            .and()
            .httpBasic();                           // Use HTTP Basic Authentication

        http
            .sessionManagement()
                .sessionCreationPolicy(SessionCreationPolicy.STATELESS);
    }

    @Override
    protected void configure(AuthenticationManagerBuilder auth) throws Exception {
        String userQuery = "SELECT username, password, enabled FROM User WHERE username=?";
        String authQuery = "SELECT username, role FROM User WHERE username=?";
        auth
            .jdbcAuthentication()
                .dataSource(dataSource)
                .usersByUsernameQuery(userQuery)
                .authoritiesByUsernameQuery(authQuery)
                .passwordEncoder(encoder);
    }
}
```
It's not complete output and always same as well, the length is less than 2k this is gpt formatting at work

PR #1055, opened 2024-03-14
formatted member count

Taz03:
* Format member count with proper commas
* For member count as `300000` it will be formatted to `300,000`(groups of thousand).
* update predicate to match category including commas in category name.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/basic/MemberCountDisplayRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/basic/MemberCountDisplayRoutine.java
@@ -6,6 +6,7 @@ import net.dv8tion.jda.api.entities.channel.concrete.Category;
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.features.Routine;
 
+import java.text.NumberFormat;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/basic/MemberCountDisplayRoutine.java
```diff
@@ -23,17 +24,16 @@ public class MemberCountDisplayRoutine implements Routine {
      */
     public MemberCountDisplayRoutine(Config config) {
         memberCountCategoryPredicate =
-                Pattern.compile(config.getMemberCountCategoryPattern() + "( - \\d+ Members)?")
+                Pattern.compile(config.getMemberCountCategoryPattern() + "( - [\\d,]+ Members)?")
```
ankitsmt211, line 27: would be nice to write a comment for what this predicate would match as an example.

Taz03: image

ankitsmt211: i meant inside code itself 
//matches likes of `Info-1,000 Members`
or something like that

Taz03: nah

ankitsmt211: I can do it if it's a bother, problem with regex is it looks like magical stuff so you have to either use a third party tool like a platform to validate it which you don't need to if a comment mentions what sort of pattern we're trying to match which can ofc save everyone a bit of time and make it more readable.

christolis (approved):
LGTM

surajkumar (approved):

tj-wazei (approved):

Issue #1011, opened 2024-01-22
Title for transfer questions generated by ai


Taz03:
When we want to transfer a question we have to give it a suitable title, manually

We can ask chatgpt to generate a suitable title for the question.

Additional Context:
Method `createTitle` is responsible for generating modal title in `TransferQuestionCommand.java`

vishv843:
Hi I wanna work on this

ankitsmt211:
Sure thing, welcome to the project 😊

PR #1059, opened 2024-03-19
Title generation by ChatGPT for transfer-questions command

vishv843:
Previously, the title was simply created by putting the message when /transfer-questions command was invoked. The title had to be written by a human for clarity. This PR solves that issue by using ChatGPT for title generation. #1011 

question
response

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -32,6 +32,7 @@ import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.features.BotCommandAdapter;
 import org.togetherjava.tjbot.features.CommandVisibility;
 import org.togetherjava.tjbot.features.MessageContextCommand;
+import org.togetherjava.tjbot.features.chatgpt.ChatGptService;
 import org.togetherjava.tjbot.features.utils.StringDistances;
 
 import java.awt.Color;
@@ -64,20 +65,23 @@ public final class TransferQuestionCommand extends BotCommandAdapter
     private static final int INPUT_MIN_LENGTH = 3;
     private final Predicate<String> isHelpForumName;
     private final List<String> tags;
+    private final ChatGptService chatGptService;
 
 
     /**
      * Creates a new instance.
      *
      * @param config to get the helper forum and tags
+     * @param chatGptService the service used to ask ChatGPT questions via the API.
      */
-    public TransferQuestionCommand(Config config) {
+    public TransferQuestionCommand(Config config, ChatGptService chatGptService) {
         super(Commands.message(COMMAND_NAME), CommandVisibility.GUILD);
 
         isHelpForumName =
                 Pattern.compile(config.getHelpSystem().getHelpForumPattern()).asMatchPredicate();
 
         tags = config.getHelpSystem().getCategories();
+        this.chatGptService = chatGptService;
     }
 
     @Override
@@ -92,12 +96,20 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
+        String chatGptPrompt =
+                "generate title for this question %s, it shouldn't be longer than 4-5 words with no header"
+                    .formatted(originalMessage);
+        Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
+        String title = chatGptResponse.orElse(createTitle(originalMessage));
+        if (title.length() > TITLE_MAX_LENGTH) {
+            title = title.substring(0, TITLE_MAX_LENGTH);
+        }
 
         TextInput modalTitle = TextInput.create(MODAL_TITLE_ID, "Title", TextInputStyle.SHORT)
             .setMaxLength(TITLE_MAX_LENGTH)
             .setMinLength(TITLE_MIN_LENGTH)
             .setPlaceholder("Describe the question in short")
-            .setValue(createTitle(originalMessage))
+            .setValue(title)
             .build();
 
         Builder modalInputBuilder =
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -97,7 +97,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
         String chatGptPrompt =
-                "generate title for this question %s, it shouldn't be longer than 4-5 words with no header"
+                "Summarize the following text into a concise title or heading not more than 4-5 words, remove quotations if any: %s"
                     .formatted(originalMessage);
         Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
         String title = chatGptResponse.orElse(createTitle(originalMessage));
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -64,20 +65,22 @@ public final class TransferQuestionCommand extends BotCommandAdapter
...
+    private final ChatGptService chatGptService;
 
 
     /**
      * Creates a new instance.
      *
      * @param config to get the helper forum and tags
      */
-    public TransferQuestionCommand(Config config) {
+    public TransferQuestionCommand(Config config, ChatGptService chatGptService) {
```
ankitsmt211, line 76: document the param please

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -92,12 +95,20 @@ public void onMessageContext(MessageContextInteractionEvent event) {
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
+        String chatGptPrompt = "Generate title for question:"
+                                + originalMessage + ". Minimum " + TITLE_MIN_LENGTH
+                                + " and maximum " + TITLE_MAX_LENGTH +" characters. Return only the title.";
+        Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
```
ankitsmt211, line 101: do you have access to thread tag here? might be a good idea to use it as context instead of empty string

ankitsmt211 (changes requested):

vishv843:
Hi @ankitsmt211. I have made the necessary changes. Please review it.

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -92,12 +96,18 @@ public void onMessageContext(MessageContextInteractionEvent event) {
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
+        String chatGptPrompt = String.format(
+                "Generate topic for question: %s. Minimum %s and maximum %s characters.",
+                originalMessage, TITLE_MIN_LENGTH, TITLE_MAX_LENGTH);
+        Optional<String> chatGptResponse =
+                chatGptService.ask(chatGptPrompt, String.join(",", tags));
```
ankitsmt211, line 103: please just use the first tag in applied ones on thread, that should be enough

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -92,12 +96,18 @@ public void onMessageContext(MessageContextInteractionEvent event) {
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.get(0);
+        String chatGptPrompt = String.format(
+                "Generate topic for question: %s. Minimum %s and maximum %s characters.",
```
ankitsmt211, line 100: I don't know if it's a good idea to rely on GPT for char limit for title, might be a good idea to check the title length before setting title, if title generated is greater than our limit you can trim the title using `String#substring` instead as backup.

ankitsmt211 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
vishv843: I did that because chatgpt was giving long topic names which really wasn't suitable. Trimming that long response generate bad output overall. I have attached test responses for your reference.
image
image

vishv843: I have addressed rest of your comments

ankitsmt211: can we use both? Like request gpt to generate shorter responses and check if it's actually within limit, if yes then go ahead otherwise we should trim it.

vishv843: Yes that can be done

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -92,12 +96,20 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
+                originalMessage, TITLE_MIN_LENGTH, TITLE_MAX_LENGTH);
```
ankitsmt211, line 101:
```suggestion
        String chatGptPrompt =
                "generate title for this question %s, it shouldn't be longer than 4-5 words with no header".formatted(originalMessage);
```

I suggest using `formatted` instead, it's same thing but for consistency across project it's a bit better.

Also made slight variations in prompt, gpt isn't too great with character limits so there was no point in using Max or min limit here. When i tried running the prompt you added, it would add `Title: whatever title here` as header sometimes also the title was for some reasons inside double quotes.

i tested this prompt with couple of questions and generates more cleaner ones, you can try it out as well.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -92,12 +96,20 @@ public void onMessageContext(MessageContextInteractionEvent event) {
...
+        Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, mostCommonTag);
```
ankitsmt211, line 102: your earlier approach of not passing any tags was right since i just realised it's directly transfering a text so it's impossible to know what tag would be here. My bad for suggesting you otherwise, instead of mostCommonTag you can pass empty string 👍

ankitsmt211 (changes requested):
This is it from my side, good work 
Thanks 👍

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
ankitsmt211: for reference this is how earlier title was generated with older prompt
image

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
vishv843: I had thought the same thing but I thought that you guys must rely on the config tags to get responses related to the topic. I should have clarified it with you.

vishv843:
I tested the prompt you gave on some questions, worked pretty well. I have added it. Thanks!!

ankitsmt211:
@vishv843 everything looks great, i wish i should have tested the prompt more before suggesting it. I tested it on a couple of more questions and it would return title in quotation marks sometimes, so i finally decided to ask gpt itself for a better prompt. Now this one i validated against more than 4-5 questions from server itself so seems to be solid overall.

image

```
String chatGptPrompt = "Create a concise title for the question: '%s'. The title should be 4-5 words maximum and presented without quotation marks."
    .formatted(originalMessage);
```

I'd be fine with doing the changes myself if it's a bother 😅

Taz03:
@ankitsmt211 imo its better to just check for quotes and remove them manually, gpt sometimes does things even when told not to

```java
String title = gptTitle.matches("['\"]\\w+['\"]") ? gptTitle.substring(1, gptTitle.length() - 1) : gptTitle
```
where `gptTitle` is the title suggested by chatgpt

ankitsmt211:
> @ankitsmt211 imo its better to just check for quotes and remove them manually, gpt sometimes does things even when told not to
> 
> ```java
> if (title.matches("['\"]\\w+['\"]")) title = title.substring(1, gptTitle.length() - 1);
> ```

this is very specific and if GPT is unreliable then we can't really handle all other cases, so imo just trying for best prompt that we can should be a priority, if you're specific with prompt it does work as expected specially when generating simple text is what it's known about.

ankitsmt211 (approved):

PR #1062, opened 2024-03-20
made tags command ephemeral

### LINKED ISSUE #958: tag qol alternative?
Saw a PR  #602  , seems like it's not relevant anymore. But there's still a problem when it comes to using `/tags`, often have to go back and forth from `#bots` to `#questions` unless you remember most of em. 

**Solution:**
A slightly different approach instead of "getting the closest match" and all the heavy lifting, why not make `/tags` as ephemeral, that way you can use it wherever you want without swapping or memorizing any. Which will also prevent any clutter in the channel.

To educate the users about `/tags`, we can modify the response for each `/tag tagNamehere`, append it with something like "to list all commands you can use /tags".

**Caveat:**
This could lead to curious users following `/tags` with `/tag tagNameHere` in respective channel, for that we can put a restriction for for usage of `/tag` everywhere for normal users apart from `Helper Threads` and `#bots`.

Taz03:
resolves #958 
closes #983 

- Made tags command ephemeral
- Removed footer
- Removed delete button
- Removed tests for delete button

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/tags/TagsCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/tags/TagsCommand.java
@@ -1,9 +1,7 @@
 package org.togetherjava.tjbot.features.tags;
 
 import net.dv8tion.jda.api.EmbedBuilder;
-import net.dv8tion.jda.api.Permission;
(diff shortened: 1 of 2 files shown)


Taz03:
@ankitsmt211 

image

SquidXTV (approved):

ankitsmt211 (approved):
looks good 👍

PR #1065, opened 2024-03-21
fixed dismiss buttons

Taz03:
after the latest jda-upgrade, deleting self message gives "Unkown Webhook" Error, to fix we first need to defer reply to the event
 
 replacing `getMessage().delete()` with `deferEdit().deleteOriginal()` in
 * Tex command
 * tags command
 * File Sharing Listener


Taz03:
@ankitsmt211 after the latest jda-upgrade, deleting self message gives "Unkown Webhook" Error, to fix we first need to defer reply to the event

ankitsmt211:
@Taz03 it seems `getMessage.delete` is present in other commands as well, assuming it would lead to issues there too?
 some of the test cases also has it

couple of em i could find
* tags command
* Tex command
* some tests around media listener/ above mentioned commands

Taz03:
blocked by #1062

SquidXTV (approved):

PR #1068, opened 2024-03-21
handle threads when authorId found

### LINKED ISSUE #1067: HelpThreadAutoArchiver can't find author from database
# About

Since the new release, this error comes up a lot of times. 
Might be worth looking into.

stacktrace

ankitsmt211:
Resolves #1067 
Error:
when retrieving threads from DB, helper method would try to get active thread channel and match it's ID with thread record from DB. If a match happens, it would then grab authorID from that record and go on with rest of archive flow (retrieve user and then chooses to archive thread with or without mentioning author based on results)

Changes:
* Dividing archive logic into two parts, if a record match happens in DB aka authorId is found or not
    * `triggerArchiveFlow`
    * `triggerAuthorIdNotFoundArchiveFlow`

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
@@ -125,14 +125,21 @@ public final class HelpThreadAutoArchiver implements Routine {
     }
 
     private void handleArchiveFlow(ThreadChannel threadChannel, MessageEmbed embed) {
+        helper.getAuthorByHelpThreadId(threadChannel.getIdLong()).ifPresentOrElse(authorId -> {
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadAutoArchiver.java
```diff
@@ -151,4 +156,11 @@ private void handleArchiveFlow(ThreadChannel threadChannel, MessageEmbed embed)
             .flatMap(any -> threadChannel.getManager().setArchived(true))
             .queue();
     }
+
+    private void triggerAuthorIdNotFoundArchiveFlow(ThreadChannel threadChannel,
```
Taz03, line 160: maybe also log the thread channel id

Taz03 (approved):

SquidXTV (approved):

PR #1078, opened 2024-03-26
refactor HelpThreadCreatedListener

ankitsmt211:
Resolving error:
image

more context for bugfix : [discord conversation](https://discord.com/channels/272761734820003841/895717328359153664/1221775843286843402)

replacing `onChannelCreate` with `onMessageRecieved` for `HelpThreadCreatedListenerFlow` this also ensures that we no longer need `additional delay` in creating automated response or gpt response. Not only makes code a bit less complex but also possibly resolves this error around `unknown message` 
* move write to db in a method
* use messageRecieved instead of onChannelCreate
* rename createMessages to generateAutomatedResponse
* chain generateAutomatedResponse and pinOriginal to retreiveMessage rest action

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
@@ -9,10 +9,12 @@ import net.dv8tion.jda.api.entities.Role;
 import net.dv8tion.jda.api.entities.User;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.entities.channel.forums.ForumTag;
-import net.dv8tion.jda.api.events.channel.ChannelCreateEvent;
+import net.dv8tion.jda.api.entities.channel.unions.IThreadContainerUnion;
 import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.dv8tion.jda.api.requests.RestAction;
+import org.jetbrains.annotations.NotNull;
 
 import org.togetherjava.tjbot.features.EventReceiver;
 import org.togetherjava.tjbot.features.UserInteractionType;
@@ -56,20 +58,19 @@ public final class HelpThreadCreatedListener extends ListenerAdapter
     }
 
     @Override
-    public void onChannelCreate(ChannelCreateEvent createEvent) {
-        if (!createEvent.getChannelType().isThread()) {
-            return;
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        if (event.isFromThread()) {
+            IThreadContainerUnion parentChannel =
+                    event.getChannel().asThreadChannel().getParentChannel();
+            if (helper.isHelpForumName(parentChannel.getName())) {
+                ThreadChannel threadChannel = event.getChannel().asThreadChannel();
+                int messageCount = threadChannel.getMessageCount();
+                if (messageCount > 1 || wasThreadAlreadyHandled(threadChannel.getIdLong())) {
+                    return;
+                }
+                handleHelpThreadCreated(threadChannel);
+            }
         }
-        ThreadChannel threadChannel = createEvent.getChannel().asThreadChannel();
-
-        if (wasThreadAlreadyHandled(threadChannel.getIdLong())) {
-            return;
-        }
-
-        if (!helper.isHelpForumName(threadChannel.getParentChannel().getName())) {
-            return;
-        }
-        handleHelpThreadCreated(threadChannel);
     }
 
     private boolean wasThreadAlreadyHandled(long threadChannelId) {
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -56,20 +58,19 @@ public HelpThreadCreatedListener(HelpSystemHelper helper) {
...
-        if (!helper.isHelpForumName(threadChannel.getParentChannel().getName())) {
-            return;
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        if (event.isFromThread()) {
+            IThreadContainerUnion parentChannel =
+                    event.getChannel().asThreadChannel().getParentChannel();
+            if (helper.isHelpForumName(parentChannel.getName())) {
+                ThreadChannel threadChannel = event.getChannel().asThreadChannel();
+                int messageCount = threadChannel.getMessageCount();
+                if (messageCount > 1 || wasThreadAlreadyHandled(threadChannel.getIdLong())) {
```
ankitsmt211, line 68: to check if this is thread creation event instead of any thread message event

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -56,20 +58,19 @@ public HelpThreadCreatedListener(HelpSystemHelper helper) {
...
-        if (wasThreadAlreadyHandled(threadChannel.getIdLong())) {
-            return;
-        }
-
-        if (!helper.isHelpForumName(threadChannel.getParentChannel().getName())) {
-            return;
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        if (event.isFromThread()) {
+            IThreadContainerUnion parentChannel =
+                    event.getChannel().asThreadChannel().getParentChannel();
```
Taz03, line 64: store it in a Channel type variable, for simplicity

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
ankitsmt211: the return type is `IthreadContainerUnion`, need that to get parent channel

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
ankitsmt211: i feel like maybe there should be a comment here about message count logic it can be a bit confusing out of context

tj-wazei (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -56,20 +58,19 @@ public HelpThreadCreatedListener(HelpSystemHelper helper) {
...
-
-        if (wasThreadAlreadyHandled(threadChannel.getIdLong())) {
-            return;
-        }
-
-        if (!helper.isHelpForumName(threadChannel.getParentChannel().getName())) {
-            return;
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        if (event.isFromThread()) {
+            IThreadContainerUnion parentChannel =
```
SquidXTV, line 63: you can make this variable of type `Channel` instead, `IThreadContainerUnion` looks weird in the first place and is not needed as you only use `Channel#getName` from it

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
SquidXTV: I think this can be exploited no? I am not that familiar with JDA but will `#getMessageCount` call some API or is this information already available when the event triggers? If it calls some API then someone could spam a second message fast enough to return a message count `> 1` though it didn't get handled yet

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -82,23 +83,10 @@ private boolean wasThreadAlreadyHandled(long threadChannelId) {
...
-
-            helper.writeHelpThreadToDatabase(authorId, threadChannel);
-        });
-
-        // The creation is delayed, because otherwise it could be too fast and be executed
-        // after Discord created the thread, but before Discord send OPs initial message.
-        // Sending messages at that moment is not allowed.
-        createMessages(threadChannel).and(pinOriginalQuestion(threadChannel))
-            .queueAfter(5, TimeUnit.SECONDS);
+        threadChannel.retrieveMessageById(threadChannel.getIdLong()).flatMap(message -> {
```
SquidXTV, line 86: We upgraded JDA right? Why not using `ThreadChannel#retrieveStartMessage` then?

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
ankitsmt211: sure thing

SquidXTV (approved):

tj-wazei (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -56,20 +58,18 @@ public HelpThreadCreatedListener(HelpSystemHelper helper) {
...
-
-        if (!helper.isHelpForumName(threadChannel.getParentChannel().getName())) {
-            return;
+    public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+        if (event.isFromThread()) {
+            Channel parentChannel = event.getChannel().asThreadChannel().getParentChannel();
+            if (helper.isHelpForumName(parentChannel.getName())) {
+                ThreadChannel threadChannel = event.getChannel().asThreadChannel();
+                int messageCount = threadChannel.getMessageCount();
+                if (messageCount > 1 || wasThreadAlreadyHandled(threadChannel.getIdLong())) {
```
Taz03, line 67: messageCount is reliable, remove wasThreadAlreadyHandled

Taz03 (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
Taz03, line 67: messageCount is reliable, remove wasThreadAlreadyHandled

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
ankitsmt211: It's not, I already had a chit-chat with waz regarding another solution where we have to pull an updated history of thread on creation to rely on message count. However since wasAlreadHandled also works, I decided to not change it. 

You can try it out locally by removing it.

Taz03 (approved):

PR #1086, opened 2024-04-13
chore: add safety checks to NoteCommand

### LINKED ISSUE #1085: Note command lacks safety checks in onSlashCommand like BanCommand
**Describe the bug**
Note command lacks safety checks in onSlashCommand like BanCommand

**Expected behavior**
For the checks to be present

**To Reproduce**

**Screenshots**

**Additional context**

SimonVerhoeven:
Resolves #1085

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/NoteCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/NoteCommand.java
@@ -51,10 +51,11 @@ public final class NoteCommand extends SlashCommandAdapter {
 
     @Override
     public void onSlashCommand(SlashCommandInteractionEvent event) {
-        OptionMapping targetOption = event.getOption(USER_OPTION);
-        Member author = event.getMember();
-        Guild guild = event.getGuild();
-        String content = event.getOption(CONTENT_OPTION).getAsString();
+        OptionMapping targetOption =
+                Objects.requireNonNull(event.getOption(USER_OPTION), "The user is null");
+        Member author = Objects.requireNonNull(event.getMember());
+        Guild guild = Objects.requireNonNull(event.getGuild());
+        String content = Objects.requireNonNull(event.getOption(CONTENT_OPTION), "The content is null").getAsString();
 
         if (!handleChecks(guild.getSelfMember(), author, targetOption.getAsMember(), content,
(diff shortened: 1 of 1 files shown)


Taz03:
It is intentionally kept like that, because the values can never be null

Zabuzard (approved):
this change is unnecessary but also okay.

both approaches are totally fine and valid. intentionally triggering a NPE by using a possibly-null-but-required-not-null object versus explicitly using Objects.requireNonNull are both valid and almost identical.

we had a lengthy discussion on this matter in the past and decided to accept both ways.

generally, id suggest to not run through the entire code base and substitute one by one. but i have nothing against this PR, nor if ud use that style while writing a feature.

just don't make PRs only for that (opinionated) change, i guess.

fine for now though 👍

application/src/main/java/org/togetherjava/tjbot/features/moderation/NoteCommand.java
```diff
@@ -51,10 +51,13 @@ public NoteCommand(ModerationActionsStore actionsStore) {
...
-        Member author = event.getMember();
-        Guild guild = event.getGuild();
-        String content = event.getOption(CONTENT_OPTION).getAsString();
+        OptionMapping targetOption =
+                Objects.requireNonNull(event.getOption(USER_OPTION), "The user is null");
+        Member author = Objects.requireNonNull(event.getMember());
+        Guild guild = Objects.requireNonNull(event.getGuild());
+        String content =
+                Objects.requireNonNull(event.getOption(CONTENT_OPTION), "The content is null")
+                    .getAsString();
```
Taz03, line 60:
```suggestion
        OptionMapping targetOption = event.getOption(USER_OPTION);
        Member author = Objects.requireNonNull(event.getMember());
        Guild guild = Objects.requireNonNull(event.getGuild());
        String content = event.getOption(CONTENT_OPTION).getAsString();
```

restore `targetOption` and `content` as discussed in discord, or don't idek

surajkumar (approved):

tj-wazei (approved):

tj-wazei:
Approving + merging as I have tested the behavior. An exception is fine even though unlikely to occur. The exception won't leak out any important information so this is good to go.

PR #1090, opened 2024-04-16
Hotfix: Rewrite logic for acquiring RSS target channels

christolis:
This commit attempts to change the way a text channel list for RSS feeds gets generated, as it would originally not consider any channels that matched the fallback channel pattern and it would constantly log a warning that would clutter the log channels and skip sending any RSS posts as a result.

The method which is responsible for finding the text channels from a given RSS feed configuration now focuses on collecting a list of text channels that match the target channel pattern from the configuration, and if no channels are found, the same collection attempt happens with the fallback channel pattern. In case an empty list is still yielded, a now-improved and more accurate warning message gets logged.


Serters (approved):
Great job!

ankitsmt211 (approved):
looks good

ankitsmt211:
going to sudo merge this, PR is roughly an old month at this point.

PR #1098, opened 2024-04-22
Proper RSS default config

Zabuzard:
working config for RSS feed that fits majority of users


Zabuzard:
(@christolis for reference)

PR #1105, opened 2024-05-07
Fix: Remove pinned eclipse jdt formatter Resolves #605

### LINKED ISSUE #605: Update eclipse jdt formatter configuration
We should update our formatting configuration. It's based on [old google style](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml), and it lacks all the new features.

It also has a weird issue with formatting a specific record. Issue noticed first here: https://github.com/Together-Java/TJ-Bot/pull/140. Investigated by spotless devs here: https://github.com/diffplug/spotless/issues/951

image

After investigation, the conclusion is that our configuration is the issue. And our band-aid solution was to pin the eclipse jdt formatter version to the latest known working, `eclipse('4.19.0')`.

Partly because of that pinning, we don't have `@snippet` support. https://github.com/Together-Java/TJ-Bot/issues/591. And we will continue experiencing more and more issues down the line. 

* * *

It's also a good opportunity to rethink our formatter choice, use better and smarter configuration etc.

thrashy:
Fix: Resolves https://github.com/Together-Java/TJ-Bot/issues/605

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/bookmarks/BookmarksCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/bookmarks/BookmarksCommand.java
@@ -82,12 +82,12 @@ public final class BookmarksCommand extends SlashCommandAdapter {
 
         OptionData addNoteOption = new OptionData(OptionType.STRING, ADD_BOOKMARK_NOTE_OPTION,
                 "Your personal comment on this bookmark")
-                    .setMaxLength(BookmarksSystem.MAX_NOTE_LENGTH)
-                    .setRequired(false);
+            .setMaxLength(BookmarksSystem.MAX_NOTE_LENGTH)
+            .setRequired(false);
 
         SubcommandData addSubCommand = new SubcommandData(SUBCOMMAND_ADD,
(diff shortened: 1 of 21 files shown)


Taz03 (approved):

ankitsmt211:
@marko-radosavljevic you mind looking into the fix?

marko-radosavljevic:
Thanks for your first contribution @thrashy. Welcome to the project :relaxed: :heart:

Not sure what introduced these changes, ideally upinning the formatter version doesn't change our formatting at all, but changes themselves are not bad. Finally unpinning the formatter has a lot of benefits on it's own, so I'm fine with it.

I just dislike incosistencies with formattig, and whole point of spotless automatization and formattig configuration was to have deterministic formatting, exactly how we want it, full cotrol to modify the behaviour. This is kinda not that, we just go with the flow what spotless spits out, because configuring it is not trivial. 

Also not sure if it fixes the original reason we pinned the version in the first place, with that Token record, so our config might still have some issues. But that part of the codebase is refactored, so let's merge this and see how it goes.

marko-radosavljevic (approved):

PR #1108, opened 2024-05-12
Create .gitattributes

surajkumar:
Created a .gitattributes to enforce LF encoding. On Windows when I ran spotless, it changes the line encoding to CFRL which is incorrect.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


tj-wazei (approved):

Zabuzard (changes requested):
What about existing code? If this is ticked, will tooling suddenly report that some existing code has the wrong file endings?
If so, it has to be changed in the same PR as well.

Zabuzard (approved):
(wazei says on discord that all files are already LF)

PR #1112, opened 2024-05-13
Refactor help thread creation workflow

### LINKED ISSUE #1111: refactor thread creation flow

current thread creation flow involves retrieving the same message at least times at top level, which is lets just say not very optimal. 

The idea is to avoid retrieving same message again and again and pass it further down the chain so it can be later used inside pinned logic. 
**Additional context**

The chain looks much simpler at top level, but it has a lot of moving parts.

HelpThreadCreatedListener-> `application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java`
method-> `HelpThreadCreatedListener#handleHelpThreadCreated`

SquidXTV:
resolves #1111 

# About 
This slightly improves the help thread creation workflow by reusing already gathered `Message`.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
@@ -60,9 +60,9 @@ public final class HelpThreadCreatedListener extends ListenerAdapter
     @Override
     public void onMessageReceived(@NotNull MessageReceivedEvent event) {
         if (event.isFromThread()) {
-            Channel parentChannel = event.getChannel().asThreadChannel().getParentChannel();
+            ThreadChannel threadChannel = event.getChannel().asThreadChannel();
+            Channel parentChannel = threadChannel.getParentChannel();
             if (helper.isHelpForumName(parentChannel.getName())) {
-                ThreadChannel threadChannel = event.getChannel().asThreadChannel();
                 int messageCount = threadChannel.getMessageCount();
                 if (messageCount > 1 || wasThreadAlreadyHandled(threadChannel.getIdLong())) {
                     return;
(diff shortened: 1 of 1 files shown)


ankitsmt211 (approved):
looks solid, good job flow is much easier to look at with refactor.

SquidXTV:
yeah kinda annoying with all these `any` in the lambdas. 
I hope we update to java 22 soon, which lets us replace them with `_`

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -84,8 +84,11 @@ private boolean wasThreadAlreadyHandled(long threadChannelId) {
     private void handleHelpThreadCreated(ThreadChannel threadChannel) {
         threadChannel.retrieveStartMessage().flatMap(message -> {
             registerThreadDataInDB(message, threadChannel);
-            return generateAutomatedResponse(threadChannel);
-        }).flatMap(message -> pinOriginalQuestion(threadChannel)).queue();
+            return sendHelperHeadsUp(threadChannel)
+                .flatMap(any -> HelpThreadCreatedListener.isContextSufficient(message),
+                        any -> createAIResponse(threadChannel, message))
+                .flatMap(any -> pinOriginalQuestion(message));
```
Taz03, line 90:
```suggestion
                .flatMap(Message::pin);
```

Taz03 (approved):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
SquidXTV: This would change the behavior, because now it would pin the `any` Message from the `RestAction` call before, which in this case is the message returned by `createAIRsponse()`.

SquidXTV: But technically, I could do this:
```java
.flatMap(any -> message.pin());
```
This way I can remove the method as well

Taz03: Sry i mean `message::pin`

SquidXTV: well that doesnt work, it gives error, because it tries to give `any` as the method parameter:
image

PR #1114, opened 2024-05-14
Dynamically create and delete voice channels based on activity

### LINKED ISSUE #1113: Dynamically create and delete voice channels based on activity
Lately, the community has been more active in both text channels and in voice channels, and a noticeable problem that community members have stumbled upon is the fact that voice channels get frequently full, to the point where multiple discussion topics tend to be discussed in the same channel, leading to people moving to off-topic voice channels (like `Gaming`, `Chit Chat`, or even `java-21-launch`) which does in fact work for the most part but it clearly highlights a demand for more voice channels of the same topic.

One way to counteract this issue is to dynamically create and delete voice channels with the same name based on the amount of members connected to a voice channel with a particular topic. The overall goal should be to **always** have one empty voice channel per topic. For instance, if a member connects to a voice channel called `Support/Studying Room`, a new voice channel with the same name and its number concatenated should be created right under the original text channel. If any member disconnects from a particular voice channel and it is detected that it is now empty, it should be deleted.

**Additional context**
Inspired from [this](https://discord.com/channels/272761734820003841/480829495486447627/1239295694297104505) server suggestion from the community.

christolis:
Closes #1113.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/Features.java
@@ -20,6 +20,7 @@ import org.togetherjava.tjbot.features.chatgpt.ChatGptService;
 import org.togetherjava.tjbot.features.code.CodeMessageAutoDetection;
 import org.togetherjava.tjbot.features.code.CodeMessageHandler;
 import org.togetherjava.tjbot.features.code.CodeMessageManualDetection;
+import org.togetherjava.tjbot.features.dynamicvc.DynamicVoiceListener;
 import org.togetherjava.tjbot.features.filesharing.FileSharingMessageListener;
 import org.togetherjava.tjbot.features.github.GitHubCommand;
 import org.togetherjava.tjbot.features.github.GitHubReference;
@@ -145,6 +146,9 @@ public class Features {
         features.add(new CodeMessageManualDetection(codeMessageHandler));
         features.add(new SlashCommandEducator());
         features.add(new PinnedNotificationRemover(config));
+
+        // Voice receivers
+        features.add(new DynamicVoiceListener());
 
         // Event receivers
         features.add(new RejoinModerationRoleListener(actionsStore, config));
--- a/application/src/main/java/org/togetherjava/tjbot/features/system/BotCore.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/system/BotCore.java
@@ -2,6 +2,12 @@ package org.togetherjava.tjbot.features.system;
 
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.channel.Channel;
+import net.dv8tion.jda.api.entities.channel.unions.AudioChannelUnion;
+import net.dv8tion.jda.api.events.guild.voice.GuildVoiceDeafenEvent;
+import net.dv8tion.jda.api.events.guild.voice.GuildVoiceMuteEvent;
+import net.dv8tion.jda.api.events.guild.voice.GuildVoiceStreamEvent;
+import net.dv8tion.jda.api.events.guild.voice.GuildVoiceUpdateEvent;
+import net.dv8tion.jda.api.events.guild.voice.GuildVoiceVideoEvent;
 import net.dv8tion.jda.api.events.interaction.ModalInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.CommandAutoCompleteInteractionEvent;
 import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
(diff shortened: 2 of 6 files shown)


tj-wazei:
I haven't looked at the code yet but I have some suggestions for the functionality. 

* We should have a new Voice Chat that's more descriptive on "hey you're about to create a new voice channel". Maybe something like "New Voice Channel" instead of using Chit Chat. 
* The new room should be based on the "creators" name e.g. "Christolis's Room" and it should automatically move you from the "New Voice Channel" to "Christolis's Room" (people can rename after as needed)
* We should have a cleanup task e.g. if a voice channel has been empty for 5 minutes then it should be deleted. 

Importantly, the existing channels should be protected from the changes that get implemented here. 

This is just an IMO - we should wait for further feedback first.

ankitsmt211:
I love @tj-wazei suggestions, we shouldnt delete channels immediately and one more thing to add. We also wanna cap channel count to a certain limit. For example we can't have more than say 3 channels per category for now. We can put this limit in configuration so it's easily tweaked without making changes to code by maintainers.

SquidXTV (changes requested):

ZahirHumphries:
Going to help work on this.

application/src/main/java/org/togetherjava/tjbot/features/system/BotCore.java
```diff
@@ -238,6 +255,76 @@ public void onMessageDelete(final MessageDeleteEvent event) {
...
+     *         optional
+     */
+    private Optional<Channel> calculateSubscribeTarget(@Nullable AudioChannelUnion joinChannel,
+            @Nullable AudioChannelUnion leftChannel) {
+        if (joinChannel != null) {
+            return Optional.of(joinChannel);
+        }
+
+        return Optional.ofNullable(leftChannel);
+    }
```
Zabuzard, line 291: im not sure i follow what this does. who is join channel and who is left channel?

application/src/main/java/org/togetherjava/tjbot/features/Features.java
```diff
@@ -151,6 +152,9 @@ public static Collection<Feature> createFeatures(JDA jda, Database database, Con
         features.add(new SlashCommandEducator());
         features.add(new PinnedNotificationRemover(config));
 
+        // Voice receivers
+        features.add(new DynamicVoiceListener(config));
```
Zabuzard: id rename the feature to `DynamicVoiceChannelListener`, i.e. add `Channel` to signal that it deals with managing channels, not voice itself

Zabuzard:
something that is problematic is that we cannot store message content outside of discord (for legal/privacy reasons).
neither can you offer some form of export feature that sends it to the users, for example as file.

message content needs to stay inside discord. so if we want to archive something, that archive must be in discord and available to the users in a way that they can for example still delete their own messages if desired.

i dont really see a practical solution to that. so best is probably to get rid of the channels id assume

SquidXTV:
> i dont really see a practical solution to that. so best is probably to get rid of the channels id assume

I would also say that there isn't a need for that in practice. Like check the current channel's chat history. They mostly consist of convos between people using mic vs people just listening and writing messages instead, so archiving that is mostly out of context.

application/src/main/java/org/togetherjava/tjbot/features/system/BotCore.java
christolis: I have added more clarity and explained things a bit more with JavaDocs. Let me know what you think.

application/src/main/java/org/togetherjava/tjbot/features/Features.java
christolis: This has been done!

christolis:
I have added warnings to ephemeral voice chats, it looks like this:
image

application/src/main/java/org/togetherjava/tjbot/features/dynamicvc/DynamicVoiceChannelListener.java
```diff
@@ -0,0 +1,281 @@
...
+    private final Map<String, Queue<GuildVoiceUpdateEvent>> eventQueues = new HashMap<>();
+
+    /** Map to track if an event queue is currently being processed for each channel topic. */
+    private final Map<String, AtomicBoolean> activeQueuesMap = new HashMap<>();
+
+    /** Boolean to track if events from all queues should be handled at a slower rate. */
+    private final AtomicBoolean voiceActivityCongestion = new AtomicBoolean(false);
+    private final Executor eventQueueExecutor =
+            CompletableFuture.delayedExecutor(1L, TimeUnit.SECONDS);
+    private static final int CONGESTION_THRESHOLD = 5;
```
tj-wazei, line 66: Could you explain `CONGESTION_THRESHOLD`? Thanks

application/src/main/java/org/togetherjava/tjbot/features/dynamicvc/DynamicVoiceChannelListener.java
```diff
@@ -0,0 +1,281 @@
...
+ * It is designed to handle events related to voice channel updates (e.g. when users join or leave
+ * voice channels). It dynamically creates or deletes voice channels to ensure there is always
+ * <i>one</i> available empty channel for users to join, and removes duplicate empty channels to
+ * avoid clutter.
+ * <p>
+ * This feature relies on configurations provided at initialization to determine the patterns for
+ * channel names it should manage. The configuration is expected to provide a list of regular
+ * expression patterns for these channel names.
+ */
+public class DynamicVoiceChannelListener extends VoiceReceiverAdapter {
```
tj-wazei, line 48: I'm kinda struggling to follow the code here, mind adding more docs and or renaming methods to be less generic?

PR #1123, opened 2024-05-25
fixed eol config in gitattributes

Taz03:


SquidXTV (dismissed):

.gitattributes
```diff
@@ -1 +1,2 @@
 * text eol=lf
```
marko-radosavljevic, line 1:
```suggestion
* text=auto
```
This is probably better.

.gitattributes
```diff
@@ -1 +1,2 @@
...
+* binary text=auto
```
marko-radosavljevic, line 2: I don't understand what this means? Binary text? You can remove this line.

You should generally specify which files are binary, and which files are text manually, if git has problems with them. Otherwise, binary files can be corrupted.

For the most part, git is smart enough to know which is which, you may add manually files without extensions like gradlew, or other exceptions that git doesn't handle well (you probably noticed some of them already).

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

This guide has an example. You can add file extensions, or individual files, and set them to binary.

.gitattributes
Taz03: afaik this was intentionally set to lf so that all files are lf, otherwise it may differ?

.gitattributes
Taz03: changed it to `* binary` which does nothing for binary files

Taz03:
we still want all files LF so the config was fine, except we don't want to edit binary files because it'll make then corrupt. And also want .bat files to be CRLF.

And both of those changes are done in this pr

.gitattributes
marko-radosavljevic: > text
> 
> This attribute marks the path as a text file, which enables end-of-line conversion: When a matching file is added to the index, the file’s line endings are normalized to LF in the index. Conversely, when the file is copied from the index to the working directory, its line endings may be converted from LF to CRLF depending on the eol attribute, the Git config, and the platform (see explanation of eol below).
> 

Text does normalization to LF automatically.

> Set to string value "auto"
> 
> When text is set to "auto", Git decides by itself whether the file is text or binary. If it is text and the file was not already in Git with CRLF endings, line endings are converted on checkin and checkout as described above. Otherwise, no conversion is done on checkin or checkout.
> 

Auto automatically decides if text or binary, and performs normalization to LF automatically on text files, unless it's already in git with CRLF ending.

So basically, this line is everything you need, if some binaries are not recognized, or file has no extension, you can manually specify that it is binary with this pattern `*.jpg binary`

here is the example autoconfig:
```
# This will do normalization to LF on index (staging area) 
* text=auto

*.java   text
*.html   text
*.css    text

# Explicit for linux files
*.sh     text eol=lf

# Explicit stuff for win files
*.bat    text eol=crlf

# Just to make sure git doesn't corrupt our binaries accidentally, it's a good practice.
*.gif binary
*.jpg binary
*.jpeg binary
*.png binary

# No extensions, so we set them manually
/gradlew           text eol=lf
```

.gitattributes
marko-radosavljevic: Yeah, I don't think that line does what we want to do. It sets every file to binary, which is wrong. Setting text to auto should automatically ignore binaries.

And if there are some binaries that git is not smart enough to recognize, you can manually add them. See the example above.

marko-radosavljevic:
Yeah, auto normalization to LF is good @surajkumar, no worries.  :relaxed: :heart: 
But you have to configure it properly. You can't force everything to LF, some win files won't work properly, and also git is not smart enough for every type of binary, so you might need to be explicit what to with each binary. Also, if there is no file extension, you want to be explicit what you want with it.

> text
> 
> This attribute marks the path as a text file, which enables end-of-line conversion: When a matching file is added to the index, the file’s line endings are normalized to LF in the index. Conversely, when the file is copied from the index to the working directory, its line endings may be converted from LF to CRLF depending on the eol attribute, the Git config, and the platform (see explanation of eol below).
> 

Text does normalization to LF automatically.

> Set to string value "auto"
> 
> When text is set to "auto", Git decides by itself whether the file is text or binary. If it is text and the file was not already in Git with CRLF endings, line endings are converted on checkin and checkout as described above. Otherwise, no conversion is done on checkin or checkout.
> 

Auto automatically decides if text or binary, and performs normalization to LF automatically on text files, unless it's already in git with CRLF ending.

So basically, this line is everything you need, if some binaries are not recognized, or file has no extension, you can manually specify that it is binary with this pattern `*.jpg binary`


Example config, since it's smarter to be explicit, and sometimes it is needed:

```
# This will do normalization to LF on index (staging area) 
* text=auto

*.java   text
*.html   text
*.css    text

# Explicit for linux files
*.sh     text eol=lf

# Explicit stuff for win files
*.bat    text eol=crlf

# Just to make sure git doesn't corrupt our binaries accidentally, it's a good practice.
*.gif binary
*.jpg binary
*.jpeg binary
*.png binary

# No extensions, so we set them manually
/gradlew           text eol=lf
```

.gitattributes
Taz03:
```
*.java   text
*.html   text
*.css    text
```

do we explicitly need to add this

christolis (approved):

marko-radosavljevic (approved):
That was just an example, but sure.. :D 

If it works correctly, without you having to add any extra rules, I'm fine with it.

Of course test it, and apply it to everything with:
```
git rm -rf --cached .
git reset --hard HEAD```

PR #1137, opened 2024-07-02
inconsistency in titles generated by GPT #fixed

### LINKED ISSUE #1134: inconsistency in titles generated by GPT
**Describe the bug**

image

as shown in the screenshot, title is quoted sometimes when question is transfered via `TransferQuestionCommand`, mainly due to inconsistency in response from GPT. 
**Expected behavior**

Title shouldnt have any quotation marks

**Additional context**

```
 String chatGptPrompt =
                "Summarize the following text into a concise title or heading not more than 4-5 words, remove quotations if any: %s"
                    .formatted(originalMessage);
        Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
        String title = chatGptResponse.orElse(createTitle(originalMessage));
        if (title.length() > TITLE_MAX_LENGTH) {
            title = title.substring(0, TITLE_MAX_LENGTH);
        }
```
Above piece of code is responsible for generating title in `TransferQuestionCommand#onMessageContext`, to solve this issue instead of directly assigning generated text from GPT as title, it should be santized by removing quotations if present.

alisha2501:
#1134 fixed

## Pull-request

- [x] I have created a relating [issue][new_issue].

### Changes

- [x] Existing code
- [x] New feature

Closes Issue: NaN

## Description

This pull request addresses the issue where titles generated by the GPT model occasionally include quotation marks. To ensure consistency and clean titles, this PR includes changes to sanitize the generated title by removing any leading or trailing quotation marks (both single and double quotes). Additionally, the title is truncated if it exceeds the maximum allowed length (TITLE_MAX_LENGTH)

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -96,14 +96,17 @@ public final class TransferQuestionCommand extends BotCommandAdapter
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.getFirst();
+        final int TITLE_MAX_LENGTH = 50;
         String chatGptPrompt =
                 "Summarize the following text into a concise title or heading not more than 4-5 words, remove quotations if any: %s"
                     .formatted(originalMessage);
         Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
         String title = chatGptResponse.orElse(createTitle(originalMessage));
+        title = title.replaceAll("^\"|\"$", "").replaceAll("^'|'$", "");
         if (title.length() > TITLE_MAX_LENGTH) {
             title = title.substring(0, TITLE_MAX_LENGTH);
         }
+
 
         TextInput modalTitle = TextInput.create(MODAL_TITLE_ID, "Title", TextInputStyle.SHORT)
             .setMaxLength(TITLE_MAX_LENGTH)
@@ -176,8 +179,7 @@ public final class TransferQuestionCommand extends BotCommandAdapter
             .retrieveUserById(authorId)
             .flatMap(fetchedUser -> createForumPost(event, fetchedUser))
             .flatMap(createdForumPost -> dmUser(event.getChannel(), createdForumPost,
-                    event.getGuild())
-                .and(sendMessageToTransferrer.apply(createdForumPost)))
+                    event.getGuild()).and(sendMessageToTransferrer.apply(createdForumPost)))
             .flatMap(dmSent -> deleteOriginalMessage(event.getJDA(), channelId, messageId))
             .queue();
     }
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
@@ -1,5 +1,7 @@
 package org.togetherjava.tjbot.features.moderation;
 
+
+import net.dv8tion.jda.api.JDAConstants;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -96,15 +96,18 @@ public void onMessageContext(MessageContextInteractionEvent event) {
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.getFirst();
+        final int TITLE_MAX_LENGTH = 50;
         String chatGptPrompt =
                 "Summarize the following text into a concise title or heading not more than 4-5 words, remove quotations if any: %s"
                     .formatted(originalMessage);
         Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
         String title = chatGptResponse.orElse(createTitle(originalMessage));
+        title = title.replaceAll("^\"|\"$", "").replaceAll("^'|'$", "");
```
Taz03, line 105:
```suggestion
        title = title.replaceAll("^['\"]|['\"]$", "");
```
we can maybe merge the two replaceAll calls like this

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
SquidXTV: do we want that though? the logic is kinda different now.
Imagine a title like this: `"!!" in 'Kotlin'` or similar

SquidXTV, line 105: Also prefer putting the regex in a `Pattern` field to get rid of the initialization of the regex each call

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -96,15 +96,18 @@ public void onMessageContext(MessageContextInteractionEvent event) {
         String originalChannelId = event.getTarget().getChannel().getId();
         String authorId = event.getTarget().getAuthor().getId();
         String mostCommonTag = tags.getFirst();
+        final int TITLE_MAX_LENGTH = 50;
```
SquidXTV, line 99: This constant seems pretty random and I am sure there is a JDA constant already, that you should use instead.

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
Taz03: it should be removed since its already defined at
https://github.com/Together-Java/TJ-Bot/blob/0b908d82c8672a9dcbb196525b6b60dcedfdebd1/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java#L60

alisha2501: I will remove the mentioned line. I didn't check the code thoroughly.

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
```diff
@@ -84,6 +86,32 @@ public TransferQuestionCommand(Config config, ChatGptService chatGptService) {
...
+        // Define regex patterns for leading and trailing quotation marks
+        private static final Pattern LEADING_TRAILING_QUOTES = Pattern.compile("^['\"]|['\"]$");
+
+        public static String generateTitle(String originalMessage, ChatGptService chatGptService) {
+            String chatGptPrompt =
+                    "Summarize the following text into a concise title or heading not more than 4-5 words, remove quotations if any: %s"
+                        .formatted(originalMessage);
+            Optional<String> chatGptResponse = chatGptService.ask(chatGptPrompt, "");
+            String title = chatGptResponse.orElse(createTitle(originalMessage));
+            title = LEADING_TRAILING_QUOTES.matcher(title).replaceAll("");
```
surajkumar, line 101: Problem is that you're also removing intentional quotations. 

Consider the following example:
`"hello" is being printed but not "hello world"` becomes `hello" is being printed but not "hello world` 

You should probably let ChatGPT handle this entirely and update the prompt to ensure that and test it does it the way we intend it to.

surajkumar (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java
surajkumar: In sense, you're really just doing the following on-top of ChatGPT:
```
if(str.startsWith("\"") && str.endsWith("\"")) {
  str.substring(1, str.length - 1);
}
```

surajkumar: `"'hello world\""` this is also now just becoming `hello world`

ankitsmt211: We tried to deal at prompt level, unfortunately GPT is very inconsistent and sometimes would end up adding quotations regardless. The best possible way forward is to let GPT do it's thing and then like suraj mentioned look for titles that starting with quotation marks and only remove those that way we handle the edge case.

Zabuzard:
closing this for now as it seems to be inactive. the code in question also received some updates recently whose merge conflicts need to get resolved.
and yes, this should be tackled via text-code and not on the chatgpt level 👍

PR #1159, opened 2024-08-29
added help stats command

vishv843:
image
These are some of the stats I have added for now


ankitsmt211:
@vishv843 there's been a some sort of confusion, just going to rewrite the comment for visibility. The goal was to use database table `help_threads` to showcase these stats. That was the main reason we started storing some extra data per thread, if you look at the description for this issue carefully
Screenshot 2024-08-31 at 5 48 18 PM

these are the fields that are now stored as well per thread and can be utilized to showcase stats for help-threads.

However it seems you're pulling data from discord itself, which is not something we want.

Zabuzard (changes requested):
as anki explained, the data should be stored and gathered from the DB in constant time instead of querying discords api for the stats.
sorry for the confusion

christolis:
@Zabuzard Should we close this pull request?

Zabuzard:
yeah, we will close this for the time being. feel free to reopen it at any time and continue work on it if desired 😃

PR #1160, opened 2024-08-31
Bugfix/npe help thread close

### LINKED ISSUE #1157: NullPointer Exception
**Describe the bug**

Something off with the routine, 
[org.togetherjava.tjbot.features.help.MarkHelpThreadCloseInDBRoutine](https://github.com/Together-Java/TJ-Bot/blob/master/application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java)
WARN
unable to mark thread as close with id :1256940271854157855
java.lang.NullPointerException: Cannot invoke "net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel.getIdLong()" because "threadChannel" is null
at org.togetherjava.tjbot.features.help.HelpThreadLifecycleListener.handleArchiveStatus(HelpThreadLifecycleListener.java:86)
at org.togetherjava.tjbot.features.help.MarkHelpThreadCloseInDBRoutine.lambda$updateTicketStatus$1(MarkHelpThreadCloseInDBRoutine.java:66)
at java.base/java.lang.Iterable.forEach(Iterable.java:75)
at org.togetherjava.tjbot.features.help.MarkHelpThreadCloseInDBRoutine.updateTicketStatus(MarkHelpThreadCloseInDBRoutine.java:63)
at org.togetherjava.tjbot.features.help.MarkHelpThreadCloseInDBRoutine.runRoutine(MarkHelpThreadCloseInDBRoutine.java:49)
at org.togetherjava.tjbot.features.system.BotCore.lambda$scheduleRoutines$7(BotCore.java:199)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
at java.base/jav...

**Expected behavior**

**To Reproduce**

**Screenshots**

**Additional context**

ankitsmt211:
resolve #1157

Before
* when running archive help thread in DB routine, it would call a method that uses threadChannel inside a try-catch block
* It would catch Exception outside the method, even tho NPE is triggered inside it.

After
* refactor method to check of if threadChannel is NULL(if a channel is deleted when routine is triggered)
* Removes try-catch block
* conditional logic to handle if thread is NULL

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
@@ -48,6 +48,7 @@ public final class MarkHelpThreadCloseInDBRoutine implements Routine {
         updateTicketStatus(jda);
     }
 
+    @SuppressWarnings("java:S1181")
     private void updateTicketStatus(JDA jda) {
         Instant now = Instant.now();
         Instant threeDaysAgo = now.minus(3, ChronoUnit.DAYS);
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
@@ -1,7 +1,6 @@
 package org.togetherjava.tjbot.features.help;
 
 import net.dv8tion.jda.api.JDA;
-import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -59,6 +58,7 @@ public final class MarkHelpThreadCloseInDBRoutine implements Routine {
             .map(HelpThreadsRecord::getChannelId)
             .toList());
 
-        threadIdsToClose.forEach(id -> helpThreadLifecycleListener.handleArchiveStatus(now, id, jda));
+        threadIdsToClose
+            .forEach(id -> helpThreadLifecycleListener.handleArchiveStatus(now, id, jda));
     }
 }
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
```diff
@@ -59,14 +54,7 @@ private void updateTicketStatus(JDA jda) {
             .map(HelpThreadsRecord::getChannelId)
             .toList());
 
-
-        threadIdsToClose.forEach(id -> {
-            try {
```
Zabuzard, line 64: Please keep the try catch so that when one thread has a problem (whatever it might be), other threads arent stopped but still processed. every thread should get it's chance for closing.
the rest of the PR is perfect, thanks.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/MarkHelpThreadCloseInDBRoutine.java
ankitsmt211: @Zabuzard I don't understand, the only reason i earlier used try-catch was due to thread at times being nullable. Is it possible for something else to go wrong here that i can't anticipate?

Now we're handling the case where thread no longer exists in a much better way, imo exceptions should only reserved for something unexpected no?

Even if we do catch some exception is it something specific or just a general catch block?

Zabuzard: u should catch any exception happening here and log it. or collect it and rethrow it outside, after the `forEach`.

the reason is that if a thread cant be closed for any unknown reason, that should not stop the code trying to close the other threads. else they will be dangling.
like, imagine the code wants to cleanup 20 threads. thread 5 fails. what about the other 15 threads? they should also be attempted. without try-catch or similar, all 15 threads would be left opened bc the code crashed at thread 5.

so either try-catch-log or try-catch-collect+rethrow later

Zabuzard (approved):

marko-radosavljevic (approved):

Issue #1188, opened 2024-10-23
context command to pin messages in respective thread


ankitsmt211:
**Is your feature request related to a problem? Please describe.**

A suggestion from community to let OP of thread in `questions` to be able to pin helpful answers

**Describe the solution you'd like**

We can build a context menu command something like we already have [Transfer Feature] (application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java) .
It will show up in message context menu, upon clicking that command based on whether user is OP it will either pin the message in thread or reply with an ephermal message that they're not allowed to use that command.

**Additional context**

* Only OP should be able to use the command
* Put a max limit on number of pins (maybe 10? idk something that's suitable)

Zabuzard:
mh, i have mixed feelings about this one.
i think in practice it might be better to just give users the pin permission for help threads so they can use the discord native ui.

this will also open it up to users who arent OP but i honestly dont think this would cause issues in practice

PR #1194, opened 2024-11-03
context command to pin messages in respective thread #1188

andrew-09:
https://github.com/Together-Java/TJ-Bot/issues/1188

Will give a warning message when the message is not in a thread, the user invoker is not the owner of the thread, or the number of pinned answers exceeds 10

image

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
@@ -2,10 +2,13 @@ package org.togetherjava.tjbot.features.help;
 
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.User;
+import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
 import net.dv8tion.jda.api.interactions.commands.build.Commands;
 
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.config.HelpSystemConfig;
 import org.togetherjava.tjbot.features.BotCommandAdapter;
 import org.togetherjava.tjbot.features.CommandVisibility;
 import org.togetherjava.tjbot.features.MessageContextCommand;
@@ -13,10 +16,12 @@ import org.togetherjava.tjbot.features.MessageContextCommand;
 public final class PinAnswerCommand extends BotCommandAdapter implements MessageContextCommand {
     private static final String COMMAND_NAME = "pin-answer";
     private static final int MAX_PINNED_ANSWERS = 10;
-    private int count = 0;
+    private final String helpForumPattern;
 
-    public PinAnswerCommand() {
+    public PinAnswerCommand(Config config) {
         super(Commands.message(COMMAND_NAME), CommandVisibility.GUILD);
+        HelpSystemConfig helpConfig = config.getHelpSystem();
+        helpForumPattern = helpConfig.getHelpForumPattern();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,73 @@
...
+            replyMaxPinsReached(event);
+            return;
+        }
+
+        pinMessage(event, originalMessage);
+    }
+
+    private boolean isThreadCreator(User user, ThreadChannel thread) {
+        return user.getIdLong() == thread.getOwnerIdLong();
+    }
```
ankitsmt211, line 47: just a nitpick, `threadOwner` sounds a bit better to me. But it's no worries if you wanna keep it as it is.

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (!isThreadCreator(commandInvoker, threadChannel)) {
+            replyNotThreadCreator(event);
+            return;
+        }
+
+        if (count >= MAX_PINNED_ANSWERS) {
+            replyMaxPinsReached(event);
+            return;
+        }
```
ankitsmt211, line 40: `count` might not be reliable to keep track of pinned threads so far, 
* what happens when bot is offline for whatever reasons?
* what about if a mod pins a new message in that channel that's not done via this command?

So in this case it's better to rely on discord to stay upto date with current state of thread
[retrievePinnedMessages()](https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/entities/channel/middleman/MessageChannel.html#retrievePinnedMessages()) might be more helpful. Get the list, find out and then verify if you can pin a new message or not.

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,73 @@
...
+            count++;
+            event.reply("Answer pinned successfully! Pinned answers: " + count)
+                .setEphemeral(true)
+                .queue();
+        }, failure -> event.reply("Failed to pin the answer.").setEphemeral(true).queue());
+    }
+
+    private void replyNotInThread(MessageContextInteractionEvent event) {
+        event.reply("This message is not in a thread.").setEphemeral(true).queue();
+    }
```
ankitsmt211, line 60: perhaps we can improve this message for clarity...something like `This command can only be used in threads`

Although i think we should be able to limit this `questions` channel only via discord directly no ? @Zabuzard

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private void replyNotThreadCreator(MessageContextInteractionEvent event) {
+        event.reply("You are not the thread creator and cannot pin answers here.")
+            .setEphemeral(true)
```
ankitsmt211, line 64: Same here if you chose to go thread owner route, "Only thread owners/thread creatros can use this command"

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,73 @@
...
+            .queue();
+    }
+
+    private void replyMaxPinsReached(MessageContextInteractionEvent event) {
+        event.reply("Maximum pinned answers (" + MAX_PINNED_ANSWERS + ") reached.")
+            .setEphemeral(true)
```
ankitsmt211, line 70: For a better UX, user should also know what to do in order to avoid this. 
"You've reached a maximum pinned limit for threads, if you wish to pin more messages please remove a few existing pinned messages"

ankitsmt211 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,87 @@
...
+        event.reply("This command can only be used in threads").setEphemeral(true).queue();
+    }
+
+    private void replyNotInQuestionsChannel(MessageContextInteractionEvent event) {
+        event.reply("Answers can only be pinned in the questions channel")
+            .setEphemeral(true)
+            .queue();
+    }
+
+    private void replyNotThreadCreator(MessageContextInteractionEvent event) {
```
ankitsmt211, line 77: for consistency purposes, can we also change this to `replyNotThreadOwner`

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private void replyNotInThread(MessageContextInteractionEvent event) {
+        event.reply("This command can only be used in threads").setEphemeral(true).queue();
+    }
+
+    private void replyNotInQuestionsChannel(MessageContextInteractionEvent event) {
+        event.reply("Answers can only be pinned in the questions channel")
+            .setEphemeral(true)
+            .queue();
```
ankitsmt211, line 74: seems like there's been some sort of confusion, you don't really need `replyNotInQuestionsChannel`.
`replyNotInThread` is enough already to trigger when command is used outside of `questions`.

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,87 @@
...
+
+        if (!(originalMessage.getChannel() instanceof ThreadChannel threadChannel)) {
+            replyNotInThread(event);
+            return;
+        }
+
+        if (!(threadChannel.getParentChannel() instanceof ForumChannel forumChannel)
+                || !forumChannel.getName().equalsIgnoreCase(helpForumPattern)) {
+            replyNotInQuestionsChannel(event);
+            return;
```
ankitsmt211, line 40: like i mentioned in other recent comment, only `replyNotInThread` should be triggered when channel pattern does not match. If you think about it, we only have a single `happy` case that is command is used in `questions`. We don't really care about other cases so we use `replyNotInThread` for the rest of stuff.

ankitsmt211 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
andrew-09: What if its used in a thread, but in the 'projects' channel for example?

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
andrew-09: Oh right, because everything in 'questions' channel is only going to contain threads. There will be no non-threads in the 'questions' channel. That makes sense now.

andrew-09: Or I guess I can change the replyNotInThread reply to "This command can only be used in threads in the questions channel" to solve both cases.

ankitsmt211: > Or I guess I can change the replyNotInThread reply to "This command can only be used in threads in the questions channel" to solve both cases.

Yep you're right this gives more context on what went wrong than just "not in thread".

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,81 @@
...
+        Message originalMessage = event.getTarget();
+        User commandInvoker = event.getUser();
+
+        if (!(originalMessage.getChannel() instanceof ThreadChannel threadChannel)) {
+            replyNotInThread(event);
+            return;
+        }
+
+        if (!(threadChannel.getParentChannel() instanceof ForumChannel forumChannel)
+                || !forumChannel.getName().equalsIgnoreCase(helpForumPattern)) {
```
Taz03, line 38: use matches not equals

Taz03, line 38: can this all not be merged into one single if

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private boolean threadOwner(User user, ThreadChannel thread) {
+        return user.getIdLong() == thread.getOwnerIdLong();
+    }
+
+    private void pinMessage(MessageContextInteractionEvent event, Message message) {
+        message.pin()
+            .queue(success -> event.reply("Answer pinned successfully!").setEphemeral(true).queue(),
+                    failure -> event.reply("Failed to pin the answer.").setEphemeral(true).queue());
```
Taz03, line 64: you dont need to handle failure case

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/PinAnswerCommand.java
andrew-09: Right, good catch on that

Taz03 (approved):

ankitsmt211 (dismissed):
good work 👍

ankitsmt211:
@andrew-09 It seems that we can't bring this into TJ bot due to limitations of discord when it comes to content commands. I wish we could have figured that out sooner, we appreciate the efforts you put into this . Currently closing this PR, perhaps we can use this in future.

PR #1208, opened 2024-12-03
Optimize auto prune routine

SquidXTV:
# About
In order to keep helper roles below 250 users, we have a routine that removes a role from a more inactive user.
The old implementation used `Guild#findMembersWithRoles` to filter for them, which surprisingly requests **all**
members first and does the filtering on the client. This happens for each role (about 17). So we basically query the same
content multiple times. The new implementation requests **all** members beforehand and then reusing it to prune roles if needed.

This is related to the issue about a `TimeoutException` on the member chunk requests when requesting **all** members. This doesn't mean it is going to fix that issue, but it is still an optimization.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/AutoPruneHelperRoutine.java
@@ -84,23 +84,36 @@ public final class AutoPruneHelperRoutine implements Routine {
     }
 
     private void pruneForGuild(Guild guild) {
-        Instant now = Instant.now();
         TextChannel selectRoleChannel = getSelectRolesChannelOptional(guild.getJDA()).orElse(null);
+        guild.loadMembers()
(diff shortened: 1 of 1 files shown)


Zabuzard (changes requested):
Can you change the logic to fallback on the previous behavior or similar when loading fails?

I want to prevent that when "loading all users" fails for whatever reason that this means it fails for all categories.

Currently, if I get this right, when there is a problem with one category it will still attempt it for the other categories individually.

SquidXTV:
> Can you change the logic to fallback on the previous behavior or similar when loading fails?
> 
> I want to prevent that when "loading all users" fails for whatever reason that this means it fails for all categories.
> 
> Currently, if I get this right, when there is a problem with one category it will still attempt it for the other categories individually.

You sure that this is needed?
Yeah, it might be possible to fail loading all users, but I don't see how using the current, old, _bad_ code as a fallback makes sense.

Zabuzard (approved):

PR #1213, opened 2025-02-18
Improved Scam detector logic

Zabuzard:
Improved the automatic scam detector logic. Added more real scam examples (DONT CLICK on these links, they are dangerous).
Further, added dollar-sign detection as extra sign of scam. Also added `@here` to the `pingsEveryone` logic.

## Logic Review ##

Previous had a shortcut for:
```java
results.pingsEveryone && results.containsSuspiciousKeyword && results.hasUrl
```
and otherwise:
```java
results.containsSuspiciousKeyword && results.hasSuspiciousUrl
```

The new logic relaxed the shortcut heavily (since an everyone-ping is a clear sign for scam):
```java
results.pingsEveryone && (results.containsSuspiciousKeyword || results.hasUrl || results.containsDollarSign)
```
and otherwise requires 2 out of the following 3 signs:
```java
results.containsSuspiciousKeyword, results.hasSuspiciousUrl, results.containsDollarSign
```

## Config ##

When releasing this, make sure to apply the changed config as well. `scamBlocker` received a lot of new keywords.


application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
```diff
@@ -40,25 +41,36 @@ public boolean isScam(CharSequence message) {
...
             results.pingsEveryone = true;
         }
 
         if (!results.containsSuspiciousKeyword && containsSuspiciousKeyword(token)) {
             results.containsSuspiciousKeyword = true;
         }
 
+        if (!results.containsDollarSign && token.contains("$")) {
+            results.containsDollarSign = true;
+        }
```
SquidXTV, line 72: Is the first check in these ifs even needed? I don't really see a benefit other than short-circuiting the checks.

Zabuzard: probably not. i cant remember why they have been there previously already - perhaps some sonar?
dunno but i just left it 🤷

PR #1216, opened 2025-02-26
Scam Detector fix false positive

Zabuzard:
Our Scam Detector recently ran into its first false positive, a message like this:

> https://learn.microsoft.com/en-us/dotnet/csharp/fundamentali/types/anonymous-typesd

The reason it was flagged is because it has the word `learn` in it, which `contains("earn")`, one of our suspicious keywords.

This PR adds unit tests to cover false positives and fixes the above issue by allowing suspicious keywords to require a `startsWith` check instead of `contains` by starting with `^`. So `^earn` will not trigger on `learn` anymore.

Note, the logic for this was self-written. It would also be possible to make them a full regex pattern (which supports `^` as well), but I thought this would be a bit overkill and also annoying, since all other keywords would then need to be changed to `.*foo.*`.

## Config

When deploying, change the suspicious keyword `earn` in the config to `^earn`.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
@@ -116,11 +116,10 @@ public final class ScamDetector {
             .map(keyword -> keyword.toLowerCase(Locale.US))
             .anyMatch(keyword -> {
                 // Simple regex-inspired syntax "^foo"
-                if (!keyword.isEmpty() && keyword.charAt(0) == '^') {
+                if (startsWith(keyword, '^')) {
                     return preparedToken.startsWith(keyword.substring(1));
-                } else {
-                    return preparedToken.contains(keyword);
                 }
+                return preparedToken.contains(keyword);
             });
     }
 
@@ -145,6 +144,10 @@ public final class ScamDetector {
         }
 
         return false;
+    }
+
+    private static boolean startsWith(CharSequence text, char prefixToTest) {
+        return !text.isEmpty() && text.charAt(0) == prefixToTest;
     }
 
     private static class AnalyseResults {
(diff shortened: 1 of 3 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
```diff
@@ -114,7 +114,14 @@ private boolean containsSuspiciousKeyword(String token) {
         return config.getSuspiciousKeywords()
             .stream()
             .map(keyword -> keyword.toLowerCase(Locale.US))
-            .anyMatch(preparedToken::contains);
+            .anyMatch(keyword -> {
+                // Simple regex-inspired syntax "^foo"
+                if (!keyword.isEmpty() && keyword.charAt(0) == '^') {
+                    return preparedToken.startsWith(keyword.substring(1));
+                } else {
+                    return preparedToken.contains(keyword);
```
mariokhoury4, line 122: I believe that removing the else statement is cleaner and more readable. What do you think?

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
```diff
@@ -114,7 +114,14 @@ private boolean containsSuspiciousKeyword(String token) {
         return config.getSuspiciousKeywords()
             .stream()
             .map(keyword -> keyword.toLowerCase(Locale.US))
-            .anyMatch(preparedToken::contains);
+            .anyMatch(keyword -> {
+                // Simple regex-inspired syntax "^foo"
+                if (!keyword.isEmpty() && keyword.charAt(0) == '^') {
```
mariokhoury4, line 119: what do you think about `keyword.startsWith("^")` instead of `keyword.charAt(0) == '^'`?

Zabuzard: I started with `startsWith` but sonar suggested that it would be faster to do this instead.
I thought about it and decided to roll with it because this is executed on every single message, every single word, send in the entire server. So this function probably has the highest load in the entire bot, lmao.

I could perhaps wrap it in a helper function for readability. `startsWith(String, char)`

Zabuzard: Fixed :)

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
Zabuzard: Fixed :)

PR #1265, opened 2025-06-28
fix: Handle empty message content for suggestion threads

### LINKED ISSUE #1198: Server suggestion setup fails for a message without text
**Describe the bug**
The bot automatically sets up the up/down votes and a thread to discuss a suggestion.
It fails to do so if someone creates a new post in `#server-suggestions` without text.
For example: https://discord.com/channels/272761734820003841/480829495486447627/1305554470364643442

**Expected behavior**
It shouldn't fail, even for posts without text.

**To Reproduce**
1. Post a new message without text. (only an image for example)
2. See the error

**Additional context**
```
One of the EventListeners had an uncaught exception java.lang.IllegalArgumentException: Name may not be empty
  at net.dv8tion.jda.internal.utils.Checks.notEmpty(Checks.java:94)
  at net.dv8tion.jda.internal.entities.channel.mixin.attribute.IThreadContainerMixin.createThreadChannel(IThreadContainerMixin.java:69)
  at net.dv8tion.jda.internal.entities.ReceivedMessage.createThreadChannel(ReceivedMessage.java:946)
  at org.togetherjava.tjbot.features.basic.SuggestionsUpDownVoter.createThread(SuggestionsUpDownVoter.java:70)
  at org.togetherjava.tjbot.features.basic.SuggestionsUpDownVoter.onMessageReceived(SuggestionsUpDownVoter.java:52)
  at org.togetherjava.tjbot.features.system.BotCore.lambda$onMessageReceived$9(BotCore.java:221)
  at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
  at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
  at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.jav...
```

The bot tries to create a new thread and set the title of that thread using the text of the post:
```java
private static void createThread(Message message) {
    String title = message.getContentRaw();

    if (title.length() >= TITLE_MAX_LENGTH) {
        int lastWordEnd = title.lastIndexOf(' ', TITLE_MAX_LENGTH);

        if (lastWordEnd == -1) {
            lastWordEnd = TITLE_MAX_LENGTH;
        }

        title = title.substring(0, lastWordEnd);
    }

    message.createThreadChannel(title).queue();
}
```

It is missing checks if the text is empty.
What kind of default title would make sense?

billpapat:
Ensures a fallback thread title is provided when the original message content is empty. The thread title will now **default** to the message **author's nickname** in such cases, **preventing empty** or uninformative **thread titles**. Closes #1198

image


application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,26 @@ public void onMessageReceived(MessageReceivedEvent event) {
     }
 
     private static void createThread(Message message) {
-        String title = message.getContentRaw();
 
-        if (title.length() >= TITLE_MAX_LENGTH) {
-            int lastWordEnd = title.lastIndexOf(' ', TITLE_MAX_LENGTH);
+        String threadTitle;
+        String messageContent = message.getContentRaw();
```
christolis: References should have an immutable over a mutable state wherever possible and appropriate.

```suggestion
        final String threadTitle;
        final String messageContent = message.getContentRaw();
```

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,26 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
             if (lastWordEnd == -1) {
                 lastWordEnd = TITLE_MAX_LENGTH;
             }
 
-            title = title.substring(0, lastWordEnd);
+            threadTitle = messageContent.substring(0, lastWordEnd);
+        } else {
+
+            threadTitle = messageContent;
         }
```
Zabuzard: could u reorder the if-else slightly so that the big logic block is the last? makes it a bit easier to read imo. so
```java
if (messageContent.isEmpty()) {
  ...
} else if (messageContent.length() < TITLE_MAX_LENGTH) {
  ...
} else {
  ...
}
```

if you got some time i would even find it better to move this into a helper method. then u can also utilize early-return.

```java
private static String createThreadTitle(Message message) {
  String content = ...
  
  if (content.isEmpty()) {
    return ...
  }
  if (content.length() < ...) {
    return content;
  }
  
  ...
  return ...
}
```

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,26 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
     private static void createThread(Message message) {
-        String title = message.getContentRaw();
 
-        if (title.length() >= TITLE_MAX_LENGTH) {
-            int lastWordEnd = title.lastIndexOf(' ', TITLE_MAX_LENGTH);
+        String threadTitle;
+        String messageContent = message.getContentRaw();
+
+        if (messageContent.isEmpty()) {
+            threadTitle = message.getAuthor().getName();
```
Zabuzard: Can we go for not just the name but something more human readable. E.g. "Suggestion by John" or "Johns suggestion".

christolis (changes requested):
+ what @Zabuzard suggested. We should prioritize code readability.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
```diff
@@ -0,0 +1,38 @@
+
+package org.togetherjava.tjbot.features.basic;
+
+public record ThreadTitle(String value) {
```
christolis: There should be JavaDocs for this record.

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
```diff
@@ -0,0 +1,38 @@
...
+
+            threadTitle = value.substring(0, lastWordEnd);
+        } else {
+            threadTitle = value;
+        }
+
+        this.value = threadTitle;
+    }
+
+    public static ThreadTitle withFallback(String threadtitle, String fallback) {
```
christolis: Can name something like `threadtitle` as `primary`, and regardless, rename the original one to `threadTitle`.

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
Zabuzard: Mh. This class is too specific for the Suggestions-Voter feature to be put into its own class right into the `features/basic` package.

Either create a subpackage for the whole suggestions voting topic or move this back into the old class as `private record` 👍

Zabuzard: Im not sure if its worth making this a `record`. You are only using it in this one single spot. I think it might be better to just have this as simple `private static String createTitle(...)` method in your previous class and then utilize early-return to have the logic clean and readable without all that nesting :)

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,26 +54,10 @@ public void onMessageReceived(MessageReceivedEvent event) {
     }
 
     private static void createThread(Message message) {
+        ThreadTitle threadTitle = ThreadTitle.withFallback(message.getContentRaw(),
+                message.getAuthor().getName() + "'s suggestions");
```
Zabuzard: Singular I would say, `suggestion`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/basic/ThreadTitle.java
billpapat: Done

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -23,9 +23,9 @@
  */
 public final class SuggestionsUpDownVoter extends MessageReceiverAdapter {
     private static final Logger logger = LoggerFactory.getLogger(SuggestionsUpDownVoter.class);
-    private static final int TITLE_MAX_LENGTH = 60;
     private static final Emoji FALLBACK_UP_VOTE = Emoji.fromUnicode("👍");
     private static final Emoji FALLBACK_DOWN_VOTE = Emoji.fromUnicode("👎");
+    private static final int THREAD_TITLE_MAX_LENGTH = 60; // Define the max length
```
Zabuzard: that comment feels unecessary. id remove it

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,35 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+        message.createThreadChannel(threadTitle).queue();
+    }
 
-        if (title.length() >= TITLE_MAX_LENGTH) {
-            int lastWordEnd = title.lastIndexOf(' ', TITLE_MAX_LENGTH);
+    /**
+     * Generates a thread title, enforcing a maximum length of 60 characters. If an initial title
+     * exceeds this limit, it's truncated at the last word boundary before or at the 60-character
+     * mark to prevent cutting words mid-sentence. If no space is found, it truncates at 60
+     * characters. Uses a fallback title if the primary title is empty.
```
Zabuzard: If you write `60` here, this documentation will quickly get outdated when someone adjusts the constant but forgets to also adjust it down here.
You can link the value instead in the Javadoc: `maximum length of {@value #THREAD_TITLE_MAX_LENGTH}`. something like that, you can check it if you mousehover.

Imo the responsibilities for this method would be better if you give it the thread and it figures out how to build the title from that. I.e. imo the method should take `generateThreadTitle(Message message)` and extract title and fallback by itself.

I.e. a method named `generateThreadTitle` should be the one responsible for figuring out the proper title.

Further, I personally find this javadoc too implementation-specific. Perhaps just:

> Generates a title for the given message.

How it then does that is up to the method and imo the code is readable enough to understand it without being put into plain english sentences again 👍

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,33 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+    /**
+     * Generates a title for the given message. The maximum length of the title is
+     * {@value #THREAD_TITLE_MAX_LENGTH}.
+     *
+     * @param message The message for which to generate the title.
+     * @return The generated and truncated thread title.
+     */
+    private static String generateThreadTitle(Message message) {
+        String primaryTitle = message.getContentRaw();
+        String fallbackTitle = message.getAuthor().getName() + "'s suggestion";
```
Taz03: this will use their username rather than nickname, i think using their nickname is a better option

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,33 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+        String title = primaryTitle.isEmpty() ? fallbackTitle : primaryTitle;
```
Taz03, line 72: trim it before checking for isEmpty

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,33 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
 
-            if (lastWordEnd == -1) {
-                lastWordEnd = TITLE_MAX_LENGTH;
-            }
+        if (title.length() < THREAD_TITLE_MAX_LENGTH) {
```
Taz03: less than or equal to

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
```diff
@@ -55,19 +55,33 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
-            int lastWordEnd = title.lastIndexOf(' ', TITLE_MAX_LENGTH);
+    /**
+     * Generates a title for the given message. The maximum length of the title is
+     * {@value #THREAD_TITLE_MAX_LENGTH}.
+     *
+     * @param message The message for which to generate the title.
+     * @return The generated and truncated thread title.
+     */
+    private static String generateThreadTitle(Message message) {
+        String primaryTitle = message.getContentRaw();
```
Taz03: use [Message#getContentStripped](https://javadoc.io/doc/net.dv8tion/JDA/latest/net/dv8tion/jda/api/entities/Message.html#getContentStripped()) instead

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
Taz03: i think this branch is kinda useless?? not sure, if its not pls just close this conversation

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
billpapat: Done!

billpapat: It is needed to return the title as it is (<= THREAD_TITLE_MAX_LENGTH).

Zabuzard (changes requested):
could u add some screenshots to the PR please that showcase/demo the feature in practice? its always difficult to review these things when u cant see how it actually looks in practice. often one also spots improvement opportunities that way :)

application/src/main/java/org/togetherjava/tjbot/features/basic/SuggestionsUpDownVoter.java
Taz03: you are still not calling trim
```java
String primaryTitle = message.getContentStripped().trim();
```

Zabuzard: note that there is `isBlank()`. prefer that over `.trim().isEmpty()`

Zabuzard (approved):

Taz03 (approved):

Issue #1243, opened 2025-05-31
Allow to forward messages with images to media channels


Alathreon:
**Is your feature request related to a problem? Please describe.**

A media channel is a channel where you can only send messages with images.
But currently, if you use the new forward feature to forward a message with an image, i will fail, it won't notice the image.

**Describe the solution you'd like**

Allow to forward messages that contain an image to media channels.

**Describe alternatives you've considered**

Ignore the problem. Since it's niche, because you could copy an image instead, this issue could be ignored.

**Additional context**

https://support.discord.com/hc/en-us/articles/24640649961367-Message-Forwarding

billpapat:
Would like to work on it.

billpapat:
Would like to work on it.

PR #1266, opened 2025-06-28
Allow forwarding of media, to media channels

billpapat:
fix: Media forwarding #1243 

- **forwarded messages** that contain media are not deleted.
- Introduced a **1-second delay** to prevent the **unintended deletion** of subsequent forwarded comments.
- Improved **URL** pattern **recognition**.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
@@ -14,9 +14,6 @@ import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.features.MessageReceiverAdapter;
 
 import java.awt.Color;
-import java.time.Duration;
-import java.time.Instant;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
@@ -32,9 +29,6 @@ public final class MediaOnlyChannelListener extends MessageReceiverAdapter {
     private static final Pattern MEDIA_URL_PATTERN = Pattern.compile(
             ".*https?://\\S+\\.(png|jpe?g|gif|bmp|webp|mp4|mov|avi|webm|mp3|wav|ogg|youtube\\.com/|youtu\\.com|imgur\\.com/).*",
             Pattern.CASE_INSENSITIVE);
-
-    private final ConcurrentHashMap<Long, Instant> lastValidForwardedMediaMessageTime =
-            new ConcurrentHashMap<>();
 
     /**
      * Creates a MediaOnlyChannelListener to receive all message sent in MediaOnly channel.
@@ -53,22 +47,8 @@ public final class MediaOnlyChannelListener extends MessageReceiverAdapter {
         }
 
         Message message = event.getMessage();
-        long userId = event.getAuthor().getIdLong();
 
         if (!messageHasNoMediaAttached(message)) {
-            if (!message.getMessageSnapshots().isEmpty()) {
-                lastValidForwardedMediaMessageTime.put(userId, Instant.now());
-            }
-            return;
-        }
-
-        Instant lastForwardedMediaTime = lastValidForwardedMediaMessageTime.get(userId);
-        Duration gracePeriod = Duration.ofSeconds(1);
-
-        if (lastForwardedMediaTime != null
-                && Duration.between(lastForwardedMediaTime, Instant.now())
-                    .compareTo(gracePeriod) <= 0) {
-            lastValidForwardedMediaMessageTime.remove(userId);
             return;
         }
 


application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -25,6 +27,13 @@
  */
 public final class MediaOnlyChannelListener extends MessageReceiverAdapter {
 
+    private static final Pattern MEDIA_URL_PATTERN = Pattern.compile(
+            ".*https?://\\S+\\.(png|jpe?g|gif|bmp|webp|mp4|mov|avi|webm|mp3|wav|ogg|youtube\\.com/|youtu\\.com|imgur\\.com/).*",
+            Pattern.CASE_INSENSITIVE);
+
+    private final ConcurrentHashMap<Long, Long> lastValidForwardedMediaMessageTime =
+            new ConcurrentHashMap<>();
```
Zabuzard: please use `Instant` instead of raw `Long`s

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -45,19 +54,60 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+            return true;
+        }
+
+        return true;
     }
 
     private MessageCreateData createNotificationMessage(Message message) {
         String originalMessageContent = message.getContentRaw();
+        if (originalMessageContent.trim().isEmpty()) {
+            originalMessageContent = "(Original message had no visible text content)";
```
Zabuzard: that text is written with a developer as target audience, not the user. id rewrite it as `See image:` or similar.

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -45,19 +54,60 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+            for (MessageSnapshot snapshot : message.getMessageSnapshots()) {
+                if (!snapshot.getAttachments().isEmpty() || !snapshot.getEmbeds().isEmpty()
+                        || MEDIA_URL_PATTERN.matcher(snapshot.getContentRaw()).matches()) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        return true;
```
Zabuzard: im not happy with how this logic is currently structured.

the main structure should be:
```java
if (!is there any media?) {
  delete it
} else {
  forward it
}
```
the details of detecting media should be in a helper method. and the details of that cache handling as well.

i also dont quite get the point of the cache thing. what is it used for? and how does the UX look like if they run into it?
for the cache topic u should use high level classes like `Duration` and `Instant` instead.

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
billpapat: Done!

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
billpapat: Not sure what you mean by that. It checks for .isEmpty() meaning I cannot return "See image:" Since there is not one

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
billpapat: Semi-done!

billpapat: The bot previously failed to recognize forwarded media such as images, voice messages, and GIFs, processing them as plain text and subsequently deleting them. The updated version's handling of such media is illustrated in the accompanying images.

image

If the forwarded message contains no media (i.e., it's plain text), it will be automatically deleted, and the user will be notified as usual.

image

Zabuzard: okay. and what is the role of the grace period thing now?

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
Zabuzard: what i mean is that if this is a user facing text, its not written in a user friendly way.

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
billpapat: Imagine a user tries to forward a media file (like an image or video) to a "media-only" Discord channel. Discord might process upcoming (potential) additional comment added by the sender as message without content (text).

Without a grace period: If the bot immediately checks after the forwarded message, the added comment would be incorrectly deleted even though the user did forward an image or video to post media.

image
Discord's forwarding system allows for the addition of extra comment (text) that is being sent separately than the embed, meaning it will be deleted by the bot for no-media (we do not want forwarded media without context). That is what the grace period is for, waits 1 second after forwarded media (that have passed the messageHasNoMediaAttached) to allow for the comment to not be deleted.

Zabuzard: im having a super hard time following ur explanation. it sounds like ur saying there are multiple messages involved. when a user sends an image its a single message isnt it.
my advice is to please work more with screenshots or gifs/videos. make it visual and it will be easy to understand 👍

application/src/main/java/org/togetherjava/tjbot/features/mediaonly/MediaOnlyChannelListener.java
```diff
@@ -36,38 +47,76 @@ public MediaOnlyChannelListener(Config config) {
...
-        MessageEmbed originalMessageEmbed =
-                new EmbedBuilder().setDescription(originalMessageContent)
-                    .setColor(Color.ORANGE)
-                    .build();
+        MessageCreateBuilder messageBuilder = new MessageCreateBuilder();
+        messageBuilder.setContent(message.getAuthor().getAsMention()
+                + " Hey there, you posted a message without media (image, video, link) in a media-only channel. Please see the description of the channel for details and then repost with media attached, thanks 😀");
+
+        // Conditionally add the embed only if the original message content is NOT empty
+        if (!originalMessageContent.trim().isEmpty()) {
```
bowbahdoe, line 91: .isBlank is probably what you want here

PR #1268, opened 2025-06-30
Fix RSS Headers

pot:
For some reason oracle decides to block "bots" from automatically reading RSS feeds unless certain headers and user agents are provided. 🙄 

We are now sending `TJ-Bot/1.0 (+https://github.com/Together-Java/TJ-Bot)` as user-agent (conform with RFC for bots) and that works. Also checked their `robots.txt` and seems we are good to go.

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -173,12 +173,7 @@
             }
         ],
         "fallbackChannelPattern": "java-news-and-changes",
-        "pollIntervalInMinutes": 10,
-        "clientRequestHeaders": {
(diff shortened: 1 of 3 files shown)


christolis (approved):
LGTM

Zabuzard:
This is breaking ToS. Sorry, we cannot use this feed if they do not allow it to be called by bot (without pretending to not be a bot).

Zabuzard (approved):

christolis:
Have we tested to see if it works with this user agent and the bot can successfully get a response without a 403?

SquidXTV (approved):

PR #1270, opened 2025-07-01
Notify proper post deletion

### LINKED ISSUE #1206: Notify user about properly closing thread if they delete original post
# About
It often happens that (new) people delete their original post message if they could figure it out on their own or want to close the thread in general.

Notify the user on how to properly close a help thread if they delete their original post to avoid having open threads without any content.

billpapat:
Upon the deletion of a message within a help post, a system notification will be displayed, providing guidance on the proper procedure for deleting the post.

image

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
@@ -11,9 +11,12 @@ import net.dv8tion.jda.api.entities.channel.Channel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.dv8tion.jda.api.entities.channel.forums.ForumTag;
 import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
+import net.dv8tion.jda.api.events.message.MessageDeleteEvent;
 import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.dv8tion.jda.api.requests.RestAction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.features.EventReceiver;
 import org.togetherjava.tjbot.features.UserInteractionType;
@@ -38,6 +41,9 @@ import java.util.stream.Collectors;
  */
 public final class HelpThreadCreatedListener extends ListenerAdapter
         implements EventReceiver, UserInteractor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HelpThreadCreatedListener.class);
+
     private final HelpSystemHelper helper;
 
     private final Cache<Long, Instant> threadIdToCreatedAtCache = Caffeine.newBuilder()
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
Taz03, line 1: better to create a new listener here, this one seems to handle creation stuff

Taz03 (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
Zabuzard: yeah, this is definitely the wrong class to handle this

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -71,6 +77,34 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+                long ownerId = threadChannel.getOwnerIdLong();
+                threadChannel.getGuild()
+                    .retrieveMemberById(ownerId)
+                    .queue(member -> threadChannel
+                        .sendMessage(
+                                """
+                                        It looks like a message was deleted in this help post.\s
+                                        To properly close this post, please __right-click__ on the **Created Post** __followed by__ **Delete Post**\s
+                                        instead of deleting messages. This ensures proper record-keeping and\s
+                                        helps other users find solutions.""")
```
Zabuzard, line 99: people in our server cant delete threads. u need to educate them about using `/help-thread close` and not in a text-way but as slash command. do a text search of `mentionSlashCommand`, ull find a helper for it.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -71,6 +77,34 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+                        .queue(),
+                            failure -> LOGGER.error(
+                                    "Could not retrieve thread owner for thread ID {}: {}",
+                                    threadChannel.getId(), failure.getMessage(), failure));
```
Zabuzard, line 103: `failure.getMessage()` is unecessary, ur already attaching the exception itself at the end which contains all info :)

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -71,6 +77,34 @@ public void onMessageReceived(MessageReceivedEvent event) {
         }
     }
 
+    @Override
+    public void onMessageDelete(MessageDeleteEvent event) {
+        // Check if the deleted message was from a thread
```
Zabuzard, line 82: i think u misunderstood the task. its not about deleting any message in a help thread. its about deleting the original (first) message only. and ideally also only if it was deleted by OP. otherwise it also triggers when someone else deletes it (for example a mod).

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
```diff
@@ -71,6 +77,34 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+            Channel parentChannel = threadChannel.getParentChannel();
+
+            // Check if it's a help forum thread
+            if (helper.isHelpForumName(parentChannel.getName())) {
+                // Retrieve the thread owner using the owner's ID
+                long ownerId = threadChannel.getOwnerIdLong();
+                threadChannel.getGuild()
+                    .retrieveMemberById(ownerId)
+                    .queue(member -> threadChannel
+                        .sendMessage(
```
Zabuzard, line 94: id prefer sending a DM instead of a public message. public message only as fallback if DMs are closed. see `/transfer` which has a similar flow

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
billpapat: I am not familiar with the role system in the server.

Normal user's can also created posts or is it for OP's and above?

Zabuzard: any user can create a question thread and any user can send messages in the thread.

the github-issue is about OP deleting the first message in the thread, not "any".

SquidXTV: > Normal user's can also created posts or is it for OP's and above?

Btw for clarification: OP stands for "Original Poster", so it means the person creating the question in our case

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadCreatedListener.java
SquidXTV: I think that the `hint` should be in the thread directly, because thats also the context of their actions. Just like how the notification for automatic closing works.

Zabuzard: fair. we can keep it then, i guess.

KonstantineVashalomidze:
@billpapat Are you still working on this?

PR #1279, opened 2025-07-15
Block image-based scam messages

### LINKED ISSUE #1278: Block Picture-Scam
## Problem

Recently we are facing a new type of scam message pattern that uses pictures only. Looks for example like this:

scam

The user posts a bunch of pictures and they contain the malicious scam instructions.

## Approach

We want to extend our existing `ScamBlocker` (more specifically the `ScamDetector`) to detect and handle this new scam pattern.

Therefore, the proposal is to trigger on any message ticking the following:
* no text content, just attachments
* there are 3 or more picture attachments (end with `.jpg` or `.png`) and no other attachments
* these attachments are named with increasing digits (it would also be enough to check that all of them are single or two digits only, such as `17.jpg`)

### Details

Therefore, we should modify `ScamDetector#isScam` and add an overload taking `Message`, not just `CharSequence`. This new `isScam` method then first checks for this new threat with the attachments etc and then, if that is considered safe, calls the existing `isScam(CharSequence)` overload the way it is called today in `ScamBlocker` already.

```java
// in ScamBlocker
if (isSafe && scamDetector.isScam(message)) {
  isSafe = false;
}
```
```java
// in ScamDetector
public boolean isScam(Message message) {
  // TODO check new attachment scam
  //  and then, if that is considered safe, do this:
  String content = message.getContentDisplay();
  return isScam(content);
}
```

Additionally, this should be covered with unit tests. The existing `ScamDetectorTest` class can be expanded for that. Make use of a `Message message = mock(Message.class)` for that.

Zabuzard:
## Summary

Closes #1278. Extends `ScamBlocker` to detect image-only scam messages, plus unit tests.

## Details

As described in the ticket, messages are now flagged as scam if:

* empty message (`isBlank()`)
* amount of suspicious attachments is greater than configed (`suspiciousAttachmentsThreshold`)
* an attachment is considered suspicious if it is an image (`IsImage()`) and its name matches the configured regex pattern `suspiciousAttachmentNamePattern`

The file name pattern is configured by default as desired in the ticket description. That is, it triggers on files like `image.png` or files with 1-2 digits like `1.jpg` or `23.png`.

## Config

Added two new config params (part of `"scamBlocker"` config):
```json
"suspiciousAttachmentsThreshold": 3,
"suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}"
```

## Impressions

bad post

is flagged

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
@@ -1,11 +1,16 @@
 package org.togetherjava.tjbot.features.moderation.scam;
 
+import net.dv8tion.jda.api.entities.Message;
+
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.config.ScamBlockerConfig;
 import org.togetherjava.tjbot.features.utils.StringDistances;
 
 import java.net.URI;
+import java.util.Collection;
+import java.util.List;
 import java.util.Locale;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
@@ -18,6 +23,7 @@ import java.util.stream.Stream;
 public final class ScamDetector {
     private static final Pattern TOKENIZER = Pattern.compile("[\\s,]");
     private final ScamBlockerConfig config;
+    private final Predicate<String> isSuspiciousAttachmentName;
 
     /**
      * Creates a new instance with the given configuration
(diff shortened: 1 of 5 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
```diff
@@ -123,6 +149,16 @@ private boolean containsSuspiciousKeyword(String token) {
             });
     }
 
+    private boolean areAttachmentsSuspicious(Collection<? extends Message.Attachment> attachments) {
+        long suspiciousAttachments =
+                attachments.stream().filter(this::isAttachmentSuspicious).count();
+        return suspiciousAttachments >= config.getSuspiciousAttachmentsThreshold();
```
SquidXTV, line 155: Shouldn't `config.getSuspiciousAttachmentsThreshold()` be stored in a variable like all the other config stuff extracted in the constructor?

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
Zabuzard: the other stuff isnt extracted in this class actually. only the pattern (bc its meaningful to store it as `Predicate` directly).

the rest is also accessed right through the config:
config

i personally wouldnt mind either, but i think its useful for readability if it says `config.foo()`, so u know this comes from the config. although thats probably not much of a concern for the logic itself 🤷 

but yeah. if we want to change it for the threshold, it also needs to be changed for the rest of config accesses in this class

SquidXTV: Yeah I see, I think that's just how most of the codebase does it. But I don't mind tbh

SquidXTV (approved):

PR #1282, opened 2025-07-21
Ignore "scam" send by trusted users

Zabuzard:
## Summary

This makes it so that trusted users, such as moderators, are ignored by the scam detection.

The idea here is not to let them send scam around but when the scam detector has a false positive, mods could then at least still send the message for the user.

This of course bears the risk that if a mod is hacked, they can freely send undetected scam in the server. We never had such a case though and mods are required to enable 2FA anyways.

## Details

This does not require a config change as it uses the already-existing:

```json
"softModerationRolePattern": "Moderator|Community Ambassador",
```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamDetector.java
@@ -1,6 +1,8 @@
 package org.togetherjava.tjbot.features.moderation.scam;
 
+import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.Role;
 
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.config.ScamBlockerConfig;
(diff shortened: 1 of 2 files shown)


Zabuzard:
> Do we really only include these two roles? I would say that `Expert` or `Top Helper` should also be skipped here. It is rarely (basically never) the case that these accounts get hacked and start spamming scam. And even if we can still manually handle that like we did all the years before.

I wouldnt mind. I kinda wanted to prevent adding _yet another config entry for roles_. But welp 🤷

marko-radosavljevic:
I agree with squid, we have this one in config already, which is perfect:

`"excludeCodeAutoDetectionRolePattern": "Top Helpers .+|Moderator|Community Ambassador|Expert",`

It's made for similar purpose, to not annoy trusted memebers that know how to format. It's nice to have some granularity wtih roles, those current patterns can also be renamed to something more general, and resued. Because intent behind them is same, trusted/knowledgeble member that shouldn't be automated, a member that is trusted to edit the tag, etc.
Only if our worry is that we are exploding with pattens, and that almost every feature is using unique one, otherwise I don't mind explicit naming even if we have slightly bigger configs.

(there is also `tagManageRolePattern`)

marko-radosavljevic (approved):
cute! :heart: 

I will be safe from the bloodthirsty robot! :relieved: :pray:

Issue #1283, opened 2025-07-23
Auto ban/quarantine based on similar messages


Alathreon:
**Is your feature request related to a problem? Please describe.**

TJBot has many features to detect scams, either by finding suspicious patterns in message or via the ignore-me channels, but they both have problems:
*  Patterns
* * It appears that stacktraces are often falsely flaged, which is a problem, and while there are effort to make the pattern more and more complex, there are still many false positives.
* Ignore-me
* * Regular users tend to put messages in ignore-me, usually because they are curious, and usually after a scam or other user put a message here, which made the channel white, leading other users here.
* * Ignore-me also tend to have very little true positives, here are the stats
* *  *  25% true positives, so 75% are false positives
* *  *  Each user who see this channel has 8% chance to post a message in it

The advantage of this new feature is that it should have little to no false positive and also take the same niche as the ignore-me channel in the way that it won't flag the bot at the first time, but if the bot put enough messages, in ignore-me case, the bot will reach this channel, and in this new feature, after enough messages, the bot will be flaged automatically.
This is useful since ignore-me is the biggest problem currently because it causes a lot more work for the mods while not doing much, and it would also allow the pattern feature to be less strict, falsely flaging less.

**Describe the solution you'd like**

- First of, storage, either a DB or a local cache can be used, and while the following mention a DB, it's probably better to just use a cache.
- Each time a user post a message in a channel, an entry(<ins>userId</ins>, <ins>channelId</ins>, messagehHash, timestamp) is saved in the db (and override previous entry if needed), key is underlined.
- If after saving into the db, TJBot notices that there are at least 3 entries with same userId, and messageHash, but different channelId, the user gets quarantined.
- Every hour, every entry that are older than an hour gets deleted.

**Describe alternatives you've considered**

- Don't do anything and assume that the currently situation is fine.
- Continue to improve the patterns.
- Continue to improve the patterns and archive ignore-me.
- Add a feature so it's easier to unquarantine (currently WIP by Alathreon, no Issue currently, and only for ignore-me currently), this one can be made in combination with all other choices.

**Additional context**

This feature is mostly here because of the mods and the users being both annoyed for handling the false flaging and being false flaged respectively.

PR #1292, opened 2025-07-31
feature/handle-similar-messages-as-scam

Alathreon:
From in #1283 with those differences:
- Whole feature uses its own class
- I used a local cache
- I didn't use caffeeine
- 4 new config variables
- A few fail safes
- Hash is based on SHA

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -89,7 +89,11 @@
         ],
         "isHostSimilarToKeywordDistanceThreshold": 2,
         "suspiciousAttachmentsThreshold": 3,
-        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}"
+        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}",
+        "maxAllowedSimilarMessages": 2,
+        "similarMessagesWindow": 1,
+        "similarMessageLengthIgnore": 10,
+        "similarMessagesWhitelist": []
     },
     "wolframAlphaAppId": "79J52T-6239TVXHR7",
     "helpSystem": {
--- a/application/src/main/java/org/togetherjava/tjbot/config/ScamBlockerConfig.java
+++ b/application/src/main/java/org/togetherjava/tjbot/config/ScamBlockerConfig.java
@@ -26,6 +26,10 @@ public final class ScamBlockerConfig {
     private final int isHostSimilarToKeywordDistanceThreshold;
     private final int suspiciousAttachmentsThreshold;
     private final String suspiciousAttachmentNamePattern;
+    private final int maxAllowedSimilarMessages;
+    private final int similarMessagesWindow;
+    private final int similarMessageLengthIgnore;
+    private final Set<String> similarMessagesWhitelist;
 
     @JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
     private ScamBlockerConfig(@JsonProperty(value = "mode", required = true) Mode mode,
@@ -46,7 +50,12 @@ public final class ScamBlockerConfig {
             @JsonProperty(value = "suspiciousAttachmentsThreshold",
                     required = true) int suspiciousAttachmentsThreshold,
             @JsonProperty(value = "suspiciousAttachmentNamePattern",
-                    required = true) String suspiciousAttachmentNamePattern) {
+                    required = true) String suspiciousAttachmentNamePattern,
+            @JsonProperty(value = "maxAllowedSimilarMessages") int maxAllowedSimilarMessages,
+            @JsonProperty(value = "similarMessagesWindow") int similarMessagesWindow,
+            @JsonProperty(value = "similarMessageLengthIgnore") int similarMessageLengthIgnore,
+            @JsonProperty(
+                    value = "similarMessagesWhitelist") Set<String> similarMessagesWhitelist) {
         this.mode = Objects.requireNonNull(mode);
         this.reportChannelPattern = Objects.requireNonNull(reportChannelPattern);
         this.botTrapChannelPattern = Objects.requireNonNull(botTrapChannelPattern);
@@ -59,6 +68,10 @@ public final class ScamBlockerConfig {
         this.suspiciousAttachmentsThreshold = suspiciousAttachmentsThreshold;
         this.suspiciousAttachmentNamePattern =
                 Objects.requireNonNull(suspiciousAttachmentNamePattern);
+        this.maxAllowedSimilarMessages = maxAllowedSimilarMessages;
+        this.similarMessagesWindow = similarMessagesWindow;
+        this.similarMessageLengthIgnore = similarMessageLengthIgnore;
+        this.similarMessagesWhitelist = similarMessagesWhitelist;
     }
 
     /**
@@ -165,6 +178,43 @@ public final class ScamBlockerConfig {
      */
     public String getSuspiciousAttachmentNamePattern() {
         return suspiciousAttachmentNamePattern;
+    }
+
+    /**
+     * Gets the maximum amount of allowed messages before it gets flagged by the scam detector.
+     * 
+     * @return the maximum amount of allowed messages
+     */
+    public int getMaxAllowedSimilarMessages() {
+        return maxAllowedSimilarMessages;
+    }
+
+    /**
+     * Gets the window in minutes to which messages are kept in the similar messages feature.
+     * 
+     * @return the window in minutes to keep the messages
+     */
+    public int getSimilarMessagesWindow() {
+        return similarMessagesWindow;
+    }
+
+    /**
+     * Gets the maximum length allowed before the message gets monitored by the similar message
+     * feature.
+     * 
+     * @return maximum length allowed
+     */
+    public int getSimilarMessageLengthIgnore() {
+        return similarMessageLengthIgnore;
+    }
+
+    /**
+     * Gets the set of messages that are allowed to be spammed in the similar messages feature.
+     * 
+     * @return set of whitelisted messages
+     */
+    public Set<String> getSimilarMessagesWhitelist() {
+        return similarMessagesWhitelist;
     }
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/MessageInfo.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/MessageInfo.java
@@ -1,0 +1,28 @@
+package org.togetherjava.tjbot.features.moderation.scam;
+
+
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * Information about a message, used to detect spam of the same message by the same user in
+ * different channels.
+ * 
+ * @param userId the id of the user
+ * @param channelId the channel where the message was posted
+ * @param messageHash the hash of the message
+ * @param timestamp when the message was posted
+ */
+public record MessageInfo(long userId, long channelId, String messageHash, Instant timestamp) {
+
+    @Override
+    public boolean equals(Object other) {
+        return other instanceof MessageInfo message && this.userId == message.userId
+                && this.channelId == message.channelId;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(userId, channelId);
+    }
+}
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
@@ -25,9 +25,7 @@ import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.config.ScamBlockerConfig;
-import org.togetherjava.tjbot.features.MessageReceiverAdapter;
-import org.togetherjava.tjbot.features.UserInteractionType;
-import org.togetherjava.tjbot.features.UserInteractor;
+import org.togetherjava.tjbot.features.*;
 import org.togetherjava.tjbot.features.componentids.ComponentIdGenerator;
 import org.togetherjava.tjbot.features.componentids.ComponentIdInteractor;
 import org.togetherjava.tjbot.features.moderation.ModerationAction;
@@ -38,11 +36,8 @@ import org.togetherjava.tjbot.features.utils.MessageUtils;
 import org.togetherjava.tjbot.logging.LogMarkers;
 
 import java.awt.Color;
-import java.util.Collection;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
@@ -55,7 +50,7 @@ import java.util.regex.Pattern;
  * If scam is detected, depending on the configuration, the blockers actions range from deleting the
  * message and banning the author to just logging the message for auditing.
  */
-public final class ScamBlocker extends MessageReceiverAdapter implements UserInteractor {
+public final class ScamBlocker extends MessageReceiverAdapter implements UserInteractor, Routine {
     private static final Logger logger = LoggerFactory.getLogger(ScamBlocker.class);
     private static final Color AMBIENT_COLOR = Color.decode("#CFBFF5");
     private static final Set<ScamBlockerConfig.Mode> MODES_WITH_IMMEDIATE_DELETION =
@@ -72,8 +67,8 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
     private final ModerationActionsStore actionsStore;
     private final ScamHistoryStore scamHistoryStore;
     private final Predicate<String> hasRequiredRole;
-
     private final ComponentIdInteractor componentIdInteractor;
+    private final SimilarMessagesDetector similarMessagesDetector;
 
     /**
      * Creates a new listener to receive all message sent in any channel.
@@ -103,6 +98,7 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
         hasRequiredRole = Pattern.compile(config.getSoftModerationRolePattern()).asMatchPredicate();
 
         componentIdInteractor = new ComponentIdInteractor(getInteractionType(), getName());
+        similarMessagesDetector = new SimilarMessagesDetector(config.getScamBlocker());
     }
 
     @Override
@@ -141,6 +137,10 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
             isSafe = false;
         }
 
+        if (isSafe && similarMessagesDetector.doSimilarMessageCheck(event)) {
+            isSafe = false;
+        }
+
         if (isSafe) {
             return;
         }
@@ -151,6 +151,16 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
         }
 
         takeAction(event);
+    }
+
+    @Override
+    public Schedule createSchedule() {
+        return new Schedule(ScheduleMode.FIXED_RATE, 1, 1, TimeUnit.MINUTES);
+    }
+
+    @Override
+    public void runRoutine(JDA jda) {
+        similarMessagesDetector.runRoutine();
     }
 
     private void takeActionWasAlreadyReported(MessageReceivedEvent event) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
@@ -1,0 +1,109 @@
+package org.togetherjava.tjbot.features.moderation.scam;
+
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+
+import org.togetherjava.tjbot.config.ScamBlockerConfig;
+import org.togetherjava.tjbot.features.utils.Hashing;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Class which tries to detect scams by monitoring similar messages.
+ */
+public class SimilarMessagesDetector {
+    private static final String HASH_METHOD = "SHA";
+
+    private final ScamBlockerConfig scamBlockerConfig;
+    private final Set<MessageInfo> messageCache;
+    private final Set<Long> alreadyFlaggedUsers;
+
+    /**
+     * Creates an instance of this class by using the given config.
+     * 
+     * @param scamBlockerConfig the scam config
+     */
+    public SimilarMessagesDetector(ScamBlockerConfig scamBlockerConfig) {
+        this.scamBlockerConfig = scamBlockerConfig;
+        this.messageCache = new HashSet<>();
+        this.alreadyFlaggedUsers = new HashSet<>();
+    }
+
+    private boolean shouldIgnore(Message message) {
+        if (!message.getAttachments().isEmpty()) {
+            return false;
+        }
+        if (message.getContentRaw().length() <= scamBlockerConfig.getSimilarMessageLengthIgnore()) {
+            return true;
+        }
+        return scamBlockerConfig.getSimilarMessagesWhitelist()
+            .contains(message.getContentRaw().toLowerCase());
+    }
+
+    private MessageInfo addToMessageCache(MessageReceivedEvent event) {
+        long userId = event.getAuthor().getIdLong();
+        long channelId = event.getChannel().getIdLong();
+        String messageHash = getHash(event.getMessage());
+        Instant timestamp = event.getMessage().getTimeCreated().toInstant();
+        MessageInfo messageInfo = new MessageInfo(userId, channelId, messageHash, timestamp);
+        messageCache.add(messageInfo);
+        return messageInfo;
+    }
+
+    private String getHash(Message message) {
+        String wholeText = message.getContentRaw() + message.getAttachments()
+            .stream()
+            .map(Message.Attachment::getFileName)
+            .collect(Collectors.joining());
+        return Hashing.bytesToHex(Hashing.hashUTF8(HASH_METHOD, wholeText));
+    }
+
+    private boolean hasPostedTooManySimilarMessages(long userId, String messageHash) {
+        long similarMessageCount = messageCache.stream()
+            .filter(m -> m.userId() == userId && m.messageHash().equals(messageHash)
+                    && !isObsolete(m))
+            .count();
+        return similarMessageCount > scamBlockerConfig.getMaxAllowedSimilarMessages();
+    }
+
+    private boolean isObsolete(MessageInfo messageInfo) {
+        return messageInfo.timestamp()
+            .plus(scamBlockerConfig.getSimilarMessagesWindow(), ChronoUnit.MINUTES)
+            .isBefore(Instant.now());
+    }
+
+    /**
+     * Stores message data and if many messages of same author, different channel and same content
+     * is posted several times, returns true.
+     *
+     * @param event the message event
+     * @return true if the user spammed the message in several channels, false otherwise
+     */
+    public boolean doSimilarMessageCheck(MessageReceivedEvent event) {
+        long userId = event.getAuthor().getIdLong();
+        if (alreadyFlaggedUsers.contains(userId)) {
+            return true;
+        }
+        if (shouldIgnore(event.getMessage())) {
+            return false;
+        }
+        String hash = addToMessageCache(event).messageHash();
+        if (hasPostedTooManySimilarMessages(userId, hash)) {
+            alreadyFlaggedUsers.add(userId);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Has to be called often to clear the cache.
+     */
+    public void runRoutine() {
+        messageCache.removeIf(this::isObsolete);
+    }
+}


application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
```diff
@@ -0,0 +1,26 @@
+package org.togetherjava.tjbot.config;
```
Zabuzard: this file shouldnt be in the config package.
in fact, id make it a package private helper class for scamblocker and move it into its scam package instead.

should it turn out that some other feature needs it as well, it can get elevated. its just a quick record anyway and other classes most likely need specialized solutions with slightly different values anyways 👍

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+     * 
+     * @param event the message event
+     * @return true if the user spammed the message in several channels, false otherwise
+     */
+    private boolean doSimilarMessageCheck(MessageReceivedEvent event) {
+        long userId = event.getAuthor().getIdLong();
+        long channelId = event.getChannel().getIdLong();
+        int messageHash = getHash(event.getMessage());
+        Instant timestamp = event.getMessage().getTimeCreated().toInstant();
+        messageCache.add(new MessageInfo(userId, channelId, messageHash, timestamp));
```
Zabuzard: the cache needs a max capacity feature, otherwise users could blow up our VPS by spam messaging the server in the night. our RAM is quite limited. 10k should be a good value

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+    }
+
+    /**
+     * Stores message data and if many messages of same author, different channel and same content
+     * is posted several times, returns true.
+     * 
+     * @param event the message event
+     * @return true if the user spammed the message in several channels, false otherwise
+     */
+    private boolean doSimilarMessageCheck(MessageReceivedEvent event) {
```
Zabuzard: the method should be split into two for better readability. one method that adds the message `addToMessageCache(message)` and a second method that does the check and then has a typical bool-method-name like `hasPostedTooManySimilarMessages(message)` or something like that.

(if doing the factory method from the next comment one could also ommit the extra method for `addToMessageCache` and simply do `messageCache.add(MessageInfo.fromMessageEvent(event))`. that brings the advantage that the `MessageInfo.fromMessageEvent(event)` could be put into a variable for reuse on the second method call `hasPostedTooManySimilarMessage(messageInfo)`, so that a lookup isnt necessary anymore 👍)

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+        long userId = event.getAuthor().getIdLong();
+        long channelId = event.getChannel().getIdLong();
+        int messageHash = getHash(event.getMessage());
+        Instant timestamp = event.getMessage().getTimeCreated().toInstant();
```
Zabuzard: i would move that into the record. sth like `static MessageInfo fromMessageEvent(MessageReceivedEvent event) {...}`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+        messageCache.add(new MessageInfo(userId, channelId, messageHash, timestamp));
+        return config.getScamBlocker().getMaxSimilarMessages() < messageCache.stream()
```
Zabuzard: NIT: if u dont need the rest of the `config` u might change the field to save the `ScamBlockerConfig` instead

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+            .filter(m -> m.userId() == userId && m.messageHash() == messageHash)
+            .count();
```
Zabuzard: potentially its more readable to put the logic into a variable:
```java
int similarMessages = messageCache.stream()...
return similarMessages >= config.getMaxSimilarMessages();
```

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Information about a message, used to detect spam of the same message by the same user in
+ * different channels.
+ * 
+ * @param userId the id of the user
+ * @param channelId the channel where the message was posted
+ * @param messageHash the hash of the message
+ * @param timestamp when the message was posted
+ */
+public record MessageInfo(long userId, long channelId, int messageHash, Instant timestamp) {
```
Zabuzard: it might not be clear what the hash is referring to since "message" is ambigous (for example, maybe it includes the author or its the id, ...). perhaps rename to `contentHash`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -136,6 +136,42 @@ public void onMessageReceived(MessageReceivedEvent event) {
...
+    }
+
+    private int getHash(Message message) {
+        return message.getContentRaw().hashCode() + message.getAttachments()
+            .stream()
+            .mapToInt(a -> a.getFileName().hashCode())
+            .reduce(1, (a, b) -> a * b);
+    }
```
Zabuzard: i would move that into the record as well. (if u go with the suggested factory method, it can be right there)
for the hash, prefer using SHA on the content instead of string hashcode.

we have code for that already available somewhere. just do a quick text search for SHA or hash 👍

application/src/main/java/org/togetherjava/tjbot/config/ScamBlockerConfig.java
```diff
@@ -111,6 +114,15 @@ public int getIsHostSimilarToKeywordDistanceThreshold() {
...
     }
 
+    /**
+     * Gets the maximum amount of allowed messages before it gets flagged by the scam detector.
+     * 
+     * @return the maximum amount of allowed messages
+     */
+    public int getMaxSimilarMessages() {
+        return maxSimilarMessages;
+    }
```
Zabuzard: id rename it to `maxAllowedSimilarMessages` or `allowedSimilarMessages` so the consequence is a bit clearer

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
+rem the _cmd.exe /c_ return code!
+set EXIT_CODE=%ERRORLEVEL%
+if %EXIT_CODE% equ 0 set EXIT_CODE=1
+if not ""=="%GRADLE_EXIT_CONSOLE%" exit %EXIT_CODE%
+exit /b %EXIT_CODE%
+
+:mainEnd
+if "%OS%"=="Windows_NT" endlocal
+
+:omega
```
Zabuzard: u prob didnt touch this file, so the diff shouldnt be here 😬

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -124,7 +124,7 @@ public void onMessageReceived(MessageReceivedEvent event) {
 
         Message message = event.getMessage();
         String content = message.getContentDisplay();
-        if (!scamDetector.isScam(content)) {
+        if (!scamDetector.isScam(content) && !doSimilarMessageCheck(event)) {
```
Zabuzard: would be neat if we cover this new feature with a unit test with some mockito to create the message (see `ScamDetectorTest`), or alternatively using `jdaTestHelper` as used by a lot of other tests.

i understand it though if u rather want to skip this for now. then someone else can maybe do it eventually. just make sure its tested locally then though

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
Alathreon: I have no idea why I put it here lol

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: What about I create Set<Long> which I fill with every user that was flagged with the max message feature, and I check this set before the map ? It would solve the ram problem while not blocking the feature if we reach the 10k you mentioned.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: I kept the current code which was saving the whole config, I'll change it if you want.

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
Alathreon: I made it ambigous on purpose since it's not its responsability to know how the hash was made.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: So a private static method in MessageInfo ?

Alathreon: Also, does SHA have perf problems ? Since we need to handle the hash of many attachment, a user could attempt a DOS attack by putting many attachments and sending many messages.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: i dont fully follow. but if it serves the RAM issue, sounds good.
FYI Caffeeine supports time based eviction and max limit out of the box. One line of code and you do not have to care about it anymore, also no need for `Routine` then.

Perhaps its the better choice given that we already have caffeeine in use in a lot of classes. Up to you

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: oh, i see. i dont mind. in `ScamDetector` i only memorized the scamblocker config. i personally would find it cleaner

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
Zabuzard: thats a fair argument

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: yeah, private static in MessageInfo sounds good. or directly in the factory method `fromMessage`

im not aware of SHA having perf issues. our bot so far barely scratches the 5% CPU afaik, we should probably be good.
i dont recall where exactly but i think we already compute sha hashes (or was it MD5?) on all messages send.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: Actually, Config is indeed used somewhere, solely because of config.quarantinedRolePattern, what do I do ?
Also note that the field `botTrapChannelPattern` is useless, do you want me to remove it ?
Or do I create a new issue for those two things ? If we find a solution, we could easily mark it as good first issue.

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
Alathreon: Done

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: I still prefer my new solution, review it again if you think it can be done in a more easier way with caffeine, also note that I will migrate the max message into a separate class I think

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: Done

Alathreon: Also, why did you change the < into a >= ?

application/src/main/java/org/togetherjava/tjbot/config/MessageInfo.java
Alathreon: Won't do then

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: > Actually, Config is indeed used somewhere, solely because of config.quarantinedRolePattern, what do I do ?

thats fair, then keep it as is.

>  Also note that the field botTrapChannelPattern is useless, do you want me to remove it ?

please keep the feature for now, at least until we feel comfortable with the new solution and its battle-tested. but yeah, can be removed then, i guess

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: bc order changed. imo its more readable to have the `expected >= actual` than `actual < expected`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: > please keep the feature for now, at least until we feel comfortable with the new solution and its battle-tested. but yeah, can be removed then, i guess

I am not talking about the feature but the field.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: Marked it as resolved since it's outdated, please review again.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: 2 < 3 =! 3 >= 2
Those are not equivalent, we are just mirroring left and right side.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: I did it
Marked it as resolved since it's outdated, please review again.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Alathreon: I tested it locally, I will make a test later.
I wish that this gets merged first.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
SquidXTV: I would still agree to Zabus point and say it is more readable

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: fair

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
Zabuzard: uh, yeah. i had a logic twist in my head. u can figure it out. but id prefer changing the order

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        if (shouldIgnore(event.getMessage())) {
+            return false;
+        }
+        String hash = addToMessageCache(event).messageHash();
+        if (hasPostedTooManySimilarMessages(userId, hash)) {
+            alreadyFlaggedUsers.add(userId);
+            return true;
+        } else {
+            return false;
```
Zabuzard, line 99: i am not sure if that `alreadyFlaggedUsers` user thing really works.. the bot has a really long uptime usually, sometimes many months.
the way i read the code is that if a user got flagged once, they will then be immune to this check until the bot is restarted again. u could argue that its probably not an issue in practice but it smells a bit. you would need a time based cleanup for `alreadyFlaggedUsers` as well.

at which point u have like 30 lines of code dealing with what caffeeine gets done in one line, i guess.
```java
private final Cache<String, Instant> userIdToAskedAtCache = Caffeine.newBuilder()
  .maximumSize(1_000)
  .expireAfterWrite(Duration.of(10, ChronoUnit.SECONDS))
  .build();
```
the interface is essentially that of a `Map`, so you have various `get` and `put` variants. internally its a LRU map, so it kicks out the entries that werent touched the longest when it reaches the max limit.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/Hashing.java
```diff
@@ -57,4 +57,16 @@ public static byte[] hash(String method, byte[] data) {
...
+
+    /**
+     * Hashes the given UTF 8 text using the given method, see {@link Hashing#hash(String, byte[])}.
+     *
+     * @param method the method to use for hashing, must be supported by {@link MessageDigest}, e.g.
+     *        {@code "SHA"}
+     * @param text the UTF 8 text to hash
+     * @return the computed hash
+     */
+    public static byte[] hashUTF8(String method, String text) {
```
Zabuzard, line 69: NIT: should probably be renamed into `hashUtf8` for better camelCase readability

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private MessageInfo addToMessageCache(MessageReceivedEvent event) {
+        long userId = event.getAuthor().getIdLong();
+        long channelId = event.getChannel().getIdLong();
+        String messageHash = getHash(event.getMessage());
+        Instant timestamp = event.getMessage().getTimeCreated().toInstant();
+        MessageInfo messageInfo = new MessageInfo(userId, channelId, messageHash, timestamp);
+        messageCache.add(messageInfo);
+        return messageInfo;
```
Zabuzard, line 54: id still refactor this and move all the stuff dealing with `MessageInfo` into the `record`. so:
```java
record MessageInfo(...) {
  static MessageInfo fromMessageEvent(MessageReceivedEvent event) {...}
}
```
then this `addToMessageCache` method becomes obsolete and can be replaced with just `messageCache.add(MessageInfo.fromMessageEvent(event));` at the call-site

the hash creation would then also move into the records `fromMessageEvent` method (or as `private static` helper in the record)

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Stores message data and if many messages of same author, different channel and same content
+     * is posted several times, returns true.
+     *
+     * @param event the message event
+     * @return true if the user spammed the message in several channels, false otherwise
+     */
+    public boolean doSimilarMessageCheck(MessageReceivedEvent event) {
```
Zabuzard, line 86: could u do some reordering so the flow reads top to bottom. this method would then be at the top, as entry-point. and the private methods then below it in the order they are used. the `runRoutine` can stay at the very bottom, i guess, since its a detached flow.

application/config.json.template
```diff
@@ -89,7 +89,11 @@
         ],
         "isHostSimilarToKeywordDistanceThreshold": 2,
         "suspiciousAttachmentsThreshold": 3,
-        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}"
+        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}",
+        "maxAllowedSimilarMessages": 2,
+        "similarMessagesWindow": 1,
```
Zabuzard, line 94: missing time unit. please add it to the name.

application/config.json.template
```diff
@@ -89,7 +89,11 @@
...
+        "similarMessageLengthIgnore": 10,
+        "similarMessagesWhitelist": []
```
Zabuzard, line 96: i dont think we need that, please remove.

the UX for it for maintainers would be really weird, having to bloat the config with 100 character long messages and whatnot. and then it didnt work bc of a smiley or other stuff that ends up slightly different.

application/config.json.template
```diff
@@ -89,7 +89,11 @@
         ],
         "isHostSimilarToKeywordDistanceThreshold": 2,
         "suspiciousAttachmentsThreshold": 3,
-        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}"
+        "suspiciousAttachmentNamePattern": "(image|\\d{1,2})\\.[^.]{0,5}",
+        "maxAllowedSimilarMessages": 2,
```
Zabuzard, line 93: lets start with `3`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/MessageInfo.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public boolean equals(Object other) {
+        return other instanceof MessageInfo message && this.userId == message.userId
+                && this.channelId == message.channelId;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(userId, channelId);
+    }
```
Zabuzard, line 27: i find that quite confusing, not what someone using this record would expect.
it says that two completely different messages posted in the same channel by the same author are considered identical `MessageInfo`.

if you need this to determine duplicates in ur `Set` it would probably be better to go for something like `record MessageLocation(long userId, long channelId) {}` and then go for `Map<MessageLocation, MessageInfo>` instead.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
```diff
@@ -141,6 +137,10 @@ public void onMessageReceived(MessageReceivedEvent event) {
             isSafe = false;
         }
 
+        if (isSafe && similarMessagesDetector.doSimilarMessageCheck(event)) {
```
Zabuzard, line 140: `doSimilarMessageCheck` should be renamed. maybe `hasPostedSimilarMessageTooOften(event)`

if you dont want it to start with a question-word bc of the side effects, maybe `handleHasPostedSimilarMessagesTooOften(event)`.

but the name should clarify what the `boolean` it returns means and how it fits into the logic of `ScamBlocker`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Has to be called often to clear the cache.
+     */
+    public void runRoutine() {
```
Zabuzard, line 106: rename to `clearCache` or `cleanupCache` or something.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private boolean hasPostedTooManySimilarMessages(long userId, String messageHash) {
+        long similarMessageCount = messageCache.stream()
+            .filter(m -> m.userId() == userId && m.messageHash().equals(messageHash)
+                    && !isObsolete(m))
+            .count();
+        return similarMessageCount > scamBlockerConfig.getMaxAllowedSimilarMessages();
+    }
+
+    private boolean isObsolete(MessageInfo messageInfo) {
```
Zabuzard, line 73: rename to `isOutdated` or `isExpired`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private boolean shouldIgnore(Message message) {
+        if (!message.getAttachments().isEmpty()) {
+            return false;
+        }
+        if (message.getContentRaw().length() <= scamBlockerConfig.getSimilarMessageLengthIgnore()) {
+            return true;
+        }
+        return scamBlockerConfig.getSimilarMessagesWhitelist()
+            .contains(message.getContentRaw().toLowerCase());
```
Zabuzard, line 44: `.toLowerCase` needs `Locale.US`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
```diff
@@ -0,0 +1,109 @@
...
+            .collect(Collectors.joining());
+        return Hashing.bytesToHex(Hashing.hashUTF8(HASH_METHOD, wholeText));
+    }
+
+    private boolean hasPostedTooManySimilarMessages(long userId, String messageHash) {
+        long similarMessageCount = messageCache.stream()
+            .filter(m -> m.userId() == userId && m.messageHash().equals(messageHash)
+                    && !isObsolete(m))
+            .count();
+        return similarMessageCount > scamBlockerConfig.getMaxAllowedSimilarMessages();
```
Zabuzard, line 70: NIT: reorder to match `expected ... actual`

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Alathreon: how do i make it a set ? if the cache already handles he cleanup, then storing the instant is pointless.
edit: i saw your other comment, i'll see that

application/config.json.template
Alathreon: All 4? or just the whitelist ?

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/MessageInfo.java
Alathreon: hmm ill try that

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Alathreon: it was expected...actual previously, but you asked me to reverse it to actual...expected

application/config.json.template
Alathreon: I want to do strictly higher than, and so > 3 means at least 3 channels. I can't go for higher or equal because i use the word max.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Alathreon: I don't quite like adding a unnecessary dependencies to a record, and even less logic like how to create a hash or how to create it from an event imo.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Zabuzard: caffeine only has a `Map`. sometimes in the past we also only needed a `Set` and then mapped it to something. for example to itself, i.e. key == value.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Zabuzard: mh... but its a factory to create instances of this record. so imo this is exactly the place it should be. what do others here think?

application/config.json.template
Zabuzard: sorry, the preview didnt turn out the way i wanted. i was referring to the message whitelist.

application/config.json.template
Zabuzard: mh. not a fan of the name but i dont have a better suggestion right now. so yeah, its fine. thanks

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/SimilarMessagesDetector.java
Zabuzard: lol, now im confused 😆 sorry.
its fine, keep it as is

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/MessageInfo.java
Alathreon: So I tried and the problem is that by using a map it just makes the code less readable.
Do you have another alternative ? I could maybe not use a record.

PR #1293, opened 2025-08-01
Block CDN attachment scam

Zabuzard:
## Summary

This extends the scam blocker to also detect CDN based attachment scam. That is, people posting discord-provided image links containing scam, such as:

example

We already had similiar logic but only for actual attachments. These are "just links" in the message and not technicially attachments from Discords perspective.

The new case is also covered by unit tests and was tested locally.

## Details

The existing code for handling suspicious attachments was tweaked so it can also be used for this scenario. Therefore, we created a small wrapper for `Attachment` that can now either represent an actual Discord attachment or just an URL. The logic then uses this wrapper instead.

The URL analyse code received some rework, in particular the introduction of `AnalyseUrlResult`. This allows us to hold additional details for each URL (for example the contained attachment, if any).

These data classes received a couple of helper methods that mostly deal with string extraction (some `indexOf` and `substring` fun).

All this was done to keep the main flow of how the analyse works intact and not add a bunch of extra if-else branching just for this new edge case.

## Config

No changes are required.

The host-whitelist addition `"cdn.discordapp.com"` was already added to the TEST and PROD bots config.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
@@ -1,0 +1,157 @@
+package org.togetherjava.tjbot.features.moderation.scam;
+
+import org.togetherjava.tjbot.config.ScamBlockerConfig;
+import org.togetherjava.tjbot.features.utils.StringDistances;
+
+import java.net.URI;
+import java.util.Locale;
+
+/**
+ * Analyzes a given text token. Populates various metrics regarding the token possibly being
+ * suspicious, returning back results of the token analyze.
+ *
+ * Highly configurable, using {@link ScamBlockerConfig}. Entry point to use is
+ * {@link #analyze(String, AnalyseResults)}.
+ */
+final class TokenAnalyse {
+    private final ScamBlockerConfig config;
+
+    TokenAnalyse(ScamBlockerConfig config) {
+        this.config = config;
+    }
+
+    /**
+     * Analyzes the given token about being suspicious.
+     * 
+     * @param token the token to analyze
+     * @param results metrics representing how suspicious the token is
+     */
+    void analyze(String token, AnalyseResults results) {
+        if (token.isBlank()) {
+            return;
+        }
+
+        if (!results.pingsEveryone()
+                && ("@everyone".equalsIgnoreCase(token) || "@here".equalsIgnoreCase(token))) {
+            results.markPingsEveryone();
+        }
+
+        if (!results.containsSuspiciousKeyword() && containsSuspiciousKeyword(token)) {
+            results.markContainsSuspiciousKeyword();
+        }
+
+        if (!results.containsDollarSign()
+                && (token.contains("$") || "usd".equalsIgnoreCase(token))) {
+            results.markContainsDollarSign();
+        }
+
+        if (token.startsWith("http")) {
+            analyzeUrl(token, results);
+        } else {
+            results.markNonUrlTokenFound();
+        }
+    }
+
+    private boolean containsSuspiciousKeyword(String token) {
+        String preparedToken = token.toLowerCase(Locale.US);
+
+        // Checks the token against various keywords from the config
+        // The keywords support some regex-inspired syntax
+        return config.getSuspiciousKeywords()
+            .stream()
+            .map(keyword -> keyword.toLowerCase(Locale.US))
+            .anyMatch(keyword -> {
+                // Exact match "^foo$"
+                if (startsWith(keyword, '^') && endsWith(keyword, '$')) {
+                    return preparedToken.equals(keyword.substring(1, keyword.length() - 1));
+                }
+                // Simple regex-inspired syntax "^foo"
+                if (startsWith(keyword, '^')) {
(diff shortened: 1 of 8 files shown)


SquidXTV (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
```diff
@@ -0,0 +1,157 @@
...
+            results.markNonUrlTokenFound();
+        }
+    }
+
+    private boolean containsSuspiciousKeyword(String token) {
+        String preparedToken = token.toLowerCase(Locale.US);
+
+        // Checks the token against various keywords from the config
+        // The keywords support some regex-inspired syntax
+        return config.getSuspiciousKeywords()
```
SquidXTV, line 60: Why arent we using actual regex here?

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param results metrics representing how suspicious the token is
+     */
+    void analyze(String token, AnalyseResults results) {
+        if (token.isBlank()) {
+            return;
+        }
+
+        if (!results.pingsEveryone()
+                && ("@everyone".equalsIgnoreCase(token) || "@here".equalsIgnoreCase(token))) {
+            results.markPingsEveryone();
```
SquidXTV, line 36: These pre-checks really needed? They seem more like optimization, but would make the code a lot cleaner without them.

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
Zabuzard: mh yeah... originally it wasnt needed and what i wanted to avoid is having to make a readable list of keywords ugly by adding `.*foo.*` around all of them.
originall it was just a list of keywords which the code then essentially checks with `.contains` (i.e. implicit `.*foo.*`).

now, time has passed and we needed explicit checks for "the token begins with this" or "the token ends with this".

full regex still isnt needed and i would argue that it would make the keyword list in the config uglier to read and maintain. but it would have the benefit to make the code easier to read, thats absolutely right.

like, this
```json
"keywords": [ "foo", "bar", "^baz", "boo$", "beep"]
```
i find easier to read than
```json
"keywords": [ ".*foo.*", ".*bar.*", "^baz.*", ".*boo$", ".*beep.*"]
```

although u could add as counterargument that "inventing a new syntax" is equally confusing to a maintainer. tricky situation. perhaps it should be changed.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
Zabuzard: u mean that `if (!flag is set && actual checks) { set flag }` could be changed to just `if (actual checks) { set flag }`.

yeah, absolutely. i think i did it bc this code actually runs on every single message send in any channel on the entirey server. and not just "when help thread is created" or similar like most other stuff.

the everyone-ping-check isnt expensive, but some of the other checks are non trivial. in particular, the keyword iteration shouldnt be executed on every single token of every single message on the entire server if we already know that there was a sus keyword.

maybe its premature optimization

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
SquidXTV: I am fine with it for now because the "new syntax" is kinda just a subset of regex.

application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/TokenAnalyse.java
SquidXTV: Hm I was thinking about maybe putting the check in the implementation instead, but that would require `AnalyseResult` as an Argument. For example:
```java
if (!results.containsSuspiciousKeyword() && containsSuspiciousKeyword(token)) {
            results.markContainsSuspiciousKeyword();
}
```
vs
```java
if (containsSuspiciousKeyword(results, token)) {
            results.markContainsSuspiciousKeyword();
}
```
and
```java
private boolean containsSuspiciousKeyword(AnalyseResults results, String token) {
        if (results.containsSuspiciousKeyword()) {
                return true;
        }
        // actual impl
        String preparedToken = token.toLowerCase(Locale.US);
        ...
}
```

Zabuzard: great idea. has the risk to confuse in regards to who is responsible for also setting the flag though. but that would be a heavy side effect for a method called `contains...`, so it should be okay.

i like the idea, lets do it like that :)

Zabuzard: (ran into a logic issue doing that, which turned out to be less readable/preferable. we decided to keep it as-is for now)

Issue #1307, opened 2025-08-24
Decouple the forum post creation from TransferQuestionCommand into a more accessible space


discorddioxin:
## Motivation
Forum Channels are common in Discord. By having a centralized API to perform forum actions, developers would be more encouraged to make use of forum posting for bot actions. The API could be scaled to include utility which eases the use of forums in code, as well as to help ease migration-based breakage.

This new Forum Posting API will start extremely lightweight & streamlined, but could be used as a foundation for improvement (such as acting as a facade for complex forum-based operations)

## Goals
- Allow access to a forum post creation system, which currently lives in [`TransferQuestionCommand`](https://github.com/Together-Java/TJ-Bot/blob/f206bd44593b2ad18471852fcb8fe209808f4f5e/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java)
- Centralize forum posting, allowing for cross-cutting management over forums (JDA API breakage, styling, etc..)

## Risks and Assumptions
- Increasing visibility of an API can bloat the API of the code base
- Any changes to a code base could encounter problems if not properly assessed/tested

## Alternatives
- Use non-forum based channels for bot posting
- Manually write-out the forum post creation code each time forum post creation needed

### Creating a new post

```java
var forumChannel = ...; // could use channel or channelId (in String form)
var forumTitle = "My First Post";
var messageData = MessageCreationData.fromContent("Hello world!");

ForumPoster.using(jda)
    .createPost(forumChannel, forumTitle, messageData)
    .setTags(...) // access to ForumPostAction methods
    .addFiles(...)
    .queue();
```

### Sending a message to a forum thread
```java
var threadChannel = ...; // could use channel or channelId (in String form)
var messageData = MessageCreationData.fromContent("How's it going?");

ForumPoster.using(jda)
    .sendPost(threadChannel, messageData)
    .addFiles(...)  // access to MessagePostAction methods
    .queue();
```

Zabuzard:
Im generally against investing time into making functionality more modular and abstract as long as there arent multiple users.

Like, this functionality is currently only used and needed by a single part in the code and chances are that this might potentially stay like that for a few more years. So if we now invest time into breaking this out into a sophisticated extra-API because we envision potential multiple use-cases "in the future", we make the code harder to maintain for today.

That said, this part of the API sounds useful to break out. But I would probably still wait with it until there are multiple users.

Zabuzard:
[...]

PR #1308, opened 2025-08-25
Introduce Forum Management API

discorddioxin:
Introduces [`ForumPoster`](https://github.com/discorddioxin/TJ-Bot/blob/develop/application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java) and updates [`TransferQuestionCommand`](https://github.com/Together-Java/TJ-Bot/blob/f206bd44593b2ad18471852fcb8fe209808f4f5e/application/src/main/java/org/togetherjava/tjbot/features/moderation/TransferQuestionCommand.java#L236-L238) to make use of it.

Addresses issue #1307

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
@@ -22,14 +22,14 @@ public final class ForumPoster {
 
     private ForumChannel findForumChannel(String channelId) {
         return Optional.ofNullable(jda.getForumChannelById(channelId))
-            .orElseThrow(() -> new IllegalStateException(
+            .orElseThrow(() -> new IllegalArgumentException(
                     "Did not find a forum channel with ID %s while trying to create a forum post. Make sure the config is setup properly."));
     }
 
     private ThreadChannel findForumPost(String postId) {
         return Optional.ofNullable(jda.getThreadChannelById(postId))
-            .orElseThrow(() -> new IllegalStateException(
-                    "Did not find the forum post with ID %s while trying to reply to a post. Make sure the config is setup properly."));
+            .orElseThrow(() -> new IllegalArgumentException(
+                    "Did not find the forum post with ID %s while trying to reply to a post."));
 
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
@@ -1,0 +1,52 @@
+package org.togetherjava.tjbot.features.utils;
+
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
+import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
+import net.dv8tion.jda.api.requests.restaction.ForumPostAction;
+import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
+import net.dv8tion.jda.api.utils.messages.MessageCreateData;
+
+import java.util.*;
+
+public final class ForumPoster {
+    private final JDA jda;
+
+    private ForumPoster(JDA jda) {
+        this.jda = jda;
+    }
+
+    public static ForumPoster using(JDA jda) {
+        return new ForumPoster(jda);
+    }
+
+    private ForumChannel findForumChannel(String channelId) {
+        return Optional.ofNullable(jda.getForumChannelById(channelId))
+            .orElseThrow(() -> new IllegalStateException(
+                    "Did not find a forum channel with ID %s while trying to create a forum post. Make sure the config is setup properly."));
+    }
+
+    private ThreadChannel findForumPost(String postId) {
+        return Optional.ofNullable(jda.getThreadChannelById(postId))
+            .orElseThrow(() -> new IllegalStateException(
+                    "Did not find the forum post with ID %s while trying to reply to a post. Make sure the config is setup properly."));
+
+    }
+
+    public MessageCreateAction sendPost(String postId, MessageCreateData messageData) {
+        return sendPost(findForumPost(postId), messageData);
+    }
+
+    public MessageCreateAction sendPost(ThreadChannel post, MessageCreateData messageData) {
+        return post.sendMessage(messageData);
+    }
+
+    public ForumPostAction createPost(String channelId, String title, MessageCreateData message) {
+        return createPost(findForumChannel(channelId), title, message);
+    }
+
+    public ForumPostAction createPost(ForumChannel channel, String title,
+            MessageCreateData message) {
+        return channel.createForumPost(title, message);
+    }
+}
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public static ForumPoster using(JDA jda) {
+        return new ForumPoster(jda);
+    }
+
+    private ForumChannel findForumChannel(String channelId) {
+        return Optional.ofNullable(jda.getForumChannelById(channelId))
+            .orElseThrow(() -> new IllegalStateException(
+                    "Did not find a forum channel with ID %s while trying to create a forum post. Make sure the config is setup properly."));
```
SquidXTV: I personally prefer an `IllegalArgumentException`.
Why is this exception message mentioning config? That looks like guessing how this method is used.

Same for `#findForumPost`.

SquidXTV (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
discorddioxin: It was basically a copy & paste from the original code. For channels, it's possible the issue could be from a config issue, so it's nice to let the user know. I removed that part from `findForumPost`, because it's more common for that to be used for temporary stuff

EDIT: Seems like I didn't push the `findForumPost` message change. I'll fix all that up

I agree, if it was a bad ID given as an argument, it should be `IllegalArgumentException`. Good catch, I'll update it

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
+import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
+import net.dv8tion.jda.api.requests.restaction.ForumPostAction;
+import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
+import net.dv8tion.jda.api.utils.messages.MessageCreateData;
+
+import java.util.*;
+
+public final class ForumPoster {
```
Zabuzard, line 12: i find the name of the class confusing. for quite some time it sounded to me like its referring to a person posting in the forum, i.e. `Poster` being something like `OP`, author of a post.

but it is referring to this being an utility to post stuff in forums. i would probably have renamed it to `Forums`, but since the API is builder-like and wants to create instances, a more suitable name is perhaps `ForumPostBuilder`, also similar to JDAs `MessageCreateBuilder` and `MessageEditBuilder` that we are all used to.

or alternatively `ForumPost`, since its not actually a builder in the classic sense.

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private ForumPoster(JDA jda) {
+        this.jda = jda;
+    }
+
+    public static ForumPoster using(JDA jda) {
+        return new ForumPoster(jda);
+    }
+
+    private ForumChannel findForumChannel(String channelId) {
```
Zabuzard, line 23: all these `channelId` and `postId`s in the class should probably be `long` instead

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+            .orElseThrow(() -> new IllegalArgumentException(
+                    "Did not find a forum channel with ID %s while trying to create a forum post. Make sure the config is setup properly."));
+    }
+
+    private ThreadChannel findForumPost(String postId) {
+        return Optional.ofNullable(jda.getThreadChannelById(postId))
+            .orElseThrow(() -> new IllegalArgumentException(
+                    "Did not find the forum post with ID %s while trying to reply to a post."));
+
+    }
```
Zabuzard, line 34: do we want this to throw if not found or return `Optional` instead, so the caller can deal with how to properly display error messages and the like to the user.

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+        return Optional.ofNullable(jda.getForumChannelById(channelId))
+            .orElseThrow(() -> new IllegalArgumentException(
+                    "Did not find a forum channel with ID %s while trying to create a forum post. Make sure the config is setup properly."));
+    }
+
+    private ThreadChannel findForumPost(String postId) {
+        return Optional.ofNullable(jda.getThreadChannelById(postId))
+            .orElseThrow(() -> new IllegalArgumentException(
+                    "Did not find the forum post with ID %s while trying to reply to a post."));
+
```
Zabuzard, line 33: empty line, can delete

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public MessageCreateAction sendPost(String postId, MessageCreateData messageData) {
+        return sendPost(findForumPost(postId), messageData);
+    }
```
Zabuzard, line 38: i dont quite understand this. does this reply to a given message? or is `postId` reffering to a channel and not a "message"?

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public ForumPostAction createPost(String channelId, String title, MessageCreateData message) {
+        return createPost(findForumChannel(channelId), title, message);
+    }
+
+    public ForumPostAction createPost(ForumChannel channel, String title,
+            MessageCreateData message) {
+        return channel.createForumPost(title, message);
+    }
```
Zabuzard, line 51: potentially a more convenient API that also aligns more with what JDA already provides and all contributors are used to, would be to create an actual builder. so something where you can do
```java
ForumPost.using(jda)
  .forum(channel)
  .title(title)
  .message(messageCreateData)
  .post();
```
given that you need at least one object coming from JDA within this chain, you could even drop the `using(jda)`. For example, it could extract it from `messageCreateData` or from the `channel` it finds.

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
Zabuzard, line 12: (public classes and methods require meaningful/useful JavaDoc in this project)

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/ForumPoster.java
discorddioxin: > potentially a more convenient API that also aligns more with what JDA already provides and all contributors are used to, would be to create an actual builder. so something where you can do
> 
> ```java
> ForumPost.using(jda)
>   .forum(channel)
>   .title(title)
>   .message(messageCreateData)
>   .post();
> ```
> 
> given that you need at least one object coming from JDA within this chain, you could even drop the `using(jda)`. For example, it could extract it from `messageCreateData` or from the `channel` it finds.

If you'd prefer that, I can update the code

Issue #1315, opened 2025-08-29
GitHubCommand Autocomplete Cache is blocking


Zabuzard:
## Problem

The buildup of the autocomplete-cache in `GitHubCommand` is currently coded blocking during construction. The method takes long (around 30s) because it queries APIs ([GHRepository#getIssues() takes 30s](https://github.com/hub4j/github-api/issues/2128)) and this causes the bot-startup to now take over 30s.

That is not okay and also totally unecessary as this autocomplete-cache is an optional feature that is not needed for the bot to work. It should be triggered non-blocking (async) and be used whenever it is ready.

## Details

The issue starts here:

```java
public GitHubCommand(GitHubReference reference) {
  ...
  updateCache();
}
```

That method looks as follows:

```java
private void updateCache() {
  autocompleteGHIssueCache = reference.getRepositories().stream().map(repo -> {
    ...
    return repo.getIssues(GHIssueState.ALL)
    ...
  })
  ...
}
```

The culprit is the call `reference.getRepositories()` plus `repo.getIssues(...)`, which takes 30s because it needs to query the GitHub API.

The thing is, `autocompleteGHIssueCache` is only needed for the autocompletion of the `/github-search` slashcommand in `onAutoComplete`:

![autocomplete](https://i.imgur.com/HF22rCO.png)

This is an optional feature and it would not be a problem for anyone if the autocompletion simply is not available during the first 30s of the bot being startup.

## Solution

The `updateCache()` method should be triggered async, not blocking the constructor. While it is then being build in the background, the `List<String> autocompleteGHIssueCache` needs to simply be an empty list. That way, autocompletion will naturally not suggest anything until the list is then swapped out with the real list from `updateCache` once it is ready.

So:

1. In the constructor (`GitHubCommand#69`) swap out `updateCache();` by something like:
```java
CompletableFuture.runAsync(this::updateCache);
```
To not miss any potential errors from this async path we can also instead use:
```java
CompletableFuture.runAsync(() -> {
    try {
        updateCache();
    } catch (Exception e) {
        logger.error("Unknown error updating the GitHub cache", e);
    }
});
```
2. Ensure the `List<String> autocompleteGHIssueCache` has a proper value for while the cache is building up, i.e. change its declaration in `GitHubCommand#48` by
```java
private List<String> autocompleteGHIssueCache = List.of();
```

[...]

PR #1317, opened 2025-08-30
GitHub API increased page size (30 to 1000), reduces load time (30s to 1s)

Alathreon:
As an alternative and add on https://github.com/Together-Java/TJ-Bot/issues/1315
By using the advice from https://github.com/hub4j/github-api/issues/2128

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/github/GitHubCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/github/GitHubCommand.java
@@ -4,7 +4,6 @@ import net.dv8tion.jda.api.events.interaction.command.CommandAutoCompleteInterac
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
 import net.dv8tion.jda.api.interactions.commands.OptionType;
 import org.kohsuke.github.GHIssue;
-import org.kohsuke.github.GHIssueState;
 
 import org.togetherjava.tjbot.features.CommandVisibility;
 import org.togetherjava.tjbot.features.SlashCommandAdapter;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/github/GitHubCommand.java
```diff
@@ -124,7 +123,7 @@ private ToIntFunction<String> suggestionScorer(String title) {
     private void updateCache() {
         autocompleteGHIssueCache = reference.getRepositories().stream().map(repo -> {
             try {
-                return repo.getIssues(GHIssueState.ALL);
+                return repo.queryIssues().pageSize(1000).list().toList();
```
SquidXTV, line 126: Why do we use page size `1000` here? Wouldn't that break with our repo?

Alathreon: Why would it break ?

SquidXTV: Because we have more than 1000 issues?

Alathreon: Ah you mean why not more ? Afaik 1000 is the max allowed by the github API.

SquidXTV: Oh, but does it make multiple queries then and merge them at the end?

Alathreon: It should yes, by default it is 30 (which means 44 requests...).

SquidXTV: Ah cool :)

PR #1325, opened 2025-09-12
Fix: audit message being empty for attachment-only scam posts (#1300)

### LINKED ISSUE #1300: Audit message for attachment-scam is empty
## Problem

When a scam-bot posts scam that consists only of attachments and not actual message content, our bot correctly detects it and sends a message for the staff to audit.

However, in this case this message is empty because it is coded to show message content, which is empty in this case:

message

This is only semi helpful. Ideally it should mention that there were attachments as well. Possibly even the names of the attachments (or their content, but that is optional).

## How to reproduce

Make sure you set your scam configuration to something that does not immediately put your user under quarantine, for example the default:

```json
"scamBlocker": {
  "mode": "AUTO_DELETE_BUT_APPROVE_QUARANTINE",
```

Then post a message containing 3 or more picture attachments and no text content. The attachments need to be named something like `image.jpg` or `1.jpg`, `2.jpg`, ...
This should be the default behavior in Discord if you just drag&drop an image file into ur chat window. So that should be the easiest way to trigger it.

The bot will now detect it as scam and send the confirmation message in the audit channel. The channel is set in the config, the default is:

```json
"scamBlocker": {
  "reportChannelPattern": "commands",
```

## Solution

The relevant code (`ScamBlocker.java#reportScamMessage`) needs to be adjusted to also include attachment details, not just the message content. This is the current code that makes this (and similar messages):

```java
MessageEmbed embed = new EmbedBuilder()
  .setDescription(event.getMessage().getContentStripped())
  .setTitle(reportTitle)
  .setAuthor(author.getName(), null, avatarOrDefaultUrl)
  .setTimestamp(event.getMessage().getTimeCreated())
  .setColor(AMBIENT_COLOR)
  .setFooter(author.getId())
  .build();
```

One can get hands on the attachments by doing:

```java
List<Message.Attachment> attachments = event.getMessage().getAttachments();
```

### Quick text adjustment

So we either adjust the first call, `.setDescription(...)` to also show something like `(and 3 attachments (foo.png, bar.jpg, baz.exe)` to keep it simple.

(the file name is available via `attachment.getFileName()`)

### Multiple embeds

Or even better, we add multiple embeds to the message, one per attachment. The message is build right below:
```java
MessageCreateBuilder messageBuilder = new MessageCreateBuilder().setEmbeds(embed);
```
Instead of only giving it a single embed, we could give it a list of embeds.

#### Simple

Each embed can just contain the full name of the attachment as text (the file name is available via `attachment.getFileName()`).

Maybe also add the URL so people can click on it to view it (`attachment.getProxyUrl()`).

#### As actual attachment (optional)

The best UX would be to actually upload the attachment as attachment again. This can be done by using

```java
MessageEmbed embed = new EmbedBuilder(...)
  .setThumbnail("attachment://" + name)
  .build();
```
and then linking that `name` to an actual data content in the message:
```java
MessageCreateBuilder messageBuilder = new MessageCreateBuilder()
  .setEmbeds(embed)
  .setFiles(...);
```
that `setFiles` expects a `List<FileUpload>`, which links names to data.

`FileUpload` objects can be obtained from `Attachment` by using something like:

```java
FileUpload fileUpload = attachment.getProxy().downloadAsFileUpload(100, 100);
```

That `fileUpload` then also has the name of the attachment which can then be set in the embed (`fileUpload.getName()`).

JankiGadhiya:
Fixed #1300

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
@@ -257,7 +257,7 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
                 .collect(Collectors.joining(", "));
             content += "%s(The message has %d attachment%s: %s)".formatted(
                     content.isBlank() ? "" : "\n", attachments.size(),
-                    attachments.size() > 1 ? "s " : " ", attachmentInfo);
+                    attachments.size() > 1 ? "s " : "", attachmentInfo);
         }
         MessageEmbed embed = new EmbedBuilder().setDescription(content)
             .setTitle(reportTitle)
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
@@ -255,8 +255,9 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
             String attachmentInfo = attachments.stream()
                 .map(Message.Attachment::getFileName)
                 .collect(Collectors.joining(", "));
-            content += (content.isEmpty() ? "" : " ") + "and " + attachments.size() + " attachment"
-                    + (attachments.size() > 1 ? "s " : " ") + "(" + attachmentInfo + ")";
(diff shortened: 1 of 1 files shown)


Zabuzard:
`license/cla` was signed (probably with wrong account but its fine)

Zabuzard:
Local testing:

### No text
empty message

### No attachments
no attachment

### Both
both

PR #1346, opened 2025-11-17
Java 25 update and other deps (Gradle, ...)

Zabuzard:
Updates from:

* Java 24 to Java 25 and
* Gradle 8.14.2 to 9.2.0

The gradle update brings some breaking changes that required updating a bunch of plugins as well.


SquidXTV (changes requested):
Probably best to update `wiki/Setup-project-locally.md` in this PR as we have it now in the repo :)

PR #1353, opened 2025-11-28
Fixes /reminder list fails if no reminders

### LINKED ISSUE #1350: /reminder list fails if no reminders
## Issue

When using the command `/reminder list` to show all current pending reminders for the user, the command throws an exception when the user has no pending reminders at the moment.

### StackTrace

```java
java.lang.IllegalArgumentException: 1 > 0
  at java.base/java.lang.Math.clamp(Math.java:2221)
  at org.togetherjava.tjbot.features.reminder.ReminderCommand.createPendingRemindersPage(ReminderCommand.java:223)
  at org.togetherjava.tjbot.features.reminder.ReminderCommand.handleListCommand(ReminderCommand.java:207)
  at org.togetherjava.tjbot.features.reminder.ReminderCommand.onSlashCommand(ReminderCommand.java:119)
  at org.togetherjava.tjbot.features.system.BotCore.lambda$onSlashCommandInteraction$0(BotCore.java:260)
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
  at java.base/java.lang.Thread.run(Thread.java:1474)
```

### Reproduce

Start the bot locally, do not use `/reminder create`. Instead, use `/reminder list` right away. Thats all, the exception should be thrown now.

## Bug

The issue is that this code section in `ReminderCommand.java:223`

clamp

calls `Math.clamp(x, 1, 0)`, i.e. `min = 1` and `max = 0`. That makes little sense, hence the crash.

## Fix

Funnily enough the code is already prepared to handle the case correctly, just a few lines later it has

```java
if (pendingReminders.isEmpty()) {
  remindersEmbed.setDescription("No pending reminders");
```

We just have to make sure that `totalPages` is at least `1`. That can also be justified logically as even with nothing to show, we at least want to show a single page. So we just change

```java
int totalPages = Math.ceilDiv(pendingReminders.size(), REMINDERS_PER_PAGE);
```

into

```java
int totalPages = Math.ceilDiv(pendingReminders.size(), REMINDERS_PER_PAGE);
totalPages = Math.max(1, totalPages);
```

which ensures it is at least `1`. The rest of the code should then work out of the box already.

Zachdehooge:
Fixes the `/reminder list` issue and returns the expected result of:

Screenshot 2025-11-28 at 15 46 33

Closes #1350

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/reminder/ReminderCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/reminder/ReminderCommand.java
@@ -219,6 +219,7 @@ public final class ReminderCommand extends SlashCommandAdapter {
             List<PendingRemindersRecord> pendingReminders, int pageToShow) {
         // 12 reminders, 10 per page, ceil(12 / 10) = 2
         int totalPages = Math.ceilDiv(pendingReminders.size(), REMINDERS_PER_PAGE);
+        totalPages = Math.max(1, totalPages);
 
         pageToShow = Math.clamp(pageToShow, 1, totalPages);
 


Zabuzard (approved):

Zabuzard:
@suryatejess the other way around. u got assigned to the issue first but @Zachdehooge just went ahead and made a PR, ignoring that you already "took the issue" 🙂

PR #1361, opened 2025-12-21
Feat/help thread stats

ankitsmt211:
continues work of #1159 with /help-stats slash command, refactors to fetch stats from database instead of discord APIs. Also adds stats to embed for better UX.

image

embed color has three states
* GREEN (high response rates)
* ORANGE (medium response rates)
* RED(low response rates)

edited: also adds per channel limit, can be used once per minute.


application/src/main/java/org/togetherjava/tjbot/features/Features.java
```diff
@@ -23,6 +23,7 @@
 import org.togetherjava.tjbot.features.filesharing.FileSharingMessageListener;
 import org.togetherjava.tjbot.features.github.GitHubCommand;
 import org.togetherjava.tjbot.features.github.GitHubReference;
+import org.togetherjava.tjbot.features.help.*;
```
tj-wazei: No star imports please.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,239 @@
...
+import org.togetherjava.tjbot.db.Database;
+import org.togetherjava.tjbot.features.CommandVisibility;
+import org.togetherjava.tjbot.features.SlashCommandAdapter;
+
+import java.awt.Color;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Objects;
+
+import static org.jooq.impl.DSL.*;
```
tj-wazei: No star imports please.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,243 @@
...
+
+        getData().addOptions(durationOption);
+        this.database = database;
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        long days = event.getOption(DURATION_OPTION) != null
+                ? Objects.requireNonNull(event.getOption(DURATION_OPTION)).getAsLong()
+                : 1;
```
SquidXTV: Can't you use [CommandInteractionPayload.html#getOption(java.lang.String,T,java.util.function.Function)](https://docs.jda.wiki/net/dv8tion/jda/api/interactions/commands/CommandInteractionPayload.html#getOption(java.lang.String,T,java.util.function.Function)) here instead?


```suggestion
        long days = event.getOption(DURATION_OPTION, 1L, OptionMapping::getAsLong)
```

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,243 @@
...
+                    .as(OPEN_NOW_ALIAS),
+                        count().filterWhere(HELP_THREADS.PARTICIPANTS.eq(1)).as(GHOST_NOW_ALIAS),
+                        avg(HELP_THREADS.PARTICIPANTS).as(AVERAGE_PARTICIPANTS_ALIAS),
+                        avg(HELP_THREADS.MESSAGE_COUNT).as(AVERAGE_MESSAGE_COUNT_ALIAS),
+                        avg(durationInSeconds(HELP_THREADS.CLOSED_AT, HELP_THREADS.CREATED_AT))
+                            .as(AVERAGE_THREAD_DURATION_IN_SECONDS_ALIAS),
+                        min(durationInSeconds(HELP_THREADS.CLOSED_AT, HELP_THREADS.CREATED_AT))
+                            .as(MINIMUM_THREAD_DURATION_IN_SECONDS_ALIAS),
+                        max(durationInSeconds(HELP_THREADS.CLOSED_AT, HELP_THREADS.CREATED_AT))
+                            .as(MAXIMUM_THREAD_DURATION_IN_SECONDS_ALIAS))
```
SquidXTV, line 129: Not really important as a review for this PR because the stats collecting is already in the code, but do we have a reliable way of getting that closed_at? What about reopened questions? Does it update it accordingly?

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,243 @@
...
+                    "Created: `%d`%nCurrently Open: `%d`%nResponse Rate: %.1f%%%nPeak Hours: `%s`"
+                        .formatted(totalCreated, openThreads, rawResRate, peakHourRange),
+                    false);
+
+            embed.addField(EMOJI_SPEECH_BUBBLE + WHITESPACE + "ENGAGEMENT",
+                    "Avg Messages: `%s`%nAvg Helpers: `%s`%nUnanswered (Ghost): `%d`".formatted(
+                            formatDouble(Objects
+                                .requireNonNull(statsRecord.get(AVERAGE_MESSAGE_COUNT_ALIAS))),
+                            formatDouble(Objects
+                                .requireNonNull(statsRecord.get(AVERAGE_PARTICIPANTS_ALIAS))),
```
SquidXTV: Isn't the number of helpers equal to participants - 1? Because the owner of the question doesn't count as a helper.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,243 @@
...
+            .where(HELP_THREADS.CREATED_AT.ge(start))
+            .and(HELP_THREADS.TAGS.ne("none"))
+            .groupBy(HELP_THREADS.TAGS)
+            .orderBy(order)
+            .limit(1)
+            .fetchOptional(HELP_THREADS.TAGS)
+            .orElse("N/A");
+    }
+
+    private String getPeakHour(DSLContext context, Instant start) {
```
SquidXTV: This is really nice idea and probably can be expanded into an actual activity graph over the whole 24h in a future PR

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
ankitsmt211: HelpThreadLifecycleListener#handleThreadStatus updated closed_at each time a thread is archived, so looking from a thread's lifecycle

closed for first time -> update closed_at, message count and participants at DB level
opened again -> mark thread as active at DB level
closed again due to inactivity/command -> update closed_at, message count and participants at DB level 

.....so on

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
ankitsmt211: that is correct, we already do that when saving data in HelpThreadLifecycleListener class

```
 int participantsExceptAuthor = threadChannel.getMemberCount() - 1; //here

        database.write(context -> context.update(HELP_THREADS)
            .set(HELP_THREADS.CLOSED_AT, closedAt)
            .set(HELP_THREADS.TICKET_STATUS, HelpSystemHelper.TicketStatus.ARCHIVED.val)
            .set(HELP_THREADS.MESSAGE_COUNT, messageCount)
            .set(HELP_THREADS.PARTICIPANTS, participantsExceptAuthor)
            .where(HELP_THREADS.CHANNEL_ID.eq(threadId))
            .execute());
  ```

File: application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java (hunk 3 above)
Comment (ankitsmt211) [reply application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java]: i think we can, will do that, thanks

File: application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java (hunk 5 above)
Comment (SquidXTV) [reply application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java]: Well actually, what about `count().filterWhere(HELP_THREADS.PARTICIPANTS.eq(1)).as(GHOST_NOW_ALIAS),` then?

Comment (ankitsmt211) [reply application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java]: ah good spot,  fixed now

File: application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java [hunk 7]
```diff
@@ -0,0 +1,272 @@
...
+            .addChoice("30 Days", 30)
+            .addChoice("90 Days", 90)
+            .addChoice("180 Days", 180);
+
+        getData().addOptions(durationOption);
+        this.database = database;
+        this.cooldownCache = Caffeine.newBuilder()
+            .maximumSize(500)
+            .expireAfterWrite(COOLDOWN_VALUE, TimeUnit.of(COOLDOWN_UNIT))
+            .build();
```
SquidXTV: Use `Duration` here instead, that way you only need one field

ankitsmt211: makes sense, took care of it.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,270 @@
...
+    private static final String MINIMUM_THREAD_DURATION_IN_SECONDS_ALIAS = "min_sec";
+    private static final String MAXIMUM_THREAD_DURATION_IN_SECONDS_ALIAS = "max_sec";
+
+    private static final String EMOJI_CHART = "\uD83D\uDCCA";
+    private static final String EMOJI_MEMO = "\uD83D\uDCDD";
+    private static final String EMOJI_SPEECH_BUBBLE = "\uD83D\uDCAC";
+    private static final String EMOJI_LABEL = "\uD83C\uDFF7\uFE0F";
+    private static final String EMOJI_LIGHTNING = "\u26A1";
+
+    private static final String EMBED_BLANK_LINE = "\u200B";
```
Zabuzard: please prefer having the actual UTF-8 emoji characters here, so the source code is readable.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,270 @@
...
+            .addChoice("180 Days", 180);
+
+        getData().addOptions(durationOption);
+        this.database = database;
+        this.cooldownCache =
+                Caffeine.newBuilder().maximumSize(500).expireAfterWrite(COOLDOWN_DURATION).build();
+    }
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
```
Zabuzard: that method is quite long. think about breaking it into maybe 2 or 3 helper methods? looks like there is a clear structure with: handle cooldown, database management, create and send response

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -0,0 +1,270 @@
...
+        double rawResRate =
+                totalCreated > 0 ? ((double) (totalCreated - ghostThreads) / totalCreated) * 100
+                        : -1;
+
+        if (rawResRate >= 70)
+            return Color.GREEN;
+        if (rawResRate >= 30)
+            return Color.YELLOW;
+        if (rawResRate >= 0)
+            return Color.RED;
```
Zabuzard: what is `Res`? improve name

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
ankitsmt211: done

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
ankitsmt211: thanks, refactored

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
ankitsmt211: Updated to a more suitable name

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadLifecycleListener.java
```diff
@@ -96,8 +103,13 @@ void handleArchiveStatus(Instant closedAt, long id, JDA jda) {
         }
 
         long threadId = threadChannel.getIdLong();
-        int messageCount = threadChannel.getMessageCount();
-        int participantsExceptAuthor = threadChannel.getMemberCount() - 1;
+        int messageCount = threadChannel.getMessageCount(); // TODO: to be replaced with participant
+                                                            // message count
```
Zabuzard, line 107: dangling todo. will this be done in this PR or in another? consider maybe making a gh issue if latter

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -129,108 +119,157 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
+    }
 
-            double rawResRate =
+
+    @Override
+    public void onSlashCommand(SlashCommandInteractionEvent event) {
+        long channelId = event.getChannel().getIdLong();
+        Instant now = Instant.now();
+        long secondsSinceLastUsage = getSecondsSinceLastUsage(channelId, now);
+        if (secondsSinceLastUsage != 0L) {
```
Zabuzard, line 200: i dont quite get this. feels to me like ur abusing `0` as magic value to indicate something.

like, lets say i used the command 2 days ago, then a method named `getSecondsSinceLastUsage` should return me the seconds equivalent for 2 days and not 0.

consider renaming this to maybe `getSecondsStillOnCooldown` or something similar

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -129,108 +119,157 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
 
-    private static Color getStatusColor(int totalCreated, long ghostThreads) {
-        double rawResRate =
-                totalCreated > 0 ? ((double) (totalCreated - ghostThreads) / totalCreated) * 100
-                        : -1;
+    private static Color getStatusColor(int totalHelpThreadsCreated, long ghostThreads) {
+        double helpThreadInteractionRate = totalHelpThreadsCreated > 0
+                ? ((double) (totalHelpThreadsCreated - ghostThreads) / totalHelpThreadsCreated)
+                        * 100
+                : -1;
```
Zabuzard, line 247: whats this? definitely needs a good name (or helper method) so this "formula" is easy to understand at first glance.

application/src/main/java/org/togetherjava/tjbot/features/help/HelpThreadStatsCommand.java
```diff
@@ -129,108 +119,157 @@ public void onSlashCommand(SlashCommandInteractionEvent event) {
...
-                .where(HELP_THREADS.CREATED_AT.ge(startDate))
+                .where(HELP_THREADS.CREATED_AT.ge(statsDurationStartDate))
                 .fetchOne();
 
             if (statsRecord == null || statsRecord.get(TOTAL_CREATED_FIELD, Integer.class) == 0) {
-                event.getHook()
-                    .editOriginal("No stats available for the last " + days + " days.")
-                    .queue();
-                return null;
+                return Optional.empty();
```
Zabuzard, line 135: im pretty sure there is some variant of that `fetchOne()` that also returns an `Optional` already. in which case the code below could be simplified by using the optional-api instead

PR #1362, opened 2025-12-21
fix: rss failures with exponential blacklist

ankitsmt211:
context: error logs being spammed by intermittent 404 status code on rss feed urls

image

failed rss feed urls are marked as blacklisted upto 24 hours, if it's a dead url it notifies with error log for admin to remove it.

How the fix works

* Failed RSS feed URLs are tracked in an in-memory cache along with their failure count and last failure timestamp.

* Each failure triggers an exponential backoff before the next retry: 1h, 2h, 4h, 8h, 16h, capped at 24 hours maximum.

* While a URL is within its backoff window, it is skipped and not fetched.

* If a fetch succeeds, the failure state is removed from cache and the URL resumes normal polling.

* Cache entries automatically expire after 7 days, preventing stale failures from persisting indefinitely.

* If a feed fails 15 times, it is considered potentially dead and an error is logged to notify admins to remove it from configuration.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
@@ -72,9 +72,6 @@ import static org.togetherjava.tjbot.db.generated.tables.RssFeed.RSS_FEED;
  */
 public final class RSSHandlerRoutine implements Routine {
 
-    private record FailureState(int count, ZonedDateTime lastFailure) {
-    }
-
     private static final Logger logger = LoggerFactory.getLogger(RSSHandlerRoutine.class);
     private static final int MAX_CONTENTS = 1000;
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
```diff
@@ -70,6 +72,9 @@
  */
 public final class RSSHandlerRoutine implements Routine {
 
+    private record FailureState(int count, ZonedDateTime lastFailure) {
+    }
```
christolis: Perhaps we could put this as its own file.

christolis (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
ankitsmt211: done

PR #1365, opened 2025-12-29
refactor(RSSHandlerRoutine): switch to Instant

christolis:
From a recent [conversation](https://discord.com/channels/272761734820003841/895717328359153664/1455140764903538881):

```log
[13:09]Zabuzard: while at it, ZonedDateTime should be replaced by Instant in that file
[13:10]Zabuzard: does someone want to tackle that quickly or else we need to make a github issue 🙂
[13:56]christolis: what makes this change worthwhile?
[14:01]Zabuzard: its just the wrong class for the job
[14:01]Zabuzard: there is no reason to use ZDT in this case
[14:01]Zabuzard: Instant is the correct class
[14:02]Zabuzard: shouldnt require any deeper changes in this case, just a quick replacement should work```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
@@ -303,7 +303,7 @@ public final class RSSHandlerRoutine implements Routine {
             throws DateTimeParseException {
         Optional<String> pubDate = item.getPubDate();
 
-        return pubDate.map(s -> getInstantTimeFromFormat(s, dateTimeFormat)).orElse(Instant.MIN);
+        return pubDate.map(s -> parseDateTime(s, dateTimeFormat)).orElse(Instant.MIN);
 
     }
 
@@ -325,7 +325,7 @@ public final class RSSHandlerRoutine implements Routine {
             // If this throws a DateTimeParseException then it's certain
             // that the format pattern defined in the config and the
             // feed's actual format differ.
-            getInstantTimeFromFormat(firstRssFeedPubDate.get(), feedConfig.dateFormatterPattern());
+            parseDateTime(firstRssFeedPubDate.get(), feedConfig.dateFormatterPattern());
         } catch (DateTimeParseException _) {
             return false;
         }
@@ -374,8 +374,8 @@ public final class RSSHandlerRoutine implements Routine {
 
         // Set the item's timestamp to the embed if found
         item.getPubDate()
-            .ifPresent(date -> embedBuilder
-                .setTimestamp(getInstantTimeFromFormat(date, feedConfig.dateFormatterPattern())));
+            .ifPresent(dateTime -> embedBuilder
+                .setTimestamp(parseDateTime(dateTime, feedConfig.dateFormatterPattern())));
 
         embedBuilder.setTitle(title, titleLink);
         embedBuilder.setAuthor(item.getChannel().getLink());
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by tj-wazei — DISMISSED** [review]
Looks reasonable to me.

File: application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java [hunk 1]
```diff
@@ -223,16 +219,13 @@ private Optional<RssFeedRecord> getRssFeedRecordFromDatabase(RSSFeed feedConfig)
...
      */
-    private Optional<ZonedDateTime> getLastSavedDateFromDatabaseRecord(RssFeedRecord rssRecord,
-            String dateFormatterPattern) throws DateTimeParseException {
+    private Optional<Instant> getLastSavedDateFromDatabaseRecord(RssFeedRecord rssRecord)
+            throws DateTimeParseException {
         try {
-            ZonedDateTime savedDate =
-                    getZonedDateTime(rssRecord.getLastDate(), dateFormatterPattern);
-            return Optional.of(savedDate);
+            return Optional.of(Instant.parse(rssRecord.getLastDate()));
```
ankitsmt211: this will throw `java.time.format.DateTimeParseException`, we should handle it instead of just silenty eating the exception. Something like
```
catch (DateTimeParseException ignored) {
        try {
            //TODO: Should be removed once we cycle through all stored ZDT string instances
            Instant parsed_instant = ZonedDateTime.parse(raw).toInstant();
            //INFO log here converted from raw -> instant
            // save parsed_instant to database so this doesnt trigger exception on next run
            return Optional.of(parsed_instant); //
        } catch (DateTimeParseException e) {
            return Optional.empty();
        }
    }
    
```

ankitsmt211 (changes requested):
everything looks good but we should handle the outdated timestamps stored in DB

application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
```diff
@@ -441,21 +432,19 @@ private List<Item> fetchRSSItemsFromURL(String rssUrl) {
...
+     * Helper function for parsing a given date value to a {@link Instant} with a given format.
      *
      * @param date the date value to parse, can be null
-     * @param format the format pattern to use for parsing
-     * @return the parsed {@link ZonedDateTime} object
+     * @return the parsed {@link Instant} object
      * @throws DateTimeParseException if the date cannot be parsed
      */
-    private static ZonedDateTime getZonedDateTime(@Nullable String date, String format)
+    private static Instant getInstantTimeFromFormat(@Nullable String date, String datePattern)
```
Zabuzard: btw, having the need to rename this on a type-change indicates that the method name is flawed.

it should be renamed into `getTimeFromFormat` or more idiomatic `parseTime`.

Also note the inconsistency with the parameter names called `date` instead of `time`. To avoid confusion it should proibably be renamed into `parseDateTime` and `String dateTime`, String pattern`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/rss/RSSHandlerRoutine.java
```diff
@@ -191,17 +189,24 @@ private Optional<Predicate<Item>> prepareItemPostPredicate(RSSFeed feedConfig,
...
+        } catch (DateTimeParseException _) {
+            Optional<Instant> convertedDate = convertDateTimeToInstant(feedConfig);
 
-        if (lastSavedDate.isEmpty()) {
-            return Optional.empty();
+            if (convertedDate.isEmpty()) {
+                return Optional.empty();
+            }
+
+            lastSavedDate = convertedDate.get();
```
ankitsmt211, line 202: this wouldnt be raised, if there's an error then database is corrupted maybe just log this and move on (silent handling is no longer needed)

ankitsmt211 (changes requested):
It seems my assumptions were wrong, time parsing wouldn't blow up
I have looked at the database, rss_feed stores Instant friendly timestamps for our active rss_feeds using "yyyy-MM-dd'T'HH:mm:ssXXX" datetime format from configuration.

So I have pointed out the error handling that we wouldn't need anymore. Apologies for the confusion, should have looked into this more carefully before putting out any comments.

Issue #1276, opened 2025-07-11
Replace broken links send by ChatGpt


Zabuzard:
## Problem

ChatGpt attempts to answer each question posted in `#question` and it provides "useful links", which is great.
However, and especially because ChatGpt runs with old models, it sometimes posts dead links. That leads to unecessary user frustration.

For example in this post:

![post](https://i.imgur.com/AWJINNX.png)

The first and last link are both 404 NOT FOUND.

## Proposed Solution

We should write code that either removes those links or at least replaces them, for example by `(broken link)`.

For that, we should add some helper methods to the already existing `LinkDetection` class and also use one of the already existing methods there:
* Add a method like `CompletableFuture<Boolean> isLinkDead(String url)` to it (which uses `HttpClient#sendAsync` to check the status. Dont download the full content, just check the status code, it must be 2xx or 3xx. Make sure this runs non-blocking, i.e. with a `CompletableFuture`.
* Add a method like `CompletableFuture<String> replaceDeadLinks(String text, String replacement)` that uses above methods and the existing `extractLinks` method to replace all dead links in the given text with the given replacement. Since `isLinkDead` runs async, this will run async as well with all futures combined

Once that is done, use that method for the ChatGpt responses. They are constructed in `HelpSystemHelper#145`. Ideally we send out the chatgpt response without delay, check the links async and then, when the result for that is available through the future, we **edit** the existing chatgpt response.

This can be done by adding more actions onto the queue in `HelpThreadCreatedListener#createAIResponse`. `RestAction` allows adding something like "and then do this" to it.
If we have trouble with this, we can also try to do the link replacement in a blocking way directly, without any edits. But that as a fallback.

Ideally all the helper methods added are all unit tested properly. For the `isLinkDead` we can mock the `HttpClient`, so no actual HTTP requests are send out.

Zabuzard:
@tferdous17 Any progress? Otherwise we would probably unassign you for now, so someone else could take it 🙂

barsh404error:
@christolis i will gladly contribute on this, is there anything i can do or just review it and add some changes if i can?

vlaym:
I've opened a PR that implements the fallback approach from the issue description - dead links are replaced before the embed is built, so the user never sees them.

The ideal flow would be: send the ChatGPT response immediately, check links in the background, then edit the already-sent message with clean links. But these two steps use different async mechanisms:
[...]

PR #1366, opened 2025-12-30
Add utilities to detect and replace broken links.

barsh404error:
Thanks alot to @christolis for helping me out on making this pull request.
Added two utulity methods `isLinkBroken` and `replaceDeadLinks`

-`isLinkBroken`(String url)  checks the link availability using a HEAD request
I used HEAD request instead of GET request to check link availability without downloading the response body, reducing bandwidth and improving the performance.

-`replaceDeadLinks` (String text, String replacement)  replaces unreachable/broken links asynchronously. 

This change does not have any behavior changes to the existing code.

Part of #1276, implements the mentioned utility but doesnt apply it.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
@@ -29,7 +29,6 @@ import java.util.concurrent.CompletableFuture;
  * It is intentionally stateless and uses asynchronous HTTP requests to avoid blocking calling
  * threads.
  */
-
 public class LinkDetection {
     private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
 
@@ -40,7 +39,6 @@ public class LinkDetection {
      * Links to intentionally ignore in order to reduce false positives when scanning chat messages
      * or source-code snippets.
      */
-
     private static final Set<LinkFilter> DEFAULT_FILTERS =
             Set.of(LinkFilter.SUPPRESSED, LinkFilter.NON_HTTP_SCHEME);
 
@@ -89,7 +87,6 @@ public class LinkDetection {
      * @param filter a set of filters controlling which detected links are returned
      * @return a list of extracted links in the order they appear in the text
      */
-
     public static List<String> extractLinks(String content, Set<LinkFilter> filter) {
         return new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect()
             .stream()
@@ -121,7 +118,6 @@ public class LinkDetection {
      * @param content the text to scan
      * @return {@code true} if at least one URL-like pattern is detected
      */
-
     public static boolean containsLink(String content) {
         return !(new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect().isEmpty());
     }
@@ -145,7 +141,6 @@ public class LinkDetection {
      * @return a {@code CompletableFuture} completing with {@code true} if the link is broken,
      *         {@code false} otherwise
      */
-
     public static CompletableFuture<Boolean> isLinkBroken(String url) {
         // Try HEAD request first (cheap and fast)
         HttpRequest headRequest = HttpRequest.newBuilder(URI.create(url))
@@ -210,7 +205,6 @@ public class LinkDetection {
      * @return a {@code CompletableFuture} that completes with the modified text, or the original
      *         text if no broken links were found
      */
-
     public static CompletableFuture<String> replaceBrokenLinks(String text, String replacement) {
         List<String> links = extractLinks(text, DEFAULT_FILTERS);
 
@@ -254,7 +248,6 @@ public class LinkDetection {
      * @return an {@link Optional} containing the normalized link, or {@code Optional.empty()} if
      *         the link should be filtered out
      */
-
     private static Optional<String> toLink(Url url, Set<LinkFilter> filter) {
         String raw = url.getOriginalUrl();
         if (filter.contains(LinkFilter.SUPPRESSED) && raw.contains(">")) {


application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -57,6 +63,54 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
     public static boolean containsLink(String content) {
         return !(new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect().isEmpty());
     }
+    public static CompletableFuture<Boolean> isLinkBroken(String url) {
+        HttpClient client = HttpClient.newHttpClient();
```
tj-wazei: Can you define the `HttpClient` as a static reference in this class please? 

i.e. 

```java
private static final HttpClient CLIENT = HttpClient.newHttpClient();

public static CompletableFuture<Boolean> isLinkBroken(String url) {
    HttpRequest request = HttpRequest.newBuilder(URI.create(url))
            .method("HEAD", HttpRequest.BodyPublishers.noBody())
            .build();

    return CLIENT.sendAsync(request, HttpResponse.BodyHandlers.discarding())
            .thenApply(response -> response.statusCode() >= 400)
            .exceptionally(ignored -> true);
}
```
Creating a new HttpClient per link is unnecessary and inefficient. HttpClient is thread-safe and designed to be reused across concurrent requests.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -57,6 +63,54 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
...
+
+        return client.sendAsync(request, HttpResponse.BodyHandlers.discarding())
+                .thenApply(response -> response.statusCode() >= 400)
+                .exceptionally(ignored -> true);
+    }
+    public static CompletableFuture<String> replaceDeadLinks(
+            String text,
+            String replacement
+    ) {
+        Set<LinkFilter> filters = Set.of(
```
tj-wazei: There's no need to always re-create this `Set` as it never changes, we can make this static-final.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -57,6 +63,54 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
...
+                LinkFilter.NON_HTTP_SCHEME
+        );
+
+        List<String> links = extractLinks(text, filters);
+
+        if (links.isEmpty()) {
+            return CompletableFuture.completedFuture(text);
+        }
+
+        StringBuilder result = new StringBuilder(text);
```
tj-wazei: This implementation is not thread-safe. StringBuilder is mutated from multiple async callbacks, causing data races and invalid index calculations. Async HTTP completion order is nondeterministic, so index-based replacements are unsafe.

The correct approach is to separate concurrent link checks from sequential string mutation: perform all HTTP checks first, then apply replacements in a single thread once all futures complete.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -57,6 +63,54 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
     public static boolean containsLink(String content) {
         return !(new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect().isEmpty());
     }
+    public static CompletableFuture<Boolean> isLinkBroken(String url) {
+        HttpClient client = HttpClient.newHttpClient();
+
+        HttpRequest request = HttpRequest.newBuilder(URI.create(url))
+                .method("HEAD", HttpRequest.BodyPublishers.noBody())
```
tj-wazei: While, it's a good idea to use `HEAD` requests, a lot of HTTP servers do not implement this properly, you might commonly get 405, 404 etc., and CDNs like Cloudflare actually block it as suspicious. It's best to use a hybrid approach here, try `HEAD` first and if that fails try `GET` before entirely failing.

tj-wazei (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Thanks a lot for the detailed review, I really appreciate it @tj-wazei 

I’ve updated the code to reuse a `private static final HttpClient`

I see the issue with mutating a shared `StringBuilder` from async callbacks — I’m currently reworking `replaceDeadLinks(..)` to separate the async link checks from the sequential text replacement so it’s thread-safe.

I’ll push an updated version shortly when am done. once again thanks for pointing those out!

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,59 @@ public static boolean containsLink(String content) {
...
+                }
+                HttpRequest getRequest = HttpRequest.newBuilder(URI.create(url)).GET().build();
+
+                return HTTP_CLIENT.sendAsync(getRequest, HttpResponse.BodyHandlers.discarding())
+                    .thenApply(resp -> resp.statusCode() >= 400)
+                    .exceptionally(ignored -> true);
+            });
+    }
+
+    public static CompletableFuture<String> replaceDeadLinks(String text, String replacement) {
```
christolis: It does not seem that we are using this method anywhere. Perhaps remove it or call it from somewhere.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,59 @@ public static boolean containsLink(String content) {
         return !(new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect().isEmpty());
     }
 
+    @SuppressWarnings("java:S2095")
```
christolis: What do we need this here for?

surajkumar: I'm going to assume it's on the HttpClient

barsh404error: > What do we need this here for?

i did that bc of the sonarlint thing for the first time i made the pr, i had to do checks i couldnt run the sonarling and spotlessapply it told me to make changes and i added that to solve it.

barsh404error: i removed that line, so the problem is solved :]

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: > It does not seem that we are using this method anywhere. Perhaps remove it or call it from somewhere.

yes i understand but removing it causes other classes to get ruined so i can't remove it i will just keep it.

tj-wazei: @barsh404error can you explain? It's a completely unused method.

image

In your PR you are adding this code but it's not actually being used anywhere.

tj-wazei: At first I was under the impression the code changes in this PR are being consumed but it seems like it's just adding helpers and that's it?

barsh404error: @tj-wazei answered your question on discord hope you understand,
i will want this to be merged :)

barsh404error: > > It does not seem that we are using this method anywhere. Perhaps remove it or call it from somewhere.
> 
> yes i understand but removing it causes other classes to get ruined so i can't remove it i will just keep it.

sorry i checked the wrong method not ReplaceDeadLinks

christolis: > answered your question on discord hope you understand,
> i will want this to be merged :)

@tj-wazei @barsh404error What was the resolution? Please either link to the discussion you have had or hold discussions relevant to the code reviews in GitHub to maintain a single source of truth and to make it easier for other contributors to catch up.

tj-wazei: He said he'll work on it.

barsh404error: > It does not seem that we are using this method anywhere. Perhaps remove it or call it from somewhere.

`replaceDeadLinks` is part of the solution for issue #1276 and is intended to be used when processing chatGPT responses that may contain outdated or broken links.
so it's needed to be kept @christolis it's linked to the issue. And The issue should be linked to the pr which @tj-wazei already did that for me :]

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+        HttpRequest headRequest = HttpRequest.newBuilder(URI.create(url))
+            .method("HEAD", HttpRequest.BodyPublishers.noBody())
+            .build();
+
+        return HTTP_CLIENT.sendAsync(headRequest, HttpResponse.BodyHandlers.discarding())
+            .thenApply(response -> {
+                int status = response.statusCode();
+                return status < 200 || status >= 400;
+            })
+            .exceptionally(ignored -> true)
```
Zabuzard: the idiomatic name for something you ignore is `_`

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+        }
+
+        List<CompletableFuture<String>> deadLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link)
+                .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? link : null))
+
+            .toList();
+
+        return CompletableFuture.allOf(deadLinkFutures.toArray(new CompletableFuture[0]))
```
Zabuzard: `toArray(new Foo[0])` is an anti-pattern. use method references instead `toArray(Foo[]::new)`

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+
+        if (links.isEmpty()) {
+            return CompletableFuture.completedFuture(text);
+        }
+
+        List<CompletableFuture<String>> deadLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link)
+                .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? link : null))
+
```
Zabuzard, line 221: empty line in the middle of the flow

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -4,14 +4,24 @@
...
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Utility class to detect links.
  */
 public class LinkDetection {
+    private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+
+    private static final Set<LinkFilter> DEFAULT_FILTERS =
+            Set.of(LinkFilter.SUPPRESSED, LinkFilter.NON_HTTP_SCHEME);
```
Zabuzard, line 43: improve name. what does this filter, what does it mean, what is it used for?

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
         return !(new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect().isEmpty());
     }
 
+    /**
+     * Checks whether the given URL is considered broken.
+     *
+     * <p>
+     * A link is considered broken if:
+     * <ul>
+     * <li>The URL is invalid or malformed</li>
```
Zabuzard: this is incorrect, isnt it? did you test it? `URI.create(url)` throws on malformed URLs, doesnt it. so your documentation for that edge case is wrong then.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+     * <li>Status code {@code 200} is considered valid, even if the response body is empty</li>
+     * <li>The response body content is not inspected</li>
+     * </ul>
+     *
+     * @param url the URL to check
+     * @return a future completing with {@code true} if the link is broken
+     */
+
+    public static CompletableFuture<Boolean> isLinkBroken(String url) {
+        HttpRequest headRequest = HttpRequest.newBuilder(URI.create(url))
```
Zabuzard, line 146: improve name `headRequest` documents the "how" not the "what/why". a better name would be `checkLinkHeadRequest`

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+            .exceptionally(ignored -> true)
+            .thenCompose(result -> {
+                if (!Boolean.TRUE.equals(result)) {
+                    return CompletableFuture.completedFuture(false);
+                }
+                HttpRequest getRequest = HttpRequest.newBuilder(URI.create(url)).GET().build();
+                return HTTP_CLIENT.sendAsync(getRequest, HttpResponse.BodyHandlers.discarding())
+                    .thenApply(resp -> resp.statusCode() >= 400)
+                    .exceptionally(ignored -> true); // still never null
+            });
```
Zabuzard, line 168: the logic doesnt correctly handle HTTP MOVED, it needs to follow up a MOVED resource to check the real resource for being valid or not

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+            .thenApply(response -> {
+                int status = response.statusCode();
+                return status < 200 || status >= 400;
+            })
+            .exceptionally(ignored -> true)
+            .thenCompose(result -> {
+                if (!Boolean.TRUE.equals(result)) {
+                    return CompletableFuture.completedFuture(false);
+                }
+                HttpRequest getRequest = HttpRequest.newBuilder(URI.create(url)).GET().build();
```
Zabuzard: same name issue with that one. better is `checkLinkGetRequest`

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard, line 146: I find the approach used here confusing to read top to bottom. a comment would be super helpful to make it easier to understand:
```java
// Quick request with HEAD, then GET to make sure
```
otherwise u read from top to bottom and get super confused what the second request is and why its there and what the difference is

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+     * broken
+     * http://workinglink/1
+     * }</pre>
+     *
+     * @param text the input text containing URLs
+     * @param replacement the string to replace broken links with
+     * @return a future containing the modified text
+     */
+
+    public static CompletableFuture<String> replaceDeadLinks(String text, String replacement) {
```
Zabuzard: improve naming. avoid multiple terms for the same think. you called it `isLinkBroken` and now you call it `replaceDeadLinks`. either dead or broken, not both. pick one and align the other

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+        List<String> links = extractLinks(text, DEFAULT_FILTERS);
+
+        if (links.isEmpty()) {
+            return CompletableFuture.completedFuture(text);
+        }
+
+        List<CompletableFuture<String>> deadLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link)
+                .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? link : null))
```
Zabuzard, line 219: `isBroken` cant be `null`, so just `isBroken ? link : null` works and is easier to read

Zabuzard: better: instead of `map(foo -> ...thenApply(...))`, use `.map(foo -> ...).filter(...)` then you also dont have all these `null` items in ur list, polluting it

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+
+            .toList();
+
+        return CompletableFuture.allOf(deadLinkFutures.toArray(new CompletableFuture[0]))
+            .thenApply(ignored -> deadLinkFutures.stream()
+                .map(CompletableFuture::join)
+                .filter(Objects::nonNull)
```
Zabuzard, line 225: `.filter(Objects::nonNull)` that one isnt needed anymore with the above fix

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+
+        List<CompletableFuture<String>> deadLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link)
+                .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? link : null))
+
+            .toList();
+
+        return CompletableFuture.allOf(deadLinkFutures.toArray(new CompletableFuture[0]))
+            .thenApply(ignored -> deadLinkFutures.stream()
```
Zabuzard: (`_` instead of `ignored`)

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -58,6 +68,106 @@ public static boolean containsLink(String content) {
...
+                .map(CompletableFuture::join)
+                .filter(Objects::nonNull)
+                .toList())
+            .thenApply(deadLinks -> {
+                String result = text;
+                for (String deadLink : deadLinks) {
+                    result = result.replace(deadLink, replacement);
```
Zabuzard: performance trap. have a quick check if `StringBuilder` provides `replace`. if not, u can keep it, its not that big of a deal in this case, i guess.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -74,39 +73,46 @@ public static boolean containsLink(String content) {
...
      * A link is considered broken if:
      * <ul>
-     * <li>The URL is invalid or malformed</li>
      * <li>An HTTP request fails</li>
      * <li>The HTTP response status code is outside the 200–399 range</li>
      * </ul>
      *
      * <p>
+     * The method first performs an HTTP {@code HEAD} request and falls back to an HTTP {@code GET}
+     * request if the {@code HEAD} request indicates a failure.
```
Zabuzard: implementation detail like that goes into a comment inside the method, not in the javadoc 👍

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -135,29 +145,28 @@ public static CompletableFuture<Boolean> isLinkBroken(String url) {
      * http://workinglink/1
      * }</pre>
      *
-     * @param text the input text containing URLs
-     * @param replacement the string to replace broken links with
+     * @param text the input text containing URLs (must not be {@code null})
+     * @param replacement the string to replace broken links with (must not be {@code null})
```
Zabuzard: u dont need to write that as ur params are all (implicitly) annotated with `@NonNull` already :)

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -135,29 +145,28 @@ public static CompletableFuture<Boolean> isLinkBroken(String url) {
...
             return CompletableFuture.completedFuture(text);
         }
 
-        List<CompletableFuture<String>> deadLinkFutures = links.stream()
+        List<CompletableFuture<Optional<String>>> deadLinkFutures = links.stream()
             .distinct()
             .map(link -> isLinkBroken(link)
-                .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? link : null))
-
+                .thenApply(isBroken -> isBroken ? Optional.of(link) : Optional.<String>empty()))
```
Zabuzard: u misunderstood me. its better if u simply filter out these, skip them here instead of later.
`.filter(...)` so they are not even part of the list

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Renamed ignored lambda parameters to `_` where applicable (e.g. `exceptionally(_ -> ...)`, t`henApply(_ -> ...)`) to clearly indicate intentional non-usage.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Replaced `toArray(new CompletableFuture[0])` with `toArray(CompletableFuture[]::new)` as suggested :]

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: done 👍

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Renamed to DEFAULT_EXTRACT_FILTERS and added a short comment explaining that it defines the default filters used during link extraction

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Fixed the Javadoc to no longer claim malformed URLs are handled as “broken links” The documentation now reflects that `URI.create(url)` throws for invalid URLs.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Renamed getRequest to getFallbackRequest 👍

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
barsh404error: Added an inline comment explaining the HEAD → GET fallback strategy to make the control flow easier to read and understand :]

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -17,9 +17,13 @@
  * Utility class to detect links.
  */
 public class LinkDetection {
-    private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+    private static final HttpClient HTTP_CLIENT =
+            HttpClient.newBuilder().followRedirects(HttpClient.Redirect.NORMAL).build();
 
-    private static final Set<LinkFilter> DEFAULT_FILTERS =
+    /**
+     * Default filters used when extracting links: skip suppressed URLs and non-http schemes.
```
Zabuzard: You are repeating what the code already says, this javadoc isnt helpful. Instead, make an example so people understand the **why** 👍

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -164,16 +167,20 @@ public static CompletableFuture<String> replaceDeadLinks(String text, String rep
...
             .thenApply(deadLinks -> {
-                String result = text;
+                StringBuilder sb = new StringBuilder(text);
                 for (String deadLink : deadLinks) {
-                    result = result.replace(deadLink, replacement);
+                    int idx = sb.indexOf(deadLink);
+                    while (idx != -1) {
+                        sb.replace(idx, idx + deadLink.length(), replacement);
+                        idx = sb.indexOf(deadLink, idx + replacement.length());
+                    }
```
Zabuzard: i find that code a bit too complex to just put it between everything else. move it into a quick `private static` helper method. sth like `private static void replace(StringBuilder haystack, String needle, String replacement)`

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: you forgot one:
```java
.exceptionally(ignored -> true); // still never null
```

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -55,7 +68,120 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
...
+            return CompletableFuture.completedFuture(text);
+        }
+
+        List<CompletableFuture<Optional<String>>> deadLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link)
+                .thenApply(isBroken -> isBroken ? Optional.of(link) : Optional.<String>empty()))
+            .toList();
+
+
```
Zabuzard: double empty line

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -4,27 +4,69 @@
...
+     * <li>Non-HTTP(S) schemes such as {@code ftp://} or {@code file://}</li>
+     * </ul>
      *
-     * @see LinkDetection
+     * <p>
+     * This reduces false positives when scanning chat messages or source-code snippets.
+     */
+
+    private static final Set<LinkFilter> DEFAULT_FILTERS =
+            Set.of(LinkFilter.SUPPRESSED, LinkFilter.NON_HTTP_SCHEME);
```
Zabuzard, line 43: dont list the current content of the filters. the problem with such a doc is that it easily rots.

like, in two years someone will have added a new filter and removed another filter, without adjusting the javadoc... oops.

its great that you have explained the individual filters a few lines below, thats enough. for the `DEFAULT_FILTERS` i would instead just describe:

> Links to intentionally ignore in order to reduce false positives when scanning chat messages or source-code snippets.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -34,12 +76,26 @@ private LinkDetection() {
     }
 
     /**
-     * Extracts all links from the given content.
+     * Extracts HTTP(S) links from the given text.
```
Zabuzard: too specific. this assumes that the link-filter remains setup as it currently is. what if in the future its also running on ftp links, then this javadoc will definitely be forgotten to be adjusted and be wrong then.

just keep it as

> Extracts links from the given text

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -34,12 +76,26 @@ private LinkDetection() {
...
+     * }</pre>
+     *
+     * @param content the text to scan for links
+     * @param filter a set of filters controlling which detected links are returned
+     * @return a list of extracted links in the order they appear in the text
      */
+
     public static List<String> extractLinks(String content, Set<LinkFilter> filter) {
         return new UrlDetector(content, UrlDetectorOptions.BRACKET_MATCH).detect()
             .stream()
```
Zabuzard, line 92: make an overload of this method that uses the default filter right away.

```java
public static List<String> extractLinks(String content) {
  return extractLinks(content, DEFAULT_LINK_FILTER);
}
```

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -49,15 +105,157 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
...
 
+    /**
+     * Asynchronously checks whether a URL is considered broken.
+     *
+     * <p>
+     * The check is performed in two steps:
+     * <ol>
+     * <li>A {@code HEAD} request is sent first (cheap and fast)</li>
+     * <li>If that fails or returns an error, a {@code GET} request is used as a fallback</li>
+     * </ol>
```
Zabuzard: implementation detail. irrelevant for the user of the method. relevant for the maintainers of the code. so move this part as comment into the method instead.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -49,15 +105,157 @@ public static List<String> extractLinks(String content, Set<LinkFilter> filter)
...
+
+    /**
+     * Converts a detected {@link Url} into a normalized link string.
+     *
+     * <p>
+     * Applies the provided {@link LinkFilter}s:
+     * <ul>
+     * <li>{@link LinkFilter#SUPPRESSED} - filters URLs wrapped in angle brackets</li>
+     * <li>{@link LinkFilter#NON_HTTP_SCHEME} - filters non-HTTP(S) schemes</li>
+     * </ul>
```
Zabuzard: this assumes the content of the filter to remain the same over the years, dont do that. instead just link the variable.

in fact, the doc is already incorrect as it does currently NOT apply these filters. it applies what the user gives as parameter.

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
Zabuzard: not done yet

application/src/main/java/org/togetherjava/tjbot/features/utils/LinkDetection.java
```diff
@@ -210,31 +211,31 @@ public static CompletableFuture<Boolean> isLinkBroken(String url) {
...
         }
 
-        List<CompletableFuture<Optional<String>>> brokenLinkFutures = links.stream()
-                .distinct()
-                .map(link -> isLinkBroken(link)
-                        .thenApply(isBroken -> Boolean.TRUE.equals(isBroken) ? Optional.of(link) : Optional.<String>empty()))
-                .toList();
+        List<CompletableFuture<String>> brokenLinkFutures = links.stream()
+            .distinct()
+            .map(link -> isLinkBroken(link).thenApply(isBroken -> isBroken ? link : null))
```
Zabuzard: ur still throwing trash into the list. dont put nulls in there. filter it out RIGHT HERE so it does not even end up in the list from line 222.

once u correctly did that, line 230 where u filter for nulls will be completely obsolete because your list then does not even contain trash to BEGIN WITH :)

Zabuzard (approved):

PR #1370, opened 2026-01-04
feat(DynamicVoiceChat): implement main logic

### LINKED ISSUE #1113: Dynamically create and delete voice channels based on activity

christolis:
Closes #1113.

## Showcase

https://github.com/user-attachments/assets/2ae3b2d8-4653-47de-9492-2a6d1ba09c78

# Configuration changes
| Property 	| Description 	| Type |
|---	|---	|---
| `dynamicChannelPatterns` 	| List of voice channel patterns to manage dynamically. 	| List[String] |
| `archiveCategoryPattern` 	| The archive category name in which archived voice chats will go under. 	| String |
| `cleanChannelsAmount` 	| When an archive cleanup is triggered, how many of the oldest channels to remove. 	| int |
| `minimumChannelsAmount` 	| How many channels the archive category should have before a cleanup is triggered. 	| int |

## ⚠️
The category should obviously only be visible by moderators, so maintainer please make sure that's the case.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
@@ -17,11 +17,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.config.DynamicVoiceChatConfig;
 import org.togetherjava.tjbot.features.VoiceReceiverAdapter;
 
-import java.util.List;
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 /**
  * Handles dynamic voice channel creation and deletion based on user activity.
@@ -33,18 +32,15 @@ import java.util.regex.Pattern;
 public final class DynamicVoiceChat extends VoiceReceiverAdapter {
     private static final Logger logger = LoggerFactory.getLogger(DynamicVoiceChat.class);
 
-    private static final String ARCHIVE_CATEGORY_NAME = "Voice Channel Archives";
-    private static final int CLEAN_CHANNELS_AMOUNT = 25;
-    private static final int MINIMUM_CHANNELS_AMOUNT = 50;
-
     private final VoiceChatCleanupStrategy voiceChatCleanupStrategy;
-    private final List<Pattern> dynamicVoiceChannelPatterns;
+    private final DynamicVoiceChatConfig dynamicVoiceChannelConfig;
 
     public DynamicVoiceChat(Config config) {
-        this.dynamicVoiceChannelPatterns =
-                config.getDynamicVoiceChannelPatterns().stream().map(Pattern::compile).toList();
+        this.dynamicVoiceChannelConfig = config.getDynamicVoiceChatConfig();
+
         this.voiceChatCleanupStrategy =
-                new OldestVoiceChatCleanup(CLEAN_CHANNELS_AMOUNT, MINIMUM_CHANNELS_AMOUNT);
+                new OldestVoiceChatCleanup(dynamicVoiceChannelConfig.cleanChannelsAmount(),
+                        dynamicVoiceChannelConfig.minimumChannelsAmount());
     }
 
     @Override
@@ -74,7 +70,8 @@ public final class DynamicVoiceChat extends VoiceReceiverAdapter {
     }
 
     private boolean eventHappenOnDynamicRootChannel(AudioChannelUnion channel) {
-        return dynamicVoiceChannelPatterns.stream()
+        return dynamicVoiceChannelConfig.dynamicChannelPatterns()
+            .stream()
             .anyMatch(pattern -> pattern.matcher(channel.getName()).matches());
     }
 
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
```diff
@@ -0,0 +1,114 @@
...
+                    channel.getName(), memberCount);
+            return;
+        }
+
+        channel.delete()
+            .queue(_ -> logger.trace("Deleted dynamically created voice channel: {} ",
+                    channel.getName()),
+                    error -> logger.error("Failed to delete dynamically created voice channel: {} ",
+                            channel.getName(), error));
+    }
```
SquidXTV: I think I would kinda prefer if this workflow will post a warning in the voice channel chat with a specific timer for deletion (30sec) or whatever, then checking again for `memberCount == 0` and then actually deleting or otherwise stopping the workflow

tj-wazei: I thought about this when reviewing and decided to skip past a timer. There is a message sent when the channel is created and there's really no need to keep the channel longer. It can be recreated right away anyway. 

Having said this, I am now thinking about moderation... 

Temp voice chat -> users posting nonsense -> evidence deleted ...? 

Perhaps instead of deleting the channel, we can move it to a private/mod-only `voice-chat-archive` channel so only we can still check chats. Followed by a daily clean up task. 

What do you think?

christolis: That sounds great, we need to keep track of what people post in these ephemeral channels. Perhaps instead of deleting, send a log of all the messages sent as one embed.

tj-wazei: You'll hit the character limit very quickly with an embed. It's best to just hide the channel so only moderators can see this and clean up after some time period.

Zachdehooge:
Would love to carry this feature forward, got permission from @christolis

application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
SquidXTV: I would say it is best if we keep it as it is now and add such functionality later on

christolis: I will try to push a temporary solution today so that moderation is still possible with these ephemeral voice channels. Feel free to resolve this thread once you are fine with the pushed implementation.

application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
```diff
@@ -0,0 +1,161 @@
...
+            MessageHistory messageHistory = channelLeft.asVoiceChannel().getHistory();
+            messageHistory.retrievePast(2).queue(messages -> {
+                // Don't forget that there is always one
+                // embed message sent by the bot every time.
+                if (messages.size() > 1) {
+                    archiveDynamicVoiceChannel(channelLeft);
+                } else {
+                    channelLeft.delete().queue();
+                }
+            });
```
SquidXTV, line 68: Can't you simply use `MessageHistory#size` instead?

SquidXTV (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
christolis: In order to make use of `MessageHistory#size`, a method like `MessageHistory#retrievePast` needs to be invoked first because otherwise the size method will always return 0 regardless of the channel's history.

application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
```diff
@@ -0,0 +1,161 @@
...
+ * Handles dynamic voice channel creation and deletion based on user activity.
+ * <p>
+ * When a member joins a configured root channel, a temporary copy is created and the member is
+ * moved into it. Once the channel becomes empty, it is archived and further deleted using a
+ * {@link VoiceChatCleanupStrategy}.
+ */
+public final class DynamicVoiceChat extends VoiceReceiverAdapter {
+    private static final Logger logger = LoggerFactory.getLogger(DynamicVoiceChat.class);
+
+    // @christolis: Unless somebody is willing to make the category name configurable,
```
tj-wazei: Please remove the comment and adds these configurations into the config file. Thanks

tj-wazei (changes requested):

PR #1371, opened 2026-01-04
VoiceReceiverAdapter.java: add default missing doc

christolis:
# Why
Sonarlint wants every public class to have a JavaDoc.


tj-wazei (approved):

tj-wazei:
>  * {@link #onVoiceUpdate(GuildVoiceUpdateEvent)} like the other provided methods can be overridden
>  * if desired. The default implementation is empty, the adapter will not react to such events.

Seems unnecessary but approved anyway, the intent is there.  

But if you want to make it better...

> This class follows the <em>adapter pattern</em>: it provides empty (no-op) implementations for all {@link VoiceReceiver} callback methods, allowing subclasses to override only the events they are interested in.

ankitsmt211 (approved):

PR #1374, opened 2026-01-06
Extract secrets out of config.json and into a secrets.json

tj-wazei:

## Pull-request

### Changes

- [X] Existing code
- [X] New feature

Closes Issue: NaN

## Description
The problem we currently have is that whenever there is a config change, somebody with access to the VPS has to manually update the `config.json` so that we can add config the bot needs to start.

This for obvious reasons doesn't make sense for non-secret values (e.g. API URLs, channel names etc.) thus this PR separates anything secretive like API keys into a `secrets.json`.

`config.json` has been moved under `application/src/main/resources`

Goal here is that we only touch the VPS when needed and not for trivial changes that can be git-ops. 

Before this PR can be merged we need to:

* Rename `config.json.template` to `config.json` and copy over the values from the VPS to this repository.

This has not yet been done as I want confirmation that nothing else secretive remains in the `config.json.template`!

Post merge:

* Create a `secrets.json` on the VPS containing just the secrets that were removed from the `config.json` in this PR.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/Application.java
+++ b/application/src/main/java/org/togetherjava/tjbot/Application.java
@@ -36,7 +36,8 @@ public class Application {
     }
 
     private static final Logger logger = LoggerFactory.getLogger(Application.class);
-    private static final String DEFAULT_CONFIG_PATH = "config.json";
+    private static final String DEFAULT_CONFIG_PATH_DISK = "config.json";
+    private static final String DEFAULT_CONFIG_PATH_RESOURCES = "/config.json";
     private static final String DEFAULT_SECRETS_PATH = "secrets.json";
 
     /**
@@ -64,9 +65,9 @@ public class Application {
         if (args.length > 0) {
             configPath = args[0];
         } else if (useIncludedConfig) {
-            configPath = "/" + DEFAULT_CONFIG_PATH;
+            configPath = DEFAULT_CONFIG_PATH_RESOURCES;
         } else {
-            configPath = DEFAULT_CONFIG_PATH;
+            configPath = DEFAULT_CONFIG_PATH_DISK;
         }
 
         Config config;


Zabuzard (changes requested):
The general motivation and idea for this makes sense, however:

- iirc there is a precommit hook or sth that attempts to create the config.json out of the template on first launch, prob need to adjust that
- the documenation (wiki) needs to be adjusted. since the wiki is part of the code base, can do that right away in this PR as well
- one main reason we had the config outside of github control so far was that we are able to change it without needing to do a PR, merging that through branches and making a release. We must remain able to "oh shit, we need to deactivate this feature, its on rampage!" without having to go through a release. Im not sure if this new approach has that in mind. Maybe a hybrid would work as well. I guess you could argue that in such a case you would just "force" push a commit straight on the master branch, forcing the config change without proper a proper release flow. That technically works but then develop and master history diverge, which will cause unecessary trouble. Anyways, sth to think about.

application/src/main/resources/config.json
```diff
@@ -1,111 +1,45 @@
 {
-    "databasePath": "local-database.db",
+    "databasePath": "/home/bot/database/database.db",
```
Zabuzard, line 2: problematic for users local setup? especially if not linux id imagine

application/src/main/resources/config.json
```diff
@@ -1,111 +1,45 @@
...
+        "channelPattern": "server-suggestions",
+        "upVoteEmoteName": "upvote",
+        "downVoteEmoteName": "downvote"
     },
     "quarantinedRolePattern": "Quarantined",
     "scamBlocker": {
-        "mode": "AUTO_DELETE_BUT_APPROVE_QUARANTINE",
-        "reportChannelPattern": "commands",
-        "botTrapChannelPattern": "bot-trap",
+        "mode": "AUTO_DELETE_AND_QUARANTINE",
```
Zabuzard, line 18: mh, that one doesnt fit the local use-case that well. namely people working on the feature, for example testing new scam code, they shouldnt get autoquarantined by the bot

tj-wazei:
Updated the config.json to match what's on the VPS. 

And on the VPS I have added the secrets.json for dev/master

application/src/main/resources/config.json
tj-wazei: How do you suggest doing this? Updating docs? Changing the path? With a git-ops route, whatever is versioned controlled is what will be run on the VPS. Unless, I move that into secrets since that's technically that's different between environments.

tj-wazei: Right here's what I've done. I've tried to make this backward compatible so you can still use a disk based config file (which fits your local dev example). 

This feels like a sweet spot/middle ground for what you're asking. Let me know what you think.

PR #1376, opened 2026-01-08
Update ChatGPT model to accept ChatModel & use 4.1-mini for questions

tj-wazei:

## Pull-request

### Changes

- [X] Existing code
- [ ] New feature

Closes Issue: NaN

## Description

`GPT-5-NANO` works but it's slow and can take around 30 seconds to generate a response. `GPT-5-MINI` while more expensive (still cheaper but closer to the old 3.5-turbo) halves that time to around 15 seconds. 

3.5-turbo is around 5 seconds .
4.1-mini is actually not that bad, 6 seconds. 

In this PR I have added the ChatModel as a parameter to the ask function so that it can be configured fairly easily. 

Using 4-1-mini for questions, 3.5-turbo for transfer and 5-mini for the slash command.  

Lastly, in the AI responses footer, the model used has been added.


application/src/main/java/org/togetherjava/tjbot/features/chatgpt/ChatGptService.java
```diff
@@ -51,11 +51,12 @@ public ChatGptService(Config config) {
      * @param question The question being asked of ChatGPT. Max is {@value MAX_TOKENS} tokens.
      * @param context The category of asked question, to set the context(eg. Java, Database, Other
      *        etc).
+     * @param chatModel The AI model to use for this request.
      * @return response from ChatGPT as a String.
      * @see <a href="https://platform.openai.com/docs/guides/chat/managing-tokens">ChatGPT
      *      Tokens</a>.
      */
-    public Optional<String> ask(String question, @Nullable String context) {
+    public Optional<String> ask(String question, @Nullable String context, ChatModel chatModel) {
```
Zabuzard: sth to consider: wrapping this enum with our own to avoid the dependency bleeding into all call-sites. similar to an ACL (anti corruption layer) in the world of backend services. otherwise, if they make breaking changes, they bleed into more than just a single file on our end and are then potentially more annoying to fix.

and if we are doing that, maybe its better UX to have that enum represent the use case instead of the model. so sth like "FAST" versus "HIGH_QUALITY".

that said, might be overkill, so feel free to neglect this comment.

tj-wazei: Done

PR #1384, opened 2026-01-23
Hotfix: Don't care about reactions in the Quotes Board itself

christolis:
Great, we recently made a quotes board, and if a message gets 5 stars, then it's added to the quotes board.  What happens if someone reacts from _within_ the quotes board itself?

Add an additional condition so that message reaction additions are not counted within this feature if the reaction is in the quotes board itself.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
@@ -89,7 +89,7 @@ public final class QuoteBoardForwarder extends MessageReceiverAdapter {
             return;
         }
 
-        TextChannel boardChannel = boardChannelOptional.get();
+        TextChannel boardChannel = boardChannelOptional.orElseThrow();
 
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
```diff
@@ -80,22 +80,28 @@ public void onMessageReactionAdd(MessageReactionAddEvent event) {
...
-        if (boardChannel.isEmpty()) {
+        if (boardChannelOptional.isEmpty()) {
             logger.warn(
                     "Could not find board channel with pattern '{}' in server with ID '{}'. Skipping reaction handling...",
                     this.config.channel(), guildId);
             return;
         }
 
-        logger.debug("Forwarding message to quote board channel: {}", boardChannel.get().getName());
+        TextChannel boardChannel = boardChannelOptional.get();
```
Zabuzard: NIT: prefer `orElseThrow()`, dont use Optional#get

PR #1389, opened 2026-01-24
Quotes board: Allow reactions with any emoji

christolis:
# Problem

Recently from the [server suggestions](https://discord.com/channels/272761734820003841/480829495486447627/1464256506617659554), many people have been slightly upset that the only reaction emoji considered for adding to the quotes board is the default star emoji.

Some members have suggested specific additional emojis to be considered, [others](https://ptb.discord.com/channels/272761734820003841/1464328383864242392/1464343398893752414) suggested that the star emoji should have a weight of `1.0` while the rest of the emojis should have a weight of `0.5`. While both solutions can work, all emojis can have a custom weight for the purpose of customizability.

# Solution

Introduce a scoring concept for each emoji, configurable for each particular one, and provide the ability to set a default value if an emoji is not defined in the configuration file.

For those who are wondering, _any_ kind of emoji that Discord can handle is able to be added in the configuration, including custom emojis in the server.  JDA can keep track of those.

For adding a unicode emoji, the actual unicode value has to be provided, like it has been done in `config.json.template`.

For adding a guild emoji, a "code" for the emoji has to be provided, for instance:

    youtube:1464573182206804010

Which stands for the friendly name of the emoji, a colon right after, and finally the ID of the custom emoji.

The "config.json.template" is _NOT_ exhaustive, more emojis have to be added and some others removed according to preference.

# Configuration changes

## quoteBoardConfig
|---	|---	|---	|---	|
| `quoteBoardConfig.minimumScoreToTrigger` 	| The minimum amount of reaction score for a message to be quoted. 	| float 	| `5.0` 	|
| `quoteBoardConfig.channel` 	| The pattern for the board channel. 	| String 	| `"quotes"` 	|
| `quoteBoardConfig.botEmoji` 	| The emoji with which the bot will mark quoted messages. 	| String 	| `⭐` 	|
| `quoteBoardConfig.defaultEmojiScore` 	| The default score of an emoji if it's not in the emojiScores map. 	| float 	| `0.5` 	|
| `quoteBoardConfig.emojiScores` 	| A map of each emoji's custom score. 	| Map<String, Float> 	|  _see below_ 	|

## emojiScores Map
⚠️ Note: This is just an example for you to get an idea of how you could utilize this configuration property and you are _encouraged_ to edit it to the server's needs.
```json
{
  "😬": -0.5,
  "💔": -0.5,
  "🖕": -0.5,
  "👎": -0.5,
  "🤨": -0.5,

  "🇷🇺": 0.0,
  "🇵🇸": 0.0,
  "🇮🇱": 0.0,

  "⭐": 1.0,

  /* Custom emoji */
  "youtube:1464573182206804010": 0.0
}

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -195,9 +195,34 @@
         "pollIntervalInMinutes": 10
     },
     "quoteBoardConfig": {
-        "minimumReactionsToTrigger": 5,
+        "minimumScoreToTrigger": 5.0,
         "channel": "quotes",
-        "reactionEmoji": "⭐"
+        "botEmoji": "⭐",
+        "defaultEmojiScore": 0.5,
+        "emojiScores": {
+            "😬": -0.5,
+            "💔": -0.5,
+            "😐": -0.5,
+            "😊": -0.5,
+            "🖕": -0.5,
+            "👎": -0.5,
+            "💩": -0.5,
+            "🤢": -0.5,
+            "🤮": -0.5,
+            "🤬": -0.5,
+            "😡": -0.5,
+            "😒": -0.5,
+            "🤨": -0.5,
+
+            "🇷🇺": 0.0,
+            "🇵🇸": 0.0,
+            "🇮🇱": 0.0,
+            "🏳️‍🌈": 0.0,
+
+            "⭐": 1.0,
+
+            "youtube:1464573182206804010": 0.0
+        }
     },
     "memberCountCategoryPattern": "Info",
     "topHelpers": {
--- a/application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
@@ -6,6 +6,7 @@ import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.MessageReaction;
 import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
 import net.dv8tion.jda.api.entities.emoji.Emoji;
+import net.dv8tion.jda.api.entities.emoji.EmojiUnion;
 import net.dv8tion.jda.api.events.message.react.MessageReactionAddEvent;
 import net.dv8tion.jda.api.requests.RestAction;
 import org.slf4j.Logger;
@@ -36,7 +37,7 @@ import java.util.regex.Pattern;
 public final class QuoteBoardForwarder extends MessageReceiverAdapter {
 
     private static final Logger logger = LoggerFactory.getLogger(QuoteBoardForwarder.class);
-    private final Emoji triggerReaction;
+    private final Emoji botEmoji;
     private final Predicate<String> isQuoteBoardChannelName;
     private final QuoteBoardConfig config;
 
@@ -48,7 +49,7 @@ public final class QuoteBoardForwarder extends MessageReceiverAdapter {
      */
     public QuoteBoardForwarder(Config config) {
         this.config = config.getQuoteBoardConfig();
-        this.triggerReaction = Emoji.fromUnicode(this.config.reactionEmoji());
+        this.botEmoji = Emoji.fromUnicode(this.config.botEmoji());
 
         this.isQuoteBoardChannelName = Pattern.compile(this.config.channel()).asMatchPredicate();
     }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: application/config.json.template [hunk 1]
```diff
@@ -195,9 +195,34 @@
...
+            "🤮": -0.5,
+            "🤬": -0.5,
+            "😡": -0.5,
+            "😒": -0.5,
+            "🤨": -0.5,
+
+            "🇷🇺": 0.0,
+            "🇵🇸": 0.0,
+            "🇮🇱": 0.0,
+            "🏳️‍🌈": 0.0,
```
SquidXTV, line 220: I don't think we should keep this politically influenced example in the repo.

SquidXTV (changes requested):

application/config.json.template
christolis: As [discussed](https://discord.com/channels/272761734820003841/895717328359153664/1464594233338888193) in Discord, it's an example, it's _made_ to prove a point for the person setting the configuration, to give them an idea on how to provide scores for each emoji. It's intentionally there to evoke an emotion, and all of these flags have a flag of `0` to maintain _neutrality_. No flag is more powerful than the other, at least none of the flags commonly used as reactions to offend in 2026.

firasrg: My tip: Avoid flag reactions

application/config.json.template
```diff
@@ -195,9 +195,34 @@
         "pollIntervalInMinutes": 10
     },
     "quoteBoardConfig": {
-        "minimumReactionsToTrigger": 5,
+        "minimumScoreToTrigger": 5.0,
```
firasrg, line 198: There is no need to put it decimal

application/config.json.template
```diff
@@ -195,9 +195,34 @@
         "pollIntervalInMinutes": 10
     },
     "quoteBoardConfig": {
```
firasrg, line 197: I think that we should ignore some text channels like `announcements`, and `hall-of-fame`, `github-activity-logs`, `vps-status`, `bots` and the likes, because they are not intended for such a feature, consider to put a config attribute to list them here

application/config.json.template
```diff
@@ -195,9 +195,34 @@
...
+            "🤨": -0.5,
+
+            "🇷🇺": 0.0,
+            "🇵🇸": 0.0,
+            "🇮🇱": 0.0,
+            "🏳️‍🌈": 0.0,
+
+            "⭐": 1.0,
+
+            "youtube:1464573182206804010": 0.0
```
firasrg, line 224: While I know this is an example, it's better to add a custom emoji ID from the server TJ for more consistency

application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
```diff
@@ -96,21 +84,27 @@ public void onMessageReactionAdd(MessageReactionAddEvent event) {
...
 
-        logger.debug("Forwarding message to quote board channel: {}", boardChannel.getName());
+        event.retrieveMessage().queue(message -> {
+            float emojiScore = calculateMessageScore(message.getReactions());
 
-        event.retrieveMessage()
-            .queue(message -> markAsProcessed(message).flatMap(v -> message.forwardTo(boardChannel))
-                .queue(_ -> logger.debug("Message forwarded to quote board channel: {}",
-                        boardChannel.getName())),
+            if (emojiScore < config.minimumScoreToTrigger()) {
```
firasrg, line 90: a log.debug can help here

firasrg (commented):
@christolis Thank you for this update, please check my reviews, thank you

application/config.json.template
christolis: @firasrg Just curious: what happens if we want half-point scoring later (like the 0.5 for non-star emojis)?

application/config.json.template
christolis: Could you clarify what you mean by "for more consistency"?

application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
christolis: How do you choose when to log a debug message?

application/config.json.template
christolis: Considering that the bot has no read access to those channels ([source](https://discord.com/channels/272761734820003841/895717328359153664/1473945463978983426)), what makes such configuration necessary?

firasrg: Since access is possible, adding these controls makes sense, it significantly improves the feature's security

application/config.json.template
firasrg: in that case change it to decimal ?

application/config.json.template
firasrg: I mean compatibility with the emojis on the server

application/src/main/java/org/togetherjava/tjbot/features/basic/QuoteBoardForwarder.java
firasrg: When I find a place in code potential for bugs

SquidXTV (approved):

surajkumar (approved):

PR #1395, opened 2026-01-26
fix: DynamicVoiceChat deleting channel without checking if members are in there

tj-wazei:
## Description

Dynamic voice chats are currently being deleted while members are still in there. This PR adds a check before doing anything.


application/src/main/java/org/togetherjava/tjbot/features/voicechat/DynamicVoiceChat.java
```diff
@@ -81,6 +81,12 @@ private void handleVoiceChannelLeave(AudioChannelUnion channelLeft) {
         if (!eventHappenOnDynamicRootChannel(channelLeft)) {
             logger.debug("Event happened on left channel {}", channelLeft);
 
+            if (!hasMembers(channelLeft)) {
+                logger.debug("Voice channel {} not empty, so not doing anything.",
+                        channelLeft.getName());
+                return;
+            }
```
christolis: I think the conditions are inverted aren't they?

tj-wazei: yes

PR #1422, opened 2026-02-28
build: include git-describe version as part of the bot in /ping command

christolis:
It is uncertain on which version the bot is running in production, and those with a lack of access to the production infrastructure have almost no way to get such information.

Expand the "/ping" slash command to actually mention which version it is running as well as the date on which it was built, more specfifically it now says:

    > Pong! Running on version v1.2.3-g1a2b3c (2026-02-28)

That way we don't have to approach the lucky few folks with access and badger them with primitive questions. :)

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -93,3 +93,23 @@ application {
     mainClass = 'org.togetherjava.tjbot.Application'
     applicationDefaultJvmArgs = ["--enable-native-access=ALL-UNNAMED"]
 }
+
+def generatedBuildPropertiesDir = file("build/resources/main/")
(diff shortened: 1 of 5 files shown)


Zabuzard (changes requested):
the `/ping` command acts as minimal and simplest introduction for learners.
i would appreciate not cluttering up its code and its unit-test for this feature.

we can add a `/version` command instead. i like the idea in general but have some doubts whether this will work properly in all environments, on the VPS, with docker and whatnot. so would be great if @marko-radosavljevic or @tj-wazei will review this

PR #1439, opened 2026-03-13
Unit Tests for Metrics

firasrg:
## What
Adds the first unit test for the analytics system.

- Added `MetricsTests` to verify `Metrics#count(...)` persists a metric event;
- Asserted both the stored event name and the recorded timestamp;
- Added async waiting/timeout handling for reliable verification of background writes;
- Exposed the `Metrics` executor service so tests can shut it down cleanly;
- Enabled debug logging for `org.togetherjava.tjbot` in test runs;
- Added [AssertJ](https://assertj.github.io/) for test assertions

## Why

`Metrics` performs asynchronous database writes, so having direct test coverage here improves confidence in analytics persistence and makes future refactors safer.

## Validation

- ran `:application:test --tests org.togetherjava.tjbot.features.MetricsTests`


application/src/test/resources/log4j2.xml
```diff
@@ -13,6 +13,8 @@
         </RollingFile>
     </Appenders>
     <Loggers>
+        <Logger name="org.togetherjava.tjbot" level="debug"/>
+
```
Zabuzard: undo

application/build.gradle
```diff
@@ -80,6 +80,7 @@ dependencies {
     implementation 'org.apache.commons:commons-text:1.15.0'
     implementation 'com.apptasticsoftware:rssreader:3.12.0'
 
+    testImplementation 'org.assertj:assertj-core:3.27.7'
```
Zabuzard: undo. we want to use raw junit in this project for now as it is easier for beginners and reflects more what they will see in the wild and might already be familiar with from their own projects and university etc

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
```diff
@@ -53,4 +53,7 @@ private void processEvent(String event, Instant happenedAt) {
             .insert());
     }
 
+    public ExecutorService getExecutorService() {
+        return service;
+    }
```
Zabuzard: undo, this won't be needed (see other comment why)

application/src/test/java/org/togetherjava/tjbot/features/MetricsTests.java
```diff
@@ -0,0 +1,118 @@
...
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.concurrent.locks.LockSupport;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.within;
+import static org.junit.jupiter.api.Assertions.fail;
+
+final class MetricsTests {
```
Zabuzard: this test is too much in terms of complexity for no tangible benefit.

lets step back and create something simpler that beginners can understand and maintain while providing pretty much the same test-safety in terms of relevant coverage.

1. create a package private method in Metrics called `countBlocking`. its identical to `count` but it won't use the executor service. so it creates the instant and then calls `processEvent` directly
2. in ur unit test this will be the method ur testing against now. without the async stuff the test will simplify **a lot**
3. the test should use a GWT structure (see other examples in the codebase, alternative patterns that share the same idea: 4-phases SEVT or AAA). try to aim for something like this:
```java
@Test
void basicEventCount() {
  // GIVEN a test event
  String expectedEvent = "test";
  Instant expectedHappenedAt = Instant.now();

  // WHEN counting the event
  metrics.countBlocking(expectedEvent);

  // THEN the event was saved in the DB
  Foo entry = database...
  String actualEvent = entry.event();
  Instant actualHappenedAt = entry.happenedAt();

  assertEquals(expectedEvent, actualEvent);
  assertCloseEnough(expectedHappenedAt, actualHappenedAt);
} 

private static void assertCloseEnough(Instant expected, Instant actual) {
 ... // assert that its within 1min difference
} 
```
(double check the correct order of expected vs actual in assert, i never can remember it, lol)

Zabuzard (changes requested):

application/src/test/java/org/togetherjava/tjbot/features/MetricsTests.java
firasrg: Thanks for the guide. However, i don't see this a good idea: creating new method inside the `Metrics`, while it's only for tests. Also putting such method in the test class, doesn't really test the real one `count`

Zabuzard: It is not meaningful to test the true `count` method. The async nature makes it very difficult to test, too difficult for what we want right now. It is more meaningful to instead split the method into parts so that your test can at least cover 98% of its code. Thats good and also standard practice.

Please do it, thanks.

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
```diff
@@ -44,6 +44,26 @@ public void count(String event) {
...
+    /**
+     * Track an event execution with flag for async execution.
+     * 
+     * @param event the event to save
+     * @param doAsync the async flag
+     */
+    public void count(String event, boolean doAsync) {
+        count(event, Map.of(), doAsync);
+    }
+
```
Zabuzard: remove that, we dont want to expose that `doAsync` stuff to the outside. ur test can use the `void count(String event, Map<String, Object> dimensions, boolean doAsync) {`, tests are in the same package so package-private works for them.

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
```diff
@@ -44,6 +44,26 @@ public void count(String event) {
         count(event, Map.of());
     }
 
+    /**
+     * Track an event execution with dimensions provided.
+     * 
+     * @param event the event to save
+     * @param dimensions the dimensions to save
```
Zabuzard, line 51: please dont change the javadoc, it was this:

```java
     * Track an event execution with additional contextual data.
     *
     * @param event the name of the event to record (e.g. "user_signup", "purchase")
     * @param dimensions optional key-value pairs providing extra context about the event. These are
     *        often referred to as "metadata" and can include things like: userId: "12345", name:
     *        "John Smith", channel_name: "chit-chat" etc. This data helps with filtering, grouping,
     *        and analyzing events later. Note: A value for a metric should be a Java primitive
     *        (String, int, double, long float).
     *
```

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
```diff
@@ -72,12 +97,6 @@ private static String serializeDimensions(Map<String, Object> dimensions) {
         }
     }
 
-    /**
-     *
-     * @param event the event to save
-     * @param happenedAt the moment when the event is dispatched
-     * @param dimensionsJson optional JSON-serialized dimensions, or null
-     */
```
Zabuzard, line 80: dont remove that

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
Zabuzard: (not done yet)

application/src/test/java/org/togetherjava/tjbot/features/MetricsTests.java
Zabuzard: (not done yet)

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
firasrg: ok and I'll add a description

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
firasrg: what do you mean by ??
> tests are in the same package so package-private works for them.

application/src/test/java/org/togetherjava/tjbot/features/MetricsTests.java
firasrg: Why do we have to create an assertion method `assertCloseEnough` ?

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
firasrg: ~~There was absolutely no javadoc~~

application/src/test/java/org/togetherjava/tjbot/features/MetricsTests.java
Zabuzard: (talked about it in discord)

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
Zabuzard: (talked about it in discord)

application/src/main/java/org/togetherjava/tjbot/features/analytics/Metrics.java
Zabuzard: There was. The method that was previously public facing is now not public facing anymore and the method you made public facing has a different, less detailled, javadoc:
grafik

firasrg: Alright then, all i have to do is to make the new one has better javadoc and done

PR #1458, opened 2026-04-13
Blacklistable AI Help-Answer

Alathreon:
Made so it is possible to blacklist the auto ai response + made it default in config template

### DIFF
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -179,6 +179,7 @@
         "normal": [
         ],
         "special": [
+            "AutoAIResponse"
         ]
     },
     "selectRolesChannelPattern": "select-your-roles",
--- a/application/config.json.template
+++ b/application/config.json.template
@@ -179,7 +179,6 @@
         "normal": [
         ],
         "special": [
-            "AutoAIResponse"
         ]
     },
     "selectRolesChannelPattern": "select-your-roles",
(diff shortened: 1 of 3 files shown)


Zabuzard (changes requested):
Please get rid of the default. The default bot setup should be similar to that of the server and the AI response is a feature that we also often work with in the code. It is important (especially for frictionless onbkarding with newcomers) that its there automatically.

Alathreon:
> Please get rid of the default. The default bot setup should be similar to that of the server and the AI response is a feature that we also often work with in the code. It is important (especially for frictionless onbkarding with newcomers) that its there automatically.

Do you know where I could put this constant ? Maybe in the wiki ? Because this constant isn't the name of any class.

surajkumar:
This isn't really something that needs a feature flag?

Zabuzard:
> > Please get rid of the default. The default bot setup should be similar to that of the server and the AI response is a feature that we also often work with in the code. It is important (especially for frictionless onbkarding with newcomers) that its there automatically.
> 
> Do you know where I could put this constant ? Maybe in the wiki ? Because this constant isn't the name of any class.

Mh yeah. Id say the javadoc of the helper class is a good start and maybe we could also start adding them to the Config class next to the property dealing with these flags. Wiki is also good but might rot more easily.

PR #1471, opened 2026-04-29
Fix dead links in ChatGPT responses (#1276)

### LINKED ISSUE #1276: Replace broken links send by ChatGpt

vlaym:
Closes #1276

### What was done
- Integrated LinkDetection.replaceDeadLinks() (added in #1366) into HelpSystemHelper#generateGptResponseEmbed(). The replacement runs before the embed is built, so both the auto-AI response in help threads and /chatgpt command are covered since they share the same embed builder.
- If the link check fails (e.g. network error), the original answer is kept as-is - no broken behavior.
- Added two unit tests in HelpSystemHelperTest: one verifying dead link replacement, one verifying fallback when replacement fails.

### Approach
The issue proposes two options: ideally sending the response immediately and then editing it after async link checking, or as a fallback - replacing links before sending.

### Current (fallback) flow:
post created → ChatGPT generates answer → dead links replaced → embed built → embed sent to Discord

### Ideal flow (not implemented):
post created → ChatGPT generates answer → embed sent immediately → links checked in the background → already-sent message edited with clean links

The fallback was chosen because it keeps the change in one place (HelpSystemHelper) and is easy to test. The ideal flow would require mixing two different async mechanisms: checking links uses Java's CompletableFuture (because it makes plain HTTP requests to external sites — JDA is not involved here at all), while sending and editing Discord messages uses JDA's RestAction. These two do not compose directly, so some extra glue code is needed to chain them together correctly.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
@@ -29,6 +29,7 @@ import org.togetherjava.tjbot.features.chatgpt.ChatGptModel;
 import org.togetherjava.tjbot.features.chatgpt.ChatGptService;
 import org.togetherjava.tjbot.features.componentids.ComponentIdInteractor;
 import org.togetherjava.tjbot.features.utils.Guilds;
+import org.togetherjava.tjbot.features.utils.LinkDetection;
 
 import java.awt.Color;
 import java.time.Instant;
@@ -41,6 +42,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.StringJoiner;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -57,6 +59,7 @@ import static org.togetherjava.tjbot.features.utils.MessageUtils.mentionGuildSla
 public final class HelpSystemHelper {
     private static final Logger logger = LoggerFactory.getLogger(HelpSystemHelper.class);
     private static final ChatGptModel CHAT_GPT_MODEL = ChatGptModel.FAST;
+    private static final String BROKEN_LINK_REPLACEMENT = "(broken link removed)";
 
     static final Color AMBIENT_COLOR = new Color(255, 255, 165);
 
@@ -74,6 +77,7 @@ public final class HelpSystemHelper {
 
     private final Database database;
     private final ChatGptService chatGptService;
+    private final Function<String, CompletableFuture<String>> brokenLinkReplacer;
     private static final int MAX_QUESTION_LENGTH = 200;
     private static final int MIN_QUESTION_LENGTH = 10;
     private static final String CHATGPT_FAILURE_MESSAGE =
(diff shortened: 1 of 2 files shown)


surajkumar (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
```diff
@@ -199,12 +211,21 @@ public MessageEmbed generateGptResponseEmbed(String answer, SelfUser selfUser, S
...
-            .setDescription(answer)
+            .setDescription(sanitizedAnswer)
             .setColor(Color.pink)
             .setFooter(responseByGptFooter)
             .build();
     }
 
+    private String sanitizeBrokenLinks(String answer) {
+        try {
+            return brokenLinkReplacer.apply(answer).join();
```
Zabuzard: join without timeout isnt good

Zabuzard (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/help/HelpSystemHelper.java
vlaym: Thank you. Added a 10-second timeout via `orTimeout()` so the thread cannot block indefinitely.

```diff
+private static final int LINK_CHECK_TIMEOUT_SECONDS = 10;

 private String sanitizeBrokenLinks(String answer) {
     try {
-        return brokenLinkReplacer.apply(answer).join();
+        return brokenLinkReplacer.apply(answer)
+            .orTimeout(LINK_CHECK_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+            .join();
     } catch (RuntimeException runtimeException) {
         logger.debug("Failed to replace broken links in ChatGPT response", runtimeException);
         return answer;
     }
 }
```

If the check does not finish in time, `join()` throws a `CompletionException` (wrapping `TimeoutException`), which is caught by the existing `RuntimeException` catch block — so the original answer is returned as a fallback. Since `replaceDeadLinks` checks all links in parallel, the timeout applies to the slowest single request, not the sum of all of them.

surajkumar (dismissed):

Zabuzard (approved):

PR #1484, opened 2026-05-22
feat: create command to purge a given channel

surajkumar:

## Pull-request

### Changes

- [ ] Existing code
- [X] New feature

Closes Issue: #912 

## Description

This PR introduces a new slash command for moderators called `/purge`. 
The command takes 2 required params:

1. `channel`: so the bot knows which channel to purge
2. `message-id`: The message to start the purge from

A third optional `amount` to limit how many messages are deleted. When absent all messages after the given `message-id` are deleted.

### Why?
There have been cases were we've needed to purge a channel but haven't had the tools to automatically do this. Instead, mods and admins manually delete messages which is painful. 

Example of when this would have been useful: "the `@everyone` ping". 

---

See comments for more context.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/Features.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/Features.java
@@ -67,6 +67,8 @@ import org.togetherjava.tjbot.features.moderation.scam.ScamHistoryPurgeRoutine;
 import org.togetherjava.tjbot.features.moderation.scam.ScamHistoryStore;
 import org.togetherjava.tjbot.features.moderation.temp.TemporaryModerationRoutine;
 import org.togetherjava.tjbot.features.projects.ProjectsThreadCreatedListener;
+import org.togetherjava.tjbot.features.purge.PurgeCommand;
+import org.togetherjava.tjbot.features.purge.PurgeMessagesByUserCommand;
 import org.togetherjava.tjbot.features.reminder.RemindRoutine;
 import org.togetherjava.tjbot.features.reminder.ReminderCommand;
 import org.togetherjava.tjbot.features.roleapplication.CreateRoleApplicationCommand;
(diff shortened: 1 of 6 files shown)


Zabuzard (changes requested):
imo needs some sort of confirmation dialog including the amount of messages that will get deleted.
otherwise all it takes is someone making a mistake and the entire channel history is gone.

a better and safer ux might be something like "purge all messages from the last x hours/minutes" or alternatively a safety check that the message to purge from isnt older than two days. 

the real command we need, i think, is purging messages from a single user in all channels (like what /ban does). but difficult to do discord api wise.

ankitsmt211:
https://github.com/Together-Java/TJ-Bot/issues/912 not around channels but perhaps relevant for the PR or discussion, seems like there's a graveyard of PRs and issues around this

> a better and safer ux might be something like "purge all messages from the last x hours/minutes" or alternatively a safety check that the message to purge from isnt older than two days.

agreed around UX feels better

surajkumar:
No worries, I'm currently tackling this and I'll do it piece by piece.

First is the confirmation:
image

Then the completion: 
image

and the audit log (#mod-audit-log):
image

surajkumar:
next is time based:

image

image

surajkumar:
Usage now:
  - `/purge channel:#foo minutes:30` purge last 30 minutes
  - `/purge channel:#foo message-id:123…`  purge after a specific message
  - `/purge channel:#foo minutes:30 amount:50` purge last 30 minutes, capped at 50

surajkumar:
Added another command: `/purge-messages-by-user`

Takes 2 params:
- user: the user
- minutes: time range capped to 6 hours
- amount: optional global amount of messages to delete 

Confirmation:
image

image

Audit log:
image

PR #1485, opened 2026-05-23
feat: add support for specifying response schema to AI response

surajkumar:

## Pull-request

### Changes

- [X] Existing code
- [X] New feature

Closes Issue: NaN

## Description

The OpenAI API supports structured JSON responses which we are currently not taking advantage of.

The way it works is that with each prompt, you can specify a schema and the API will respond with only that. This is useful in a fair bit of scenarios where we want structured output that we can parse in code. There are some places were we are trying to achieve this via prompt injection however we then gamble on the AI producing a response we want. 

Usage example:

```java
  ResponseSchema schema = new ResponseSchema(
          Map.of(
                  "summary", Property.of(Type.STRING),
                  "tags",    Property.array(Property.of(Type.STRING)),
                  "author",  Property.object(
                          Map.of(
                                  "name", Property.of(Type.STRING),
                                  "ts",  Property.of(Type.INTEGER)),
                          List.of("name", "ts"))),
          List.of("summary", "tags", "author"));

  Optional<String> json = chatGptService.askRaw(
          "Summarize this thread and extract its tags and author.",
          ChatGptModel.GPT_4_MINI,
          schema);
```

Example AI output:
```json
  {
    "summary": "User asks about generics in Java.",
    "tags": ["java", "generics", "type-erasure"],
    "author": { "name": "alice", "ts": 123456789 }
  }
```

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Property.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Property.java
@@ -1,0 +1,117 @@
+package org.togetherjava.tjbot.features.chatgpt.schema;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents a single property in an OpenAI JSON schema, used to describe the shape of one field
+ * inside a {@link ResponseSchema}.
+ * <p>
+ * The hierarchy is sealed and mirrors the JSON Schema specification supported by OpenAI's
+ * structured outputs:
+ * <ul>
+ * <li>{@link Primitive} – scalar values ({@code string}, {@code number}, {@code integer},
+ * {@code boolean}, {@code null}).</li>
+ * <li>{@link ArrayProperty} - an array whose elements conform to a nested {@link Property}.</li>
+ * <li>{@link ObjectProperty} - a nested object with its own {@code properties} and {@code required}
+ * list.</li>
+ * </ul>
+ * Prefer the static factory methods ({@link #of}, {@link #array}, {@link #object}) for readable
+ * construction.
+ *
+ * @see ResponseSchema
+ * @see <a href="https://platform.openai.com/docs/guides/structured-outputs">OpenAI Structured
+ *      Outputs</a>
+ */
+public sealed interface Property
+        permits Property.Primitive, Property.ArrayProperty, Property.ObjectProperty {
+
+    /**
+     * The JSON schema {@link Type} of this property.
+     *
+     * @return the type this property declares
+     */
+    Type type();
+
+    /**
+     * Creates a primitive property of the given scalar type.
+     *
+     * @param type a scalar type such as {@link Type#STRING} or {@link Type#INTEGER}
+     * @return a new {@link Primitive} property
+     */
+    static Primitive of(Type type) {
+        return new Primitive(type);
+    }
+
+    /**
+     * Creates an array property whose elements conform to the given item schema.
+     *
+     * @param items the schema each element must satisfy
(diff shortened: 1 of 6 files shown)


application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Type.java
```diff
@@ -0,0 +1,29 @@
...
+    NULL;
+
+    /**
+     * Returns the lowercase JSON representation of this type. Used by Jackson via
+     * {@link JsonValue}, so the enum serializes to {@code "string"} rather than {@code "STRING"}.
+     *
+     * @return the JSON Schema type name in lowercase
+     */
+    @JsonValue
+    public String jsonValue() {
```
SquidXTV, line 26: I kinda prefer `json()`, not sure if that 'value' in the name gives much info

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Type.java
```diff
@@ -0,0 +1,29 @@
...
+ * matching the JSON Schema specification.
+ */
+public enum Type {
+    STRING,
+    NUMBER,
+    INTEGER,
+    BOOLEAN,
+    OBJECT,
+    ARRAY,
+    NULL;
```
SquidXTV, line 17: Any reason why it doesn't support `ENUM` and `anyOf`?

I quickly checked and refer to this:
https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/package-info.java
```diff
@@ -0,0 +1,7 @@
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+package org.togetherjava.tjbot.features.chatgpt.schema;
```
SquidXTV, line 3: Maybe put quick package-level javadocs for this one, we usually do have some quick introduction

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Property.java
```diff
@@ -0,0 +1,117 @@
...
+    record ArrayProperty(Type type, Property items) implements Property {
+        /**
+         * Convenience constructor that fixes {@link #type} to {@link Type#ARRAY}.
+         *
+         * @param items the schema each element must satisfy
+         */
+        public ArrayProperty(Property items) {
+            this(Type.ARRAY, items);
+        }
+    }
```
SquidXTV, line 91: Couldn't you remove the type from the record fields and simply implement `type()` with hardcoded value? Otherwise it would be possible to create an "invalid" array property that has a different type. Same for object type down below and `ResponseSchema`.

SquidXTV (approved):

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Type.java
surajkumar: Ah yeah - in a follow up PR - we don't need it right now.

application/src/main/java/org/togetherjava/tjbot/features/chatgpt/schema/Type.java
surajkumar: nit - might just ignore, i did think about this and ended up asking AI on it's opinion. json() might feel like full json, jsonValue kinda implies is a value of JSON, toString() was considered but I didn't.

christolis (approved):
LGTM

PR #1486, opened 2026-05-23
Add member id at the end of the role application embed

christolis:
This adds a member ID at the end of the role application embed.

# Image

image


surajkumar (approved):
Thank you

PR #1505, opened 2026-06-19
Adds "this-is-scam" message context command

Zabuzard:
## Overview

This adds a new message context command (**right click message**) called `this-is-scam`. It acts as the primary way for users to report scam to mods.

Mods then get presented a quick way to handle the scam properly by automatically deleting all messages from the user and putting them under quarantine through a single button press!

grafik

Internally this works by **quarantine, ban (delete messages from one day), unban** in that order. The user ends up being kicked out of the guild (due to the ban) but can immediately rejoin (due to the unban), but remains **quarantined**.

## Other

The following extra situations are handled gracefully via in-memory caches:

* multiple users report the same message as scam: only one is being processed
* a user tries to abuse the command by using it constantly: users are on cooldown for 1 minute

Also, an appropriate entry in the audit-log of the user (`/audit`) is created.

Attachments are spelled out so it becomes easier for devs later on to edit `ScamBlocker` accordingly.

## Code

The integration is minimal and simple, just a straightforward single class is added `ThisIsScamCommand.java`, thats it. No config changes, everything easy.

### Context Menu
grafik

### Response
grafik

### What mods see
grafik

grafik

### No scam

grafik

### Yes scam

grafik

### DM

grafik

### Quarantined

grafik

### Audit Log

grafik

### Edge cases

grafik

grafik

### Error
grafik

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -190,22 +190,20 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
         String content = target.getContentStripped();
         String description = content.isBlank() ? "(message had no text content)" : content;
 
+        String attachmentText = "_none_";
         List<Message.Attachment> attachments = target.getAttachments();
         if (!attachments.isEmpty()) {
-            String attachmentInfo = attachments.stream()
+            attachmentText = attachments.stream()
                 .map(Message.Attachment::getFileName)
-                .collect(Collectors.joining("\n"));
-            description += """
-
-
-                    Attachments:
-                    """ + attachmentInfo;
+                .collect(Collectors.joining("\n- ", "\n- ", ""));
         }
 
         description += """
 
 
-                [Go to message](%s)""".formatted(target.getJumpUrl());
+                Attachments: %s
+                [Go to message](%s)
+                """.formatted(attachmentText, target.getJumpUrl());
         return description;
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -61,16 +61,16 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
     private static final Duration USER_COMMAND_COOLDOWN = Duration.ofMinutes(1);
     private static final Color AMBIENT_COLOR = Color.decode("#CFBFF5");
 
+    private final Config config;
+    private final ModerationActionsStore actionsStore;
+    private final Predicate<String> isModAuditLogChannel;
+
     private final Cache<Long, Instant> reportedMessageToTimestamp =
             Caffeine.newBuilder().maximumSize(10_000).expireAfterWrite(Duration.ofDays(1)).build();
     private final Cache<Long, Instant> userToLastCommandUse = Caffeine.newBuilder()
         .maximumSize(10_000)
         .expireAfterWrite(USER_COMMAND_COOLDOWN)
         .build();
-
-    private final Config config;
-    private final ModerationActionsStore actionsStore;
-    private final Predicate<String> isModAuditLogChannel;
 
     /**
      * Creates a new instance.
@@ -96,7 +96,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
             return;
         }
 
-        Optional<TextChannel> modAuditLog = getModAuditLogChannel(event);
+        Optional<TextChannel> modAuditLog = findModAuditLogChannel(event);
         if (modAuditLog.isEmpty()) {
             event.reply(FAILED_MESSAGE).setEphemeral(true).queue();
             return;
@@ -114,9 +114,9 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
     }
 
     private boolean handleIsOnCooldown(MessageContextInteractionEvent event) {
-        Instant lastCommandUse = userToLastCommandUse.getIfPresent(event.getUser().getIdLong());
-        Runnable isNotOnCooldownAction =
-                () -> userToLastCommandUse.put(event.getUser().getIdLong(), Instant.now());
+        long userId = event.getUser().getIdLong();
+        Instant lastCommandUse = userToLastCommandUse.getIfPresent(userId);
+        Runnable isNotOnCooldownAction = () -> userToLastCommandUse.put(userId, Instant.now());
 
         if (lastCommandUse == null) {
             isNotOnCooldownAction.run();
@@ -147,7 +147,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
         return false;
     }
 
-    private Optional<TextChannel> getModAuditLogChannel(MessageContextInteractionEvent event) {
+    private Optional<TextChannel> findModAuditLogChannel(MessageContextInteractionEvent event) {
         Guild guild = Objects.requireNonNull(event.getGuild());
         Optional<TextChannel> modAuditLogChannel =
                 Guilds.findTextChannel(guild, isModAuditLogChannel);
@@ -181,20 +181,26 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
                     Button.danger(generateComponentId(args), "No"));
     }
 
-    @SuppressWarnings("squid:S3457") // %n is wrong, markdown must use \n
     private static String createDescription(Message target) {
         String content = target.getContentStripped();
-        String description = content.isBlank() ? "(empty message)" : content;
+        String description = content.isBlank() ? "(message had no text content)" : content;
 
         List<Message.Attachment> attachments = target.getAttachments();
         if (!attachments.isEmpty()) {
             String attachmentInfo = attachments.stream()
                 .map(Message.Attachment::getFileName)
                 .collect(Collectors.joining("\n"));
-            description += "\n\nAttachments:\n" + attachmentInfo;
+            description += """
+
+
+                    Attachments:
+                    """ + attachmentInfo;
         }
 
-        description += "\n\n[Go to message](%s)".formatted(target.getJumpUrl());
+        description += """
+
+
+                [Go to message](%s)""".formatted(target.getJumpUrl());
         return description;
     }
 
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -182,6 +182,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
                     Button.danger(generateComponentId(args), "No"));
     }
 
+    @SuppressWarnings("squid:S3457") // %n is wrong, markdown must use \n
     private static String createDescription(Message target) {
         String content = target.getContentStripped();
         String description = content.isBlank() ? "(empty message)" : content;
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -285,7 +285,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
                         This means you can no longer interact with anyone in the server until you have been unquarantined again.
 
                         Potentially your account got compromised.
-                        After you regained control of your account make sure you secure it properly, for example using 2FA.
+                        After you regained control of your account make sure you secure it properly, for example by using 2FA.
                         You can then join back our server and contact the mods to get unquarantined.""";
 
         return ModerationUtils.sendModActionDm(ModerationUtils.getModActionEmbed(guild,
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -27,6 +27,7 @@ import org.togetherjava.tjbot.config.Config;
 import org.togetherjava.tjbot.features.BotCommandAdapter;
 import org.togetherjava.tjbot.features.CommandVisibility;
 import org.togetherjava.tjbot.features.MessageContextCommand;
+import org.togetherjava.tjbot.features.utils.AmbientColors;
 import org.togetherjava.tjbot.features.utils.Guilds;
 import org.togetherjava.tjbot.features.utils.MessageUtils;
 import org.togetherjava.tjbot.logging.LogMarkers;
@@ -62,7 +63,6 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
     private static final String FAILED_MESSAGE =
             "Sorry, there was an issue forwarding your scam report to the moderators. We are investigating.";
     private static final Duration USER_COMMAND_COOLDOWN = Duration.ofMinutes(1);
-    private static final Color AMBIENT_COLOR = Color.decode("#CFBFF5");
 
     private final Config config;
     private final ModerationActionsStore actionsStore;
@@ -173,7 +173,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
                     MessageUtils.abbreviate(description, MessageEmbed.DESCRIPTION_MAX_LENGTH))
             .setAuthor(author.getName(), null, author.getEffectiveAvatarUrl())
             .setTimestamp(message.getTimeCreated())
-            .setColor(AMBIENT_COLOR)
+            .setColor(AmbientColors.MODERATION_SCAM)
             .setFooter("%s - account age %d days".formatted(author.getId(), accountAgeDays))
             .build();
 
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -15,7 +15,6 @@ import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionE
 import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
 import net.dv8tion.jda.api.exceptions.ErrorHandler;
 import net.dv8tion.jda.api.interactions.commands.build.Commands;
-import net.dv8tion.jda.api.interactions.components.ActionRow;
 import net.dv8tion.jda.api.interactions.components.buttons.Button;
 import net.dv8tion.jda.api.interactions.components.buttons.ButtonStyle;
 import net.dv8tion.jda.api.requests.ErrorResponse;
@@ -217,10 +216,7 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
         List<MessageEmbed> embeds = new ArrayList<>(event.getMessage().getEmbeds());
         embeds.add(resultEmbed);
 
-        List<Button> disabledButtons =
-                event.getMessage().getButtons().stream().map(Button::asDisabled).toList();
-
-        event.editMessageEmbeds(embeds).setComponents(ActionRow.of(disabledButtons)).queue();
+        event.editMessageEmbeds(embeds).setComponents().queue();
 
         if (!isScam) {
             return;
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -281,8 +281,12 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
     private RestAction<Boolean> sendQuarantineDm(User target, Guild guild) {
         String description =
                 """
-                        Hey there, sorry to tell you but unfortunately you have been put under quarantine.
-                        This means you can no longer interact with anyone in the server until you have been unquarantined again.""";
+                        Hey there, sorry to tell you but unfortunately you have been put under quarantine after sending scam.
+                        This means you can no longer interact with anyone in the server until you have been unquarantined again.
+
+                        Potentially your account got compromised.
+                        After you regained control of your account make sure you secure it properly, for example using 2FA.
+                        You can then join back our server and contact the mods to get unquarantined.""";
 
         return ModerationUtils.sendModActionDm(ModerationUtils.getModActionEmbed(guild,
                 ACTION_TITLE, description, ACTION_REASON, true), target);
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -175,10 +175,11 @@ public final class ThisIsScamCommand extends BotCommandAdapter implements Messag
 
         long guildId = message.getGuild().getIdLong();
         long authorId = author.getIdLong();
-        String componentId = generateComponentId(String.valueOf(guildId), String.valueOf(authorId));
+        String[] args = {String.valueOf(guildId), String.valueOf(authorId)};
 
         return auditChannel.sendMessageEmbeds(reportEmbed)
-            .addActionRow(Button.success(componentId, "Yes"), Button.danger(componentId, "No"));
+            .addActionRow(Button.success(generateComponentId(args), "Yes"),
+                    Button.danger(generateComponentId(args), "No"));
     }
 
     private static String createDescription(Message target) {
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
@@ -1,0 +1,302 @@
+package org.togetherjava.tjbot.features.moderation;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.Permission;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.User;
+import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
+import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
+import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
+import net.dv8tion.jda.api.exceptions.ErrorHandler;
+import net.dv8tion.jda.api.interactions.commands.build.Commands;
+import net.dv8tion.jda.api.interactions.components.ActionRow;
+import net.dv8tion.jda.api.interactions.components.buttons.Button;
+import net.dv8tion.jda.api.interactions.components.buttons.ButtonStyle;
+import net.dv8tion.jda.api.requests.ErrorResponse;
+import net.dv8tion.jda.api.requests.RestAction;
+import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.togetherjava.tjbot.config.Config;
+import org.togetherjava.tjbot.features.BotCommandAdapter;
+import org.togetherjava.tjbot.features.CommandVisibility;
+import org.togetherjava.tjbot.features.MessageContextCommand;
+import org.togetherjava.tjbot.features.utils.Guilds;
+import org.togetherjava.tjbot.features.utils.MessageUtils;
+import org.togetherjava.tjbot.logging.LogMarkers;
+
+import java.awt.Color;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Allows users to report a message as potential scam. Moderators can confirm the report from the
+ * audit log, causing the author to be quarantined plus message history getting deleted.
+ */
+public final class ThisIsScamCommand extends BotCommandAdapter implements MessageContextCommand {
+    private static final Logger logger = LoggerFactory.getLogger(ThisIsScamCommand.class);
+
+    private static final String COMMAND_NAME = "this-is-scam";
+
+    private static final String ACTION_TITLE = "Quarantine";
+    private static final String ACTION_REASON = "Message was reported and confirmed as scam";
+
+    private static final String FAILED_MESSAGE =
+            "Sorry, there was an issue forwarding your scam report to the moderators. We are investigating.";
+    private static final Duration USER_COMMAND_COOLDOWN = Duration.ofMinutes(1);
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        long authorId = author.getIdLong();
+        String[] args = {String.valueOf(guildId), String.valueOf(authorId)};
+
+        return auditChannel.sendMessageEmbeds(reportEmbed)
+            .addActionRow(Button.success(generateComponentId(args), "Yes"),
+                    Button.danger(generateComponentId(args), "No"));
+    }
+
+    @SuppressWarnings("squid:S3457") // %n is wrong, markdown must use \n
+    private static String createDescription(Message target) {
```
SquidXTV: Couldn't you easily use multiline strings here to make it cleaner and get rid of the suppress warnings?

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+                    ModerationUtils.getQuarantinedRole(guild, config).orElseThrow())
+            .reason(ACTION_REASON);
+    }
+
+    private RestAction<Void> deleteMessagesByBanAndUnban(Guild guild, User target) {
+        return guild.ban(target, 1, TimeUnit.DAYS)
+            .reason(ACTION_REASON)
+            .flatMap(_ -> guild.unban(target).reason(ACTION_REASON));
+    }
+}
```
SquidXTV, line 313: Can't all of this be reused from the existing quarantine/moderation logic?

SquidXTV (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: not easily. but yeah, some of it can probably be moved around.

the main issue is that we only want part of the behavior but not everything.

so moving the code around could possibly result in making it too complex to understand.

but its a good comment, ill have a second look :)

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: great idea

surajkumar (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Allows users to report a message as potential scam. Moderators can confirm the report from the
+ * audit log, causing the author to be quarantined plus message history getting deleted.
+ */
+public final class ThisIsScamCommand extends BotCommandAdapter implements MessageContextCommand {
```
firasrg, line 55: I'd use a name like `ScamReportCommand` for the class, and for the command `scam-report`

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ThisIsScamCommand.class);
+
+    private static final String COMMAND_NAME = "this-is-scam";
+
+    private static final String ACTION_TITLE = "Quarantine";
```
firasrg, line 60: Since this constant is used once at line 277, I think it's better to put its value directly there where without declaring this

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+    private static final String ACTION_REASON = "Message was reported and confirmed as scam";
+
+    private static final String FAILED_MESSAGE =
+            "Sorry, there was an issue forwarding your scam report to the moderators. We are investigating.";
+    private static final Duration USER_COMMAND_COOLDOWN = Duration.ofMinutes(1);
+    private static final Color AMBIENT_COLOR = Color.decode("#CFBFF5");
```
firasrg: - Same as above, about removing the constant, since the value is used only in line 171.
- I'd make an enum for colors (e,g: `AppColors`), rather than declaring colors literally like it's happening here and in other classes (e.g `CodeMessageHandler`), or simply use the `java.awt.Color`, like you did already at line 212

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+
+    private final Cache<Long, Instant> reportedMessageToTimestamp =
+            Caffeine.newBuilder().maximumSize(10_000).expireAfterWrite(Duration.ofDays(1)).build();
+    private final Cache<Long, Instant> userToLastCommandUse = Caffeine.newBuilder()
```
firasrg, line 73: This other constant has similar builder chain as `reportedMessageToTimestamp` , why not making a helper function to avoid redundency: 
```java
private static Cache<Long, Instant> createTempCache(){
   // you can make the values 10_000 and duration of days = 1 as constants at the top so anybody can see that when they reach this class
   return Caffeine.newBuilder().maximumSize(10_000).expireAfterWrite(Duration.ofDays(1)).build();
}
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        .maximumSize(10_000)
+        .expireAfterWrite(USER_COMMAND_COOLDOWN)
+        .build();
+
+    private final Config config;
+    private final ModerationActionsStore actionsStore;
```
firasrg: I'd put these 2 fields `config` and `actionsStore` above `reportedMessageToTimestamp` and `userToLastCommandUse`, because they represent a higher level

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+    private final Predicate<String> isModAuditLogChannel;
```
firasrg: I don't understand the role of this field, it's not obvious and maybe useless. It's only being used in constructor and one method, so why not simply removing it, then you do this : 
```java
private Optional<TextChannel> getModAuditLogChannel(MessageContextInteractionEvent event) {
       // some code ..
        Optional<TextChannel> modAuditLogChannel =
                Guilds.findTextChannel(guild, Pattern.compile(config.getModAuditLogChannelPattern()).asMatchPredicate());
       // some code ..
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * Creates a new instance.
+     *
+     * @param config to resolve the moderation audit log channel and quarantined role
+     * @param actionsStore used to store issued quarantine actions
+     */
+    public ThisIsScamCommand(Config config, ModerationActionsStore actionsStore) {
+        super(Commands.message(COMMAND_NAME), CommandVisibility.GUILD);
+
+        this.config = Objects.requireNonNull(config);
```
firasrg, line 87: I think `config` cannot be null, since it's loaded at startup and it has validations inside of it

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        this.actionsStore = Objects.requireNonNull(actionsStore);
+        isModAuditLogChannel =
+                Pattern.compile(config.getModAuditLogChannelPattern()).asMatchPredicate();
+    }
+
+    @Override
+    public void onMessageContext(MessageContextInteractionEvent event) {
+        if (handleIsOnCooldown(event)) {
+            return;
```
firasrg, line 96: `log.debug` can be helpful before that `return;`, same for other `if`s below

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+                .setEphemeral(true)
+                .queue();
+            return true;
+        }
+
+        reportedMessageToTimestamp.put(messageId, Instant.now());
+        return false;
+    }
+
+    private Optional<TextChannel> getModAuditLogChannel(MessageContextInteractionEvent event) {
```
firasrg: Since the return type is `Optional<>`, I'd name it `findModAuditLogChannel()`. From my experience especially with JPA, there is a difference between Get and Find, it comes down to semantic expectations regarding predictability, presence of data, and execution cost

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        Optional<TextChannel> modAuditLog = getModAuditLogChannel(event);
+        if (modAuditLog.isEmpty()) {
+            event.reply(FAILED_MESSAGE).setEphemeral(true).queue();
+            return;
+        }
+
+        Message message = event.getTarget();
+        reportToMods(message, modAuditLog.orElseThrow()).mapToResult().map(result -> {
```
firasrg, line 109: You did check on Optional value already, why doing `.orElseThrow()`?

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+            if (result.isFailure()) {
+                logger.warn("Unable to forward a scam report to the mod audit log channel.",
+                        result.getFailure());
+                return FAILED_MESSAGE;
+            }
+            return "Thank you for your report, a moderator will take care of it 👍";
```
firasrg, line 115: I'm not sure about putting emoji in code as a return, and check [this out](https://stackoverflow.com/a/60537273)

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        }).flatMap(response -> event.reply(response).setEphemeral(true)).queue();
+    }
+
+    private boolean handleIsOnCooldown(MessageContextInteractionEvent event) {
+        Instant lastCommandUse = userToLastCommandUse.getIfPresent(event.getUser().getIdLong());
+        Runnable isNotOnCooldownAction =
+                () -> userToLastCommandUse.put(event.getUser().getIdLong(), Instant.now());
+
+        if (lastCommandUse == null) {
```
firasrg, line 124: it would be good to put `log.debug` here for debugging

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+            isNotOnCooldownAction.run();
+            return false;
+        }
+        Instant momentCooldownEnds = lastCommandUse.plus(USER_COMMAND_COOLDOWN);
+        if (Instant.now().isAfter(momentCooldownEnds)) {
```
firasrg, line 129: For consistency: `Instant.now()` is duplicated between this line and 119, it can be defined once as `final Instant now = Instant.now();`, then that constant gets called instead

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+                return FAILED_MESSAGE;
+            }
+            return "Thank you for your report, a moderator will take care of it 👍";
+        }).flatMap(response -> event.reply(response).setEphemeral(true)).queue();
+    }
+
+    private boolean handleIsOnCooldown(MessageContextInteractionEvent event) {
+        Instant lastCommandUse = userToLastCommandUse.getIfPresent(event.getUser().getIdLong());
+        Runnable isNotOnCooldownAction =
+                () -> userToLastCommandUse.put(event.getUser().getIdLong(), Instant.now());
```
firasrg: For consistency: `event.getUser().getIdLong()` is duplicated/redundent between this line and 117, it can be defined as a local constant

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+                    guild.getId(), config.getModAuditLogChannelPattern());
+        }
+        return modAuditLogChannel;
+    }
+
+    private MessageCreateAction reportToMods(Message message, TextChannel auditChannel) {
+        User author = message.getAuthor();
+        String description = createDescription(message);
+
+        MessageEmbed reportEmbed = new EmbedBuilder().setTitle("Is this Scam?")
```
firasrg, line 171: I'd put as title `Scam Report`

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+            .setFooter(author.getId())
+            .build();
+
+        long guildId = message.getGuild().getIdLong();
+        long authorId = author.getIdLong();
+        String[] args = {String.valueOf(guildId), String.valueOf(authorId)};
+
+        return auditChannel.sendMessageEmbeds(reportEmbed)
+            .addActionRow(Button.success(generateComponentId(args), "Yes"),
+                    Button.danger(generateComponentId(args), "No"));
```
firasrg, line 186: `No` and `Yes` can be defined as Enum values in something cool like the code below, im seeing them duplicates across multiple classes, maybe this is the opportunity 
```java
enum YesNo {
    YES("Y", "Yes" true), 
    NO("N", "No", false);

    private final String code;
    private final String label;
    private final boolean boolValue;

   // getters and some utility methods ..
}
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        long guildId = Long.parseLong(args.get(0));
+        long targetId = Long.parseLong(args.get(1));
+
+        ButtonStyle clickedStyle = event.getButton().getStyle();
+        boolean isScam = clickedStyle == ButtonStyle.SUCCESS;
+
+        MessageEmbed resultEmbed = new EmbedBuilder()
+            .setDescription(
+                    isScam ? "This is scam. The user was quarantined and messages were deleted."
+                            : "This is not scam, no action executed.")
```
firasrg, line 221: I'd rather say `"This is NOT Scam. The user has NOT been quarantined and NO messages have been deleted."`

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+            .setColor(isScam ? Color.GREEN : Color.RED)
+            .build();
+
+        List<MessageEmbed> embeds = new ArrayList<>(event.getMessage().getEmbeds());
+        embeds.add(resultEmbed);
+
+        event.editMessageEmbeds(embeds).setComponents().queue();
+
+        if (!isScam) {
```
firasrg, line 230: add log.debug for better future debugging

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+        return description;
+    }
+
+    @Override
+    public void onButtonClick(ButtonInteractionEvent event, List<String> args) {
+        long guildId = Long.parseLong(args.get(0));
+        long targetId = Long.parseLong(args.get(1));
+
+        ButtonStyle clickedStyle = event.getButton().getStyle();
+        boolean isScam = clickedStyle == ButtonStyle.SUCCESS;
```
firasrg, line 216: You can simplify do like this: 
```java
boolean isScam = event.getButton().getStyle() == ButtonStyle.SUCCESS;
```

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,300 @@
...
+
+        return auditChannel.sendMessageEmbeds(reportEmbed)
+            .addActionRow(Button.success(generateComponentId(args), "Yes"),
+                    Button.danger(generateComponentId(args), "No"));
+    }
+
+    @SuppressWarnings("squid:S3457") // %n is wrong, markdown must use \n
+    private static String createDescription(Message target) {
+        String content = target.getContentStripped();
+        String description = content.isBlank() ? "(empty message)" : content;
```
firasrg: i'd rather say "empty description"

firasrg (commented):
~~Thank you @Zabuzard for the great work you've put into this. I'm genuinely impressed by the scope of the changes and the amount of effort behind them. I've left comments, mostly focused on long-term maintainability and readability. I didn't spend much time reviewing the logic itself, as I trust you've given that aspect the attention it deserves.~~

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zachdehooge: Would this not be the same as what he has now? Feel that would be unnecessarily verbose to add that.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: imo all that does is making it more confusing for the user and the devs. a context command needs to be a verb, especially bc they dont have any description in the discord ui

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: kinda NIT. ill keep it as is bc this is the style also used in the other mod commands

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: those are the constants that someone might want to tweak, so better to have them here.
not sure what u mean referring the color, that is an awt color already and this is the color used for mod commands

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: makes it needlessly complicated bc:
1) they are not related so they also shouldn't use the same construction mechanism. they might differ in the future on purpose
2) that makes u jump up and down through the code just to understand the details of the cache, interrupts code flow reading. not an improvement

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: maybe, ill see

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: bc now u compile the pattern on each use. u want to compile it only once and then reuse it. regex is expensive.

This is a standard technique used everywhere in the code base

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: but we want fail fast. classes shouldn't trust their callers, only themselves.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: i dont think it would be useful. remember that we already have JDA built-in logging for everything happening out the box.

a metric might be interesting if anything

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: kinda NIT but okay

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: what do u mean? this is how u get hands on the content of an optional.
ifPresent() and then orElseThrow().
alternatively the fluent-api should be used (not that elegant in this case)

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: emojis should be put into the code. our project has a fully working unicode setup and this boosts readability a lot. also, we do this everywhere in the code base

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: already covered by jda

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: mh, ill check

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
SquidXTV: >    // you can make the values 10_000 and duration of days = 1 as constants at the top so anybody can see that when they reach this class

I also don't get how you argue about removing constants because they are used once and now say he should do it here

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ill check

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: that would be worse as then its unclear what the bot wants from u. also, it might not even be scam after all. potential scam if anything.
better is if the bot talks to u: is this scam? yes/no

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: don't put together unrelated stuff. the yes/no here has nothing to do with yes/no choices elsewhere. putting such stuff in a common shared place will make stuff worse bc then u have to rip things apart again if one of them had different needs suddenly than the other. only put together related stuff

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: if anything ud use markdown highlighting by making it **bold**.
but the point of this message is to not draw the readers attention. so highlighting it in any way would miss the goal

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: jda covers that already

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: but then we don't have to opportunity anymore to give this otherwise confusing line a good variable name that helps give context

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: that would be confusing as "description" is right in the context of the embed but not in the context of the reading user in discord.
the scan message that was reported is empty, so that is what we need to tell the user.
empty message.

i might improve the wording slightly though:
(message had no text content)

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
surajkumar: Both work fine, agree it's NIT

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
surajkumar: This is fine as-is (good how it's done). 

The ACTION_TITLE gives the string some meaning otherwise you wouldn't clearly know what the text "Quarantine" is being used for in the method it's used:

```
return ModerationUtils.sendModActionDm(ModerationUtils.getModActionEmbed(guild, ACTION_TITLE, description, ACTION_REASON, true), target);
```

vs.

```
return ModerationUtils.sendModActionDm(ModerationUtils.getModActionEmbed(guild, "Quarantine", description, ACTION_REASON, true), target);
```

same with the other params, if Java had named parameters or the function was using a builder pattern, it would have been fine to inline however since we're not, it's best to keep the context clear.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
surajkumar: NIT - optional

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
firasrg: > imo all that does is making it more confusing for the user and the devs. a context command needs to be a verb, especially bc they dont have any description in the discord ui

Then "Report Scam"

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
firasrg: IMO declaring a color like that `#CFBFF5`, can lead to confusion in the future, it's better to centralize colors somewhere and reuse. But since it sounds to you as off scope this PR, ignore.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
firasrg: > I also don't get how you argue about removing constants because they are used once and now say he should do it herea

Good catch, but look the helper function I suggested, is supposed to be used **more than once**, so I thought it'd makes sense to keep constants in this case

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
christolis: I will join the "it's a NIT" team here. It's just the name at the end of the day, and the class has a JavaDoc which makes it easier to understand what this class focuses on.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
christolis: I agree that this is a NIT comment.

At the very least this should be handled by our Spotless plugin.

christolis (dismissed):
LGTM. Thank you for your contribution 👍

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: #1506

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ✅

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ✅

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ✅

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: i checked. line 119 is a lambda, the `Instant.now()` there needs to be constructed on invocation and not before. so there is no duplication here that could be get rid of.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ✅

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: ✅

Zabuzard:
CRs done, thanks 👍 Please check again.

Also, while at it, I added the account age to the footer (might be useful in some situations and was easy to add):
grafik

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: yeah, seems a bit complex. ill skip this one for now i think

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
```diff
@@ -0,0 +1,315 @@
...
+                    Attachments:
+                    """ + attachmentInfo;
+        }
+
+        description += """
+
+
+                [Go to message](%s)""".formatted(target.getJumpUrl());
+        return description;
+    }
```
SquidXTV, line 208: I would have done it a bit differently probably, so that I only have one bigger multiline string at the end and having each "component" in that like the attachments string filled using formatted

SquidXTV (dismissed):

application/src/main/java/org/togetherjava/tjbot/features/moderation/ThisIsScamCommand.java
Zabuzard: sounds good to me, gimme a sec

Zabuzard: me likes:

grafik

grafik

Zabuzard: ✅

SquidXTV (dismissed):

Zabuzard:
Given that it had 4 approves and only subtle changes since then, I am going to merge it.

PR #1519, opened 2026-07-10
Fix: ScamBlocker does not scan edited messages

### LINKED ISSUE #1516: ScamBlocker doesn't scan edited messages, allowing scam
**Describe the bug**
`ScamBlocker` only checks messages when they are first sent (`onMessageReceived`), but not when they are edited (`onMessageUpdated`). This means a user can send a clean message nd then edit it to conain scam content, bypassing detection entirely.

**Expected behavior**
Edited messages should be scanned the same way new messages are. (They should be deleted and reported)

**To Reproduce**
For example type `@everyone $` in any channel, we expect this to be deleted which gets deleted correctly since corresponding combination is considered as scam. Now type for example `Meow` and then edit it to `Meow` -> `@everyone $`. It is not getting deleted.  

**Screenshots**

Image

__NOTE__: This is my test discord server not ours. 

**Additional context**
The one possible solution is to override `onMessageUpdated` in `ScamBlocker` and run the same scam detection logic (as in `onMessageReceived`) on the edited message content.

KonstantineVashalomidze:
Fixes: #1516 

`ScamBlocker` only checked messages on creation (`onMessageReceived`), not on edit (`onMessageUpdated`). A user could send a clean message then edit it to contain scam content. 

Extracted the detection logic into a private `handleMessage(Message, Guild)` method and called it from both event handlers. Several private methods were updated to accept `Message` and `Guild` parameters directly instead of the event object. 

Tested this manually (Since there is no `ScamBlockerTest`) on a local test discord server.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/scam/ScamBlocker.java
@@ -158,14 +158,14 @@ public final class ScamBlocker extends MessageReceiverAdapter implements UserInt
         }
 
         if (scamHistoryStore.hasRecentScamDuplicate(message)) {
-            takeActionWasAlreadyReported(message, guild);
+            takeActionWasAlreadyReported(message);
             return;
         }
 
         takeAction(message, guild);
     }
 
-    private void takeActionWasAlreadyReported(Message message, Guild guild) {
+    private void takeActionWasAlreadyReported(Message message) {
         // The user recently send the same scam already, and that was already reported and handled
         addScamToHistory(message);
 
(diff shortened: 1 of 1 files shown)


Zabuzard (changes requested):
This was done on purpose as we never ever had scam that came through an edit so far.
Is this now a real use case? If not, I would probably vote to reject this PR as it would only bear risks (more false positives) than covering more (real) scam.

KonstantineVashalomidze:
@Zabuzard 
Thanks for the feedback. I see your point. If such cases hasn't happened or doesn't happen usually I will be happy to close this PR.

KonstantineVashalomidze:
Closing this pull request as it might introduce false positives of scam detection during edit, and since no scammers are trying to edit their messages with scam content that much.

PR #1530, opened 2026-07-25
Add feature to support 14 days and 1 year bans. fixes: #1527

### LINKED ISSUE #1527: Add 14 days and 1 year ban option
Bans for 7 and 30 days are possible, but we would like a middle ground of 14 days.
Similarly, we can permanently ban, but we are lacking an option to do a long but not infinite ban, so we would like a 1 year option.

Add a 14 days choice and a 1 year choice in the ban command.
This can be done simply by adding in `"14 days"` and `"1 year"`  in the field `DURATIONS ` of class `BanCommand`.
Note that TJ Bot doesn't support year based temporals, so a `case "year", "years"` in `ModerationUtils.computeTemporaryData` should be added.
Testing could be done by changing computer time.

**Additional context**

Image

KonstantineVashalomidze:
fixes: #1527
New feature was added for admins. Admins can now also ban with durations 14 days or 1 year.

The following now looks like this:
image

Testing was done manually. I invited my friend into my test server and banned him for 14 days and 1 year. He was banned and then I have advanced system time further than 14 days and 1 year. Ban was revoked and after routine run he was unbanned.

```
15:35:28.878 [pool-4-thread-4] WARN  org.togetherjava.tjbot.features.help.HelpThreadActivityUpdater - Unable to update help thread activities, did not find a help forum matching the configured pattern 'questions' for guild 'Mama africa's server'
15:38:32.984 [ForkJoinPool.commonPool-worker-2] INFO  org.togetherjava.tjbot.features.moderation.BanCommand - 'kosta4115' (1021391825677860975) banned the user 'antiriad7' (488754445673431042) for 14 days from guild 'Mama africa's server' and deleted their message history of the last 0 days, for reason 'kkk'.
15:39:28.131 [ForkJoinPool.commonPool-worker-3] INFO  org.togetherjava.tjbot.features.moderation.temp.TemporaryModerationRoutine - Revoked temporary action BAN against user 'antiriad7' (488754445673431042).
```

Note that logs are within the same console since, I have just reloaded the project with Intellij devtools feature since I had  problems with certificate when running app from scratch. 

EDIT:
I have made changes as @Zabuzard and @Alathreon suggested. Since `.plus(...)` method doesn't support years, I had to convert number of years into number of days and use unit of `DAYS` to calculate ban period.

EDIT 2:
We have decided just to have 365 days option inserted instead of one 1 year.

### DIFF
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/BanCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/BanCommand.java
@@ -50,7 +50,7 @@ public final class BanCommand extends SlashCommandAdapter {
     private static final String ACTION_TITLE = "Ban";
     @SuppressWarnings("StaticCollection")
     private static final List<String> DURATIONS = List.of(ModerationUtils.PERMANENT_DURATION,
-            "1 hour", "3 hours", "1 day", "2 days", "3 days", "7 days", "30 days");
+            "1 hour", "3 hours", "1 day", "2 days", "3 days", "7 days", "14 days", "30 days", "365 days");
     private final ModerationActionsStore actionsStore;
 
     /**
--- a/application/src/main/java/org/togetherjava/tjbot/features/moderation/BanCommand.java
+++ b/application/src/main/java/org/togetherjava/tjbot/features/moderation/BanCommand.java
@@ -49,8 +49,9 @@ public final class BanCommand extends SlashCommandAdapter {
     private static final String ACTION_VERB = "ban";
     private static final String ACTION_TITLE = "Ban";
     @SuppressWarnings("StaticCollection")
-    private static final List<String> DURATIONS = List.of(ModerationUtils.PERMANENT_DURATION,
-            "1 hour", "3 hours", "1 day", "2 days", "3 days", "7 days", "14 days", "30 days", "365 days");
+    private static final List<String> DURATIONS =
+            List.of(ModerationUtils.PERMANENT_DURATION, "1 hour", "3 hours", "1 day", "2 days",
+                    "3 days", "7 days", "14 days", "30 days", "365 days");
     private final ModerationActionsStore actionsStore;
 
     /**


application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
```diff
@@ -360,6 +360,11 @@ static Optional<TemporaryData> computeTemporaryData(String durationText) {
         String[] data = durationText.split(" ", 2);
         int duration = Integer.parseInt(data[0]);
         ChronoUnit unit = switch (data[1]) {
+            case "year", "years" -> {
+                // Represents a year (1 year = 365 days)
+                duration = duration * 365;
+                yield ChronoUnit.DAYS;
+            }
```
Zabuzard: I find it confusing that the section setting the unit manipulates the duration as side effect.

I would move the duration multiplication up to the section where the duration is set instead.
Btw, you can use `*=` as shortcut.

Zabuzard (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
KonstantineVashalomidze: @Zabuzard Yes I get your point and I agree, didn't thought about that. Since @Alathreon said that ```case "year", "years"``` in ```ModerationUtils.computeTemporaryData``` should be added, that is why I did like that. Either way plus method in Temporal utils doesn't support years at all and I think having 365 days instead of 1 year directly is the best option here. I would love to hear your thoughts about this.

Zabuzard: Yes, it is because a year isnt always 365 days. Time is complex.

In this case it is okay to have it convert to days, I would just layout the code differently:

int duration =...
if (years) {
 duration *= 365;
}

Unit unit = switch (...)
{
  case years -> DAYS
 ... 
}

KonstantineVashalomidze: @Zabuzard I have made those changes. Can you please take a look at it. I've preferred using .startsWith("year") instead of .equals("year"), .equals("years"). Would love your opinion about this.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
```diff
@@ -359,10 +360,14 @@ static Optional<TemporaryData> computeTemporaryData(String durationText) {
         // 1 minute, 1 day, 2 days, ...
         String[] data = durationText.split(" ", 2);
         int duration = Integer.parseInt(data[0]);
+        if (data[1].toLowerCase().startsWith("year")) {
+            duration = Year.now().length();
```
Alathreon: Override the duration ???

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
```diff
@@ -359,10 +360,14 @@ static Optional<TemporaryData> computeTemporaryData(String durationText) {
...
+        }
+
         ChronoUnit unit = switch (data[1]) {
             case "minute", "minutes" -> ChronoUnit.MINUTES;
             case "hour", "hours" -> ChronoUnit.HOURS;
-            case "day", "days" -> ChronoUnit.DAYS;
+            case "day", "days", "year", "years" -> ChronoUnit.DAYS;
```
Alathreon: No, it should be its own case with ChronoUnit.YEARS.

Alathreon (changes requested):

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
Alathreon: Ah I see the problem, well it's a side effect of this system being about storing the duration rather than substracting time.
If you really want to implement it precisely, you would need to loop on every year based on duration as the loop index, and add the year duration for each.
I don't think we need to be presice too. It will be correct in most cases anyways.

Alathreon: After checking how ChronoUnit works, it seems that ChronoUnit.YEARS indeed correctly handles leap years so it should be used instead.

KonstantineVashalomidze: Yah, my bad it works here accidently since we have 1 year option. You are right looping is one solution but it was 5 lines. So, I have changed the approach. I've decided to calculate days between two specific given years using this:
```java
LocalDate now = LocalDate.now();
duration *= (int) ChronoUnit.DAYS.between(now, now.plusYears(duration));
```
Which I believe counts in leap years, but I would need an approval from you about it, since I am not 100% sure.

KonstantineVashalomidze: @Alathreon

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
KonstantineVashalomidze: @Alathreon Got it. Will change this. Would like to hear why though?

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
Alathreon: Again you are duplicating code that already exist, just do what i said in the issue and above, case "year", "years" -> ChronoUnit.YEARS

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
KonstantineVashalomidze: That is what @Zabuzard told me to do.

application/src/main/java/org/togetherjava/tjbot/features/moderation/ModerationUtils.java
KonstantineVashalomidze: @Alathreon Oh I didn't read that message, sorry for that. I didn't get notification about it. I was probably typing my own thing and since comments are not real time I just commented after your message.

KonstantineVashalomidze: @Alathreon I tried switching to `case "year", "years" -> ChronoUnit.YEARS` like you suggested, but `Instant.plus(...)` does not support `YEARS` only `DAYS` and below. Calling it throws `UnsupportedTemporalTypeException`. 
If this is not what you mean, can you explain what's your suggestion in one message so that I will understand it. 
Thanks for the guidance again.

Alathreon: Ah that's a problem huh

Alathreon: case "year", "years" -> {
    LocalDateTime now = LocalDateTime.now();
    duration = (int) ChronoUnit.DAYS.between(now, now.plusYears(duration));
    yield ChronoUnit.DAYS;
}

I guess you could do something like that yea.

Alathreon: Actually maybe we are hunting the wrong bird.
@Zabuzard I noticed that for a month, instead of using "1 month", we use "30 days". So what do you think about replacing "1 year" to "365 days" ? It would dodge the problem and would be consistent with the design of the command and the util class.

KonstantineVashalomidze: @Alathreon You mean putting that if inside switch? Sure, it was @Zabuzard's Suggestion not to do that: `"I would move the duration multiplication up to the section where the duration is set instead."`
I have one question though. What is the difference between doing just `LocalDateTime` instead of `LocalDate`.

KonstantineVashalomidze: > @Zabuzard Yes I get your point and I agree, didn't thought about that. Since @Alathreon said that `case "year", "years"` in `ModerationUtils.computeTemporaryData` should be added, that is why I did like that. Either way plus method in Temporal utils doesn't support years at all and I think having 365 days instead of 1 year directly is the best option here. I would love to hear your thoughts about this.

@Alathreon This was on of my offers above. And I think this was the cleanest solution problem is just the leap years, but one day is not a big difference in 1 year ban IMO.

Alathreon: No I think that's the best solution, because that's already the solution we use for months
We don't say ban "1 month", we say ban "30 days", even if 30 days isn't always a month.
So I think "365 days" is the best compromise.
I think you should implement it like that, it would mean that you can rollback any change in the mod utils class and have an even simpler PR. I'll ask marko what he thinks since he was the one who wanted the 1 year ban.

KonstantineVashalomidze: @Alathreon Okay I have updated the code please take a look if you don't mind.

Alathreon: I meant to use "365 days" instead of "1 year".

KonstantineVashalomidze: @Alathreon Ah yeah, reverted wrong commit. Now it is like that.

Alathreon: Looks good, could you please test it ?

KonstantineVashalomidze: Yes, I could test it. I already did before but no problem I will test it again.

KonstantineVashalomidze: I don't think that 365 days test is possible though. Since I get certificate errors when running the app. So I am only testing for 14 days. Is this okay? @Alathreon

Alathreon: Could you change the date of your computer ? Or maybe edit the date in the db if that's the only way ?

KonstantineVashalomidze: That's what I am trying to say. Changing system date gives such error:
```bash
12:28:06.205 [ConcurrentSessionController-Worker-0] ERROR net.dv8tion.jda.api.utils.SessionControllerAdapter - Failed to establish connection for a node, appending to queue
java.lang.IllegalStateException: com.neovisionaries.ws.client.HostnameUnverifiedException: The certificate of the peer does not match the expected hostname (gateway.discord.gg)

```

KonstantineVashalomidze: I will change date in the db and test that way.

Alathreon: > That's what I am trying to say. Changing system date gives such error:
> 
> ```shell
> 12:28:06.205 [ConcurrentSessionController-Worker-0] ERROR net.dv8tion.jda.api.utils.SessionControllerAdapter - Failed to establish connection for a node, appending to queue
> java.lang.IllegalStateException: com.neovisionaries.ws.client.HostnameUnverifiedException: The certificate of the peer does not match the expected hostname (gateway.discord.gg)
> ```

Did you try shuting down the bot before changing the date ?

KonstantineVashalomidze: Yes of course.

KonstantineVashalomidze: @Alathreon I did the testing of both cases 14 days and 365 days. Specifically I tested the behavior like this:
I added my friend's discord account in my test server. Then I banned my friend's account for 14 days it got banned and my friend correctly received ban duration as 14 days in DM. I've looked in the `moderation_actions` table and there were a record inserted indicating banned user. Then I've changed `action_expires_at` date for that record to past and once `TemporaryModerationRoutine` was executed my friend's account got unbanned:
```bash
13:07:38.481 [ForkJoinPool.commonPool-worker-1] INFO  org.togetherjava.tjbot.features.moderation.BanCommand - 'kosta4115' (1021391825677860975) banned the user 'konstantinevashalomidze.' (1530873641397522463) for 14 days from guild 'Mama africa's server' and deleted their message history of the last 0 days, for reason '14 days ban'.
13:08:49.988 [ForkJoinPool.commonPool-worker-2] INFO  org.togetherjava.tjbot.features.moderation.temp.TemporaryModerationRoutine - Revoked temporary action BAN against user 'konstantinevashalomidze.' (1530873641397522463).
13:14:18.718 [ForkJoinPool.commonPool-worker-3] INFO  org.togetherjava.tjbot.features.moderation.BanCommand - 'kosta4115' (1021391825677860975) banned the user 'konstantinevashalomidze.' (1530873641397522463) for 365 days from guild 'Mama africa's server' and deleted their message history of the last 0 days, for reason 'no reason'.
13:14:50.221 [ForkJoinPool.commonPool-worker-3] INFO  org.togetherjava.tjbot.features.moderation.temp.TemporaryModerationRoutine - Revoked temporary action BAN against user 'konstantinevashalomidze.' (1530873641397522463).
```
 
 What I found the strange is that bot didn't inform me about ban expiration or send me a rejoin link to rejoin in server. I think this is expected behavior but it would be nicer if it included some kind of notification about it and link to joining back to server again.
 
 So basically everything works as expected. Anything else I can help with?

Alathreon (dismissed):

Zabuzard (dismissed):

Zabuzard:
Looks liike the line is too loong or something. Can you run spotless quickly, thanks.

KonstantineVashalomidze:
> Looks liike the line is too loong or something. Can you run spotless quickly, thanks.

@Zabuzard Done. spotless check passes locally.

Zabuzard (approved):

Alathreon (approved):

